Issue #18, opened 2018-03-09
Mouse clicks do not work on LG phones


Davidobot:
I am using the windows precompiled build on Windows 10 x64 and a phone running 7.0.

The program starts up fine and shows the screen perfectly, but whenever I left click on the screen with my mouse (on my PC) nothing happens.

Right-clicking does turn on the screen and all the shortcuts work. I do have two monitors on my PC, if that matters.

EDIT: I am using a LG G5 and this seems to be a common problem on LG devices?
[Reddit thread](https://www.reddit.com/r/Android/comments/834zmr/introducing_scrcpy_an_app_to_display_and_control/dvfkmva/?utm_content=permalink&utm_medium=front&utm_source=reddit&utm_name=Android)

Touch data shows correct position, but pressure is always at 1.

rom1v:
> Touch data shows correct position, but pressure is always at 1.

I don't have an LG device here, so I can't test for now.

Indeed, [the pressure is always 1](https://github.com/Genymobile/scrcpy/blob/v1.0/server/src/main/java/com/genymobile/scrcpy/EventController.java#L39-L45). Maybe you could make it work by changing some values here?

willyb321:
Using an LG K10 here, same issue. Can also confirm pressure is always 1.

DerekJedral:
Vysor also reads uses pressure, size as 1, so I don't think it's likely this is the issue (though I haven't had a chance to change the values yet to test)

DerekJedral:
Nevermind, turns out I'm wrong. A pressure of under 1 seems to fix the issues (tried with a pressure of 0.9999f)

Cascador:
Hello,

Using an LG G4 here, left click doesn't work. Bug is confirmed, same computer, left click works with an Samsung Galaxy A5 2016.

@rom1v can you create a prebuilt server with pressure of 0.9999f ?

We can see the pressure by activating Show touch data (Settings, Developer options).

Tcho !

rom1v:
If you have an LG device, could you confirm that https://github.com/Genymobile/scrcpy/issues/18 fixes the problem, please?

If you have another device, please confirm that it still works as before.

Thank you for your help.

Cascador:
How do i proceed exactly ? I use the prebuilt server.

Thx, Tcho !

rom1v:
@DerekJedral

> Vysor also reads uses pressure, size as 1, so I don't think it's likely this is the issue

[...]

PR #27, opened 2018-03-10
Fix touch events for LG devices

DerekJedral:
Haven't had the ability to test on other devices, but this solves the touch events issue on my V20

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/EventController.java
+++ b/server/src/main/java/com/genymobile/scrcpy/EventController.java
@@ -37,7 +37,7 @@ public class EventController {
 
         MotionEvent.PointerCoords coords = pointerCoords[0];
         coords.orientation = 0;
-        coords.pressure = 1;
+        coords.pressure = 0.9999f;
         coords.size = 1;
         coords.toolMajor = 1;
         coords.toolMinor = 1;


rom1v:
Wow, cool it the problem is fixed with it! :+1:

I will test on other devices before merging.

rom1v:
Could you add a comment in code to say that 1 does not work for LG devices and add a link to https://github.com/Genymobile/scrcpy/issues/18, please?

rom1v:
I'm wondering, what if instead you remove these lines:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/EventController.java b/server/src/main/java/com/genymobile/scrcpy/EventController.java
index 4a0ce33..547e20c 100644
--- a/server/src/main/java/com/genymobile/scrcpy/EventController.java
+++ b/server/src/main/java/com/genymobile/scrcpy/EventController.java
@@ -39,10 +39,6 @@ public class EventController {
         coords.orientation = 0;
         coords.pressure = 1;
         coords.size = 1;
-        coords.toolMajor = 1;
-        coords.toolMinor = 1;
-        coords.touchMajor = 1;
-        coords.touchMinor = 1;
     }
 
     private void setPointerCoords(Point point) {
```

Does it work for LG devices?

Cascador:
Works too.

Tcho !

rom1v:
OK, then I fixed it that way (it seems less hacky): https://github.com/Genymobile/scrcpy/commit/c075ad0a1e9e496bd2b0adf353d539933aa8c7ee

Thank you all.

PR #28, opened 2018-03-10
added "$" in front of terminal commands and some syntax coloring to README.md

sdushantha:
#### Before
```
sudo cd here && domagic
```
#### Now
```bash
$ sudo cd here && domagic
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,15 +38,15 @@ The client requires _FFmpeg_ and _LibSDL2_.
 #### Linux
 
 Install the required packages from your package manager (here, for Debian):
+```bash
+# runtime dependencies
+$ sudo apt install ffmpeg libsdl2-2.0.0
 
-    # runtime dependencies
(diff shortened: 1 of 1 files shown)


rom1v:
Hi, thank you for improving the README.

I agree for syntax highlighting. :+1:

However I disagree about adding `$`: I usually put them to show the result of a command (to highlight what the command is), but not for scripts/commands which could be copy/pasted (otherwise the `$` should be removed on every line before pasting).

sdushantha:
I have removed the ```$```. Also, it is MacOS not Mac OS

:smile:

rom1v:
Thanks :+1: I just merged it.

What is your text editor? It seems it inserts non-breaking-spaces at random places (even for indentation).

I fixed this, and added empty lines around code blocks: https://github.com/Genymobile/scrcpy/commit/cc4a0152567d0c7e007de140089fa0e5e139d417

PR #93, opened 2018-03-24
Add links to FFmpeg and LibSDL2 dependencies

CampbellOwen:


rom1v:
Thank you for your PR.

It's a good idea to provide links :+1:

However, I'm a bit afraid that with links to FFmpeg and LibSDL2 download pages, people will try to download and install them from here, which is not what is expected.

Instead, maybe linking to their wikipedia page would be less error-prone:
 - https://en.wikipedia.org/wiki/FFmpeg
 - https://en.wikipedia.org/wiki/Simple_DirectMedia_Layer

What do you think?

CampbellOwen:
I agree, that's a good point. Changed

PR #94, opened 2018-03-25
Add instructions to run via Docker

pierlon:
I made a [Docker image](https://github.com/pierlon/scrcpy-docker) so one could quickly try out `scrcpy` if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,10 @@ install it manually and make it available from the `PATH`:
 export PATH="$JAVA_HOME/bin:$PATH"
 ```
 
+#### Docker
+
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you for providing docker images :+1:

I merge the reference to your repo in the README.

One suggestion: for simplicity, maybe the docker files for intel, amd and nvidia could be in subdirectories instead of different branches?

pierlon:
Thanks for the suggestion. I wanted automated builds for Docker Hub but I might've overcomplicated things :sweat_smile:

qymspace:
adb devices returns "No devices". 
I have adb locally which is working okay but the one in docker doesnt.

PR #133, opened 2018-04-28
[Feature] Drag and drop to install apk files from your computer

AdoPi:
I have just added a little feature to install apk files by using drag and drop.

So, just drag and drop an apk file and wait for its installation to complete.


rom1v:
Oh, I saw the title of the PR in my mails, and thought it was a feature request, I didn't realize it was a pull request :)

That's cool, I didn't know about `SDL_DROPFILE` :+1: Thank you.

However, currently, the installation blocks the main thread (so the UI is frozen), it should be executed in a separate thread instead.

(Besides, since it is a new feature, it should target `dev` branch)

AdoPi:
Thanks :) 
I am going to work on a version which uses separate thread so.
Sorry I don't know how to change the target of an existing pull request, do you know how can I do that?

AdoPi:
[dc68d99](https://github.com/Genymobile/scrcpy/pull/133/commits/dc68d994d0ef84c34dacde6295e51d2231beaef8)  commit is a quick fix.

AdoPi:
> how to manage the separate thread? create a new one each time? reuse it?
> what happens if we start a new installation before the previous one is complete? (serialized on the same thread, in parallel…)

My idea was to make an executor thread which takes commands from a command queue and execute them (thus serialized on the same thread). 
The latest commit was just here to have something which works while I'm working on a better design which may take some time 👎 

> what happens if we close scrcpy while an installation is running?

It depends on adb, I think we can expect the same behaviour as when we disconnect the phone while adb is installing something on it.

rom1v:
> My idea was to make an executor thread which takes commands from a command queue and execute them (thus serialized on the same thread).

:+1:

The executor thread could be created only on the first time it is required (APK installations will probably be rare compared to _scrcpy_ executions).

> It depends on adb, I think we can expect the same behaviour as when we disconnect the phone while adb is installing something on it.

Yes, killing the process (like `server.c` does) is ok for me :+1:

AdoPi:
The latest commit should add this drag&drop feature in a more proper way.
What do you think? :)

rom1v:
Thank you :+1: I will review this when I have some time, hopefully this week-end.

app/src/apkinstaller.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <SDL2/SDL_mutex.h>
+#include <SDL2/SDL_stdinc.h>
+#include <SDL2/SDL_thread.h>
+#include "command.h"
+
+// NOTE(AdoPi) apk_queue and control_event can use a generic queue
+
+struct apk_queue {
+    char data[APK_QUEUE_SIZE][MAX_FILENAME_SIZE];
```
rom1v: That's wasteful and unsafe. You should use allocated strings (with `SDL_strdup`) instead (and use `char *data[APK_QUEUE_SIZE]` here).

app/src/apkinstaller.h
```diff
@@ -0,0 +1,55 @@
...
+
+#define INSTALLER_INITIALIZER {     \
+    .serial = NULL,                 \
+    .thread = NULL,                 \
+    .mutex = NULL,                  \
+    .event_cond = NULL,             \
+    .stopped = SDL_FALSE,           \
+    .initialized = SDL_FALSE,       \
+    .current_process = PROCESS_NONE \
+}
```
rom1v: To initialize an object, either we can use a static initializer or an init method, but if both are provided they should be equivalent.

app/src/apkinstaller.c
```diff
@@ -0,0 +1,157 @@
...
+SDL_bool apk_queue_take(struct apk_queue *queue, char *apk) {
+    if (apk_queue_is_empty(queue)) {
+        return SDL_FALSE;
+    }
+    strncpy(apk, queue->data[queue->tail], MAX_FILENAME_SIZE);
+    queue->tail = (queue->tail + 1) % APK_QUEUE_SIZE;
+    return SDL_TRUE;
+}
+
+SDL_bool installer_init(struct installer *installer, const char* serial) {
```
rom1v: `char* serial` → `char *serial` (there are others in the whole diff)

app/src/scrcpy.c
```diff
@@ -102,6 +104,18 @@ static void event_loop(void) {
...
                 break;
+            case SDL_DROPFILE:
+
+                if (!installer.initialized) {
+                    SDL_bool init_ok = installer_init(&installer, server.serial);
+                    if (init_ok && installer_start(&installer)) {
+                        goto push;
+                    }
+                    installer_destroy(&installer);
+                }
```
rom1v: There is a logic flaw here: if initialization fails, it still calls `installer_push_apk()`.

app/src/apkinstaller.h
```diff
@@ -0,0 +1,55 @@
...
+
+// NOTE(AdoPi) apk_queue and control_event can use a generic queue
+
+struct apk_queue {
+    char data[APK_QUEUE_SIZE][MAX_FILENAME_SIZE];
+    int tail;
+    int head;
+};
+
+struct installer {
```
rom1v: Either `installer.h` and `struct installer`, either `apkinstaller.h` and `struct apk_installer`.

app/meson.build
```diff
@@ -16,6 +16,7 @@ src = [
     'src/server.c',
     'src/strutil.c',
     'src/tinyxpm.c',
+    'src/apkinstaller.c'
```
rom1v: (move it to preserve alphabetical order – except `main.c`)
(maybe I should also sort `main.c` with the others)

app/src/apkinstaller.h
```diff
@@ -0,0 +1,55 @@
+#ifndef APK_INSTALLER_H
+#define APK_INSTALLER_H
+
+#define APK_QUEUE_SIZE 16
+#define MAX_FILENAME_SIZE 1024
+
+
+#include "apkinstaller.h"
```
rom1v: :confused:

app/src/apkinstaller.h
```diff
@@ -0,0 +1,55 @@
+#ifndef APK_INSTALLER_H
+#define APK_INSTALLER_H
+
+#define APK_QUEUE_SIZE 16
+#define MAX_FILENAME_SIZE 1024
```
rom1v: should be after the includes

app/src/apkinstaller.c
```diff
@@ -0,0 +1,157 @@
+#include "apkinstaller.h"
+
+#include "lockutil.h"
+#include "log.h"
+#include "command.h"
+#include <string.h>
```
rom1v: (nitpicking)

For consistency:
```
#include "apkinstaller.h"

#include <string.h>
#include "command.h"
#include "lockutil.h"
#include "log.h"
```
(first the associated .h, then the sorted `<>` includes, then the sorted `""` includes)

app/src/apkinstaller.h
```diff
@@ -0,0 +1,55 @@
...
+    .serial = NULL,                 \
+    .thread = NULL,                 \
+    .mutex = NULL,                  \
+    .event_cond = NULL,             \
+    .stopped = SDL_FALSE,           \
+    .initialized = SDL_FALSE,       \
+    .current_process = PROCESS_NONE \
+}
+
+
```
rom1v: double empty line

app/src/apkinstaller.c
```diff
@@ -0,0 +1,157 @@
...
+
+
```
rom1v: double newline

AdoPi:
Big thanks for your review and the time spent, I will work on it 👍

app/src/scrcpy.c
```diff
@@ -102,6 +104,18 @@ static void event_loop(void) {
             case SDL_MOUSEBUTTONUP:
                 input_manager_process_mouse_button(&input_manager, &event.button);
                 break;
+            case SDL_DROPFILE:
+
+                if (!installer.initialized) {
```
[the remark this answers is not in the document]
AdoPi: Cool! I was wondering whether or not it was an implementation detail.

app/src/apkinstaller.h
AdoPi: Sorry about that 👎

app/src/scrcpy.c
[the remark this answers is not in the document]
rom1v: You're right, it is possible not to hide the lazy initialization inside the installer: the installer could always be fully initialized on init, but the caller may call _init_ only when necessary. In that case, the _initialized_ flag would be on the caller side.

AdoPi:
Thanks for the advice, it should be better now.
Keep me informed :)

rom1v:
Thank you for you changes. I will probably not have time to review this week-end, but don't worry, I'm thinking about it :wink:

rom1v:
Hi,

Thank you for the feature.

I fixed some memory leaks and released the mutex while installing (I made me [fix the controller](https://github.com/Genymobile/scrcpy/commit/bb3a7f05ac4fe20ee0c5b1b1d1e6efa5fc7d8586) for the same reason).

Here are the changes I suggest:
```diff
diff --git a/app/src/installer.c b/app/src/installer.c
index 95a15db..2bef19d 100644
--- a/app/src/installer.c
+++ b/app/src/installer.c
@@ -44,7 +44,8 @@ SDL_bool apk_queue_take(struct apk_queue *queue, char **apk) {
     if (apk_queue_is_empty(queue)) {
         return SDL_FALSE;
     }
-    *apk = SDL_strdup(queue->data[queue->tail]);
+    // transfer ownership
+    *apk = queue->data[queue->tail];
     queue->tail = (queue->tail + 1) % APK_QUEUE_SIZE;
     return SDL_TRUE;
 }
@@ -64,9 +65,14 @@ SDL_bool installer_init(struct installer *installer, const char *serial) {
         return SDL_FALSE;
     }
 
-    installer->serial = NULL;
     if (serial) {
         installer->serial = SDL_strdup(serial);
+        if (!installer->serial) {
+            LOGW("Cannot strdup serial");
+            return SDL_FALSE;
+        }
+    } else {
+        installer->serial = NULL;
     }
 
     // lazy initialization
@@ -104,35 +110,40 @@ SDL_bool installer_install_apk(struct installer *installer, const char *apk) {
     return res;
 }
 
-static SDL_bool process_install(struct installer *installer, const char *filename) {
-    LOGI("%s will be installed",filename);
-    process_t process = adb_install(installer->serial, filename);
-    installer->current_process = process;
-    return process_check_success(process, "adb install");
-}
-
 static int run_installer(void *data) {
     struct installer *installer = data;
 
-    char *current_apk;
-    mutex_lock(installer->mutex);
     for (;;) {
+        mutex_lock(installer->mutex);
         while (!installer->stopped && apk_queue_is_empty(&installer->queue)) {
             cond_wait(installer->event_cond, installer->mutex);
         }
         if (installer->stopped) {
             // stop immediately, do not process further events
+            mutex_unlock(installer->mutex);
             break;
         }
-        while (apk_queue_take(&installer->queue, &current_apk)) {
-            SDL_bool ok = process_install(installer,current_apk);
-            if (!ok) {
-                LOGD("Error during installation");
-            }
+        char *current_apk;
+#ifdef BUILD_DEBUG
+        bool non_empty = apk_queue_take(&installer->queue, &current_apk);
+        SDL_assert(non_empty);
+#else
+        apk_queue_take(&installer->queue, &current_apk);
+#endif
+
+        LOGI("Installing %s...", current_apk);
+        process_t process = adb_install(installer->serial, current_apk);
+        installer->current_process = process;
+
+        mutex_unlock(installer->mutex);
+
+        if (process_check_success(process, "adb install")) {
+            LOGI("%s installed successfully", current_apk);
+        } else {
+            LOGE("Failed to install %s", current_apk);
         }
+        SDL_free(current_apk);
     }
-    mutex_unlock(installer->mutex);
-    SDL_free(current_apk);
     return 0;
 }
 
@@ -152,7 +163,7 @@ void installer_stop(struct installer *installer) {
     mutex_lock(installer->mutex);
     installer->stopped = SDL_TRUE;
     cond_signal(installer->event_cond);
-    if (installer->current_process == PROCESS_NONE) {
+    if (installer->current_process != PROCESS_NONE) {
         if (!cmd_terminate(installer->current_process)) {
             LOGW("Cannot terminate install process");
         }
diff --git a/app/src/scrcpy.c b/app/src/scrcpy.c
index 03bdc19..4e74bb9 100644
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -152,10 +152,10 @@ SDL_bool scrcpy(const char *serial, Uint16 local_port, Uint16 max_size, Uint32 b
         goto finally_destroy_server;
     }
 
-    if (!installer_init(&installer,server.serial)) {
+    if (!installer_init(&installer, server.serial)) {
         ret = SDL_FALSE;
         server_stop(&server);
-        goto finally_destroy_installer;
+        goto finally_destroy_frames;
     }
 
     decoder_init(&decoder, &frames, device_socket);
@@ -165,7 +165,7 @@ SDL_bool scrcpy(const char *serial, Uint16 local_port, Uint16 max_size, Uint32 b
     if (!decoder_start(&decoder)) {
         ret = SDL_FALSE;
         server_stop(&server);
-        goto finally_destroy_frames;
+        goto finally_destroy_installer;
     }
 
     if (!controller_init(&controller, device_socket)) {
```

Please review :wink: If it's ok, I'll squash all these commits into one on `dev`.

I'll also add a workaround for Windows to support spaces in path:
```diff
diff --git a/app/src/command.c b/app/src/command.c
index 463eb42..ec1674c 100644
--- a/app/src/command.c
+++ b/app/src/command.c
@@ -74,6 +74,16 @@ process_t adb_push(const char *serial, const char *local, const char *remote) {
 }
 
 process_t adb_install(const char *serial, const char *local) {
+#ifdef __WINDOWS__
+    // Windows will parse the string, so the local name must be quoted (see sys/win/command.c)
+    size_t len = strlen(local);
+    char quoted[len + 3];
+    strcpy(&quoted[1], local); // strcpy safe here, with previous strlen()
+    quoted[0] = '"';
+    quoted[len + 1] = '"';
+    quoted[len + 2] = '\0';
+    local = quoted;
+#endif
     const char *const adb_cmd[] = {"install", "-r", local};
     return adb_execute(serial, adb_cmd, ARRAY_LEN(adb_cmd));
 }
```

AdoPi:
Oh, I was not aware of the spaces problem for Windows path, that's cool to know (I don't really use Windows a lot).

Something in the mutex code was interpelling me too but I couldn't find why, nice job 💯 
It seems to be fine for me!

rom1v:
Commited as https://github.com/Genymobile/scrcpy/commit/e2a2973990ecd31b8e4c1638eed3368a96a1e9c0, and released in v1.2 :+1: :tada:

PR #142, opened 2018-05-04
Added a useful guide link for the FAQ

Samir-Rashid:
it may belong somewhere else, but I thought that this video guide (linked above) would be a better explanation for n00bs than reading a block of text. lmk..

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,5 +1,9 @@
 # Frequently Asked Questions
 
+### Troubles with the installation process?
+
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -1,5 +1,9 @@
 # Frequently Asked Questions
 
+### Troubles with the installation process?
```
rom1v: You should add *on Windows* (the xda video explains only for Windows).

Also, there is no installation involved. Maybe "Troubles to make it work on Windows?" or something similar and better. What do you think?

PR #194, opened 2018-07-10
Add Czech language characters

a1346054:
Please check code for code style and other problems. I did not compile it to see if it runs.


rom1v:
Thank you. The code seems correct, but in practice it does not work on my devices.

Maybe the only supported combinations are those described [here](https://source.android.com/devices/input/key-character-map-files#behaviors) (the ones already implemented):

> The system recognizes the following Unicode characters as combining diacritical dead key characters:
>
>  -  `'\u0300'`: Grave accent.
>  -  `'\u0301'`: Acute accent.
>  -  `'\u0302'`: Circumflex accent.
>  -  `'\u0303'`: Tilde accent.
>  -  `'\u0308'`: Umlaut accent.

a1346054:
I don't know what to say :( Sounds wrong that android does not properly support other keyboard languages than english :(

PR #204, opened 2018-07-20
Tweak the exit code

Grief:
Make scrscpy to return 0 exit code only in case when the user closes its app, otherwise, i.e. in case of connection loss, return 1

It's useful in scripts, when you'd like to restart scrcpy automatically in case of network issues.

### DIFF
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -56,7 +56,7 @@ static int event_watcher(void *data, SDL_Event *event) {
 }
 #endif
 
-static void event_loop(void) {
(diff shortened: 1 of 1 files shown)


rom1v:
Oh, sorry, I totally missed this PR. I quickly read it on the phone, looks good to me.

I'll check later on the computer and merge it :+1:

rom1v:
Oh, sorry, I totally missed this PR. I quickly read it on the phone, looks good to me.

I'll check later on the computer and merge it :+1:

Grief:
@rom1v Thanks!

PR #236, opened 2018-08-21
check adb runnable before starting scrcpy

npes87184:
There are many user who encounters missing adb.
To stop things happens again, we check it and show
sexy response to user.

In the first version, I only check in linux. If you like this idea, I will continue to finish windows part.
Thanks.

Signed-off-by: yuchenlin <npes87184@gmail.com>

### DIFF
--- a/app/src/command.c
+++ b/app/src/command.c
@@ -1,5 +1,9 @@
 #include "command.h"
 
+#ifdef __WINDOWS__
+#else
+#include <errno.h>
+#endif
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
     return adb_command;
 }
 
+static void show_err_msg(int err) {
+#ifdef __WINDOWS__
+    LOGE("Failed to execute adb.\n");
+#else
+    switch (err) {
+    case 0:
+        break;
+    case -1:
+        LOGE("Failed to execute adb.\n");
+        break;
+    case ENOENT:
+        LOGE("Missing adb. You need adb, accessible from your PATH to execute scrcpy.\n");
+        break;
+    default:
+        LOGE("Failed to execute adb, errno: [%d].\n", err);
+        break;
+    }
+#endif
+}
+
 process_t adb_execute(const char *serial, const char *const adb_cmd[], int len) {
     const char *cmd[len + 4];
-    int i;
+    int i, r;
+    process_t process;
     cmd[0] = get_adb_command();
     if (serial) {
         cmd[1] = "-s";
@@ -32,7 +57,12 @@ process_t adb_execute(const char *serial, const char *const adb_cmd[], int len) 
 
     memcpy(&cmd[i], adb_cmd, len * sizeof(const char *));
     cmd[len + i] = NULL;
-    return cmd_execute(cmd[0], cmd);
+    r = cmd_execute(cmd[0], cmd, &process);
+    if (r != 0) {
+        show_err_msg(r);
+        return -1;
+    }
+    return process;
 }
 
 process_t adb_forward(const char *serial, uint16_t local_port, const char *device_socket_name) {
--- a/app/src/command.h
+++ b/app/src/command.h
@@ -32,7 +32,7 @@
 #endif
 # define NO_EXIT_CODE -1
 
-process_t cmd_execute(const char *path, const char *const argv[]);
+int cmd_execute(const char *path, const char *const argv[], process_t *process);
 SDL_bool cmd_terminate(process_t pid);
 SDL_bool cmd_simple_wait(process_t pid, exit_code_t *exit_code);
 
--- a/app/src/sys/unix/command.c
+++ b/app/src/sys/unix/command.c
@@ -1,5 +1,7 @@
 #include "command.h"
 
+#include <errno.h>
+#include <fcntl.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <sys/types.h>
@@ -7,18 +9,62 @@
 #include <unistd.h>
 #include "log.h"
 
-pid_t cmd_execute(const char *path, const char *const argv[]) {
-    pid_t pid = fork();
-    if (pid == -1) {
-        perror("fork");
+int cmd_execute(const char *path, const char *const argv[], pid_t *pid) {
+    int fd[2];
+    int ret = 0;
+
+    if (0 > pipe(fd)) {
+        perror("pipe");
         return -1;
     }
-    if (pid == 0) {
+
+    *pid = fork();
+    if (*pid == -1) {
+        perror("fork");
+        ret = -1;
+        goto END;
+    }
+
+    if (*pid > 0) {
+        /* parent close write side. */
+        close(fd[1]);
+        fd[1] = -1;
+        /* and blocking read until child exec or write some fail
+         * reason in fd. */
+        if (read(fd[0], &ret, sizeof(int)) == -1) {
+            perror("read");
+            ret = -1;
+            goto END;
+        }
+    } else if (*pid == 0) {
+        /* child close read side. */
+        close(fd[0]);
+        if (fcntl(fd[1], F_SETFD, FD_CLOEXEC) == -1) {
+            perror("fcntl");
+	    /* To prevent parent hang in read, we should exit when error. */
+            close(fd[1]);
+	    _exit(1);
+        }
         execvp(path, (char *const *)argv);
-        perror("exec");
+        /* if child execvp failed, before exit, we should write
+         * reason into the pipe. */
+        ret = errno;
+        if (write(fd[1], &ret, sizeof(int)) == -1) {
+            perror("write");
+        }
+        /* close write side before exiting. */
+        close(fd[1]);
         _exit(1);
     }
-    return pid;
+
+END:
+    if (fd[0] != -1) {
+        close(fd[0]);
+    }
+    if (fd[1] != -1) {
+        close(fd[1]);
+    }
+    return ret;
 }
 
 SDL_bool cmd_terminate(pid_t pid) {
--- a/app/src/sys/win/command.c
+++ b/app/src/sys/win/command.c
@@ -4,7 +4,7 @@
 #include "log.h"
 #include "str_util.h"
 
-HANDLE cmd_execute(const char *path, const char *const argv[]) {
+int cmd_execute(const char *path, const char *const argv[], HANDLE *handle) {
     STARTUPINFO si;
     PROCESS_INFORMATION pi;
     memset(&si, 0, sizeof(si));
@@ -18,7 +18,8 @@ HANDLE cmd_execute(const char *path, const char *const argv[]) {
     size_t ret = xstrjoin(cmd, argv, ' ', sizeof(cmd));
     if (ret >= sizeof(cmd)) {
         LOGE("Command too long (%" PRIsizet " chars)", sizeof(cmd) - 1);
-        return NULL;
+        *handle = NULL;
+        return -1;
     }
 
 #ifdef WINDOWS_NOCONSOLE
@@ -27,10 +28,12 @@ HANDLE cmd_execute(const char *path, const char *const argv[]) {
     int flags = 0;
 #endif
     if (!CreateProcess(NULL, cmd, NULL, NULL, FALSE, flags, NULL, NULL, &si, &pi)) {
-        return NULL;
+        *handle = NULL;
+        return -1;
     }
 
-    return pi.hProcess;
+    *handle = pi.hProcess;
+    return 0;
 }
 
 SDL_bool cmd_terminate(HANDLE handle) {


rom1v:
> There are many user who encounters missing adb.
> To stop things happens again, we check it and show sexy response to user.

I agree, this is a good idea to show a better error message than:

    ERROR: "adb push" returned with value 1

However, I think the detection should not rely on the existence of the binary (iterating over `PATH` directories): it's just an heuristic which will probably fail in some corner cases.

Instead, we can report this problem properly if `exec()` fails + `errno` (which is distinguishable from the command exited with a signal or a non-zero value).

In this case, [`cmd_execute()`](https://github.com/Genymobile/scrcpy/blob/v1.3/app/src/sys/unix/command.c#L10) would need the parent process to wait for the child process until it has `exec()` (but not until it has terminated, the server is long-lived). For example using `pipe2()` with flag `O_CLOEXEC` and make the child write something if exec fails (and auto-close otherwise), and the parent executing a blocking `read()`.

Then it would report this information back to the caller in some way (its signature should be changed).

npes87184:
Hi, @rom1v 

I agree that it's just an heuristic method. However, I think the execvp is done by this method and I found the implementation of [execvp](http://cvsweb.netbsd.org/bsdweb.cgi/src/lib/libc/gen/execvp.c?rev=1.30&content-type=text/x-cvsweb-markup&only_with_tag=MAIN) which proves my assertion. The original idea is to find how does exec do and extract the part we need. If you don't like this method, I will re-write the code base on your advice. Thank you.

Sincerely.

rom1v:
> However, I think the execvp is done by this method and I found the implementation of [execvp](http://cvsweb.netbsd.org/bsdweb.cgi/src/lib/libc/gen/execvp.c?rev=1.30&content-type=text/x-cvsweb-markup&only_with_tag=MAIN) which proves my assertion.

Then, this is a good reason to use the result of `exec()` instead of reimplementing it :)

npes87184:
Thank you for your advice. I will write a prototype these days.

npes87184:
Hi, @rom1v 

I have modified the commit base on your advice.
In linux, the `cmd_execute()` will return child exec state (0: success, -1: cmd_execute fail, otherwise: errno) now.
We can check child state in `adb_execute()` and show some sexy response to user.

The result will look like:
```
ERROR: Missing adb. You need adb, accessible from your PATH to execute scrcpy.
ERROR: Could not execute "adb push"
```

Thanks.

app/src/command.c
```diff
@@ -32,7 +54,12 @@ process_t adb_execute(const char *serial, const char *const adb_cmd[], int len)
 
     memcpy(&cmd[i], adb_cmd, len * sizeof(const char *));
     cmd[len + i] = NULL;
-    return cmd_execute(cmd[0], cmd);
+    r = cmd_execute(cmd[0], cmd, &process);
+    if (r != 0) {
+        show_err_msg(r);
```
rom1v: If the error is not reported to the caller of `adb_execute()` (I'm ok with this), the relevant error message can be printed directly in the specific implementation.

app/src/command.h
```diff
@@ -32,7 +32,7 @@
 #endif
 # define NO_EXIT_CODE -1
 
-process_t cmd_execute(const char *path, const char *const argv[]);
+int cmd_execute(const char *path, const char *const argv[], process_t *process);
```
rom1v: (and if the error message is printed directly in the specific implementation, there is no need to change this signature: `PROCESS_NONE` will report the error)

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,66 @@
+#define _GNU_SOURCE
```
rom1v: Oh, I didn't realize that `pipe2()` required this. This would make _scrcpy_ non portable (it would not compile on some platforms, possibly including Mac OS X (?)).

So we need `pipe()`+`fcntl()` instead (cf below).

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,66 @@
...
+         * reason in fd. */
+        if (0 > read(fd[0], &ret, sizeof(int))) {
+            perror("read");
+            ret = -1;
+            goto END;
+        }
+    } else if (*pid == 0) {
+        /* child close read side. */
+        close(fd[0]);
+        fd[0] = -1;
```
rom1v: not necessary, you exec or `_exit()` before the end of the block.

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,66 @@
...
+            perror("write");
+        }
+        /* close write side before exiting. */
+        close(fd[1]);
         _exit(1);
     }
-    return pid;
+
+END:
+    if (fd[0] != -1) {
```
rom1v: may never happen

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,66 @@
...
+        close(fd[0]);
+    }
+    if (fd[1] != -1) {
+        close(fd[1]);
```
rom1v: can be moved to the parent block

app/src/sys/unix/command.c
rom1v: To enable CLOEXEC without `pipe2()`, add:
```c
        if (fcntl(fd[1], F_SETFD, FD_CLOEXEC) == -1) {
            perror("fcntl");
        }
```

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,66 @@
...
+        goto END;
+    }
+
+    if (*pid > 0) {
+        /* parent close write side. */
+        close(fd[1]);
+        fd[1] = -1;
+        /* and blocking read until child exec or write some fail
+         * reason in fd. */
+        if (0 > read(fd[0], &ret, sizeof(int))) {
```
rom1v: (please reverse the operands)

app/src/sys/win/command.c
```diff
@@ -4,7 +4,7 @@
 #include "log.h"
 #include "str_util.h"
 
-HANDLE cmd_execute(const char *path, const char *const argv[]) {
+int cmd_execute(const char *path, const char *const argv[], HANDLE *handle) {
```
rom1v: (same here)

app/src/sys/unix/command.c
npes87184: There is a possible path will need it, that is if we create pipe successfully but fork failed.

app/src/command.c
npes87184: In my opinion, the `cmd_exectue` is a general function. That is, it can execute any cmd. When it errors, problem should be handled by caller (print error msg in this case).

app/src/sys/unix/command.c
npes87184: Will do.

app/src/sys/unix/command.c
npes87184: Close all fds before exiting is a good habit?

app/src/sys/unix/command.c
npes87184: Will do

app/src/sys/unix/command.c
npes87184: will do

app/src/command.c
rom1v: Oh, yes, of course. I confused `adb_execute()` and `cmd_execute()` (initially, I thought that even `adb_execute()` would report error to the caller). Whatever. It's better like that.

app/src/sys/unix/command.c
rom1v: `close(fd[0])`, yes.
`fd[0] = -1` is not necessary.

app/src/sys/unix/command.c
rom1v: Yes :+1:

app/src/sys/unix/command.c
npes87184: There is a possible path will need it, that is if we create pipe successfully but fork failed. I remain it in this position.

app/src/command.c
```diff
@@ -1,5 +1,9 @@
 #include "command.h"
 
+#ifdef __WINDOWS__
+#else
```
rom1v:
```
#ifndef __WINDOWS__
```

app/src/command.c
```diff
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
     return adb_command;
 }
 
+static void show_err_msg(int err) {
+#ifdef __WINDOWS__
+    LOGE("Failed to execute adb.\n");
+#else
+    switch (err) {
+    case 0:
```
rom1v: may not happen

app/src/command.c
```diff
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
     return adb_command;
 }
 
+static void show_err_msg(int err) {
+#ifdef __WINDOWS__
+    LOGE("Failed to execute adb.\n");
```
rom1v: Mostly windows users are lost with `adb`, isn't it possible to know when the problem is a missing adb?

app/src/command.c
```diff
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
...
+#else
+    switch (err) {
+    case 0:
+        break;
+    case -1:
+        LOGE("Failed to execute adb.\n");
+        break;
+    case ENOENT:
+        LOGE("Missing adb. You need adb, accessible from your PATH to execute scrcpy.\n");
```
rom1v: I suggest:
> 'adb' command not found (make it accessible from your PATH or define its absolute path in the ADB environment variable)

app/src/command.c
```diff
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
...
 
+static void show_err_msg(int err) {
+#ifdef __WINDOWS__
+    LOGE("Failed to execute adb.\n");
+#else
+    switch (err) {
+    case 0:
+        break;
+    case -1:
+        LOGE("Failed to execute adb.\n");
```
rom1v: (for consistency, don't end with a `.`, and `\n` is not needed)

app/src/command.c
```diff
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
...
+        break;
+    case ENOENT:
+        LOGE("Missing adb. You need adb, accessible from your PATH to execute scrcpy.\n");
+        break;
+    default:
+        LOGE("Failed to execute adb, errno: [%d].\n", err);
```
rom1v: (could print a user-friendly errno message with `strerror()`)

app/src/command.c
```diff
@@ -32,7 +57,12 @@ process_t adb_execute(const char *serial, const char *const adb_cmd[], int len)
 
     memcpy(&cmd[i], adb_cmd, len * sizeof(const char *));
     cmd[len + i] = NULL;
-    return cmd_execute(cmd[0], cmd);
+    r = cmd_execute(cmd[0], cmd, &process);
+    if (r != 0) {
+        show_err_msg(r);
+        return -1;
```
rom1v:
```
return PROCESS_NONE;
```
(otherwise Windows will not be happy :wink:)

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,70 @@
...
 
-pid_t cmd_execute(const char *path, const char *const argv[]) {
-    pid_t pid = fork();
-    if (pid == -1) {
-        perror("fork");
+int cmd_execute(const char *path, const char *const argv[], pid_t *pid) {
+    int fd[2];
+    int ret = 0;
+
+    if (0 > pipe(fd)) {
```
rom1v:
```
if (pipe(fd) == -1) {
```

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,70 @@
...
+            perror("read");
+            ret = -1;
+            goto END;
+        }
+    } else if (*pid == 0) {
+        /* child close read side. */
+        close(fd[0]);
+        if (fcntl(fd[1], F_SETFD, FD_CLOEXEC) == -1) {
+            perror("fcntl");
+	    /* To prevent parent hang in read, we should exit when error. */
```
rom1v: wrong indent (tab vs space)

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,70 @@
...
+        fd[1] = -1;
+        /* and blocking read until child exec or write some fail
+         * reason in fd. */
+        if (read(fd[0], &ret, sizeof(int)) == -1) {
+            perror("read");
+            ret = -1;
+            goto END;
+        }
+    } else if (*pid == 0) {
+        /* child close read side. */
```
rom1v: (for consistency, use `//` for comments, and no ending `.`)

app/src/command.c
```diff
@@ -18,9 +22,30 @@ static inline const char *get_adb_command() {
     return adb_command;
 }
 
+static void show_err_msg(int err) {
+#ifdef __WINDOWS__
+    LOGE("Failed to execute adb.\n");
+#else
+    switch (err) {
```
rom1v: (indent cases for consistency, it makes it more readable when cases have blocks)

app/src/sys/unix/command.c
```diff
@@ -1,24 +1,70 @@
...
+        close(fd[0]);
+        if (fcntl(fd[1], F_SETFD, FD_CLOEXEC) == -1) {
+            perror("fcntl");
+	    /* To prevent parent hang in read, we should exit when error. */
+            close(fd[1]);
```
rom1v: must send a failure to the parent (e.g. `errno` set by `fcntl`), otherwise the parent will consider that the child has execed successfully

npes87184:
Hi, @rom1v 

Thank you for your effort and time. All look good to me.

btw, I am not familiar with coding in Windows. I will take a look to survey how to detect it in Windows related system recently.

Sincerely,
yuchenlin

rom1v:
Merged in `dev`: https://github.com/Genymobile/scrcpy/commit/6d2d803003c231df9df3343eb73edf97d9ac3c76
I added another commit not to handle `errno` from the common `command.c`: https://github.com/Genymobile/scrcpy/commit/bf32f25170d32a5c2f53f0fc7043841cc6d24b2b

> I will take a look to survey how to detect it in Windows related system recently.
:+1:
If you have a solution, now you can just return `PROCESS_ERROR_MISSING_BINARY` :wink:

PR #272, opened 2018-09-22
labarary.c

Ashishsharma10298:
main labarary function


npes87184:
Hi, @Chhotadon1 

In my opinion, there is no reason to include `stdio.h` in this file.
Do you meet any building warning or error?

Thanks.

PR #292, opened 2018-10-09
Record screen to file

igorinov:
This change enables writing AV packets received from the target to a video file without re-encoding.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -11,6 +11,7 @@
 struct args {
     const char *serial;
     const char *crop;
+    const char *outfilename;
     SDL_bool fullscreen;
     SDL_bool help;
     SDL_bool version;
@@ -48,6 +49,9 @@ static void usage(const char *arg0) {
         "        other dimension is computed so that the device aspect-ratio\n"
         "        is preserved.\n"
         "        Default is %d%s.\n"
+        "\n"
+        "    -o, --output\n"
+        "        Write video output to file.\n"
         "\n"
         "    -p, --port port\n"
         "        Set the TCP port the client listens on.\n"
@@ -207,6 +211,7 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {"fullscreen",   no_argument,       NULL, 'f'},
         {"help",         no_argument,       NULL, 'h'},
         {"max-size",     required_argument, NULL, 'm'},
+        {"output",       required_argument, NULL, 'o'},
         {"port",         required_argument, NULL, 'p'},
         {"serial",       required_argument, NULL, 's'},
         {"show-touches", no_argument,       NULL, 't'},
@@ -214,7 +219,7 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {NULL,           0,                 NULL, 0  },
     };
     int c;
-    while ((c = getopt_long(argc, argv, "b:c:fhm:p:s:tv", long_options, NULL)) != -1) {
+    while ((c = getopt_long(argc, argv, "b:c:fhm:o:p:s:tv", long_options, NULL)) != -1) {
         switch (c) {
             case 'b':
                 if (!parse_bit_rate(optarg, &args->bit_rate)) {
@@ -234,6 +239,9 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
                 if (!parse_max_size(optarg, &args->max_size)) {
                     return SDL_FALSE;
                 }
+                break;
+            case 'o':
+                args->outfilename = optarg;
                 break;
             case 'p':
                 if (!parse_port(optarg, &args->port)) {
@@ -310,6 +318,7 @@ int main(int argc, char *argv[]) {
         .serial = args.serial,
         .crop = args.crop,
         .port = args.port,
+        .outfilename = args.outfilename,
         .max_size = args.max_size,
         .bit_rate = args.bit_rate,
         .show_touches = args.show_touches,
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -193,11 +193,11 @@ SDL_bool scrcpy(const struct scrcpy_options *options) {
         goto finally_destroy_frames;
     }
 
-    decoder_init(&decoder, &frames, device_socket);
+    decoder_init(&decoder, &frames, device_socket, frame_size);
 
     // now we consumed the header values, the socket receives the video stream
     // start the decoder
-    if (!decoder_start(&decoder)) {
+    if (!decoder_start(&decoder, options->outfilename)) {
         ret = SDL_FALSE;
         server_stop(&server);
         goto finally_destroy_file_handler;
@@ -228,7 +228,7 @@ SDL_bool scrcpy(const struct scrcpy_options *options) {
     }
 
     ret = event_loop();
-    LOGD("quit...");
+    LOGI("quit...");
 
     screen_destroy(&screen);
 
--- a/app/src/decoder.c
+++ b/app/src/decoder.c
@@ -1,6 +1,7 @@
 #include "decoder.h"
 
 #include <libavformat/avformat.h>
+#include <libavutil/time.h>
 #include <SDL2/SDL_events.h>
 #include <SDL2/SDL_mutex.h>
 #include <SDL2/SDL_thread.h>
@@ -13,6 +14,8 @@
 #include "log.h"
 
 #define BUFSIZE 0x10000
+
+static AVRational us = {1, 1000000};
 
 static int read_packet(void *opaque, uint8_t *buf, int buf_size) {
     struct decoder *decoder = opaque;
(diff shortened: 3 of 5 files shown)


app/src/main.c
```diff
@@ -207,14 +211,15 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {"fullscreen",   no_argument,       NULL, 'f'},
         {"help",         no_argument,       NULL, 'h'},
         {"max-size",     required_argument, NULL, 'm'},
+        {"output",       required_argument, NULL, 'o'},
```
npes87184: In my opinion, the `output` is too general. In the future, we may add some feature like output format. It make a bit confusing.

app/src/main.c
```diff
@@ -235,6 +240,9 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
                     return SDL_FALSE;
                 }
                 break;
+            case 'o':
+                args->outfilename = optarg;
```
npes87184: To make consistent, use underscore as delimiter.

app/src/scrcpy.c
```diff
@@ -228,7 +228,7 @@ SDL_bool scrcpy(const struct scrcpy_options *options) {
     }
 
     ret = event_loop();
-    LOGD("quit...");
+    LOGI("quit...");
```
npes87184: In my opinion, `LOGD` is good enough.

app/src/main.c
igorinov: --output-file?

app/src/main.c
igorinov: OK

igorinov:
Thank you! I am adding muxing to the server, but it only supports Android 8.0+ because MediaMuxer(FileDescriptor fd, int format) is not available in earlier versions. While MPV correctly handles frame size changes in the same video file, creating a new file on rotation may be a better option.

rom1v:
(just acking your message, I'll try to review it next week, thank you)

rom1v:
@igorinov I just tested your `server-timestamp` branch. I think it's a good idea to retrieve the PTS like you did if recording is requested.

It records a file, but unfortunately, the timestamps are still not respected in the resulting file (like https://github.com/Genymobile/scrcpy/pull/292#pullrequestreview-162747160). IMO we need to mux in a container like mp4 or mkv.

rom1v:
> It records a file, but unfortunately, the timestamps are still not respected in the resulting file

Ah, that's because I saved to `test.h264`. If I save to `test.mp4`, timestamps are respected :tada:
However, the mp4 takes a lot of time to open in VLC. And saving to `test.mkv` does not work.

(Is it possible to specify the target container explicitly?)

igorinov:
I tried recording to .mkv on a Linux machine, and the file was played successfully with cvlc.

rom1v:
> I tried recording to .mkv on a Linux machine, and the file was played successfully with cvlc.

Arf, on Debian sid, I cannot record to mkv :(

I like your branch, and it would be great to get recording into scrcpy. But there are still some issues to solve.

If I record to mp4, the resulting mp4 file takes a very long time to load in vlc, so it is not "proper" (while a file recorded by `adb screenrecord` works fine). I don't know what the cause is. Any idea?

If I take your current branch, (after I changed `CODEC_FLAG_GLOBAL_HEADER` to `AV_CODEC_FLAG_GLOBAL_HEADER`), I get these warnings:

```
../app/src/decoder.c: In function ‘run_decoder’:
../app/src/decoder.c:145:13: warning: ‘codec’ is deprecated [-Wdeprecated-declarations]
             outstream->codec = avcodec_alloc_context3(codec);
             ^~~~~~~~~
In file included from ../app/src/decoder.c:3:
/usr/include/x86_64-linux-gnu/libavformat/avformat.h:877:21: note: declared here
     AVCodecContext *codec;
                     ^~~~~
../app/src/decoder.c:146:13: warning: ‘codec’ is deprecated [-Wdeprecated-declarations]
             outstream->codec->pix_fmt = AV_PIX_FMT_YUV420P;
             ^~~~~~~~~
In file included from ../app/src/decoder.c:3:
/usr/include/x86_64-linux-gnu/libavformat/avformat.h:877:21: note: declared here
     AVCodecContext *codec;
                     ^~~~~
../app/src/decoder.c:147:13: warning: ‘codec’ is deprecated [-Wdeprecated-declarations]
             outstream->codec->width = decoder->frame_size.width;
             ^~~~~~~~~
In file included from ../app/src/decoder.c:3:
/usr/include/x86_64-linux-gnu/libavformat/avformat.h:877:21: note: declared here
     AVCodecContext *codec;
                     ^~~~~
../app/src/decoder.c:148:13: warning: ‘codec’ is deprecated [-Wdeprecated-declarations]
             outstream->codec->height = decoder->frame_size.height;
             ^~~~~~~~~
In file included from ../app/src/decoder.c:3:
/usr/include/x86_64-linux-gnu/libavformat/avformat.h:877:21: note: declared here
     AVCodecContext *codec;
                     ^~~~~
../app/src/decoder.c:150:13: warning: ‘codec’ is deprecated [-Wdeprecated-declarations]
             outstream->codec->flags |= AV_CODEC_FLAG_GLOBAL_HEADER;
             ^~~~~~~~~
In file included from ../app/src/decoder.c:3:
/usr/include/x86_64-linux-gnu/libavformat/avformat.h:877:21: note: declared here
     AVCodecContext *codec;
                     ^~~~~
```

I have no time to investigate on this currently. I'll try to find some time later. If you have some time and find how to fix these issues, that would be awesome :wink:

Thank you.

rom1v:
> If I record to mp4, the resulting mp4 file takes a very long time to load in vlc

I just found the problem: the PTS given by `bufferInfo.presentationTimeUs` on the server is the current timestamp (since epoch). To fix this issue, the timestamp of the first frame must be 0, and all others must be relative to this one (in other words, we must substract the first `presentationTimeUs` from all PTS).

There are still little problems, for example on quit:

>    Application provided invalid, non monotonically increasing dts to muxer in stream 0: 3347160 >= 3347160

rom1v:
> I tried recording to .mkv on a Linux machine, and the file was played successfully with cvlc.

For me, it fails at `avformat_write_header(output_ctx, NULL)` which returns `AVERROR_INVALIDDATA`.

rom1v:
@igorinov I had some time today to work on it.

I started from your commit, and did these changes:
 - I created a new "class" `recorder`, so that the recording is not implemented directly by the decoder.
 - I use the non-deprecated API (`AVCodecParameters`) to pass codec params to the output `AVStream`.
 - I explicitly select the MP4 muxer (I don't let FFmpeg choose based on the filename, some muxers won't record the video properly, and for some reason I can't write header for matroska muxer, cf https://github.com/Genymobile/scrcpy/pull/292#issuecomment-435668428).
 - I use 1/1000000 as time base (since the input provide timestamps in microseconds).
 - I fixed PTS (as explained in https://github.com/Genymobile/scrcpy/pull/292#issuecomment-435663718).
 - I set both PTS/DTS to `AVPacket` before writing it (to avoid FFmpeg warnings).
 - I don't write "config" packets (when bufferInfo.flags has `MEDIA_CODEC_FLAG_CONFIG`) to output file since they contain no data and do not have any PTS/DTS.

Here is my branch: [`record`](https://github.com/Genymobile/scrcpy/commits/record).

It correctly records rotation, and plays correctly in VLC :smile: 

Remaining things I will work on:
 - have absolutely no impact on perfs if recording is not enabled (e.g. provide the original `read_packet()` function if we do not record).
 - also use `AVStream` deprecated API (the one you used) in ifdefs to support older FFmpeg versions.
 - also use deprecated `av_oformat_next()` in ifdefs (in addition to the new `av_muxer_iterate()`) to support older FFmpeg versions.
 - fix the warning on exit: `Application provided invalid, non monotonically increasing dts to muxer in stream 0: 17164020 >= 17164020`.

npes87184:
Hi, @rom1v 

Thank you for your explanation.
According to your explanation, to fix building error, I simply modify the `find_mp4_muxer` by this commit [recorder: use av_oformat_next to support older FFmpeg](https://github.com/npes87184/scrcpy/commit/8587e3c8367229585edf2ac7940e7f9e31f6b5f5) and it works.

At the first glance, it works like a charm!
However, when scrcpy receive `SIGTERM` the mp4 file cannot be played.

Thanks,
yuchenlin

rom1v:
> However, when scrcpy receive SIGTERM the mp4 file cannot be played.

Ah yes, this need to be fixed too. Thank you.

Basically, reverting f00c6c5b1321860eb6d590928448778ec8f43278 works, but we need to fix what this commit prevented.

npes87184:
Or set a flag to call `recorder_close` when we receive `SIGTERM`?

rom1v:
I think we have a version ready to be merged.

I also changed the option name:

    scrcpy --record file.mp4
    scrcpy -r file.mp4

If you want to test before the merge/release, feedbacks are welcome :)

rom1v:
I merged my `record` branch into `master`, and published v1.5 with screen recording :tada:

Thank you very much @igorinov for your initial work. Your commit is merged as d706c5d.

PR #296, opened 2018-10-14
Desktop entry file for Linux app launchers (resolves #295)

### LINKED ISSUE #295: Create desktop entry file for Linux app launchers
It'd be cool to launch scrcpy from a desktop environment's application menu. All that's needed is a `scrcpy.desktop` file that installs to `/usr/share/applications`.

See [XDG Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)

asnelling:
Adds the ability to launch scrcpy using your preferred application launcher on Linux.

See: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -143,6 +143,12 @@ endif
 
 executable('scrcpy', src, dependencies: dependencies, include_directories: src_dir, install: true, c_args: c_args, link_args: link_args)
 
+# Desktop entry file for application launchers
+if host_machine.system() == 'linux'
+    # -> /usr/share/applications/scrcpy.desktop
+    install_data('scrcpy.desktop', install_dir : join_paths(get_option('datadir'), 'applications'))
+endif
+
 
 ### TESTS
 
--- a/app/scrcpy.desktop
+++ b/app/scrcpy.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Name=scrcpy
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+Exec=scrcpy
+Icon=gparted
+Terminal=false
+Type=Application
+Categories=Utility;RemoteAccess;
+StartupNotify=false


app/meson.build
```diff
@@ -143,6 +143,12 @@ endif
 
 executable('scrcpy', src, dependencies: dependencies, include_directories: src_dir, install: true, c_args: c_args, link_args: link_args)
 
+# Desktop entry file for application launchers
+if host_machine.system() == 'linux'
+    # -> /usr/share/applications/scrcpy.desktop
```
rom1v: (by default `/usr/local/share/applications/scrcpy.desktop`, but you can keep this comment as is, it's explicit enough)

app/scrcpy.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=scrcpy
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+Exec=scrcpy
+Icon=gparted
```
rom1v: scrcpy

_gparted_ icon does not really fit _scrcpy_. Unfortunately, I didn't find any existing icon that could fit. Maybe we should add an Android icon?

app/scrcpy.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=scrcpy
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+Exec=scrcpy
```
rom1v: At first, it did not work for me.

To get the error, I changed to:
```
Exec=bash -c "scrcpy &> /tmp/log"
```
And `/tmp/log` contained:
```
exec: No such file or directory
ERROR: 'adb' command not found (make it accessible from your PATH or define its full path in the ADB environment variable)
ERROR: Could not execute "adb push"
```
In fact, the `adb` directory is added in `PATH` in my `.bashrc`, so it's not in `PATH` when the session is started, so starting it for the menu does not work in that case.

In practice, do you use `android-tools-adb`, or did you set your `adb` in `PATH` from somewhere else?

npes87184: I add the file in `/usr/share/applications/scrcpy.desktop` with content:
```
[Desktop Entry]
Name=scrcpy
GenericName=Android Remote Control
Comment=Display and control your Android device
Exec=scrcpy
Icon=gparted
Terminal=false
Type=Application
Categories=Utility;RemoteAccess;
StartupNotify=false
```
After it, click the icon from menu. Scrcpy start as expected.

npes87184: @rom1v 

Maybe you can try to change Exec to `echo $PATH > /tmp/path_out` and execute again to get more info for debugging.

app/scrcpy.desktop
srevinsaju: You can use this icon guiscrcpy, which I created for guiscrcpy. I licensed it under Creative Commons Attributions 4.0, so you can easily reuse it.

srevinsaju:
If you add this code as `scrcpy.VisualElements.xml`, you can get that customized Windows Start Menu Effect, as you see for OneNote Tiles etc.. 

This is the one I used for the Windows version of guiscrcpy
`<?xml version="1.0" encoding="utf-8"?>
<Application xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" >
  <VisualElements ShowNameOnSquare150x150Logo="off" Square150x150Logo="VisualElements\MediumIconguiscrcpy.png" Square70x70Logo="VisualElements\SmallIconguiscrcpy.png" ForegroundText="light" BackgroundColor="#00FED0" />
`

Hope it helps..

unclehowell:
In my experience its better to have the .Desktop run a script which resets network adaptors and establishes connection with the android first. In the script the correct screen size and connection speed can be set and appropriate sleep times between commands

app/scrcpy.desktop
yan12125: The Android icon at https://github.com/Genymobile/scrcpy/blob/v1.18/app/src/icon.xpm may be a better choice. Here is an improved patch:
```Diff
diff --git a/app/meson.build b/app/meson.build
index 0663c64..321ed38 100644
--- a/app/meson.build
+++ b/app/meson.build
@@ -148,6 +148,14 @@ executable('scrcpy', src,
 
 install_man('scrcpy.1')
 
+# Desktop entry file for application launchers
+if host_machine.system() == 'linux'
+    # -> /usr/share/applications/scrcpy.desktop
+    install_data('scrcpy.desktop', install_dir : join_paths(get_option('datadir'), 'applications'))
+    install_data('src/icon.xpm', rename: 'scrcpy.xpm',
+                 install_dir : join_paths(get_option('datadir'), 'icons', 'hicolor', '48x48', 'apps'))
+endif
+
 
 ### TESTS
 
diff --git a/app/scrcpy.desktop b/app/scrcpy.desktop
new file mode 100644
index 0000000..5933272
--- /dev/null
+++ b/app/scrcpy.desktop
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=scrcpy
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+Exec=scrcpy
+Icon=scrcpy
+Terminal=false
+Type=Application
+Categories=Utility;RemoteAccess;
+StartupNotify=false
```

app/scrcpy.desktop
yan12125: I guess `bash -i scrcpy` will work for `adb` not in standard paths? That bash option forces the interactive mode and `.bashrc` is sourced. I use [android-tools](https://archlinux.org/packages/community/x86_64/android-tools/) on Arch Linux, where adb is installed under /usr/bin, so I cannot test my idea.

yan12125:
> In my experience its better to have the .Desktop run a script which resets network adaptors and establishes connection with the android first. In the script the correct screen size and connection speed can be set and appropriate sleep times between commands

Hmm, scrcpy runs adb, which connects to the Android device when needed. Is there a reason to reset network adaptors? Anyway, if some preparation steps are needed, they are better to be part of the scrcpy command instead of the .desktop file, so that those who run scrcpy from the terminal can benefit from it as well.

asnelling:
Closed with #3351

PR #310, opened 2018-10-27
fix text memory leak

npes87184:
In this time, the text will be freed only when queue is destroyed.
However, the queue is rotated very quickly. The text event may be
rotated out. It causes memory leak on text. Simply free text when
we used.

In additional, fix a potential memory leak when controller_push_event
failed.

Signed-off-by: yuchenlin <npes87184@gmail.com>

### DIFF
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -144,6 +144,7 @@ void input_manager_process_text_input(struct input_manager *input_manager,
         return;
     }
     if (!controller_push_event(input_manager->controller, &control_event)) {
+        SDL_free(control_event.text_event.text);
         LOGW("Cannot send text event");
     }
 }


rom1v:
> the text will be freed only when queue is destroyed.

The text is owned by the event, and it is destroyed once the event is processed:
https://github.com/Genymobile/scrcpy/blob/v1.4/app/src/controller.c#L75

> However, the queue is rotated very quickly. The text event may be rotated out.

Only if it has been "taken" (`control_event_queue_take`), because `control_event_queue_push` fails if the queue is full: https://github.com/Genymobile/scrcpy/blob/v1.4/app/src/control_event.c#L96

> It causes memory leak on text.

So maybe I missed something, but I think there is no leak in that case. Address sanitizer (`meson configure x b_sanitize=address`) does not report memory leaks here.

> Simply free text when we used.

This would be a hack, it would break proper ownership rules (we don't expect [`control_event_serialize()`](https://github.com/Genymobile/scrcpy/blob/v1.4/app/src/control_event.c#L28), which takes a const pointer to the event, to free its internal resources).

> In additional, fix a potential memory leak when controller_push_event failed.

Thank you! It was correctly handled for [`clipboard_paste()`](https://github.com/Genymobile/scrcpy/blob/v1.4/app/src/input_manager.c#L126), but not for `input_manager_process_text_input()`.

If you agree with my analysis, you can edit your commit+commit message to fix only the memory leak when `controller_push_event()` fails, then I'll merge it.

npes87184:
Hi, @rom1v 

Thank you for your explanation. You are  right!
I have edited my commit to fix potential memory leak of text.

Cheers,
yuchenlin

PR #401, opened 2019-01-18
input_manager: don't ignore double click event when clicking inside device

npes87184:
Fix the regression from [fefb9816a99e](https://github.com/Genymobile/scrcpy/commit/fefb9816a99e6f8fa59befcfb70ba87112f90a8d).

Thanks


rom1v:
Good catch, thanks!

PR #412, opened 2019-01-27
app: add always_on_top

### LINKED ISSUE #391: Request: add command line parameter to let the window stay on top
I see that's it's hard to do it at runtime (written here: https://github.com/Genymobile/scrcpy/issues/60 ) , so maybe it's easier to do it using a command line parameter instead?

Please consider this.

npes87184:
Fix #391 

It is very convenient when I play mobile game and watch video at the
same time.

Tested on Linux mint Cinnamon as well as Windows 10.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -19,6 +19,7 @@ struct args {
     Uint16 port;
     Uint16 max_size;
     Uint32 bit_rate;
+    SDL_bool always_on_top;
 };
 
 static void usage(const char *arg0) {
@@ -64,6 +65,9 @@ static void usage(const char *arg0) {
         "    -t, --show-touches\n"
         "        Enable \"show touches\" on start, disable on quit.\n"
         "        It only shows physical touches (not clicks from scrcpy).\n"
+        "\n"
+        "    -T, --always-on-top\n"
+        "        Make scrcpy window always on top (above other windows).\n"
         "\n"
         "    -v, --version\n"
         "        Print the version of scrcpy.\n"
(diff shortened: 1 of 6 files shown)


app/src/screen.c
```diff
@@ -144,14 +144,21 @@ static inline SDL_Texture *create_texture(SDL_Renderer *renderer, struct size fr
                              frame_size.width, frame_size.height);
 }
 
-SDL_bool screen_init_rendering(struct screen *screen, const char *device_name, struct size frame_size) {
+SDL_bool screen_init_rendering(struct screen *screen,
+                               const char *device_name,
+                               struct size frame_size,
+                               const SDL_bool always_on_top) {
```
rom1v: no need for a top-level `const`

app/src/screen.h
```diff
@@ -43,7 +43,8 @@ void screen_init(struct screen *screen);
 // initialize screen, create window, renderer and texture (window is hidden)
 SDL_bool screen_init_rendering(struct screen *screen,
                                const char *device_name,
-                               struct size frame_size);
+                               struct size frame_size,
+                               const SDL_bool always_on_top);
```
rom1v: same here (no need for `const`)

app/src/main.c
```diff
@@ -206,6 +207,7 @@ static SDL_bool parse_port(char *optarg, Uint16 *port) {
 
 static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
     static const struct option long_options[] = {
+        {"always-on-top",no_argument,       NULL, 'a'},
```
rom1v: `a` is used on the `audio` branch for `--forward-audio`, this is not very important, but what do you think about using `-T`(like _top_, `t` is already taken for `--show-touches`) or `A`?

(you could also add a space after the comma and realign the others)

npes87184:
Hi,

I have pushed v2, please review. Thank you.

v1 -> v2:
* remove `const`
* replace `a` with `T`
* realign `struct option`
* add help and `README.md`

npes87184:
Thank you for your advice.

[v2 -> v3](https://github.com/Genymobile/scrcpy/compare/622922a75d1ff7b40c54827fb823a55631f7f2f0..eca82e09c3d3127cab247bfd94795da887a547cc):

* change the description

rom1v:
Perfect :+1:

PR #418, opened 2019-02-01
Added a new option : -n/--no-window

TheCapsLock:
This option allows scrcpy to be run withou showing preview window : we can now use
scrcpy only for screen recording as an example

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -13,6 +13,7 @@ struct args {
     const char *crop;
     const char *record_filename;
     SDL_bool fullscreen;
+    SDL_bool no_window;
     SDL_bool help;
     SDL_bool version;
     SDL_bool show_touches;
@@ -49,6 +50,11 @@ static void usage(const char *arg0) {
         "        other dimension is computed so that the device aspect-ratio\n"
         "        is preserved.\n"
         "        Default is %d%s.\n"
+        "\n"
+        "    -n, --no-window\n"
+        "        Do not show window. This is useful, as an example\n"
+        "        combined with -r/--record option to record screen without displaying\n"
+        "        video feedback\n"
         "\n"
         "    -p, --port port\n"
         "        Set the TCP port the client listens on.\n"
@@ -209,6 +215,7 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {"bit-rate",     required_argument, NULL, 'b'},
         {"crop",         required_argument, NULL, 'c'},
         {"fullscreen",   no_argument,       NULL, 'f'},
+        {"no-window",   no_argument,       NULL, 'n'},
         {"help",         no_argument,       NULL, 'h'},
         {"max-size",     required_argument, NULL, 'm'},
         {"port",         required_argument, NULL, 'p'},
@@ -219,7 +226,7 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {NULL,           0,                 NULL, 0  },
     };
     int c;
-    while ((c = getopt_long(argc, argv, "b:c:fhm:p:r:s:tv", long_options, NULL)) != -1) {
+    while ((c = getopt_long(argc, argv, "b:c:fnhm:p:r:s:tv", long_options, NULL)) != -1) {
         switch (c) {
             case 'b':
                 if (!parse_bit_rate(optarg, &args->bit_rate)) {
@@ -231,6 +238,9 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
                 break;
             case 'f':
                 args->fullscreen = SDL_TRUE;
+                break;
+            case 'n':
+                args->no_window = SDL_TRUE;
                 break;
             case 'h':
                 args->help = SDL_TRUE;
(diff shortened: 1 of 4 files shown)


app/src/scrcpy.c
```diff
@@ -80,7 +82,9 @@ static SDL_bool event_loop(void) {
                 if (!screen.has_frame) {
                     screen.has_frame = SDL_TRUE;
                     // this is the very first frame, show the window
-                    screen_show_window(&screen);
+                    if (!no_window) {
+                        screen_show_window(&screen);
+                    }
```
rom1v: I think you can skip code far earlier, you may not even need to initialize SDL nor decode frames at all if you have no window.

app/src/main.c
```diff
@@ -209,6 +215,7 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {"bit-rate",     required_argument, NULL, 'b'},
         {"crop",         required_argument, NULL, 'c'},
         {"fullscreen",   no_argument,       NULL, 'f'},
+        {"no-window",   no_argument,       NULL, 'n'},
```
rom1v: alignment

app/src/main.c
```diff
@@ -219,7 +226,7 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
         {NULL,           0,                 NULL, 0  },
     };
     int c;
-    while ((c = getopt_long(argc, argv, "b:c:fhm:p:r:s:tv", long_options, NULL)) != -1) {
+    while ((c = getopt_long(argc, argv, "b:c:fnhm:p:r:s:tv", long_options, NULL)) != -1) {
```
rom1v: (please keep the alphabetical order, ditto above and below)

npes87184:
Hi,

I have two simple recommendations for this patch:
* Please provide either a log in console or a toast in android side, so that we can know that the scrcpy finishes its initialization.
* Use -n alone doesn’t make sense now. IMO, we can return error if -n and -r don’t exist simultaneously.

Thanks,
Yu-Chen

app/src/main.c
```diff
@@ -232,6 +239,9 @@ static SDL_bool parse_args(struct args *args, int argc, char *argv[]) {
             case 'f':
                 args->fullscreen = SDL_TRUE;
                 break;
+            case 'n':
+                args->no_window = SDL_TRUE;
```
npes87184: Missing default value or an initialization. See [main.c](
https://github.com/Genymobile/scrcpy/blob/8d4ff03b7feab7123dd5483f16ff282fddbb425d/app/src/main.c#L291)

TheCapsLock: done

app/src/main.c
TheCapsLock: done

app/src/main.c
TheCapsLock: done; sorry for that

TheCapsLock:
>     * Use -n alone doesn’t make sense now. IMO, we can return error if -n and -r don’t exist simultaneously.

Added an error message if nor -n is provided without -r

TheCapsLock:
> Thank you for your contribution.
> 
> I like the idea to add such a flag. Maybe it could skip a lot more code (do not decode frame, do not initialize SDL…
> 
> I'll test it probably after the FOSDEM.

Thanks a lot for that ; I tried some combinations but did not get something better yet ; if you have clear idea in mind may I let you look at this specific topic ?

TheCapsLock:
>     * Please provide either a log in console or a toast in android side, so that we can know that the scrcpy finishes its initialization.

I've added a log message, thank you for this clever recommendation

rom1v:
> if you have clear idea in mind may I let you look at this specific topic ?

Yes, it requires some refactor beforehand: the decoder must be separated from the "video stream receiver", so that we can skip the decoder completely (we don't need to decode the video at all if there is no window). I will try to work on it so that it is included in the next version.

TheCapsLock:
> > if you have clear idea in mind may I let you look at this specific topic ?
> 
> Yes, it requires some refactor beforehand: the decoder must be separated from the "video stream receiver", so that we can skip the decoder completely (we don't need to decode the video at all if there is no window). I will try to work on it so that it is included in the next version.

Great to here this kind of news :) thanks a lot!

rom1v:
I refactored to split the decoder into stream+decoder: https://github.com/Genymobile/scrcpy/commit/ce5cdd877dfb9d195d0afd5be1a052b982a12a70.
Then I rebased your commit (but removed the part which disabled the window after decoding): https://github.com/Genymobile/scrcpy/commit/dbf51f15ac8ef0bae73179482ca28827aef5923b.
Then I disabled everything useless when `--no-window` is passed: https://github.com/Genymobile/scrcpy/commit/8c95835de9b05c0cee91f79016a84614ff82e226.

It's on `dev` branch for now.

I'm closing as _merged_ (even if github does see it as merged).

rom1v:
I have renamed the option `-N`/`--no-display`.

The description of scrcpy is "Display and control your Android device".

We want an option to disable display, another one to disable control, so I also implemented `-n`/`--no-control`.

TheCapsLock:
Tested on multiple devices, seems to work great, thanks :heart:

Gapplay:
@rom1v is possible to show fps when No-display is enabled?

rom1v:
Currently, no. 

A command line option (`--print-fps`) could be added to enable it on start.

rom1v:
@Gapplay Please open a new issue for it, to not forget it.

PR #426, opened 2019-02-06
server/scripts/build-wrapper.sh should not presume bash path

TheCapsLock:


rom1v:
I think I prefer to refer explicitly to /bin/bash (in case some `bash` binary is elsewhere accessible from `PATH`).

Where is your `bash` in practice?

TheCapsLock:
In my case ; it happens scrcpy is run under [NixOS](http://nixos.org/). it is an operating system where binaries path are changed after each derivation (roughly upgrade).

I know other folks like BSD users which does not have bash on /bin/bash. It seems to me that using ```/usr/bin/env``` is the most interoperable way to go when writing bash scripts

rom1v:
Do someone have a strong opinion on this?

Personally, I prefer `/bin/bash` (more standard and less "risky").

rom1v:
Merged on `dev` as cf7bf3148caca8d4428cbd70578696a8230689e0

Issue #240, opened 2018-08-23
No display when connecting a Meizu phone


FredWe:
same issue as comment in #151 

Scrcpy: v1.3
Device: Meizu Pro 6 plus
Android version: 7.0.0
ROM: Flyme 7.8.7.24 beta
Host: Ubuntu 18.04

screenrecording by adb works well as following:
```
adb shell screenrecord /sdcard/test.mp4
adb pull /sdcard/test.mp4
```

But issue appeared when executing
```
./run x        
x/server/scrcpy-server.jar: 1 file pushed. 3.0 MB/s (18570 bytes in 0.006s)
DEBUG: Starting decoder thread
DEBUG: Starting controller thread
INFO: OpenGL shaders: ENABLED
INFO: Created renderer: opengl
INFO: Initial texture: 1440x2560
```

and hold on here... no response, no display window (as i expected)

and i got some logcat related to maybe a ROM bug:
```
08-23 14:01:00.196  2650  3656 E OMXNodeInstance: getConfig(a5a00bd:Exynos.AVC.Encoder, ConfigAndroidIntraRefresh(0x6f60000a)) ERROR: UnsupportedIndex(0x8000101a)
08-23 14:01:00.200 18461 18461 E Surface : Could not create instance of Surface from IGraphicBufferProducer.
08-23 14:01:00.201 18461 18461 E Surface : java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
08-23 14:01:00.201 18461 18461 E Surface : 	at android.os.Handler.<init>(Handler.java:200)
08-23 14:01:00.201 18461 18461 E Surface : 	at android.os.Handler.<init>(Handler.java:114)
08-23 14:01:00.201 18461 18461 E Surface : 	at com.meizu.nightmode.MzNightModeManager.<init>(MzNightModeManager.java:148)
08-23 14:01:00.201 18461 18461 E Surface : 	at com.meizu.nightmode.MzNightModeManager.getDefault(MzNightModeManager.java:262)
08-23 14:01:00.201 18461 18461 E Surface : 	at android.graphics.Canvas.<init>(Canvas.java:2606)
08-23 14:01:00.201 18461 18461 E Surface : 	at android.view.Surface$CompatibleCanvas.<init>(Surface.java:595)
08-23 14:01:00.201 18461 18461 E Surface : 	at android.view.Surface.<init>(Surface.java:93)
[...]

Issue #151, opened 2018-05-18
scrcpy won't show android device


gmakacap:
hi sir,
Whenever i double click the scrcpy.exe it wont show anything i tried several times by just clicking on it but nothing happens.
![screenshot 1](https://user-images.githubusercontent.com/34020024/40248724-467c2f98-5aea-11e8-94ea-d848ee77291b.png)

I have installed adb drivers and my device is shown as authorized when typing **adb devices**.
I don;t know which error is stopping it to load....please help

rom1v:
See [here](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md#on-windows-i-have-no-output-in-the-console) to retrieve the output/error. This will help to know the issue.

rom1v:
Any news?

rom1v:
Please try v1.2.

olbb:
Same problem here with v1.2
After
>scrcpy-server.jar: 1 file pushed. 1.4 MB/s (19342 bytes in 0.013s)
>INFO: Initial texture: 1440x2560
didn't show anything.

rom1v:
Are you connected over USB or TCP?

What if you decrease the définition:

    scrcpy -m 1024

?

rom1v:
If it still does not work, are there any relevant logs in `adb logcat`?

olbb:
I'm use USB,
I try it with '-m 1024',but it does not work.
I fount something may be useful in log.

>06-01 02:55:27.916 18278 18278 E Surface : Could not create instance of Surface from IGraphicBufferProducer.
06-01 02:55:27.918 18278 18278 E Surface : java.lang.NullPointerException: Attempt to read from field 'boolean com.meizu.nightmode.MzNightModeManager.mHaveInit' on a null object reference
06-01 02:55:27.918 18278 18278 E Surface : 	at com.meizu.nightmode.MzNightModeManager.getDefault(MzNightModeManager.java:240)
06-01 02:55:27.918 18278 18278 E Surface : 	at android.graphics.Canvas.<init>(Canvas.java:2606)
06-01 02:55:27.918 18278 18278 E Surface : 	at android.view.Surface$CompatibleCanvas.<init>(Surface.java:595)
06-01 02:55:27.918 18278 18278 E Surface : 	at android.view.Surface.<init>(Surface.java:93)
06-01 02:55:27.918 18278 18278 E Surface : 	at android.media.MediaCodec.createInputSurface(Native Method)
[...]

PR #434, opened 2019-02-19
fix:Add Support For Meizu Phone

olbb:
Fix the error case by flyme nightMode.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
+++ b/server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
@@ -8,6 +8,7 @@ import android.media.MediaCodec;
 import android.media.MediaCodecInfo;
 import android.media.MediaFormat;
 import android.os.IBinder;
+import android.os.Looper;
 import android.view.Surface;
 
 import java.io.FileDescriptor;
@@ -66,6 +67,7 @@ public class ScreenEncoder implements Device.RotationListener {
                 Rect videoRect = device.getScreenInfo().getVideoSize().toRect();
                 setSize(format, videoRect.width(), videoRect.height());
                 configure(codec, format);
+                Looper.prepare();
                 Surface surface = codec.createInputSurface();
                 setDisplaySurface(display, surface, contentRect, videoRect);
                 codec.start();


olbb:
https://github.com/Genymobile/scrcpy/issues/151#issuecomment-393641540

olbb:
https://github.com/olbb/scrcpy/issues/1#issue-411715425

rom1v:
Hi,

Thank you for your report and fix.

I'm not sure this is sufficient, though:
`java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()`

If it creates an handler, it will post events that will never be consumed (the "main" thread is not an event loop).

What do you think?

olbb:
` java

/** Initialize the current thread as a looper.
   * This gives you a chance to create handlers that then reference
   * this looper, before actually starting the loop. Be sure to call
   * {@link #loop()} after calling this method, and end it by calling
   * {@link #quit()}.
   */
   public static void prepare() {
   prepare(true);
   }
`
Looper.prepare() is just set a Looper for the thread.
When  some one create a handler in the thread, then can use the handler's post() method to send message, or deal message within the handler's handleMessage() method.
So that both the post events and consumed are related to the handler.

rom1v:
> When some one create a handler in the thread, then there can use the handler's post() method to send message, or deal message within the handler's handleMessage() method.
> So that both the post events and consumed are related to the handler.

But message/events will never be processed, since we don't (and don't want to) call [`Looper.loop()`](https://developer.android.com/reference/android/os/Looper.html#loop()).

olbb:
If no one creates a Hander and uses it to send a message，there is no message in the queue.

What makes you say that ‘the messages/events will never be processed’？

rom1v:
> If no one creates a Hander and uses it to send a message，there is no message in the queue.

MzNightModeManager does (and if it creates a `Handler`, it probably uses it):


```
Surface: java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
at android.os.Handler.(Handler.java:200)
at android.os.Handler.(Handler.java:114)
at com.meizu.nightmode.MzNightModeManager.(MzNightModeManager.java:158)
at com.meizu.nightmode.MzNightModeManager.getDefault(MzNightModeManager.java:313)
```

Btw, the javadoc (that you pasted here: https://github.com/Genymobile/scrcpy/pull/434#issuecomment-465160729) says:

> Be sure to call `loop()` after calling this method

olbb:
Okay, I understand what you said.
Yes, if called Looper.loop(),the thread will entering an infinite loop, so I did not.
In fact, I just want the scrpy to work properly, but who cares about the fucking code added by the meizu's developer.
If you feel that this is not appropriate,I'll just add the code in mine brach.

amosbird:
This actually works for the Flyme OS. Good job @olbb

rom1v:
@olbb I finally applied the fix (not merged yet) to fix #921. Even if it's not very correct to prepare the looper without looping, it just makes more devices work. :+1:

I call `prepareMainLooper()` instead, and moved it out of the loop (and changed the commit message): 9da4c93582f28352a8f11acbb917d5ce9c07183e

olbb:
I'm glad this can help you.

Stanislavoloshchuk:
> @Stanislavoloshchuk The issue with `MzNightModeManager` (#240) is fixed, so your problem is probably a different one (unless there is a regression).
> 
> Your log.txt does not contain anything related to scrcpy :(
> 
> Just in case, could you test scrcpy v1.13, please?

Hello! Thanks for reply but could you kindly advice how to roll down to v1.13, please?
Thanks a lot!

rom1v:
@Stanislavoloshchuk On Windows, just take the v1.13 release.

On other system, take the prebuilt-server and build: https://github.com/Genymobile/scrcpy/blob/4e9e712312c9726824d28bed4fc85721e3a2b203/BUILD.md#prebuilt-server

ilceramista:
> Hello everybody. Already about 7 hours trying to fix the problem "Scrcpy mirror window not coming to run" on Ubuntu 20.04 with my Meizu M6Note. I still keep reading how to solve it and saw some grateful messages but cannot find solution procedure. It seems to be problems with Meizu DarkMode and it should be fixed with replacement of `scrcpy-server.jar` in the directory scrcpy 1.14 (my current version) Could anyone advice where to find this directory and finally fix it please?
> [log.txt](https://github.com/Genymobile/scrcpy/files/4761220/log.txt)
> [scrcpy.txt](https://github.com/Genymobile/scrcpy/files/4761228/scrcpy.txt)

the thread seem to be closed and i'm not an expert of github or forum generally, but i can told you my meizu m6-note work well under lubuntu 20.04 64bit with scrcpy version 1.12.1!! but just to inform, the meizu won't show a window in my old notebook bodhilinux 32.bit ubuntu 18.04 based and scrcpy version 1.16. Xiaomi mi 9t work well in both system.. 
thanks a lot for this piece of software, i can use ppowerful devices with damaged screen.. hope it can be useful ..regards

PR #442, opened 2019-03-03
server/meson.build: support relative path for prebuilt_server

npes87184:
If we don't do this trick, the prebuilt_server will be
../server/[the_user_defined_path]. In general, we will not give an relative path
based on build directory, which leads to wrong prebuilt_server path.

The building error:

ninja: error: '../scrcpy-server-v1.7.jar', needed by
'server/scrcpy-server.jar', missing and no known rule to make it

### DIFF
--- a/server/meson.build
+++ b/server/meson.build
@@ -10,6 +10,10 @@ if prebuilt_server == ''
                   install: true,
                   install_dir: 'share/scrcpy')
 else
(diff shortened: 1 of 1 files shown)


rom1v:
Ah, cool, not being able to pass a relative path was annoying :+1:

PR #522, opened 2019-05-04
Disable X11 compositor bypass

dos1:
Compositor bypass is meant for fullscreen games consuming lots of GPU resources. For a light app that will usually be windowed, this only causes unnecessary compositor suspends, especially visible (and annoying) with complying window manager like KWin.

As an added bonus, mouse focus clickthrough is fixed as it didn't work due to compat.h header not being included in scrcpy.c

### DIFF
--- a/app/src/compat.h
+++ b/app/src/compat.h
@@ -43,4 +43,9 @@
 # define SCRCPY_SDL_HAS_WINDOW_ALWAYS_ON_TOP
 #endif
 
+#if SDL_VERSION_ATLEAST(2, 0, 8)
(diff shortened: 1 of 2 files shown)


rom1v:
Thank you :+1:

I merged it (after I split this PR into 2 commits: https://github.com/Genymobile/scrcpy/commit/068253a3a299a60d78c7c1b0e760f4bbe3b75bc1 and https://github.com/Genymobile/scrcpy/commit/b941854c7375a0c9a9f3ef957ccbf659e645490c).

PR #542, opened 2019-05-19
Return success count in injectText

### LINKED ISSUE #509: Copying and Pasting from PC to Android
Before upgrading to 1.6 and then 1.8 (i think) I used to be able to copy and paste from my PC to phones easily. I would copy text from my PC and then paste and all the text  would be pasted.

eg. 

12312323
123123312
123132132
123123

However now when I copy and paste it only pastes 1 line and stops at the carriage return. 

eg
12312323

This is a new issue as it was working well before. I have tried multiple ways to resolve the issue, select all etcetc, but to no avail. I have had to email text to my phone to get it there. 

Do you have the older versions available (for PC) so I can test if they work correctly? Again this was working correctly but after upgrading it stopped working. 

I am using multiple devices with android and they all exhibit the same, new, behaviour

npes87184:
It will insert as many text as possible now.
Fix #509, tested on Windows 10 and Arch Linux.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/EventController.java
+++ b/server/src/main/java/com/genymobile/scrcpy/EventController.java
@@ -104,13 +104,15 @@ public class EventController {
         return true;
     }
 
-    private boolean injectText(String text) {
+    private int injectText(String text) {
+        int successCount = 0;
         for (char c : text.toCharArray()) {
             if (!injectChar(c)) {
-                return false;
+                continue;
             }
+            successCount++;
         }
-        return true;
+        return successCount;
     }
 
     private boolean injectMouse(int action, int buttons, Position position) {


server/src/main/java/com/genymobile/scrcpy/EventController.java
```diff
@@ -104,13 +104,15 @@ private boolean injectChar(char c) {
...
 
-    private boolean injectText(String text) {
+    private int injectText(String text) {
+        int successCount = 0;
         for (char c : text.toCharArray()) {
             if (!injectChar(c)) {
-                return false;
+                continue;
             }
+            successCount += 1;
```
rom1v: style nitpicking: `successCount++`

npes87184:
Thank you for your comment.

v1 -> v2:
* fix style

rom1v:
Thank you :+1: Merged as https://github.com/Genymobile/scrcpy/commit/1630f923ef9847142839907a9c2f3c3bde3a3369 on `dev` (instead of `master`, the idea is to get the same version on `master` with or without the prebuilt server).

PR #573, opened 2019-06-03
Add option to set the window title with the serial number

### LINKED ISSUE #120: Setting Window Title
It would be nice to have a command option to set the window title. That way having multiple windows isn't so troublesome to look through/organize on your desktop.

beango1:
I have multiple devices with the same model name.  Setting this window title with a different name is really useful.

Fixes #120

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -28,6 +28,7 @@ struct args {
     uint16_t max_size;
     uint32_t bit_rate;
     bool always_on_top;
+    bool window_serial;
 };
 
 static void usage(const char *arg0) {
@@ -91,6 +92,9 @@ static void usage(const char *arg0) {
         "\n"
         "    -v, --version\n"
         "        Print the version of scrcpy.\n"
+        "\n"
+        "    -w, --window-serial\n"
+        "        Set the Window Title with the serial number.\n"
         "\n"
         "Shortcuts:\n"
         "\n"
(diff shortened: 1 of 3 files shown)


rom1v:

I agree to add we could add an option to set the title. However, I think that the behavior of `--window-serial` is too specific.

I suggest a more general option `--window-title'. A first version could just receive a "constant" window title:

```bash
scrcpy --window-title "My device"
```

An improved version could also support "variables" dynamically retrieved from the device:

```bash
scrcpy --window-title "My Device - %m [%s]"
```

which would set the title `My Device - Nexus 5 [0123456789abcdef]`.

Note that there are two possible serial values: the one provided by the user with `-s`, and the real one [retrieved from the device](https://stackoverflow.com/a/11029386/1987178):

```
provided              real
--------              ----
""                    "0123456789abcdef"
"0123456789abcdef"    "0123456789abcdef"
"192.168.0.10:5555"   "0123456789abcdef"
```

I think it is better to expose the _real_ serial then the _provided_ one:
 - it works even if the user does not provide a serial (when there is only one device);
 - the provided one is known by the caller (it is passed to `-s`), so the value might also be passed to `--window-title`.

What do you think?

_As a side node, I would like to expose only a long option `--window-title` [without a short version](https://stackoverflow.com/a/22465538/1987178) (`-w`) for advanced options, to keep the letters available for future features. If I had implemented them now, I would not have added a short version for options like `--always-on-top` and `--show-touches`…_

beango1:
I like to keep the title short and sweet.  Long titles just are just cropped out of some characters.

However, I agree with all of your points.

rom1v:
> I agree with all of your points.

Do you plan to update this PR to implement `--window-title`? or someone else can work on this?

beango1:
real serial is not a trivial change.  I'll rather stick with what I have
and removing the short switch.

rom1v:
> real serial is not a trivial change.

As a first step, an option `--window-title` accepting a constant string would probably be sufficient.

Then you could call:

    scrcpy -s 0123456789abcdef --window-title 0123456789abcdef

What do you think?

beango1:
leave me to it

beango1:
closing this... sending another pull request

PR #574, opened 2019-06-05
Fix dependencies

crow1170:
Some missing or misspelled dependencies. Checked on Ubuntu 19.04.


rom1v:
Thank you :+1:

In fact, we can even replace `android-tools-adb` by `adb` (I just did it: a13524e7f9d696635983776c548adaa1b9c8508f).

PR #583, opened 2019-06-12
a spell mistake

zzndb:
after commented default portable option in `app/meson.build` get some error and then find this. : )


rom1v:
:scream: Thank you, merged as https://github.com/Genymobile/scrcpy/commit/bcd0a876f7bf52642330410812a6b7100cdeda91/

rom1v:
I wondered how I did not hit the error during my tests, the answer is that the default value for a boolean flag in meson option is `true`, I thought it was `false`: 53310a925a495f61d42dd90faa0a0748074d63c5

PR #595, opened 2019-06-14
The Java JDK is needed to build the server

taaem:
The relevant Fedora package is java-devel and not java


rom1v:
Merges as 0a233fd27fc9c44dc942d921f3daa70206712129.

Issue #601, opened 2019-06-17
[Suggestion] Put user-friendly name of device in window name


Plandercozeron:
For testing purposes
`shell getprop ro.product.model` returns models as **SM-G960F** (Galaxy S9) for samsung devices. Things like Motorola return readable user-friendly names, i.e. Motorola E4 Plus.
But sometimes they don't: **XT1635-02** (Motorola Z Play). I remember all models which i work with by heart. But eye catches easily when i you have window named **Samsung Galaxy S6 Edge+ (SM-G928F)**
`shell settings get secure bluetooth_name` returns almost on every device (Galaxy S4 mini is still GT-9195) it's name if not changed by user.
How reasonable is getting the name from settings?  
It's kind of slow. Or maybe i'm doing that wrong. Because if you make loop which gets info about all connected devices
```
#!/bin/bash
for serial in $(adb devices | awk 'NR>1{print $1}');
	do
		echo -e  $(adb -s $serial shell settings get secure bluetooth_name | sed "s/\\r//g")
	done
```

it would work slower than
```
#!/bin/bash
for serial in $(adb devices | awk 'NR>1{print $1}');
	do
		echo -e  $(adb -s $serial shell getprop ro.product.model | sed "s/\\r//g")
	done
```

rom1v:
In https://github.com/Genymobile/scrcpy/pull/573#issuecomment-498548993, the "bluetooth name" could be a variable (like `%b`) in the custom pattern (not implemented yet).

> It's kind of slow.

It's just because you execute `settings`, so you have the same behavior as `input`: https://github.com/Genymobile/scrcpy/issues/231#issuecomment-414111753. In _scrcpy_, it would be retrieved directly by the server via a method call.

Plandercozeron:
Right. Thanks for explanation.
Also 
```
shell settings get global wifi_p2p_device_name
```
returns device type (Phone, Tablet) on Android 6+
```
on 0215394544313237:    [Phone]  Galaxy S6 edge+
on 4200c004ecd5b361:    [Phone]  Galaxy J5 Prime
on 4ccd2337:                     Galaxy S4
on 5200b0beb46215cb:    [Phone]  Galaxy J7 Neo
on 7a58571a:            [Phone]  Galaxy A5
on abafba0b:            [Tablet] Galaxy Tab A
```
But returns `null` on Blackberry Priv. But not so many people use Blackberry devices =) so probably an option too...

PR #606, opened 2019-06-20
add --window-title options with %M and %S variable replacement

beango1:
solves issue #120 
continuation of #573

### DIFF
--- a/app/src/device.c
+++ b/app/src/device.c
@@ -2,21 +2,23 @@
 #include "log.h"
 
 bool
-device_read_info(socket_t device_socket, char *device_name, struct size *size) {
-    unsigned char buf[DEVICE_NAME_FIELD_LENGTH + 4];
+device_read_info(socket_t device_socket, char *device_name, char *device_serial, struct size *size) {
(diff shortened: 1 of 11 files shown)


rom1v:
Thank you for your PR. It works :+1:

However, I'm wondering if we shouldn't replace the variables on the server side instead:
 - their value are device-specific values, known by the device;
 - all variables are not always used (typically, they are not), so it's useless to transmit them all (more will be available in the future, see #601) just to ignore them on the client side;
 - implementing a new variable would not require to modify the protocol between the server and the client;
 - Java is much safer for string manipulation.

What do you think?

beango1:
Fair points.  Perhaps for next iterations.

After doing all of the changes, I think it would have been cleaner just to
set the window title the vanilla way.  Let the user set the it the way they
like ;)

rom1v:
> After doing all of the changes, I think it would have been cleaner just to set the window title the vanilla way.

A first separate commit doing just that would still be great :wink:

beango1:
#616

Issue #24, opened 2018-03-10
Pinch zooming support


paour:
The Android emulator uses right click and drag to simulate pinch zooming. An alternative would be ctrl-mouse wheel.

AlexHaber:
Pinch zooming with a touchscreen also doesn't work.

CodeAck:
Pinch zooming support would be a great idea!
Ctrl+← and Ctrl+→ could be used since ↓↑ are taken by volume control.
Would be great for people with low/no motor ability in their hands.
(I can't move from the shoulders down and greatly appreciate what you've done with scrcpy)

jknut:
It would be great to have the ability to pinch zoom.  Another alternative could be using the mouse wheel.

Scrcpy is really great!

Pixelstudio:
Would be great if this can be added!

apiemont:
another vote for this feature

ToggoStar:
+1! Scrcpy is awesome and this is the only feature missing for me!

Pixelstudio:
Oeh i see some references to this pr, but i cannot find if this is implemented... 
Any idea?

rom1v:
Hi, I just implemented an experimentation, on branch [`virtualfinger`](https://github.com/Genymobile/scrcpy/commits/virtualfinger).

While the Ctrl key is held, the first left-click adds an immobile "virtual finger" at this position.

Moving the mouse moves another cursor. The left click may be released and pressed as necessary to control the second finger.

Releasing Ctrl releases the virtual finger.

rom1v:
A simpler (and maybe more intuitive) alternative could be to always put the "virtual finger" in the center of the screen, so that "unzooming" is more natural.

As a drawback, it would not be possible to zoom on a widget which takes only a portion of the screen (which is not in the center), but it may be worth it.

CodeAck:
OK cool, so what do we have to do to try this virtualfinger?

srevinsaju:
Literally amazing, this fixes issue #2 https://github.com/srevinsaju/guiscrcpy/issues/2 on guiscrcpy as well

srevinsaju:
@rom1v I am having problems with the dev branch, 

```
INFO: scrcpy 1.10 <https://github.com/Genymobile/scrcpy>
/usr/local/share/scrcpy/scrcpy-server.jar: 1 file pushed. 1.1 MB/s (22546 bytes in 0.019s)
INFO: Initial texture: 720x1280
[...]

PR #676, opened 2019-07-31
Use Cmd instead of Ctrl on macOS

### LINKED ISSUE #642: Use Cmd+c/Cmd+c on Mac OS?
On Mac OS, should we use Cmd instead of Ctrl to copy-paste?

If yes, we should ifdef the same way we do for the volume:
https://github.com/Genymobile/scrcpy/blob/49612561233b3cde41339ee08c71892e42649706/app/src/input_manager.c#L293-L297

here:
https://github.com/Genymobile/scrcpy/blob/49612561233b3cde41339ee08c71892e42649706/app/src/input_manager.c#L312-L327

rom1v:
Use `Cmd` instead of `Ctrl` for all shortcuts on macOS.

Fixes #642

---

Mac users, please confirm that it works as expected :wink:

cc @npes87184 @carstenhag @zhuxianfeng13 @AoDevBlue @numbermaniac


carstenhag:
Hey, I've compiled and tried it out just now.

Cmd+M does not work as excepted, because Cmd+M is a mac (global?) shortcut to minimize the window.

The other shortcuts work fine. One thing though, which I guess is unrelated: I have no way to middle-click on a Magic Mouse. On the Macbook touchpad it's *possible* but really difficult depending on the touchpad settings. It would be good to provide an alternative shortcut for that.

rom1v:
@carstenhag Thank you very much for your tests and feedback.

So I updated this PR to keep `Ctrl` for `Ctrl`+`h` and `Ctrl`+`m`.

> I have no way to middle-click on a Magic Mouse. On the Macbook touchpad it's possible but really difficult depending on the touchpad settings. It would be good to provide an alternative shortcut for that.

What would you suggest? I will probably use `Ctrl`+click or `Shift`+click for "pinch to zoom". Maybe `Ctrl`+Right-click or something like that?

carstenhag:
Yeah, Cmd+Click seems good to me.

rom1v:
Cmd+click for HOME or pinch-to-zoom (not implemented yet)?

carstenhag:
Not really sure what you mean by pinch-to-zoom :) 

Cmd for Home, Shift for that, seems good to me. We can still test it later.

rom1v:
> Not really sure what you mean by pinch-to-zoom :)

https://github.com/Genymobile/scrcpy/issues/24

Issue #298, opened 2018-10-16
MINGW32 build failed


vivopcz:
```
$ meson x --buildtype release --strip -Db_lto=true
The Meson build system
Version: 0.48.0
Source dir: f:\code\vnc\scrcpy
Build dir: f:\code\vnc\scrcpy\x
Build type: native build
Project name: scrcpy
Project version: undefined
Native C compiler: cc (gcc 7.3.0 "cc (GCC) 7.3.0")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: d:\msys2\usr\bin\pkg-config.EXE (0.29.2)
Dependency libavformat found: NO (tried pkgconfig)

app\meson.build:25:4: ERROR:  Dependency "libavformat" not found, tried pkgconfi
g

A full log can be found at f:\code\vnc\scrcpy\x\meson-logs\meson-log.txt
```

vivopcz:
meson build failed , Dependency "libavformat" not found, but i have do 

```
pacman -S mingw-w64-x86_64-SDL2 \
          mingw-w64-x86_64-ffmpeg

pacman -S mingw-w64-x86_64-make \
          mingw-w64-x86_64-gcc \
          mingw-w64-x86_64-pkg-config \
          mingw-w64-x86_64-meson
```

rom1v:
Do you still encounter the problem?

Issue #80, opened 2018-03-18
Error during basic setup


mhdtouban:
Am getting 
Error during basic setup:

Neither directory contains a build file meson.build.

when running  `meson x --buildtype release --strip -Db_lto=tru`
tried to reinstall it but this didn't help as am getting build is already installed.

MacOS: 10.13.3

rom1v:
> Neither directory contains a build file meson.build.

See https://github.com/Genymobile/scrcpy/issues/35.

> build is already installed

Remove the build directory:

    rm -rf x

mhdtouban:
Got this error:

Dependency libavformat found: NO

app/meson.build:21:0: ERROR: Dependency 'libavformat' not found, tried Extra Frameworks and Pkg-Config:

Native dependency 'libavformat' not found

rom1v:
Did you [install the packages](https://github.com/Genymobile/scrcpy/blob/master/README.md#mac-os)?

mhdtouban:
yes
```
Warning: sdl2 2.0.8 is already installed
Warning: ffmpeg 3.4.2 is already installed
```

mhdtouban:
and here is the full log: 
```
Build started at 2018-03-18T13:29:27.492806
Main binary: /usr/local/opt/python/bin/python3.6
Python system: Darwin
The Meson build system
Version: 0.45.0
Source dir: /Users/abdullah/dev/java/scrcpy/scrcpy
Build dir: /Users/abdullah/dev/java/scrcpy/scrcpy/x
Build type: native build
Project name: scrcpy
Sanity testing C compiler: cc
Is cross compiler: False.
Sanity check compiler command line: cc /Users/abdullah/dev/java/scrcpy/scrcpy/x/meson-private/sanitycheckc.c -o /Users/abdullah/dev/java/scrcpy/scrcpy/x/meson-private/sanitycheckc.exe
Sanity check compile stdout:

-----
Sanity check compile stderr:

-----
Running test binary command: /Users/abdullah/dev/java/scrcpy/scrcpy/x/meson-private/sanitycheckc.exe
Native C compiler: cc (clang 9.0.0 "Apple LLVM version 9.0.0 (clang-900.0.39.2)")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /opt/local/bin/pkg-config (0.29.1)
Determining dependency 'libavformat' with pkg-config executable '/opt/local/bin/pkg-config'
Dependency libavformat found: NO

[...]

PR #823, opened 2019-09-27
fix scrcpy option and meson build

clang-clang-clang:


rom1v:
Thank you :+1:

I merged the first commit for now.

About the second, why is your ffmpeg installed in `/usr/local/opt/`? By default (after `apt-get install`), it is not the case. Which system do you use?

clang-clang-clang:
Thank you.

It's my bad,

I am using Mac OS, and `/usr/local/opt/` is `brew` install `ffmpeg` files location.

I suppose to add the note in section [Mac OS, System-specific steps](https://github.com/Genymobile/scrcpy/blob/master/BUILD.md#mac-os) when I met the build issue:

> libavformat found: NO

And I found the issue #80 (Mac) and issue #298 (Maybe Linux cross Win), so I decide to move it into _Common steps_. Apparently I have not checked the difference between those system's default install location.

I tested it on Mint Linux, not have this issue,`PKG_CONFIG_PATH` should be `/usr/lib/x86_64-linux-gnu/pkgconfig`.

Should we add the note into _FAQ_ or replace that path with a _PLACEHOLDER_? Seems most systems not have this issue.

Issue #679, opened 2019-08-01
Features implemented in a fork


rom1v:
Here is a specific thread for https://github.com/Genymobile/scrcpy/issues/495#issuecomment-517322589 (@Lurker00):

> Currently I've forked this project, and merged all my changes into the official 1.9 version. It is [there](https://github.com/Lurker00/scrcpy). I only have Android 8.1 devices, so I'm not sure that all my changes will work for all. It requires additional app (my dirty build of [ADBKeyBoard](https://github.com/Lurker00/ADBKeyBoard)) to type any language and to sync device's orientation by PC. In general, I'm not sure should I try a pull request... But you may build both projects yourself and give it a try...
> 
> The full list of my changes (sorry for misuse this thread!):
> User visible changes:
> - 10-fingers multitouch support.
> - Any charset input support with an aid of ADBKeyBoard.
> - Quit command (Ctrl-Q)
> - Ctrl-Shift-O turns device screen on.
> - Connection timeout detector from the device side (useful with connection over wireless network).
> - Screen density and screen size can be set on the device during the session.
> - "Tablet mode": dim device screen, synchronize device screen orientation with PC (requires additional service on the device).
> 
> Internals:
> - The server "forks" itself for full life cycle control (adb may kill it instead preventing required cleanup).
> - Full cleanup at exit.
> - Expandable parameters to the server applet, allowing compatibility between versions.
> - Java NIO to simplify network I/O and not be limited to Unix sockets.
> - Different connection type: direct connect to the device instead of "adb forward".
> - Commands: actions with no parameters

rom1v:
Hi,

Thank you for sharing all your work. :+1: You implemented a lot of things in your corner!

> In general, I'm not sure should I try a pull request

When features are "general" enough and would benefit everyone, yes you should propose to merge them (one PR per feature based on the current `dev` branch) :) (and this would minimize the amount of work for you to keep your fork up to date :wink:)

> - 10-fingers multitouch support.

This is definitely a feature to merge (#22). I investigated locally, especially to support pinch-to-zoom (#24) with `Shift`+click or something. Both features should share a lot of code.

> - Any charset input support with an aid of ADBKeyBoard.

(didn't look yet, also see https://github.com/Genymobile/scrcpy/issues/431#issuecomment-471245573)

> - Quit command (Ctrl-Q)

Is it necessary? (every system already has it own shortcut, typically Alt+F4, don't they?)

> - Ctrl-Shift-O turns device screen on.

[...]

PR #824, opened 2019-09-28
Turn screen off when closing scrcpy

### LINKED ISSUE #714: Power off screen on close
Would it be possible to have a flag power off the screen when you close the stream, like how you do with Ctrl-P, but it activates when you close the scrcpy window.

npes87184:
Fix #714

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -99,9 +99,6 @@ foreach f : check_functions
   endif
 endforeach
 
-# expose the build type
-conf.set('NDEBUG', get_option('buildtype') != 'debug')
-
 # the version, updated on release
 conf.set_quoted('SCRCPY_VERSION', meson.project_version())
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -667,6 +667,7 @@ guess_record_format(const char *filename) {
 #define OPT_FORWARD_ALL_CLICKS     1023
 #define OPT_LEGACY_PASTE           1024
 #define OPT_ENCODER_NAME           1025
+#define OPT_POWER_OFF_ON_CLOSE     1026
 
 bool
 scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
@@ -717,6 +718,8 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
         {"window-height",          required_argument, NULL, OPT_WINDOW_HEIGHT},
         {"window-borderless",      no_argument,       NULL,
                                                   OPT_WINDOW_BORDERLESS},
+        {"power-off-on-close",     no_argument,       NULL,
+                                                  OPT_POWER_OFF_ON_CLOSE},
         {NULL,                     0,                 NULL, 0  },
     };
 
@@ -884,6 +887,9 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
                 break;
             case OPT_LEGACY_PASTE:
                 opts->legacy_paste = true;
+                break;
+            case OPT_POWER_OFF_ON_CLOSE:
+                opts->power_off_on_close = true;
                 break;
             default:
                 // getopt prints the error message on stderr
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -286,7 +286,7 @@ rotate_client_right(struct screen *screen) {
     screen_set_rotation(screen, new_rotation);
 }
 
-void
+static void
 input_manager_process_text_input(struct input_manager *im,
                                  const SDL_TextInputEvent *event) {
     if (is_shortcut_mod(im, SDL_GetModState())) {
@@ -366,7 +366,7 @@ convert_input_key(const SDL_KeyboardEvent *from, struct control_msg *to,
     return true;
 }
 
-void
+static void
 input_manager_process_key(struct input_manager *im,
                           const SDL_KeyboardEvent *event) {
     // control: indicates the state of the command-line option --no-control
@@ -551,7 +551,7 @@ convert_mouse_motion(const SDL_MouseMotionEvent *from, struct screen *screen,
     return true;
 }
 
-void
+static void
 input_manager_process_mouse_motion(struct input_manager *im,
                                    const SDL_MouseMotionEvent *event) {
     if (!event->state) {
@@ -605,7 +605,7 @@ convert_touch(const SDL_TouchFingerEvent *from, struct screen *screen,
     return true;
 }
 
-void
+static void
 input_manager_process_touch(struct input_manager *im,
                             const SDL_TouchFingerEvent *event) {
     struct control_msg msg;
@@ -637,7 +637,7 @@ convert_mouse_button(const SDL_MouseButtonEvent *from, struct screen *screen,
     return true;
 }
 
-void
+static void
 input_manager_process_mouse_button(struct input_manager *im,
                                    const SDL_MouseButtonEvent *event) {
     bool control = im->control;
@@ -736,7 +736,7 @@ convert_mouse_wheel(const SDL_MouseWheelEvent *from, struct screen *screen,
     return true;
 }
 
-void
+static void
 input_manager_process_mouse_wheel(struct input_manager *im,
                                   const SDL_MouseWheelEvent *event) {
     struct control_msg msg;
@@ -745,4 +745,47 @@ input_manager_process_mouse_wheel(struct input_manager *im,
             LOGW("Could not request 'inject mouse wheel event'");
         }
     }
+}
+
+bool
+input_manager_handle_event(struct input_manager *im, SDL_Event *event) {
+    switch (event->type) {
+        case SDL_TEXTINPUT:
+            if (!im->control) {
+                return true;
+            }
+            input_manager_process_text_input(im, &event->text);
+            return true;
+        case SDL_KEYDOWN:
+        case SDL_KEYUP:
+            // some key events do not interact with the device, so process the
+            // event even if control is disabled
+            input_manager_process_key(im, &event->key);
+            return true;
+        case SDL_MOUSEMOTION:
+            if (!im->control) {
+                break;
+            }
+            input_manager_process_mouse_motion(im, &event->motion);
+            return true;
+        case SDL_MOUSEWHEEL:
+            if (!im->control) {
+                break;
+            }
+            input_manager_process_mouse_wheel(im, &event->wheel);
+            return true;
+        case SDL_MOUSEBUTTONDOWN:
+        case SDL_MOUSEBUTTONUP:
+            // some mouse events do not interact with the device, so process
+            // the event even if control is disabled
+            input_manager_process_mouse_button(im, &event->button);
+            return true;
+        case SDL_FINGERMOTION:
+        case SDL_FINGERDOWN:
+        case SDL_FINGERUP:
+            input_manager_process_touch(im, &event->tfinger);
+            return true;
+    }
+
+    return false;
 }
--- a/app/src/input_manager.h
+++ b/app/src/input_manager.h
@@ -40,28 +40,7 @@ void
 input_manager_init(struct input_manager *im,
                    const struct scrcpy_options *options);
 
-void
-input_manager_process_text_input(struct input_manager *im,
-                                 const SDL_TextInputEvent *event);
-
-void
-input_manager_process_key(struct input_manager *im,
-                          const SDL_KeyboardEvent *event);
-
-void
-input_manager_process_mouse_motion(struct input_manager *im,
-                                   const SDL_MouseMotionEvent *event);
-
-void
-input_manager_process_touch(struct input_manager *im,
-                            const SDL_TouchFingerEvent *event);
-
-void
-input_manager_process_mouse_button(struct input_manager *im,
-                                   const SDL_MouseButtonEvent *event);
-
-void
-input_manager_process_mouse_wheel(struct input_manager *im,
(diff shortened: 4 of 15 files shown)


app/src/control_msg.h
```diff
@@ -25,6 +25,7 @@ enum control_msg_type {
     CONTROL_MSG_TYPE_GET_CLIPBOARD,
     CONTROL_MSG_TYPE_SET_CLIPBOARD,
     CONTROL_MSG_TYPE_SET_SCREEN_POWER_MODE,
+    CONTROL_MSG_TYPE_TURN_SCREEN_OFF,
```
rom1v: We need to distinguish between "turn screen off but keep device on" (like [here](https://github.com/Genymobile/scrcpy/blob/513d1ac96d24d6cf07fb1850232b5c0e8f555615/app/src/scrcpy.c#L391)) and "turn screen off and device off (standby)". (There is already some naming confusion in the current codebase, if you have any suggestion…)

rom1v:
I merged the first commit fixing the log :+1:

app/src/control_msg.h
npes87184: Maybe lock screen vs turn screen off?

npes87184:
I have pushed v2:
* Let this feature be optional.
* use "lock screen"

However, I found that server may be killed before screen off.
We may need to introduce some ack mechanism from server to client.

rom1v:
> However, I found that server may be killed before screen off.
> We may need to introduce some ack mechanism from server to client.

In my TODO list, I plan to implement something similar to #679 to be able to cleanup properly:
> The server "forks" itself for full life cycle control (adb may kill it instead preventing required cleanup).

This will be a requirement to merge this PR.

rom1v:
Now that there is a real cleanup, you could implement it :wink:

(I suggest `--power-off-on-close`)

rom1v:
@npes87184 I think several people are interested:
 - #714
 - #1764
 - #2103
 - https://www.reddit.com/r/scrcpy/comments/ljzpdd/turning_off_the_display_automatically_upon/
:wink:

npes87184:
Ok, I will look it.

npes87184:
Hi, @rom1v 

There is a problem about power off screen in `CleanUp` process. In this time, `injectEvent` will consider about `displayId` to inject to correct target. However, it is diffiuclt to allocate `Device` in the process of `CleanUp`. As an result, I cannot trigger power off screen in that stage. Do you have any good advice?

Thank you.

rom1v:
As a quick-and-dirty solution, maybe you could expose a static `injectEvent()` taking a displayId:

```java
diff --git a/server/src/main/java/com/genymobile/scrcpy/Device.java b/server/src/main/java/com/genymobile/scrcpy/Device.java
index a8fdf677..9fb365ff 100644
--- a/server/src/main/java/com/genymobile/scrcpy/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
@@ -157,16 +157,20 @@ public final class Device {
         return supportsInputEvents;
     }
 
+    public static boolean injectEvent(InputEvent inputEvent, int mode, int displayId) {
+        if (displayId != 0 && !InputManager.setDisplayId(inputEvent, displayId)) {
+            return false;
+        }
+
+        return SERVICE_MANAGER.getInputManager().injectInputEvent(inputEvent, mode);
+    }
+
     public boolean injectEvent(InputEvent inputEvent, int mode) {
         if (!supportsInputEvents()) {
             throw new AssertionError("Could not inject input event if !supportsInputEvents()");
         }
 
-        if (displayId != 0 && !InputManager.setDisplayId(inputEvent, displayId)) {
-            return false;
-        }
-
-        return SERVICE_MANAGER.getInputManager().injectInputEvent(inputEvent, mode);
+        return injectEvent(inputEvent, mode, displayId);
     }
 
     public boolean injectEvent(InputEvent event) {
```

Then, you have to transmit both the displayId and the flag to NOT turn off if control is disabled.

What do you think?

npes87184:
> As a quick-and-dirty solution, maybe you could expose a static `injectEvent()` taking a displayId:
> 
> ```java
> diff --git a/server/src/main/java/com/genymobile/scrcpy/Device.java b/server/src/main/java/com/genymobile/scrcpy/Device.java
> index a8fdf677..9fb365ff 100644
> --- a/server/src/main/java/com/genymobile/scrcpy/Device.java
> +++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
> @@ -157,16 +157,20 @@ public final class Device {
>          return supportsInputEvents;
>      }
>  
> +    public static boolean injectEvent(InputEvent inputEvent, int mode, int displayId) {
> +        if (displayId != 0 && !InputManager.setDisplayId(inputEvent, displayId)) {
> +            return false;
> +        }
> +
> +        return SERVICE_MANAGER.getInputManager().injectInputEvent(inputEvent, mode);
> +    }
> +
>      public boolean injectEvent(InputEvent inputEvent, int mode) {
>          if (!supportsInputEvents()) {
>              throw new AssertionError("Could not inject input event if !supportsInputEvents()");
>          }
>  
> -        if (displayId != 0 && !InputManager.setDisplayId(inputEvent, displayId)) {
> -            return false;
> -        }
> -
> -        return SERVICE_MANAGER.getInputManager().injectInputEvent(inputEvent, mode);
> +        return injectEvent(inputEvent, mode, displayId);
>      }
>  
>      public boolean injectEvent(InputEvent event) {
> ```
> 
> Then, you have to transmit both the displayId and the flag to NOT turn off if control is disabled.
> 
> What do you think?

Oh, yes. This solution works. However, I am thinking about future. It may need to more and more things that need to be done in `CleanUp`. If we need to pass variable likes this in each time, it is very annoy and dirty.

For infrastructure, in my opinion, in `CleanUp` process, I prefer to prepare a environment similar to the main process.
It just needs to handle the `options` correctly. Maybe it can add two methods, `toFile` and `fromFile`, to the class options.

What do you think?

rom1v:
> It just needs to handle the `options` correctly.

It could be great to serialize some structure between the main process and the cleanup process indeed.

However, this structure should not be the existing `Options`, which is neither necessary (not all fields are meaningful for the cleanup process) nor sufficient (the cleanup actions may for example depend on the initial state of the device, which is not stored in the options).

(I don't know if in the end such a structure would be helpful compared to passing each option per argument.)

But yes, it would be great to share the common part of the `Device` initialization between the main and cleanup process. For now, `Device` is quite a mess, and it's not clear semantically what this class represents.

npes87184:
Ok, I will pass `displayId` in this time. The refactoring of `Device` and the environment preparation of `CleanUp` should be another problems (or features).

Thank you for your help.

npes87184:
Hi, @rom1v 

I have added two commits about this features.

First, I export a series static methods for powering off. Second, make `CleanUp` process respects the power off setting.

Thanks.

rom1v:
Thank you very much for your work :+1:

I just tested, and I found a problem (on a Nexus 5): the "power off" cleanup interacts badly with the "restore normal power mode" cleanup, which is executed whenever the screen is on on exit. As a result, `--power-off-on-close` works randomly (on this device).

I tried with these changes:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
index 7dda83fb..26012409 100644
--- a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
@@ -84,16 +84,14 @@ public final class CleanUp {
             }
         }
 
-        if (restoreNormalPowerMode) {
-            Ln.i("Restoring normal power mode");
-            if (Device.isScreenOn()) {
+        if (Device.isScreenOn()) {
+            if (powerOffScreen) {
+                Ln.i("Power off screen");
+                Device.powerOffScreen(displayId);
+            } else if (restoreNormalPowerMode) {
+                Ln.i("Restoring normal power mode");
                 Device.setScreenPowerMode(Device.POWER_MODE_NORMAL);
             }
         }
-
-        if (powerOffScreen) {
-            Ln.i("Power off screen");
-            Device.powerOffScreen(displayId);
-        }
     }
 }
```

It improves the behavior.

But I have an additional issue that can I cannot explain: if I keep touching the device (I keep one finger on the device screen) while I close the scrcpy window, it blocks and never power off the screen. It is unexpected, because if I keep one finger on the screen and press <kbd>MOD</kbd>+<kbd>p</kbd> to power off the screen, then it works. :confused: 

If I add some logs (and even if I always return `true`) in `Device.isScreenOn()`:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
index 26012409..2c54558c 100644
--- a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
@@ -84,7 +84,9 @@ public final class CleanUp {
             }
         }
 
+        Ln.i("CLEANUP #1");
         if (Device.isScreenOn()) {
+            Ln.i("CLEANUP #2");
             if (powerOffScreen) {
                 Ln.i("Power off screen");
                 Device.powerOffScreen(displayId);
@@ -93,5 +95,6 @@ public final class CleanUp {
                 Device.setScreenPowerMode(Device.POWER_MODE_NORMAL);
             }
         }
+        Ln.i("CLEANUP #3");
     }
 }
diff --git a/server/src/main/java/com/genymobile/scrcpy/Device.java b/server/src/main/java/com/genymobile/scrcpy/Device.java
index 624c9fa0..3ee248d3 100644
--- a/server/src/main/java/com/genymobile/scrcpy/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
@@ -212,7 +212,7 @@ public final class Device {
     }
 
     public static boolean isScreenOn() {
-        return SERVICE_MANAGER.getPowerManager().isScreenOn();
+        return true; //SERVICE_MANAGER.getPowerManager().isScreenOn();
     }
 
     public synchronized void setRotationListener(RotationListener rotationListener) {
```

Then it blocks:

```
02-21 11:24:23.251 21036 21056 D scrcpy  : Device message sender stopped
02-21 11:24:23.485 21054 21054 I scrcpy  : Cleaning up
02-21 11:24:23.485 21054 21054 I scrcpy  : CLEANUP #1
```

If I replace `Device.isScreenOn()` by `true`, then it prints `CLEANUP #2`, but and `Power off screen`, but does not power off the screen. :thinking:

rom1v:
In fact, it seems the cleanup process is killed (the process does not appear in `ps`). 

Btw, if I add a `SystemClock.sleep(1000);` (even if I don't touch the device), the process is closed before `CLEANUP #1`.
The fact that I touch the screen is irrelevant (it probably just change the timings and makes it more probable).

It's probably due to the `OMXNodeInstance: !!! Observer died`:
```
02-21 11:45:31.420 22574 22593 D scrcpy  : Controller stopped
02-21 11:45:32.469 22592 22592 I scrcpy  : Cleaning up
02-21 11:45:32.478 28310 29411 E OMXNodeInstance: !!! Observer died. Quickly, do something, ... anything...
02-21 11:45:32.590 28310 29411 W GraphicBufferSource: Dropped back down to Loaded without Executing
02-21 11:45:32.624 28310 29411 I OMX-VENC: Component Deinit
```

npes87184:
> In fact, it seems the cleanup process is killed (the process does not appear in `ps`).
> 
> Btw, if I add a `SystemClock.sleep(1000);` (even if I don't touch the device), the process is closed before `CLEANUP #1`.
> The fact that I touch the screen is irrelevant (it probably just change the timings and makes it more probable).
> 
> It's probably due to the `OMXNodeInstance: !!! Observer died`:
> 
> ```
> 02-21 11:45:31.420 22574 22593 D scrcpy  : Controller stopped
> 02-21 11:45:32.469 22592 22592 I scrcpy  : Cleaning up
> 02-21 11:45:32.478 28310 29411 E OMXNodeInstance: !!! Observer died. Quickly, do something, ... anything...
> 02-21 11:45:32.590 28310 29411 W GraphicBufferSource: Dropped back down to Loaded without Executing
> 02-21 11:45:32.624 28310 29411 I OMX-VENC: Component Deinit
> ```
> 
> (I don't have this problem on my OnePlus 7 Pro)

Hmm, I am not familiar with `OMXNodeInstance: !!! Observer died`.

BTW, I don't notice this problem in my note10+ with android 11.


```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
index 7dda83fb..26012409 100644
--- a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
@@ -84,16 +84,14 @@ public final class CleanUp {
             }
         }
 
-        if (restoreNormalPowerMode) {
-            Ln.i("Restoring normal power mode");
-            if (Device.isScreenOn()) {
+        if (Device.isScreenOn()) {
+            if (powerOffScreen) {
+                Ln.i("Power off screen");
+                Device.powerOffScreen(displayId);
+            } else if (restoreNormalPowerMode) {
+                Ln.i("Restoring normal power mode");
                 Device.setScreenPowerMode(Device.POWER_MODE_NORMAL);
             }
         }
-
-        if (powerOffScreen) {
-            Ln.i("Power off screen");
-            Device.powerOffScreen(displayId);
-        }
     }
 }
```

I will apply this change as it is more correct.

Thank you for your test and survey.

rom1v:
Thank you  :+1: Merged into `dev`

PR #837, opened 2019-10-02
BUILD docs: add a command to go back to the parent directory, so that…

lmanul:
… the next command on the page ("./run") works (for absent-minded people like me who will try "run" from within "x")


rom1v:
Thank you for your suggestion.

I found a bit weird to add `cd ..` here though (it's a useless command for the documented section).

Instead, I changed calls to `ninja` to not change directory: 3ea47423217775d2519fb5ac61edd55072639a64

Issue #650, opened 2019-07-18
Double input along with latin accents


dstwwy:
Hello.

I'm using the suggested keyboard app in https://github.com/Genymobile/scrcpy/issues/431 on my main phone and the stock version of LineageOS keyboard on my secondary phone, and both are able to type accented (áèîõü) characters as explained in https://blog.rom1v.com/2018/03/introducing-scrcpy/#handle-text-input.

However, when inputing [accent] then [key], scrcpy isn't able to recognize that I want a accented character only, so **áá** turns into **aáaá**. Is it possible to filter the output when I type an accent, like by just skiping the first output?

Thanks for the project, it was a lifechanger ever since its first release.

rom1v:
To fix #87, since v1.4, scrcpy handles letters differently than other text. This is the cause of this bug.

I should probably manage several "input modes" and introduce a parameter to select it, because I think there is no solution working reliably for all cases. The modes would be:
 - raw key events only (ignore text events)
 - text events (ignore raw key events which produce text) (this was th behavior until 1.3, the one you want)
 - mixed (text events only for non-letters) (this is the current behavior, a compromise to fix #87 while working relatively good for text)

dstwwy:
Started testing 1.3 and got a bunch of wildly varying results, before noticing that I was on Gboard. Thanks Google.

So far it behaves as you said on 1.3, so I'll use both according to my needs. As for your solution, it seems that every options brings issues, so I don't know if it's worth pursuing instead of waiting for a fix on how adb works (dunno if even possible) or something else.

rom1v:
This is not a bug in adb, just a conflict between injecting text and injecting press+release events.

dstwwy:
It's not caused by how adb works (https://github.com/Genymobile/scrcpy/issues/214)? Although you explained how you are handling accents on your blog, so you might be sidestepping this issue. If so I'll be here for tests, if needed.

rom1v:
> It's not caused by how adb works (#214)?

No. There are 2 issues:
 1. Android only supports ASCII for injecting text.
 2. we need different events for generating text or for pressing keys.

If you inject `à`, with the workaround explained in the blog post, we can inject the text correctly, so the first issue is not a problem here.

In itself, here, the second issue should not be a problem either: it is clear that we want to inject `à` as text.

[...]

Issue #87, opened 2018-03-21
Unable to use keyboard on Unreal Engine game


aneutron:
Hey !

First of all thanks for the amazing work. I noticed the keyboard is usable in a normal manner, and it works for apps such as Messenger, etc., and it works with Hearthstone too. However, it does not work with PUBG Mobile.

I just noticed this, and thought that maybe it's worth reporting, because on the one hand it could be something with Unreal Engine, but it could also be a bug in scrcpy.

Anyway, you're doing amazing work, thanks a lot !

rom1v:
> However, it does not work with PUBG Mobile.

I just installed the game on some device, it works for me. I can type text for example in the "team chat".

aneutron:
Yes but the controls don't work. I've seen people using OTG keyboard and mice on YouTube to play.

pete1414:
Please see screenshot below.  Top half is key events from Scrcpy, bottom half is keyboard events from bluetooth keyboard.    In both situations I held down the 'w' key'.    It seems Scrcpy sends  Key up before it is released.

![image](https://user-images.githubusercontent.com/6015306/39228948-8cc331be-482e-11e8-8f0a-7f17b5f59053.png)

pete1414:
I have a work around for this, not fully sure of the implications but it at least works for me in PUBG.
I am not too familiar with github, so I'll just explain what I did.
in scrcpy.c
 I removed SDL_TEXTINPUT from the case statement.
 I copied the call in KEYUP to KEYDOWN.
Then in Convert.c I added more key mappings to include alpha keys and number keys and a few others..

It now works in PUBG, I can also still type normally and paste text..   things that were handled in text input will not work like volume up etc, they could easily be moved to the key handling event, but I haven't bothered.

I have not thoroughly tested this.. really just wanted it to work for games and now it does..  no longer have to connect a bluetooth keyboard. 

This applies to #122 and #127  as well I believe.

rom1v:
Cool :+1:

> I am not too familiar with github, so I'll just explain what I did.

You should click on "fork", then push your branch to your github fork and post the link here :)

> Then in Convert.c I added more key mappings to include alpha keys and number keys and a few others..

What is your keyboard layout? What if you press Shift+`,` (for example, on a French keyboard)?

I initially implemented key events using keycodes only (before _scrcpy_ was published), but after reading documentation and discussions on irc #sdl, I think there is no way to work independently of your keyboard layout. Maybe I am wrong though.

[...]

PR #903, opened 2019-11-03
[WIP]Support inject unicoe text

pangliang:
Refer to the Vysor method, add a custom ime to enter unicode characters

* add IME 'ScrcpyIME' to handle InjectText
* add switch cmd 'ctrl+e' to  use ime, 'ctrl+shift+e' to use injectKeyEvent
* change original handler behavior:
    * change text_input to send all types char
    * when enable IME, injectKeyEvent skip Letter or Digit or Space char, InjectText to handler it
    * when disable IME, InjectText skip all text, injectKeyEvent to handler it

[WIP]
- [ ] ime build script
- [ ] enable/disable ime by cmd and scrcpy liefcycle
- [ ] auto install ime apk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ build/
 /dist/
 .idea/
 .gradle/
+x
+*.iml
+local.properties
--- a/app/src/control_msg.c
+++ b/app/src/control_msg.c
@@ -80,6 +80,9 @@ control_msg_serialize(const struct control_msg *msg, unsigned char *buf) {
         case CONTROL_MSG_TYPE_GET_CLIPBOARD:
             // no additional data
             return 1;
+        case CONTROL_MSG_TYPE_SET_INJECT_TEXT_MODE:
+            buf[1] = msg->set_inject_text_mode.mode;
+            return 2;
         default:
             LOGW("Unknown message type: %u", (unsigned) msg->type);
             return 0;
--- a/app/src/control_msg.h
+++ b/app/src/control_msg.h
@@ -28,12 +28,18 @@ enum control_msg_type {
     CONTROL_MSG_TYPE_GET_CLIPBOARD,
     CONTROL_MSG_TYPE_SET_CLIPBOARD,
     CONTROL_MSG_TYPE_SET_SCREEN_POWER_MODE,
+    CONTROL_MSG_TYPE_SET_INJECT_TEXT_MODE,
 };
 
 enum screen_power_mode {
     // see <https://android.googlesource.com/platform/frameworks/base.git/+/pie-release-2/core/java/android/view/SurfaceControl.java#305>
     SCREEN_POWER_MODE_OFF = 0,
     SCREEN_POWER_MODE_NORMAL = 2,
+};
+
+enum inject_text_mode {
+    USE_INPUT_MANAGER = 0,
+    USE_SCRCPY_IME = 1,
 };
 
 struct control_msg {
@@ -65,6 +71,9 @@ struct control_msg {
         struct {
             enum screen_power_mode mode;
         } set_screen_power_mode;
+        struct {
+            enum inject_text_mode mode;
+        } set_inject_text_mode;
     };
 };
 
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -174,6 +174,18 @@ set_screen_power_mode(struct controller *controller,
 }
 
 static void
+set_inject_text_mode(struct controller *controller,
+                      enum inject_text_mode mode) {
+    struct control_msg msg;
+    msg.type = CONTROL_MSG_TYPE_SET_INJECT_TEXT_MODE;
+    msg.set_inject_text_mode.mode = mode;
+
+    if (!controller_push_msg(controller, &msg)) {
+        LOGW("Could not request 'set screen power mode'");
+    }
+}
+
+static void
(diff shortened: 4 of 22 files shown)


AndroidDeveloperLB:
This won't work, because the moment you disconnect the device, the app will still be the default keyboard.
Plus, sometimes we do want to have the default keyboard...
Isn't there another workaround for this?

pangliang:
4 SDL Events here: SDL_TEXTINPUT, SDL_TEXTEDITING, SDL_KEYDOWN, SDL_KEYUP
I think inputtext event and key event are two completely different events and should not be converted to each other. 
The inputtext event is only generated when the input method (IME) window is show. 
The input method is only show when edit text area is focus

when android device edit text area is focus,  You definitely want to enter text, you care about the hold text;
otherwise, for example when you are playing a game; you want to know when the keyboard button is down and up; you will care about the SDL_KEYDOWN and SDL_KEYUP events

Scrcpy splits text into character arrays to handle:
```
private int injectText(String text) {
        for (char c : text.toCharArray()) {
            injectChar(c)
        }
}
```
but, a lot of text  is 2 or even 4 bytes of unicode string, which cannot be split into a single char. 

this commit work for me, ctrl+e to use customer IME to input chinese, ctrl+shift+e to use the original injectKeyEvent to play game

BTW,
InputMethodService. onWindowShown/Hidden  on(Un)BindInput  callback  can automatic switching of inputtext handler and keycode handler
but, it need to create a link from ime to scrcpy app, I think it is too complicated, it is enough for me to use ctrl+e

pangliang:
vysor use a Custom IME method to enter text
https://www.vysor.io/

rom1v:
Thank you for your work.

Supporting non-ASCII characters is a feature which is requested a lot (especially for Chinese or other non-ASCII languages), so that's great :+1:

I did not have time to review or test yet (I would like to finish things to be fixed for v1.11, and maybe investigate the new audio API first), but it's in my TODO list :)

> I think inputtext event and key event are two completely different events and should not be converted to each other.

Yes, I agree. It's just a constraint from key injection.

_However, on the computer, when you press `a` for example, you receive both the key events (`A` pressed, `A` released) and the text event (`"A"`). If we forward both, there will be two `a` inserted, so we still have to decide which ones must be forwarded._

> when android device edit text area is focus, You definitely want to enter text, you care about the hold text, not a single character
otherwise, for example when you are playing a game; you want to know when the keyboard button is down and up; you will care about the SDL_KEYDOWN and SDL_KEYUP events

Exactly, that's #87. And on `dev` branch, a new option had been added to "prefer text events", as a mitigation (#650): https://github.com/Genymobile/scrcpy/commit/c916af0984f72a60301d13fa8ef9a85112f54202

(Btw, you should target `dev` branch instead of `master`)

> but, a lot of text is 2 or even 4 bytes of unicode string, which cannot be split into a single char.

This is even worse than that, this method only works for ASCII text (on 7 bits) (note that `char` is an unsigned 16 bits in Java, so in theory 2 bytes unicodes could be supported). For example, injecting `é` does not work (but there is a workaround so we can [inject some accented characters](https://blog.rom1v.com/2018/03/introducing-scrcpy/#handle-accented-characters)).

Questions:
 - what happens when you unplug the device? do you need to manually change the IME?
 - does it work when you copy-paste chinese text directly (`Ctrl`+`c` on the computer, `Ctrl`+`v` in _scrcpy_) while the IME is enabled?

pangliang:
@rom1v  Thank you for your reply and endure my poor English. :)

> However, on the computer, when you press a for example, you receive both the key events (A pressed, A released) and the text event ("A"), so we still have to decide which ones must be forwarded.

Yes, I agree. The main problem is that we can't know which component(device, android ui) is being focused. 

SDL provides two methods, let developers choose whether to send inputtext event. 
* `SDL_StartTextInput()`: only InputText event, no keyEvent
* `SDL_StopTextInput()`: only keyEvent event, no InputText

But, the scrcpy app(computer side) does not have an EditText component. So, just on the scrcpy side,  it’s difficult to decide when to start and to stop textinput.
On the device(android) side, if we can know that the inputtext view is the focus, we can make a decision. Fortunately, IME can know this thing. use `InputMethodService` class `onWindowShown()` and `onWindowHidden()` callback function

But unfortunately, I am trying hard to get IME and scrcpy server to communicate directly with each other. This won't work. app_process app can send broadcast but can not recv.
So we have to find a way.  For example, create a tcp link from IME to scrcpy app (computer side).  IME send InputMethod status to app and app forward to server.
This is too complicated ! In order to enter characters, this is too expensive.

The `scrcpy` achieves great features with the most succinct code. So I think this should also be kept simple.
So, I think, let users make this choice is simple.

> what happens when you unplug the device? do you need to manually change the IME?
As it is now, yes. But as I said above, this can be smarter.

> does it work when you copy-paste chinese text directly (Ctrl+c on the computer, Ctrl+v in scrcpy) while the IME is enabled?
yes, it works

rom1v:
> Merge branch 'dev' into input_method

Oh, you should not do that, but rebase instead, to get a clean reviewable patchset in the end. :)

AndroidDeveloperLB:
Is it possible to do it using root instead of an additional keyboard app?

AndroidDeveloperLB:
Closed because implemented?

Issue #613, opened 2019-06-23
[server] ERROR


YanYangB:
I also want to get the video stream only from the android like #129 . 
But when I run the command:
```
➜ /home/yanyang > adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 0 800000 true 
[server] ERROR: Exception on thread Thread[main,5,main]
java.lang.IllegalArgumentException: Expecting 5 parameters
	at com.genymobile.scrcpy.Server.createOptions(Server.java:71)
	at com.genymobile.scrcpy.Server.main(Server.java:132)
	at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method)
	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:340)
```
I use it in Ubuntu18.04 
```
scrcpy 1.9

dependencies:
 - SDL 2.0.8
 - libavcodec 57.107.100
 - libavformat 57.83.100
 - libavutil 55.78.100
```
adb version: 1.0.41
Thanks.

YanYangB:
I also want to get the video stream only from the android like #129 .
But when I run the command:

> ➜ /home/yanyang > adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 0 800000 true [server] ERROR: Exception on thread Thread[main,5,main] java.lang.IllegalArgumentException: Expecting 5 parameters at com.genymobile.scrcpy.Server.createOptions(Server.java:71) at com.genymobile.scrcpy.Server.main(Server.java:132) at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method) at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:340) 

I use it in Ubuntu18.04

> scrcpy 1.9
dependencies:
SDL 2.0.8
libavcodec 57.107.100
libavformat 57.83.100
libavutil 55.78.100

adb version: 1.0.41
Thanks.

rom1v:
There are more parameters now:

    adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 0 8000000 false - false true

See:
https://github.com/Genymobile/scrcpy/blob/91ecb4f218f9a0f348ed8aee47919d6c2f4779a9/server/src/main/java/com/genymobile/scrcpy/Server.java#L76-L93

[...]

PR #920, opened 2019-11-09
Compare server and client version at the start of scrcpy

npes87184:
We found that some people wrong use older server with newer client. For example, #613 
To prevent this problem and let debug more easily. I add a patch to check this situation at the start of scrcpy.

It will show a warning when it finds version not matched.

### DIFF
--- a/app/src/server.c
+++ b/app/src/server.c
@@ -131,6 +131,7 @@ execute_server(struct server *server, const struct server_params *params) {
 #endif
         "/", // unused
         "com.genymobile.scrcpy.Server",
+        SCRCPY_VERSION,
         max_size_string,
         bit_rate_string,
         server->tunnel_forward ? "true" : "false",
--- a/server/src/main/java/com/genymobile/scrcpy/Server.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Server.java
@@ -67,29 +67,40 @@ public final class Server {
 
     @SuppressWarnings("checkstyle:MagicNumber")
     private static Options createOptions(String... args) {
-        if (args.length != 6) {
-            throw new IllegalArgumentException("Expecting 6 parameters");
+        if (args.length < 1) {
+            throw new IllegalArgumentException("At least client version should be specified");
+        }
+
+        String clientVersion = args[0];
+
+        if (!clientVersion.equals(BuildConfig.VERSION_NAME)) {
+            throw new java.lang.UnsupportedOperationException(
+                            "Difference version between client and server has not support yet");
+        }
+
+        if (args.length != 7) {
+            throw new IllegalArgumentException("Expecting 7 parameters");
         }
 
         Options options = new Options();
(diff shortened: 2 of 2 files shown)


rom1v:
If the client and server versions are different, the server will crash on argument parsing ("expected n arguments"), so retrieving the version on the socket is too late.

A possible solution is that the client sends in version as the first parameter when executing `app_process`, and the server checks it.

npes87184:
Like this [commit](https://github.com/npes87184/scrcpy/commit/6d0e61cdcf852b636f733a6a228722b243e28b71)?

rom1v:
> I create a commit at another branch in my repo, what about it?

Yes, looks good :+1:

rom1v:
Here is your commit with some minor changes: https://github.com/Genymobile/scrcpy/commit/f24d488ca2e7d133ebcff5fdb6ed07e602293514
OK to merge?

npes87184:
> Here is your commit with some minor changes: [f24d488](https://github.com/Genymobile/scrcpy/commit/f24d488ca2e7d133ebcff5fdb6ed07e602293514)
> OK to merge?

I am ok for this commit.

Thank you for your help.

aoy1:
> 如果客户端版本和服务器版本不同，则服务器可能在参数解析时更早崩溃（“期望的n个参数”），因此在套接字上获取版本太晚了。
> 
> 一种可能的解决方案是，客户端在执行时将其版本作为第一个参数发送`app_process`，而服务器会对其进行检查。



> 如果客户端版本和服务器版本不同，则服务器可能在参数解析时更早崩溃（“期望的n个参数”），因此在套接字上获取版本太晚了。
> 
> 一种可能的解决方案是，客户端在执行时将其版本作为第一个参数发送`app_process`，而服务器会对其进行检查helpe,

> If the client and server versions are different, the server may crash earlier on argument parsing ("expected n arguments"), so retrieving the version on the socket is too late.
> 
> A possible solution is that the client sends its version as the first parameter when executing `app_process`, and the server checks it.
AMD's CPU often flickers back

aoy1:
> 如果客户端版本和服务器版本不同，则服务器可能在参数解析时更早崩溃（“期望的n个参数”），因此在套接字上获取版本太晚了。
> 
> 一种可能的解决方案是，客户端在执行时将其版本作为第一个参数发送`app_process`，而服务器会对其进行检查。



> If the client and server versions are different, the server may crash earlier on argument parsing ("expected n arguments"), so retrieving the version on the socket is too late.
> 
> A possible solution is that the client sends its version as the first parameter when executing `app_process`, and the server checks it.

Help, amd's cpu often flashes back

aoy1:
Help, amd's cpu often flashes back(amd的cpu经常闪退）

rom1v:
Merged as b963a3b9d56f744cceba2e19cba3f9fef858c058.

PR #927, opened 2019-11-15
fix typo and link

senta2006:
I fixed a typo as an annotation of server.c and links of command.c , checkstyle.xml, gradle-wrapper.properties. Please check :)

### DIFF
--- a/app/src/server.c
+++ b/app/src/server.c
@@ -261,7 +261,7 @@ server_connect_to(struct server *server) {
 
         server->control_socket = net_accept(server->server_socket);
         if (server->control_socket == INVALID_SOCKET) {
-            // the video_socket will be clean up on destroy
+            // the video_socket will be cleaned up on destroy
             return false;
         }
 
--- a/app/src/sys/win/command.c
+++ b/app/src/sys/win/command.c
@@ -7,7 +7,7 @@
 static int
 build_cmd(char *cmd, size_t len, const char *const argv[]) {
     // Windows command-line parsing is WTF:
-    // <http://daviddeley.com/autohotkey/parameters/parameters.htm#WINPASS>
+    // <https://daviddeley.com/autohotkey/parameters/parameters.htm#WINPASS>
     // only make it work for this very specific program
     // (don't handle escaping nor quotes)
     size_t ret = xstrjoin(cmd, argv, ' ', len);
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -54,7 +54,7 @@ page at http://checkstyle.sourceforge.net/config.html -->
     <module name="SuppressWarningsHolder"/>
 
     <!-- Checks for imports                              -->
-    <!-- See http://checkstyle.sf.net/config_import.html -->
+    <!-- See http://checkstyle.sf.net/config_imports.html -->
     <module name="AvoidStarImport">
       <property name="allowStaticMemberImports" value="true" />
     </module>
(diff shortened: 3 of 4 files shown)


app/src/sys/win/command.c
```diff
@@ -7,7 +7,7 @@
 static int
 build_cmd(char *cmd, size_t len, const char *const argv[]) {
     // Windows command-line parsing is WTF:
-    // <http://daviddeley.com/autohotkey/parameters/parameters.htm#WINPASS>
+    // <https://daviddeley.com/autohotkey/parameters/parameters.htm#WINPASS>
```
rom1v: The page is not available over HTTPS (broken certificate + 404).

app/src/server.c
```diff
@@ -261,7 +261,7 @@ server_connect_to(struct server *server) {
 
         server->control_socket = net_accept(server->server_socket);
         if (server->control_socket == INVALID_SOCKET) {
-            // the video_socket will be clean up on destroy
+            // the video_socket will be cleaned up on destroy
```
rom1v: :+1:

config/checkstyle/checkstyle.xml
```diff
@@ -54,7 +54,7 @@ page at http://checkstyle.sourceforge.net/config.html -->
     <module name="SuppressWarningsHolder"/>
 
     <!-- Checks for imports                              -->
-    <!-- See http://checkstyle.sf.net/config_import.html -->
+    <!-- See http://checkstyle.sf.net/config_imports.html -->
```
rom1v: :+1:

config/checkstyle/checkstyle.xml
```diff
@@ -99,7 +99,7 @@ page at http://checkstyle.sourceforge.net/config.html -->
     <module name="WhitespaceAround" />
 
     <!-- Modifier Checks                                    -->
-    <!-- See http://checkstyle.sf.net/config_modifiers.html -->
+    <!-- See http://checkstyle.sf.net/config_modifier.html -->
```
rom1v: :+1:

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.5.1-all.zip
+distributionUrl=https://services.gradle.org/distributions/gradle-5.5.1-all.zip
```
rom1v: This is auto-generated, so `:` must probably be kept escaped.

senta2006: But it doesn’t work when I escape :  
Then, is it better not to fix this one ?-?

rom1v: > But it doesn’t work when I escape

What's your error message?

app/src/sys/win/command.c
senta2006: Okay, then it would be better keeping original one. Thanks!!

gradle/wrapper/gradle-wrapper.properties
senta2006: Actually it’s not error, it’ s not connected right away.

rom1v: What is "not connected"?

senta2006: Umm I mean, when I type https\://services.gradle.org/distributions/gradle-5.5.1-all.zip , it is just searched on the web. But when I type https://services.gradle.org/distributions/gradle-5.5.1-all.zip , it is downloaded directly.

rom1v: This file is parsed and interpreted by gradle, some characters are escaped probably to work with same separators accepted by the format, but in the end they are unescaped to retrieve the content.

senta2006: Then, is it better to useoriginal one?

rom1v: Yes.

rom1v:
Merged as 7aed5d5b60c7ae82f5ca2353c294076a2e0ffc2c.

Issue #893, opened 2019-10-29
request: translate README.md and FAQ.md into Korean.


seoyeonK:
Hello~ My name is Kim Seoyeon, a university student from Republic of Korea.

I hope you know that although a lot of Koreans use Android OS, it is not easy for us to 
 install this program because the README.md and FAQ.md is only written in English.

If you don't mind, I would like to contribute to your project, scrcpy by translating those two files that are related Installation, into Korean.

In fact, I am aware that you have declined a similar issue last year, but  I will do my best to update the translations monthly.

So, if you allow me to contribute to scrcpy by translating some manual to Korean. I will finish it as soon as I can.

I look forward to hearing from you. 
Thank you for your time.

rom1v:
Hi,

After all, we could add a translation, and add a note that only the README.md in english is expected to be up to date.

Thank you for your proposal.

cc @LimSH0731
ref #160

seoyeonK:
Thank you for replying! 
I will start the translation with a note about the monthly update.
Once again thank you for your suggestion!

seoyeonK:
Could you check the pull request I posted? #934

PR #928, opened 2019-11-16
translate README.md and FAQ.md into Korean

seoyeonK:
finished translating README.md and FAQ.md
#893


seoyeonK:
Added the notes about frequent update as requested at the issue

PR #965, opened 2019-11-29
Update BUILD.md to add ADB install

zadi15:


zadi15:
Some people may not have adb installed by default and may not read the part that says that adb is required, so this just covers that area

BUILD.md
```diff
@@ -45,7 +45,7 @@ sudo apt install ffmpeg libsdl2-2.0-0
 # client build dependencies
 sudo apt install gcc git pkg-config meson ninja-build \
                  libavcodec-dev libavformat-dev libavutil-dev \
-                 libsdl2-dev
+                 libsdl2-dev adb
```
rom1v: `adb` is a runtime dependency (not a "build dependency"), so it should be added few lines above instead :wink:

zadi15: Ah, my apologies, I've just ammended it

zadi15:
Moved `adb` into the runtime dependency area

rom1v:
Rebased, squashed and merged on `master`: 684e0abb74cf2f5291d2c52bbf758e85f917a4d6

Thank you :+1:

PR #988, opened 2019-12-06
Update README.md

thon-ju:
change Wireless step order, test on huawei phone


thon-ju:
Maybe this beacause 
1) turn on "Allow USB debugging", 
2) turn on "Allow ADB debugging in charge only mode"
3) turn off "Always prompt when connecting to USB".

then
1. adb tcpip 5555
2. adb connect <ip>:5555
3. unplug phone

Issue #397, opened 2019-01-16
Can it support another screenshot stream on android - secondary screen


chaooe:
1. Normaly we can capture screen 0 by 'adb shell screencap -p -d 0 /sdcard/sc.png', so if we have two or more display, we can use 'adb shell screencap -p -d ${displayid} /sdcard/sc.png', so can we have this supported?

chaooe:
Yes, i want to mirror a secondary screen. I am a QA to test Android Device, and we have some future running on another screen.. we can just use screenshot to capture another screen. It is slowly and unstable.

webian:
I'm another one that will find this feature precious to command Samsung Dex from my linux box without the need to switch mouse/keyboard/monitor.

I'm not an android dev but I found some little things that I hope can help:
1. if I run this apk https://github.com/mportuesisf/ScreenInfo starting from the smartphone display it returns me the smartphone display resolution, if I run it from the Dex monitor it gives me the resolution of the monitor. So, maybe in the code of ScreenInfo there are some hints on how to achieve this.
2. AnyDesk can reproduce Dex screen but only if started from the Dex monitor, if started from the smartphone monitor it reproduce the smartphone's one. Also, mouse click are always made on the smartphone even if started from the Dex monitor.

webian:
Is there maybe a way to make Android to think that the scrcpy server is started on the external (Dex) monitor?

webian:
Other info that maybe can help (see "Detect Samsung DeX"):
https://developer.samsung.com/samsung-dex/modify-optimizing

tchangeme:
请问，这个怎么实现呢，手机上有另外一个屏幕的时候

nakele:
@rom1v  btw, if the server was wrapped in an apk, and the apk was launched in Samsung Dex mode, it might mirror the second display.

chaooe:
@rom1v  I found minitouch have soultion to touch on second screen. before it use, we need to setup minitouch server on device.
`minitouch -d /dev/input/event0` /dev/input/event0 is the first screen id. and `minitouch -d /dev/input/event1` is the second display id
Does this scrcpy use minitouch to touch the screen?

rom1v:
No, it does not use minitouch. I guess it requires root access on the device to write to `/dev/input/event*`?

chaooe:
No, the /dev/input/event is the display id. I don't know where to configure the /dev/input/event*
[...]

PR #1177, opened 2020-02-25
Add feature: select display to mirror

peanutwolf:
Hello I've added a small feature that allows to configure scrcpy server with the optional parameter "display id" that should be mirrored.
It simply gets layerStack field from DisplayInfo for appropriate dispayId and passes it "setDisplayLayerStack" method
Hope it could be useful for others.


rom1v:
Hi,

Thank you for the PR.

Could you explain what it allows? Is it the feature request #397, or something different?

How do you get the list of available display ids?

Thank you

peanutwolf:
Yes, probably it could help with resolution of https://github.com/Genymobile/scrcpy/issues/397
In my case i have android device with several hardware displays. And i need to mirror them, but scrcpy allows to mirror only default buitin display
I do get information about displays by running `adb shell dumpsys display` and than I search for the 'Logical Displays' sections. It has some kind of that information:
image

mDisplayId=x is the corresponding parameter that can be passed to '$ ./scrcpy --display x' to mirror appropriate display.
Please note that display should have flag `FLAG_SUPPORTS_PROTECTED_BUFFERS`.

rom1v:
> Yes, probably it could help with resolution of #397

If you take a capture of display 1:
```
adb shell screencap -p -d 1 > /tmp/screenshot.png
```

Is it a screenshot of the display mirrored by `scrcpy --display 1`?

(and the same for any other display id)?

rom1v:
In developer options, I enabled "Simulate secondary displays":
secondary_display

It is listed by `adb shell dumpsys display`:
```
I  Display 1:
    mDisplayId=1
    mLayerStack=1
```

When I run `scrcpy --display 1`, the video has the correct dimensions (the one of the secondary display I enabled), but it always stays black :confused:

rom1v:
> but it always stays black

> Please note that display should have flag `FLAG_SUPPORTS_PROTECTED_BUFFERS`.

OK, that's the problem:
```
    mOverrideDisplayInfo=DisplayInfo{"Overlay #1", uniqueId "overlay:1", app 720 x 480, real 720 x 480, largest app 720 x 480, smallest app 720 x 480, mode 6, defaultMode 6, modes [{id=6, width=720, height=480, fps=60.0}], colorTransformId 0, defaultColorTransformId 0, supportedColorTransforms [], rotation 0, density 142 (142.0 x 142.0) dpi, layerStack 5, appVsyncOff 0, presDeadline 29333333, type OVERLAY, state ON, FLAG_PRESENTATION}
```

How to have this flag? :)

peanutwolf:
Unfortunately I don't know who is responsible for setting this flag. I can dig deeper and come back when i find some more information.
Do you or someone who started #397 have an opportunity  to check the feature on device with two hardware displays?

rom1v:
@peanutwolf https://github.com/Genymobile/scrcpy/issues/397#issuecomment-590898733

rom1v:
@peanutwolf There is a hidden `setDisplayId()` method on `KeyEvent` and `MotionEvent`.

peanutwolf:
Hi @rom1v yeah i've found this hidden method, but it is accessible only since API 29.
Moreover Google source doc explicitly says that several touch screens are unsupported until android 10 https://source.android.com/devices/tech/display/multi_display/input-routing
I can try to make a clause that if we are launched on API > 29 than MotionEvent.setDisplayId() is used otherwise no touch and input events are available for screen with FLAG_PRESENTATION.

rom1v:
This PR is still marked `[WIP]`. Is it ready?

> I can try to make a clause that if we are launched on API > 29 than MotionEvent.setDisplayId() is used 

:+1:

peanutwolf:
Hello. Yes, i think now its ready. I've added some meaningful prompts for display ids and restricted input events for presentation displays with API < 29

app/src/cli.c
```diff
@@ -369,8 +384,8 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
         {"window-y",              required_argument, NULL, OPT_WINDOW_Y},
         {"window-width",          required_argument, NULL, OPT_WINDOW_WIDTH},
         {"window-height",         required_argument, NULL, OPT_WINDOW_HEIGHT},
-        {"window-borderless",     no_argument,       NULL,
-                                                     OPT_WINDOW_BORDERLESS},
+        {"window-borderless",     no_argument,       NULL, OPT_WINDOW_BORDERLESS},
+        {"display",               required_argument, NULL, 'd'},
```
rom1v: This is an advanced option, I'd prefer to not assign a short option (`-d`).

server/src/main/java/com/genymobile/scrcpy/wrappers/DisplayManager.java
```diff
@@ -1,28 +1,54 @@
...
             // width and height already take the rotation into account
             int width = cls.getDeclaredField("logicalWidth").getInt(displayInfo);
             int height = cls.getDeclaredField("logicalHeight").getInt(displayInfo);
             int rotation = cls.getDeclaredField("rotation").getInt(displayInfo);
-            return new DisplayInfo(new Size(width, height), rotation);
+            int layerStack = cls.getDeclaredField("layerStack").getInt(displayInfo);
+            int flags = cls.getDeclaredField("flags").getInt(displayInfo);
+            return new DisplayInfo(displayId, new Size(width, height), rotation, layerStack, flags);
+        } catch (Exception e) {
+            if (e instanceof NullPointerException) suggestFix(displayId, getDisplayIds());
```
rom1v: Which variable is supposed to be `null`? Please use an explicit check `!= null` rather than a `catch NullPointerException` (or equivalent).

server/src/main/java/com/genymobile/scrcpy/wrappers/InputManager.java
```diff
@@ -37,4 +37,12 @@ public boolean injectInputEvent(InputEvent inputEvent, int mode) {
             return false;
         }
     }
+
+    public static void setDisplayId(InputEvent inputEvent, int displayId) {
+        try {
+            inputEvent.getClass().getMethod("setDisplayId", int.class).invoke(inputEvent, displayId);
+        } catch (InvocationTargetException | IllegalAccessException | NoSuchMethodException e) {
+            Ln.e("Could not invoke method", e);
```
rom1v: If this fails, this should prevent the event to be sent to the wrong display.

server/src/main/java/com/genymobile/scrcpy/DisplayInfo.java
```diff
@@ -16,5 +32,17 @@ public Size getSize() {
     public int getRotation() {
         return rotation;
     }
+
+    public int getLayerStack() {
+        return layerStack;
+    }
+
+    public boolean isPresentation() {
+        return (flags & FLAG_PRESENTATION) == FLAG_PRESENTATION;
```
rom1v: (nitpick)

```java
return (flags & FLAG_PRESENTATION) != 0;
```

(to avoid duplication)

server/src/main/java/com/genymobile/scrcpy/Device.java
```diff
@@ -22,11 +23,13 @@
 
     private final ServiceManager serviceManager = new ServiceManager();
 
+    private final DisplayInfo displayInfo;
```
rom1v: The `DisplayInfo` data is not valid after initialization (for example, the `rotation` field is obsolete after a rotation). Therefore, I think it should not be kept for the lifetime of the `Device` instance.

Maybe only its relevant fields (used in this class) could be computed initially and kept (for example `supportsInputEvents`, `layerStack`…)?

rom1v:
@peanutwolf https://github.com/Genymobile/scrcpy/issues/1288#issuecomment-613328125

I would like to change the condition to enable input events on secondary display to simply Android >= 10. Any objection?

IndustrialMuffin:
Digging deeper into the "Simulate Secondary Displays" thing rom1v mentioned, I checked logcat in adb to see how it was creating the display. It creates an OverlayDisplayAdapter
[https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/core/java/com/android/server/display/OverlayDisplayAdapter.java](url)
These adapters can also be created from adb shell with the settings command with the settings formatted as described in OverlayDisplayAdapter.java like this for example:
`settings put Global overlay_display_devices 1280x720/213,secure`

FLAG_SUPPORTS_PROTECTED_BUFFERS cannot be set for these overlay devices. I have been able to click to open apps from the overlay display adapter, and interact with them, but they have to be fully closed in the main display, otherwise they just pop up in the main display rather than the overlay display. I'm sure adb commands could be used to specify the chosen display, rather than relying on closing in the main display, and clicking in the overlay.

Something I tried doing that seems to be supported by OverlayDisplayAdapter.java is adding more than one display with the settings separated by a semicolon, like:
`settings put Global overlay_display_devices 640x480/130;640x480/150`
I only get the first device, and for the second device, I just get a notice in adb of:
`/system/bin/sh: 640x480/150: inaccessible or not found`
This seems to happen for any chosen size and density.

Using OverlayDisplayAdapter.java might be troublesome, and having the overlay displaying on top of the main display is annoying, but looking at the code for OverlayDisplayAdapter.java could reveal some useful ideas.

Issue #1048, opened 2020-01-01
number pad issues


KIMYONGUP:
When Num Lock is off, the keys are supposed to function to left, right, up or down. But they aren't working.

rom1v:
Fixed by #1188.

lollita:
It not work also for me now.

rom1v:
It's fixed on `dev` branch. You could take this build to test: https://github.com/Genymobile/scrcpy/issues/898#issuecomment-610932491

lollita:
Ty.
dev branch is accessible via Github?

rom1v:
The take these binaries: https://github.com/Genymobile/scrcpy/issues/898#issuecomment-610932491

lollita:
I test dev binaries but key pad not work again.

rom1v:
@KIMYONGUP Could you test, please?

PR #1188, opened 2020-02-29
Handle events on NumPad when NumLock is disabled

jaafarbi:
Hello, I tried to fix the issue #1048. 
I mapped the NumPad keys as follow when both SHIFT and NUM aren't set:
- SDLK_KP_6 --> AKEYCODE_DPAD_RIGHT
- SDLK_KP_4 --> AKEYCODE_DPAD_LEFT
- SDLK_KP_2 --> AKEYCODE_DPAD_DOWN
- SDLK_KP_8 --> AKEYCODE_DPAD_UP
- SDLK_KP_7 --> AKEYCODE_MOVE_HOME
- SDLK_KP_1 --> AKEYCODE_MOVE_END
- SDLK_KP_3 --> AKEYCODE_PAGE_DOWN
- SDLK_KP_9 --> AKEYCODE_PAGE_UP
- SDLK_KP_0 --> AKEYCODE_INSERT
- SDLK_KP_PERIOD --> AKEYCODE_FORWARD_DEL

### DIFF
--- a/app/src/event_converter.c
+++ b/app/src/event_converter.c
@@ -105,7 +105,7 @@ convert_keycode(SDL_Keycode from, enum android_keycode *to, uint16_t mod,
 
     if (!(mod & (KMOD_NUM | KMOD_SHIFT))) {
         // handling Numpad events when Num Lock is disabled
-        switch(from){
+        switch(from) {
             MAP(SDLK_KP_6,            AKEYCODE_DPAD_RIGHT);
             MAP(SDLK_KP_4,            AKEYCODE_DPAD_LEFT);
             MAP(SDLK_KP_2,            AKEYCODE_DPAD_DOWN);
--- a/app/src/event_converter.c
+++ b/app/src/event_converter.c
@@ -102,6 +102,23 @@ convert_keycode(SDL_Keycode from, enum android_keycode *to, uint16_t mod,
     if (mod & (KMOD_LALT | KMOD_RALT | KMOD_LGUI | KMOD_RGUI)) {
         return false;
     }
+
+    if (!(mod & (KMOD_NUM | KMOD_SHIFT))) {
+    	//	handling Numpad events when Num Lock is disabled
+        switch(from){
+			MAP(SDLK_KP_6,            AKEYCODE_DPAD_RIGHT);
+			MAP(SDLK_KP_4,            AKEYCODE_DPAD_LEFT);
+			MAP(SDLK_KP_2,            AKEYCODE_DPAD_DOWN);
+			MAP(SDLK_KP_8,            AKEYCODE_DPAD_UP);
+			MAP(SDLK_KP_7,            AKEYCODE_MOVE_HOME);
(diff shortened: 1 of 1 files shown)


npes87184:
Hi,

Why the condition including not shift?

Thank you.

app/src/event_converter.c
```diff
@@ -102,6 +102,23 @@ convert_keycode(SDL_Keycode from, enum android_keycode *to, uint16_t mod,
     if (mod & (KMOD_LALT | KMOD_RALT | KMOD_LGUI | KMOD_RGUI)) {
         return false;
     }
+
+    if (!(mod & (KMOD_NUM | KMOD_SHIFT))) {
+    	//	handling Numpad events when Num Lock is disabled
+        switch(from){
```
npes87184: I believe that there are some coding style issues.

jaafarbi:
Hi, I tried to mimic the behavior of the keyboard. If you disable the NumLock, and try the NumPad numbers with SHIFT, it'll work as the NumLock was enabled.

npes87184:
> Hi, I tried to mimic the behavior of the keyboard. If you disable the NumLock, and try the NumPad numbers with SHIFT, it'll work as the NumLock was enabled.
> 
> EDIT: At least, on Ubuntu system

I try it on my windows 10 ver 1909, the shift + numpad when num lock off, it works like shift + arrow, insert, home, etc.

app/src/event_converter.c
jaafarbi: I did some modifications regarding the coding style issues. Can you tell me what was exactly the issue?

Thank you

jaafarbi:
Should we go for Windows behavior? or the Ubuntu's one?

Thanks

app/src/event_converter.c
npes87184: Great! From now, I only notice that we need a space after switch.

npes87184:
> Should we go for Windows behavior? or the Ubuntu's one?
> 
> Thanks

The behavior problem should contact @rom1v .
In my opinion, we can use different behavior in windows and ubuntu.

Thank you.

rom1v:
Thank you for your contribution (and sorry for the delay).

I think checking SHIFT like you do produces the correct behavior: if shift is enabled, a text event will be generated, and managed as a text, so we must absolutely not send key events in that case.

IMO, the block should be moved few lines above so that it works even if `--prefer-text` is passed.

I squashed your commits, rebased on `dev`, and applied the changes I suggest: cd69eb4a4fecf8167208399def4ef536b59c9d22 (I also reordered the lines to sort by number, but it's just personal preference)

Please review and if it's ok I'll merge into `dev`.

jaafarbi:
Thank you for your answer. Since it is one of my first contributions, yes everything is fine for me.

rom1v:
> it is one of my first contributions

Congrats :tada:

Merged in `dev`. It will be in the next release.

PR #1224, opened 2020-03-19
Encoding profile option

tzah4748:
There are some decoders like [Broadway](https://github.com/emersion/Broadway) that can accept only H.264 Baseline profile streams.


rom1v:
When you create the PR, you must select the target branch `dev` :)

I just merged the 4th commit into `dev`: a0af402d96849f1573b1361523ac96772952e480

tzah4748:
Fixed (i think), sorry for making a mess ^^

PR #1226, opened 2020-03-19
Encoding profile option

tzah4748:
There are some decoders like [Broadway](https://github.com/emersion/Broadway) that can accept only H.264 Baseline profile streams.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -431,6 +431,7 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
         {"window-height",          required_argument, NULL, OPT_WINDOW_HEIGHT},
         {"window-borderless",      no_argument,       NULL,
                                                   OPT_WINDOW_BORDERLESS},
+        {"codec-profile",          required_argument, NULL, 'P'},
         {NULL,                     0,                 NULL, 0  },
     };
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -130,6 +130,12 @@ scrcpy_print_usage(const char *arg0) {
         "        Set the initial window width.\n"
         "        Default is 0 (automatic).\n"
         "\n"
+        "    -P, --codec-profile value\n"
+        "        Request specific encoding codec profile, see AVC profiles at:\n"
+        "        https://developer.android.com/reference/android/media/MediaCodecInfo.CodecProfileLevel\n"
+        "        for valid codec profile values.\n"
+        "        Default is 0 (automatic).\n"
+        "\n"
         "Shortcuts:\n"
         "\n"
         "    " CTRL_OR_CMD "+f\n"
@@ -368,6 +374,35 @@ parse_record_format(const char *optarg, enum recorder_format *format) {
     return false;
 }
 
+static bool
+parse_codec_profile(const char *optarg, uint32_t *codec_profile) {
+    long value;
+    // long may be 32 bits (it is the case on mingw), so do not use more than
+    // 31 bits (long is signed)
+    bool ok = parse_integer_arg(optarg, &value, true, 0, 0x7FFFFFFF, "codec-profile");
+    if (!ok) {
+        return false;
+    }
+
+    switch (value) {
+        case 0:         // Automatic
+        case 0x01:      // AVCProfileBaseline
+        case 0x02:      // AVCProfileMain
+        case 0x04:      // AVCProfileExtended
+        case 0x08:      // AVCProfileHigh
+        case 0x10:      // AVCProfileHigh10
+        case 0x20:      // AVCProfileHigh422
+        case 0x40:      // AVCProfileHigh444
+        case 0x10000:   // AVCProfileConstrainedBaseline
+        case 0x80000:   // AVCProfileConstrainedHigh
+            *codec_profile = (uint32_t) value;
+            return true;
+        default:
+            LOGE("Invalid codec-profile, Use -h for help");
+            return false;
+    }
+}
+
 static enum recorder_format
 guess_record_format(const char *filename) {
     size_t len = strlen(filename);
@@ -440,7 +475,7 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
     optind = 0; // reset to start from the first argument in tests
 
     int c;
-    while ((c = getopt_long(argc, argv, "b:c:fF:hm:nNp:r:s:StTv", long_options,
+    while ((c = getopt_long(argc, argv, "b:c:fF:hm:nNp:r:s:StTv:P:", long_options,
                             NULL)) != -1) {
         switch (c) {
             case 'b':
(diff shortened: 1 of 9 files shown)


app/src/cli.c
```diff
@@ -368,6 +374,35 @@ parse_record_format(const char *optarg, enum recorder_format *format) {
...
+        case 0:         // Automatic
+        case 0x01:      // AVCProfileBaseline
+        case 0x02:      // AVCProfileMain
+        case 0x04:      // AVCProfileExtended
+        case 0x08:      // AVCProfileHigh
+        case 0x10:      // AVCProfileHigh10
+        case 0x20:      // AVCProfileHigh422
+        case 0x40:      // AVCProfileHigh444
+        case 0x10000:   // AVCProfileConstrainedBaseline
+        case 0x80000:   // AVCProfileConstrainedHigh
```
rom1v: If only some specific values are authorized, maybe we should pass the strings to the command line parameter instead of an integer?

```
scrcpy --codec-profile=AVCProfileBaseline
```

(the server would still receive an integer)

What do you think?

app/src/cli.c
```diff
@@ -431,6 +466,7 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
         {"window-height",          required_argument, NULL, OPT_WINDOW_HEIGHT},
         {"window-borderless",      no_argument,       NULL,
                                                   OPT_WINDOW_BORDERLESS},
+        {"codec-profile",          required_argument, NULL, 'P'},
```
rom1v: This is an advanced option, I would prefer not to consume a "short" option (`-P`) for that.

app/src/cli.c
tzah4748: So you want to make a dictionary to map the strings to the integer values?
I just thought it makes the code messier but i guess its better for the users.
I agree.

app/src/cli.c
tzah4748: Yes that's totally fine, at first i had a bit of a problem understanding the code so i didn't know if it will work without a short version of the option, its ok to remove it -P and stay with --codec-profile only

app/src/cli.c
rom1v: I don't know which is better.

Maybe passing an integer (or hex value) is better in the end (so we can pass anything, that will also automatically work when later scrcpy support AV1 or other codecs).

But in that case, the value must be forwarded as-is, without checks.

rom1v: (oh but passing another value than the "valid" ones will cause problems because level is selected automatically)

Maybe we could just pass `--codec-profile=value` and `--codec-level=value`?

tzah4748:
Passing an invalid value will not start, and the level is selected only if
the profile is valid.
The possible levels of the codec are decided on runtime depending on the
screen size (i think, not sure) so i dont think its something you need to
pick anyway.

rom1v:
What if you don't set the level value at all on the `MediaFormat` (only the profile)?

tzah4748:
I didn't test but maybe it will be choosen automatically? The level option
is only from api 23 and not 21 so its something that was added after the
profile

rom1v:
If the level is chosen automatically when not set, then maybe an option `--codec-profile` accepting any integer and a `--codec-level` accepting any integer (ignored if API < 23) would be better?

tzah4748:
And in case that the profile is found and the level isn't it will be
choosen automatically, seems right.

rom1v:
Oh, maybe we could do it very generically: instead of adding codec options when people need them, we could just pass a string of `key=value`.

For example:

```
scrcpy --codec-options profile=1,level=2
```

The String would be passed as-is from the client to the server, and the server would parse it and set the media options.

There is an additional detail: the type depends on the key. Some expect an integer value, some others a long, some others a string.

Since almost all are integers, I suggest to use integer by default.
Otherwise, we could specify the type:
```
--codec-options somekey:string=value
```

What do you think?

tzah4748:
And what if a person doesn't want to specify a level ? Just the profile?
I personally liked it better when it's separated even though they are
related.

rom1v:
> And what if a person doesn't want to specify a level ? Just the profile?

Yes, the idea is to pass anything you want from [`MediaFormat`](https://developer.android.com/reference/android/media/MediaFormat).

```
scrcpy --codec-options key1:value1,key2=value2,...,keyn=valuen
```

Scrcpy would not even "understand" that it's a profile or a level, it just sets the value to the `MediaFormat` instance.

tzah4748:
I don't think i understood what you meant but if you see it fit and more
logical that way its fine :)

rom1v:
> I don't think i understood what you meant

Instead of `--codec-level`, `--codec-profile`, --codec-something` options, we could have only one option: `--codec-options` passing all key/values as a single string.

That way, one could pass any present and future codec options as they need.

rom1v:
@tzah4748 Are you ok for implementing such a `--codec-options` parameter instead of a specific option `--codec-profile`?

tzah4748:
Yes i'll do it, it will take some time I am quite busy with other work for
now.
I'll commit the changes once ill get to it :)

tzah4748:
@rom1v Opened a new PR with what we talked about.

rom1v:
Reopen (see https://github.com/Genymobile/scrcpy/pull/1325#pullrequestreview-404667615)

rom1v:
On my device (after fixing compilation errors), it does not work as is (of course, it works without `--codec-profile`):

```
$ scrcpy --codec-profile 2
[server] ERROR: Exception on thread Thread[main,5,main]
android.media.MediaCodec$CodecException: Error 0xfffffc0e
	at android.media.MediaCodec.native_configure(Native Method)
	at android.media.MediaCodec.configure(MediaCodec.java:1778)
	at com.genymobile.scrcpy.ScreenEncoder.configure(ScreenEncoder.java:191)
	at com.genymobile.scrcpy.ScreenEncoder.streamScreen(ScreenEncoder.java:76)
	at com.genymobile.scrcpy.Server.scrcpy(Server.java:35)
	at com.genymobile.scrcpy.Server.main(Server.java:170)
	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:251)
```

Computed max level is 65536 (`0x10000`, `AVCLevel52`). If I programmatically set level to 4096 (`0x1000`, `AVCLevel41`), it works.

Therefore, I think using the max available level for the given profile is not a good strategy :confused:

rom1v:
Oh I guess it requires changes from #1296?

tzah4748:
@rom1v Yes Rom, in https://github.com/Genymobile/scrcpy/pull/1296 I've added logic to choose the right level depending on the device screen size.
It should depend on the bit-rate as well but i ignored it for now (since i wasn't sure if its really needed) and it worked just fine.
In short, it has suggested level logic

rom1v:
OK, thank you for your answer. :+1:

If selecting a correct level based on the profile was straightforward, I would be in favor on doing it automatically.

However, it seems complicated, so it would add complex code to maintain, and to modify when scrcpy will support other codecs, for a very advanced option. Therefore, I think I prefer that the user provide both profile and level (like #1325).

tzah4748:
@rom1v are you gonna do it yourself or you are waiting a new pull request?

rom1v:
> are you gonna do it yourself or you are waiting a new pull request?

I have a working branch (`codec_options`), I might merge it before the next release.

tzah4748:
@rom1v That's great news!
Do you maybe have an ETA when is the next release?

rom1v:
Merged #1325 into `dev`.

PR #1296, opened 2020-04-17
Encoding options

tzah4748:
Introduced options arguments for the Android MediaCodec library.

There are some decoders like Broadway that can accept only H.264 Baseline profile streams.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/CodecOptions.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CodecOptions.java
@@ -1,0 +1,25 @@
+package com.genymobile.scrcpy;
+
+import java.util.HashMap;
+
+public class CodecOptions {
+    static final String PROFILE_OPTION = "profile";
+    static final String LEVEL_OPTION = "level";
+
+    private HashMap<String, String> options;
+
+    CodecOptions(HashMap<String, String> options) {
+        this.options = options;
+    }
+
+    Object parseValue(String profileOption) {
+        String value = options.get(profileOption);
+        switch (profileOption) {
+            case PROFILE_OPTION:
+            case LEVEL_OPTION:
+                return NumberUtils.tryParseInt(value);
+            default:
+                return null;
+        }
+    }
+}
--- a/server/src/main/java/com/genymobile/scrcpy/CodecOptions.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CodecOptions.java
@@ -1,10 +1,46 @@
 package com.genymobile.scrcpy;
 
+import android.media.MediaCodecInfo;
+import android.os.Build;
+
 import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 public class CodecOptions {
-    static final String PROFILE_OPTION = "profile";
-    static final String LEVEL_OPTION = "level";
+
+    public static final String PROFILE_OPTION = "profile";
+    public static final String LEVEL_OPTION = "level";
+
+    private static final LinkedHashMap<Integer, String> levelsTable = new LinkedHashMap<Integer, String>() {
+        {
+            // Adding all possible level and their properties
+            // 3rd property, bitrate was added but not sure if needed for now.
+            // Source: https://en.wikipedia.org/wiki/Advanced_Video_Coding#Levels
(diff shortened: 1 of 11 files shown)


server/src/main/java/com/genymobile/scrcpy/CodecOptions.java
```diff
@@ -0,0 +1,85 @@
+package com.genymobile.scrcpy;
+
+import android.media.MediaCodecInfo;
+import android.os.Build;
+
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class CodecOptions {
```
rom1v, line 10: (so this class would not be needed at all)

rom1v:
(I applied this quick-change from your PR: 125c5561e8f4e169fe52884e47fae1f3b9ffb9f7)

server/src/main/java/com/genymobile/scrcpy/CodecOptions.java
tzah4748: I don't think its a smart idea to let the user pass down any key=value pair he wants...
He can totally mess up and give random values and things that might even break the encoding and he won't be aware that he was giving something wrong.

I don't know about the other options but in case of profile and level it is suggested that if you provide a profile, you will also set a level, this is why when only profile (without level) is provided by the user, there is a calculation based on the screen size that should decide the level.

My goal was not to allow the user to pass down any key=value pair he wants, but to allow you or whoever wants to, to expand on the current CodecOptions and easily add new options that he can choose how they will be parsed and what will happen with them and where in the code.

rom1v: > I don't think its a smart idea to let the user pass down any key=value pair he wants...

It would allow to configure the codec generically, without having to build the app.


> He can totally mess up and give random values and things that might even break the encoding and he won't be aware that he was giving something wrong.

At least, we can warn if the provided key is not supported: [`containsKey()`](https://developer.android.com/reference/android/media/MediaFormat#containsKey(java.lang.String)).

User can already pass random values (a bitrate of 1Mbps for instance), which is "ignored" by the codec without warnings.



> I don't know about the other options but in case of profile and level it is suggested that if you provide a profile, you will also set a level, this is why when only profile (without level) is provided by the user, there is a calculation based on the screen size that should decide the level.

Is it done automatically if you don't provide the level? https://github.com/Genymobile/scrcpy/pull/1226#issuecomment-601596309

Btw, contrary to what I said [here](https://github.com/Genymobile/scrcpy/pull/1226#issuecomment-601612659), there is no need to specify the type in the string, it can be requested: [`getValueTypeForKey`](https://developer.android.com/reference/android/media/MediaFormat#getValueTypeForKey(java.lang.String))

tzah4748: OK, it will be done your way :)
Thanks for the good lesson.

tzah4748: @rom1v 
Just a quick question, what if the key is supported but value is not supported by the device?
If we are assuming the codec ignores bad values shouldn't the user know about it somehow?

tzah4748: getValueTypeForKey is API29+ ...
What do you suggest?
We can go back to what you said, assuming int unless specified key:type=value

tzah4748: @rom1v Ping

rom1v: > what if the key is supported but value is not supported by the device?

The device will just ignore it I guess. Like if for now you pass `--max-fps` but the encoder does not support it.

> getValueTypeForKey is API29+

:(

> We can go back to what you said, assuming int unless specified key:type=value

Hmm, yes... I have no better solution then :(

tzah4748: OK, so I've made the changes on our side, ill need to rebase my dev branch and make a new PR.
@rom1v 
Just to make things clear before i make a new PR, the expected value types are
Integer/int - String/str - Float/float - Long/long
Yes?

PR #1303, opened 2020-04-20
Added to Wireless section in README for hotspot/Bluetooth connections

flying-press:
- I added a new first step because I felt it wasn't obvious that the device should be first plugged *in order* begin using scrcpy wirelessly, especially to those who aren't very familiar with adb.

- scrcpy is possible over hotspot/Bluetooth. I thought to add that section for those who might not be sure if scrcpy in those situations could work at all.
--- Expanded third step so the reader knows that they have to be looking for the default gateway IP address and *not* for an IP address of a device host, and for letting even the inexperienced find that IP address.
- Added a small troubleshooting section since wireless might add some complexity, and to lessen incoming relevant issue posts.

The README might be more dense, but I thought it would help new users know they can use it via hotspot and Bluetooth, while giving them some guidance regardless of skill level.


flying-press:
> --- Expanded third step so the reader knows that they have to be looking for the default gateway IP address and not for an IP address of a device host, and for letting even the inexperienced find that IP address.

That third step is now the fifth step.

From my addition:
> If you see something like "ADB over network" in the developer options, you can toggle that option on to bypass having to physically connect your device directly to your computer. Step 5 can be done anytime during the process before reaching Step 6.

Finding out about this and taking advantage of it eliminates the (minor) annoyance of having to connect my device to a computer before being able to do use scrcpy. Everyone's situation is different, but mine uses this remedy to make life a bit easier. Figured others would like this as well.

rom1v:
Hi,

Sorry for the (huge) delay.

I like the improvements, but I think there is no need for a separate bluetooth section: whether the device is connected via bluetooth or wifi (hotspot or not), the steps are the same, and I think that the IP can always be retrieved by `adb shell ip route` (so there is no need to retrieve the bluetooth gateway on the computer).

I attempted to split your commits and rebase, and propose some changes (for example, link to the "wireless debugging" section for Android 11 directly). Here is a temporary (minimal) branch: [wireless-section](https://github.com/Genymobile/scrcpy/commits/wireless-section) ([rendered](https://github.com/Genymobile/scrcpy/tree/wireless-section#wireless))

Please review (and feel free to edit/rebase the commits to propose changes).

---

If I may, to make your proposals more quickly accepted, I'd suggest to propose one commit per change, especially when some of them are simple and uncontroversial.

The huge delay is my fault (and I have always postponed testing over bluetooth), but here is what can happen concretely: the reviewers (in that case, me) read the PR, some parts look ok, but they could not be merged immediately because they are entangled with other changes which need more time (to be tested) or more controversial. In that case, it's easy to say "ok, will do that later" again and again (this is what happened, I'm sorry).

Diamantaeres:
Hello!

Sorry, is it something possible in the end, to use Bluetooth (even in USB mode)?

Was considering the option to use Bluetooth as I can't use cable and I would prefer to leave the traffic outside of the router (because I'm using the same for other things, so the less data passes, the better it is).

Regards

rom1v:
> I attempted to split your commits and rebase, and propose some changes (for example, link to the "wireless debugging" section for Android 11 directly).

Oh, it wasn't merged yet. Just merged it into `dev`.

Diamantaeres:
Hello!

I want to stream at 854x480, would Bluetooth handle it?

Also, with merged you mean it's going to be available in the latest release only? I'm currently using a draft (issue 99, #2130).

So in conclusion I would like to use Bluetooth with the build 2130... I hope I'm not asking for too much! If possible, I would be grateful if you can help to get both the features combined together.

Regards

rom1v:
> I want to stream at 854x480, would Bluetooth handle it?

I don't know, you could test.

> Also, with merged you mean it's going to be available in the latest release only?

This PR is just a documentation PR, this has no impact on the features.

Scrcpy does nothing special for bluetooth: if you connect your computer to the device bluetooth network, it should work as usual.

Diamantaeres:
Hello,

Okay so I can simply connect my android device and PC via Bluetooth with scrcpy? How? Sorry but it's not clear for me.

You are talking about device Bluetooth network, is it sufficient to pair the Android device and the PC via Bluetooth or does it mean something more?

Thank you for your time

Diamantaeres:
Hello,

managed to do it but doesn't perform well at all, even with low resolution such as 854*480.

Regards

rom1v:
> managed to do it but doesn't perform well at all, even with low resolution such as 854*480.

Not surprised: https://github.com/Genymobile/scrcpy/pull/1303#issuecomment-1048040505

However, what matters here is the bitrate, not resolution:

```
scrcpy -m1024 -b1M
```

Diamantaeres:
Hello and thank you for the suggestion.

I tried that, doesn't make so much difference. I will simply give up this option.

Thank you anyway for helping me to try it.

Regards

PR #1307, opened 2020-04-21
[Request for tests] Refactor rendering and events position/scaling

### LINKED ISSUE #15: Touch input offset on secondary monitor

rom1v:
This patchset removes calls to `SDL_RenderSetLogicalSize()` and position and scale the content (and the events) manually.

This has several benefits:
 - it allows to avoid rounding inconsistencies between internal SDL logical size and the window size (causing 1 unexpected row/column of pixels blacks in the "optimal size")
 - it will make possible to draw visual feedback for #24 in the future
 - it can possibly help to fix #15

It's been a long time since I want to do this change, but it is quite "dangerous" as it impacts heavily the rendering and the event forwarding.

Therefore, I'm very interested in testing from users. Especially on macOS (but not only) (because I don't have any macOS to test), especially with HiDPI (but not only), especially with a secondary monitor (#15) (but not only).

Please also test clicks after window rotation (Ctrl+Left and Ctrl+Right, a new feature in the next release).

In theory, this is the last remaining change I would like to merge before a new release.

Thank you for your feedbacks.

EDIT: please test `logical_size.8` instead

### DIFF
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -7,33 +7,6 @@
 #include "util/lock.h"
 #include "util/log.h"
 
-// Convert window coordinates (as provided by SDL_GetMouseState() to renderer
-// coordinates (as provided in SDL mouse events)
-//
-// See my question:
-// <https://stackoverflow.com/questions/49111054/how-to-get-mouse-position-on-mouse-wheel-event>
-static void
-convert_to_renderer_coordinates(SDL_Renderer *renderer, int *x, int *y) {
-    SDL_Rect viewport;
-    float scale_x, scale_y;
-    SDL_RenderGetViewport(renderer, &viewport);
-    SDL_RenderGetScale(renderer, &scale_x, &scale_y);
-    *x = (int) (*x / scale_x) - viewport.x;
-    *y = (int) (*y / scale_y) - viewport.y;
-}
-
-static struct point
-get_mouse_point(struct screen *screen) {
-    int x;
-    int y;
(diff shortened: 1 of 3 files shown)


r4dixx:
Thank you for this! Works as expected for me here. Don't have an external monitor at home though.

esnaultdev:
With a setup similar to my initial post in #15

Launching scrcpy on the macOS (retina, hidpi) side, everything works well:
- high quality image
- no touch offset

However, the transition between screens breaks the display of scrcpy.

Moving from the retina screen to the normal screen results in a zoomed out UI
scrcpy_1_small

In that state, the touch input are actually correct based on the window size.
Resizing/rotating the window fixes the display.

Moving back the "fixed" window from the normal screen to the retina screen results in a zoomed in UI
scrcpy_2_small

Once again, rotating/resizing the window fixes the display.

-----

Using Cmd + Left / Cmd + Right to rotate the window works well on both screens! :+1:

rom1v:
There are still issues, and it's too risky for a release… I didn't merge it for now.

Thank you for your feedbacks :)

PR #1325, opened 2020-04-26
Codec options

tzah4748:
Introduced options arguments for the Android MediaCodec library.

**The original need:**
There are some decoders like Broadway that can accept only H.264 Baseline profile streams.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
+++ b/server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
@@ -25,15 +25,17 @@ public class ScreenEncoder implements Device.RotationListener {
     private final AtomicBoolean rotationChanged = new AtomicBoolean();
     private final ByteBuffer headerBuffer = ByteBuffer.allocate(12);
 
+    private String codecOptions;
     private int bitRate;
     private int maxFps;
     private boolean sendFrameMeta;
     private long ptsOrigin;
 
-    public ScreenEncoder(boolean sendFrameMeta, int bitRate, int maxFps) {
+    public ScreenEncoder(boolean sendFrameMeta, int bitRate, int maxFps, String codecOptions) {
         this.sendFrameMeta = sendFrameMeta;
         this.bitRate = bitRate;
         this.maxFps = maxFps;
+        this.codecOptions = codecOptions;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


tzah4748:
@rom1v ping :)

rom1v:
@tzah4748 Don't worry, I haven't forgotten, it the next task in my scrcpy TODO list :)

I just took a look, globally, it seems ok :+1:

When I have some time, I will look in detail and test. Sorry for the delay.

server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
```diff
@@ -155,6 +157,33 @@ private static MediaFormat createFormat(int bitRate, int maxFps, int iFrameInter
             // <https://github.com/Genymobile/scrcpy/issues/488#issuecomment-567321437>
             format.setFloat(KEY_MAX_FPS_TO_ENCODER, maxFps);
         }
+        // Adding additional options specified by the user
+        if(!"-".equals(codecOptions)) {
+            for (String pair : codecOptions.split(",")) {
+                String[] option = pair.split("=");
+                String key = option[0];
+                if(format.containsKey(key)) {
```
rom1v, line 165: Oh, contrary to what I said in a previous PR, this does not do what I expected (in fact it just test whether we already set the key, so in practice this is always false).

server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
```diff
@@ -155,6 +157,33 @@ private static MediaFormat createFormat(int bitRate, int maxFps, int iFrameInter
...
+                    if (option.length < 2) {
+                        Ln.w("No value specified for codec option - " + key);
+                        continue;
+                    }
+                    String[] valueAndType = option[1].split(":");
+                    String value = valueAndType[0];
+                    String type = valueAndType.length < 2 ? "" : valueAndType[1].toLowerCase();
+                    if (type.contains("str")) {
+                        format.setString(key, value);
```
rom1v, line 174: We could not pass a String containing a `,`.

rom1v:
Merged into `dev`: 080a4ee3654a9b7e96c8ffe37474b5c21c02852a

Thank you :+1:

kass507:
Can you explain what this function does?
I'm not an expert in codec and settings. but I would love to be able to give more quality to the window and send it to the OBS

rom1v:
@kass507 To increase the quality, just increase the bitrate:

    scrcpy -b20m

(This feature just passes raw parameters to MediaFormat, for advanced usage.)

kass507:
> 
> 
> @kass507 To increase the quality, just increase the bitrate:
> 
> ```
> scrcpy -b20m

This function only works when using low qualities. I have placed above 15 and no more than 12 passes.
And using the software that the factory team brings, it has reached 20 (Xiaomi Pocophone F1)

maddinmcfly:
somehow the intel-encoder crashes all the time on my panasonic FZ-A2 and i use the google encoder for that. the intel one is way better. How can i install the broadway encoder to my panasonic tablet to test it?

rom1v:
You can't "install" an encoder, it typically just implements an API to give instructions to an hardware component on your device.

PR #1368, opened 2020-05-08
Add a note about using hidden Android Framework APIs without reflection

S-trace:
Add mention about the android-hidden-api project and its modified android.jar
file, which allows writing the code which uses the hidden APIs directly in the
Android Studio with all the benefits like code completion and arguments
verifications without using any reflection tricks, which simplifies things a
lot and can even speed up resulting code by direct calls without the reflection.

However, it's not possible to access blacklisted APIs (starting from Android 9)
directly in this way, but there is a double-reflection trick to allow the
application to access those hidden APIs.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -64,9 +64,26 @@ another).
 They can be called using reflection though. The communication with hidden
 components is provided by [_wrappers_ classes][wrappers] and [aidl].
 
+Note: It's also sometimes possible to use hidden Android framework methods just by
+linking to the modified android.jar file available from the [android-hidden-api]
+project - just place android.jar which matches compileSdkVersion of the project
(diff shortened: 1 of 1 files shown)


rom1v:
Hi,

Thank you for the doc, that's interesting.

However, there is currently no performance problems calling private API methods by reflections occasionally (i.e. not 1000× per second) (and it would add some complexity to the build system).

IMO, this would be a good solution if all the methods would be guaranteed to exist, but it is not always the case, and some of them differ depending on the Android version, known at runtime. For example (but there are others):

https://github.com/Genymobile/scrcpy/blob/82446b6b0db284e9f9a125cd3e4785586f6845f3/server/src/main/java/com/genymobile/scrcpy/wrappers/PowerManager.java#L22-L24

Since scrcpy doesn't use this mechanism, I won't merge this documentation.

PR #1408, opened 2020-05-18
Rendering rework

### LINKED ISSUE #15: Touch input offset on secondary monitor

rom1v:
I reworked many things related to the window/rendering.

Firstly, I improved the maximized/fullscreen handling to avoid weird behaviors on some platforms (see commit messages 6dc113e67b4484fa402daeeb6e6ccacb021a9eb2 and eb189ceaab80433f682e847ab5d297edde3c52a6).

Then, (as a followup of #1307 that I closed because it was not ready) I reworked the rendering to position and scale the content (and the events) manually (i.e. without using the SDL "logical size" feature). (commit a6dcbf34bade4d8d71bdac15c04ff1cd12ce9450)

It has several benefits:
 - it allows to avoid rounding inconsistencies between internal SDL logical size and the window size (causing 1 unexpected row/column of pixels blacks in the "optimal size")
 - it will make possible to draw visual feedback at the expected size (for #24 for example) in the future
 - above all, it fixes #15 :tada:

I am now quite confident that it works correctly.

But since it impacts heavily the rendering and the event forwarding (which must NOT be broken at all), more tests from users are welcome:
 - if you can, also test with a secondary screen
 - check that clicks are correctly applied at the right position
   - even if you resize so that there are black borders
   - both in fullscreen or maximized or normal state
   - with or without content rotation (`Ctrl`+`←` / `Ctrl`+`→`)
 - try to maximize, then fullscreen (both `Ctrl`+`f` and "expand to fullscreen", which behaves differently), rotate device, then unfullscreen (check that the window is correctly restored in maximized state), then unmaximize (check that the window is correctly restored at the correct size).
 - use the app normally, you could find other edge cases :)

## Binaries

To simplify testing, here are binaries.

For windows users, take both `scrcpy.exe` and `scrcpy-server`, and replace them in your scrcpy v1.13 release.

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1408/scrcpy.exe)
  _SHA256: 7d61f31c8e33aa1b70a43b38b7c40c7e5289683a959e8426b34c3a488a385186_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1408/scrcpy-server)
  _SHA256: 6d124307e081b8bbccc478dcf82731aad3cc994e7502197c9bae3847d3169dc7_

### DIFF
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -7,33 +7,6 @@
 #include "util/lock.h"
 #include "util/log.h"
 
-// Convert window coordinates (as provided by SDL_GetMouseState() to renderer
-// coordinates (as provided in SDL mouse events)
-//
-// See my question:
-// <https://stackoverflow.com/questions/49111054/how-to-get-mouse-position-on-mouse-wheel-event>
-static void
-convert_to_renderer_coordinates(SDL_Renderer *renderer, int *x, int *y) {
-    SDL_Rect viewport;
-    float scale_x, scale_y;
-    SDL_RenderGetViewport(renderer, &viewport);
-    SDL_RenderGetScale(renderer, &scale_x, &scale_y);
-    *x = (int) (*x / scale_x) - viewport.x;
-    *y = (int) (*y / scale_y) - viewport.y;
-}
-
-static struct point
-get_mouse_point(struct screen *screen) {
-    int x;
-    int y;
-    SDL_GetMouseState(&x, &y);
-    convert_to_renderer_coordinates(screen->renderer, &x, &y);
-    return (struct point) {
-        .x = x,
-        .y = y,
-    };
-}
-
 static const int ACTION_DOWN = 1;
 static const int ACTION_UP = 1 << 1;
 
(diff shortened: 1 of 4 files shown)


carstenhag:
Tried out all cases which you described. In this one I wasn't sure what it was about though: 

>try to maximize, then fullscreen (both Ctrl+f and "expand to fullscreen", which behaves differently), rotate device, then unfullscreen (check that the window is correctly restored in maximized state), then unmaximize (check that the window is correctly restored at the correct size).

What did you mean with "expand to fullscreen"? I tried starting it with `-f`, I used the `cmd+f` shortcut, the yellow mac maximize button, all of those worked fine.

rom1v:
> Tried out all cases which you described. Seems like the issue is fixed on all of them!

:tada:

> What did you mean with "expand to fullscreen"?

https://support.apple.com/fr-fr/guide/mac-help/mchl9c21d2be/mac

carstenhag:
Yeah I did that, haha. All fine

rom1v:
It seems to work as expected (I also did more tests and received private feedbacks).

So I just merged it into `dev` :rocket:

PR #1418, opened 2020-05-22
please surpport send input event to extented display on smartisan nut pro3

gfzhy:
adb shell input --ext-display tap/swipe... This commend can sending event to extented display,
so it is very useful for scrcpy --display 2 .
it also can solve the issue "Input events are not supported for secondary displays before Android 10".
Because "adb shell input --ext-display keyevent " can send input to secondary displays on android 9.
it is very useful  for  samrtisan   tnt.
Thank you very much.


gfzhy:
@rom1v   Thank you very much.

Issue #1159, opened 2020-02-18
[Work in progress] Adding an option --serve tcp:localhost:port


Darkroll76:
Hi everyone,

Following this post: #1073 and suggested by @rom1v, I started to work on an option `--serve`. This option would allow forwarding the video stream to any clients connected. It could be a C# Unity App, Java App or C# UWP App. 

The command would be `--serve tcp:localhost:1234`

It could be a great solution for people like me, who wants to display the video stream inside a desktop application without redeveloping and rebuilding the scrcpy client in the desired language. 

I create this post to give some updates and request some help in C development. 

Following these indications :
> Basically, when you receive a packet, it is decoded + recorded (if necessary):
> https://github.com/Genymobile/scrcpy/blob/39356602ed472cc3f533e36ae04a110b247c29e0/app/src/stream.c#L81-L97
>
> 
> We can, in addition, write it to a socket.
> 
> But in fact, it could even be written earlier (without even parsing it):
> 
> https://github.com/Genymobile/scrcpy/blob/39356602ed472cc3f533e36ae04a110b247c29e0/app/src/stream.c#L229
> 
> _Originally posted by @rom1v in https://github.com/Genymobile/scrcpy/issues/1073#issuecomment-577636642_

So, we create another socket (for testing, we're using localhost and port 27015) 
In `stream.c -> run_stream(void *data)`
```
//open another socket on port 27015
socket_t socket;
socket = net_listen(IPV4_LOCALHOST, 27015, 1);

// We must only pass complete frames to av_parser_parse2()!
// It's more complicated, but this allows to reduce the latency by 1 frame!
stream->parser->flags |= PARSER_FLAG_COMPLETE_FRAMES;

int cpt = 0;

for (;;) {
	AVPacket packet;
	bool ok = stream_recv_packet(stream, &packet);

	if (!ok) {
		// end of stream
		break;
	}

	ok = stream_push_packet(stream, &packet);

	if (socket == INVALID_SOCKET) {
		LOGI("Error");
		return false;
	}
	else {
		LOGI("socket available");
	}

	cpt++;
	net_send(socket, packet.buf, packet.size);
	printf("Send : %d", cpt);
	
        av_packet_unref(&packet);
	if (!ok) {
		// cannot process packet (error already logged)
		break;
	}
}
```

We 're not sure that we are sending the right thing yet but this is our first step. 
Everyone who wants to help is welcomed 😃  !

Thanks,

Darkroll76:
[...]

Issue #1073, opened 2020-01-10
Is it possible to redirect the stream to another application ? 


Darkroll76:
Hi,

Currently, _scrpcy_ opens independent windows to render video stream but I'm trying to integrate it inside an application. Is there any way to do it? By redirecting the stream? 

It could be a Unity App or Java App or C# UWP App but I would like to be able to share android-based VR headset screen (Oculus Go, Quest, PICO, etc) inside a desktop application. 

Thanks for your time,
Killian

iiasceri:
![72148353-7c333e00-3398-11ea-8db7-cac9a1826762](https://user-images.githubusercontent.com/34134411/72155067-249dce00-33aa-11ea-98f2-02becd949890.png)
https://github.com/barry-ran/QtScrcpy

Darkroll76:
> ![72148353-7c333e00-3398-11ea-8db7-cac9a1826762](https://user-images.githubusercontent.com/34134411/72155067-249dce00-33aa-11ea-98f2-02becd949890.png)
> https://github.com/barry-ran/QtScrcpy

Thanks, I will try if I could find what I need with this project. For now, it just seems to offer the same features.

rom1v:
> Currently, scrpcy opens independent windows to render video stream but I'm trying to integrate it inside an application. Is there any way to do it?

The proper way would be to split scrcpy in two parts: libscrcpy and the UI (https://github.com/Genymobile/scrcpy/issues/707#issuecomment-520941669). But this is not done.

Another way could be to use only the .jar on your side, and write your client which mimic the protocol used between the client and the server (but this protocol should be considered [private](https://github.com/Genymobile/scrcpy/issues/673#issuecomment-516360374), i.e. it may change at any time).

Alternatively, you could fork current master and replace some parts…

LBensman:
That's pretty much what I did - just take the Java portion that already sends back H.264 stream (with initial custom header), and just handle H.264 stream inside your desktop application (or where ever you want to consume it).  Basically reuse the server jar, but dump viewer in favor of your own stream handling and visualization.

On a separate note, @rom1v , did you notice from screenshots above that QtScrcpy seems to be supporting audio?  I know you tried USBAudio (I'm looking into it now myself), but quick look (really, probably not even 2 min spent on it) at QtScrcpy source, I couldn't see much for audio handling.  Have you looked into it and see if it something more solid than USBAudio?

rom1v:
[...]

Issue #684, opened 2019-08-02
Is there a way to capture only a frame or screenshot rather than a video stream?


toddsierens:
I am looking for a minimalized version of what the video stream part of scrcpy does. 

It is apparently not so straightforward to take a screenshot of an android device outside of the android context. I imagine similar frustrations occurred while learning how to attain the video stream. Is there a way to only attain one frame or a screenshot at the time of a request?

hasansidd:
does ascreenshot not accomplish what you need?

rom1v:
> Is there a way to only attain one frame or a screenshot at the time of a request?

It requires some investigations to do it in Java without an Android context (I have no time to do this now).

Sometimes, the implementation of an existing command may help:

     adb exec-out screencap -p > file.png

But in this case, it's implemented in C++: https://github.com/aosp-mirror/platform_frameworks_base/blob/master/cmds/screencap/screencap.cpp

toddsierens:
> does ascreenshot not accomplish what you need?

This is what I am wanting, I am not sure what you are suggesting.

> ` adb exec-out screencap -p > file.png`

This is similar to what I was already using, but with the stdout captured by the process and handled without writing to disk. This method, however does take about 1-2 seconds to work.

I did find a method that allows me to access a screenshot in about 0.07s. Which is much better, giving me about 14 fps maximum. Not ideal, but much better than using the screencap shell command.

`SurfaceControl.screenshot` method gives a `HARDWARE.Bitmap`, which unfortunately must be copied before it is read. Once copied, it can be converted into a bytebuffer and passed along the socket.

hasansidd:
> This is what I am wanting, I am not sure what you are suggesting.

Sorry, I thought I was in the wrong repo. I know that https://github.com/IntergalacticPenguin/mobile-toolkit has a feature to take screenshots that works pretty well for me. But I believe they are using screencap, so if that does not work for your use case then it may not work for you.

toddsierens:
I looked into this a little bit more, and found that `SurfaceControl` invokes a native method, and this native method essentially executes `"/system/bin/screencap"`. Now I wanted to see if a java program executing `"/system/bin/screencap"` as a subprocess would give any speed-up.

Something of this sort:
```
            try{
                Process process = Runtime.getRuntime().exec("/system/bin/screencap");
                BufferedReader ob = new BufferedReader(
                        new InputStreamReader(process.getInputStream()));
                int currentChar;
[...]

PR #1419, opened 2020-05-22
Adding an option --serve tcp:localhost:port 

Darkroll76:
Hi everyone,

Following this posts: #1073 and #1159  and suggested by  @rom1v , I worked on an option `--serve`. This option would allow forwarding the video stream to any client connected. It could be a C# Unity App, Java App or C# UWP App or whatever.

The command is like below :
```
--serve tcp:[localhost|IP]:[port]
# For example
--serve tcp:localhost:27115 
```

It could be a great solution for people like me, who wants to display the video stream inside a desktop application without redeveloping and rebuilding the scrcpy client in the desired language.


rom1v:
Thank you for the PR.

However, it seems there's a rebase issue, your commit https://github.com/Genymobile/scrcpy/pull/1419/commits/17947e175c2fc79c8a6e1746d9e684af852f4155 contains a lot of code which already exists on `master` (and which is not related to the feature you implement).

overvox:
Hey @Darkroll76,
first of all big thanks for your pull request and added features. (My usecase is similar to yours, as I want to feed the stream in realtime to Unity) I checked out your latest commit and managed to build it for windows. In general the serving feature works, however I noticed a very noticable lag as well as a lot of artifacts when testing the tcp stream in the vlc player (all on the same host machine).

I tested multiple downgrades in terms of bitrate and resolution without any improvements.
Is this an issue with vlc or just the nature of the tcp stream?

Darkroll76:
Hi @overvox ,

Thanks! Apparently it's due to VLC buffering. See this [post](https://github.com/Genymobile/scrcpy/issues/1159#issuecomment-589282647) 

Since I'm working on a big project, I didn't find the time to remove all the code which already exists to have a clean commit.

Darkroll76:
@overvox I understood that I was doing the wrong thing. When you use the --serve command, it waits for the client to connect to start the stream. But during this time, scrcpy is already receiving packets from the scrcpy server and apparently, it keeps the frames until it can send them. I used the same behaviour as the decoder and the recorder, but I should rather look for the same thing as the display window.

This is why there is a time lag between the action we perform on our device and the frame received and displayed. Because it sends the frames it kept since the start of scrcpy.

hgedik:
hi @Darkroll76 I have a question. Will be able to use it in webrtc after --serve update? 
thanks

rom1v:
> I'm trying to remove the time lag between the action we perform on our device and the frame displayed. Is ScrCpy keep each frame since the server is started? How could I clean the pending frames?

IMO, the server should not be started until a client connects to the socket.

Maciejszuchta:
Any updates on this feature. Rly looking forward to it.

Darkroll76:
Hi @Maciejszuchta, I'm trying to fix the rebase issue on commit 17947e1 like @rom1v said :

> it contains a lot of code which already exists on master (and which is not related to the feature you implement).

But I'm a bit lost in how to do it properly without breaking anything

andreluisos:
This feature would be great.

PranshuTG:
where is this serve repo?
i wanna use it

intergratedsolutions (commented):
Not really sure

DreamInjection:
Hi, Any update on this serve repo? I also would like to use the --serve command but I somehow can't even find the branch on SourceTree. Am I just noob?

rom1v:
Hi,

A quick follow up.

Despite I initially suggested this feature, I'm finally not convinced a `--serve` argument is a good idea now.

Firstly, it adds complexity:
 - it impacts scrcpy starting
 - it is limited to video, and could possibly constrain what can be done about audio in the future (if there is a real solution, which is not the case for now)

But above all, the same feature (receive the video stream for another app) can now be achieved differently (it was not the case when this PR was created):
 1. on Linux, via [V4L2](https://github.com/Genymobile/scrcpy#v4l2loopback)
 2. using scrcpy-server directly

For using `scrcpy-server`, recent changes (in particular 04e5537f8ce21dbae365dcd980efda911188c945 and cabcbc2b151255b49a442a1c43e7db44432eb58e) facilitates reusing the server for receiving the raw video.

Here is a sample script which starts the server:

```bash
adb forward tcp:1234 localabstract:scrcpy
adb push scrcpy-server /data/local/tmp/scrcpy-server.jar
adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar \
    app_process / com.genymobile.scrcpy.Server 1.21 \
    send_frame_meta=false tunnel_forward=true control=false max_size=1920
```

From another terminal:

```bash
# skip the 69 first bytes (1 zero-byte, 64-bytes device name string, and 4 bytes for the device size)
nc localhost 1234 | tail -c+70 | vlc --demux h264 -
```

This will not be real time at all because VLC will add buffering, but this is to illustrate that it is possible to get the raw H.264 stream.
Instead of running `nc` and `vlc`, you probably use your own client.

Moreover, that way, the client can decide if it wants to receive the "frame meta" in front of each Annex B H.264 packet or not, to avoid an additional 1-frame latency (see #646).

rom1v:
See #2971.

Maciejszuchta:
Hello, is there a possibility to dynamically take a screenshot (encode it to png or jpg) of a live stream from the device? What I need is to take a screenshot of a device screen as fast as it is possible.

rom1v:
> Hello, is there a possibility to dynamically take a screenshot (encode it to png or jpg)

Currently, no: #684 #2040.

But you could [record the video stream](https://github.com/Genymobile/scrcpy#recording) and extract frames later:

```
scrcpy --record file.mp4
```

DreamInjection:
> For using `scrcpy-server`, recent changes (in particular [04e5537](https://github.com/Genymobile/scrcpy/commit/04e5537f8ce21dbae365dcd980efda911188c945) and [cabcbc2](https://github.com/Genymobile/scrcpy/commit/cabcbc2b151255b49a442a1c43e7db44432eb58e)) facilitate reusing the server for receiving the raw video.
> 
> Here is a sample script which starts the server:
> 
> ```shell
> adb forward tcp:1234 localabstract:scrcpy
> adb push scrcpy-server /data/local/tmp/scrcpy-server.jar
> adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar \
>     app_process / com.genymobile.scrcpy.Server 1.21 \
>     send_frame_meta=false tunnel_forward=true control=false max_size=1920
> ```

I am unable to run these commands. It seems that my phone needs to be rooted in order for me to be able to push the scrcpy server to the data folder. I found a workaround by drag-and-dropping my file to a folder I had access to and changing the path in the third command. However, that command fails as well, saying that app_process cannot be found, even though it is in the folder it should be in, and included in the PATH as well. I'm wondering if it has to do with the phone not being rooted again? Or does the scrcpy-server absolutely have to be in the data/local/temp folder?

rom1v:
> I am unable to run these commands.

What is the exact output for each one?

DreamInjection:
For the adb push command, I was getting something like push: permission denied. I tried again just now and can't reproduce the error, the push command now somehow works even though I still don't have root access. For the adb shell command, I realized I hadn't deleted the \ from the command, which gave me /system/bin/sh: app_process: inaccessible or not found. But now that this 3rd command runs, it simply outputs "Aborted".

I'm sorry for bothering you with this, this is something I'm fairly new at and I don't quite understand everything.

varundtsfi:
> For the adb push command, I was getting something like push: permission denied. I tried again just now and can't reproduce the error, the push command now somehow works even though I still don't have root access. For the adb shell command, I realized I hadn't deleted the \ from the command, which gave me /system/bin/sh: app_process: inaccessible or not found. But now that this 3rd command runs, it simply outputs "Aborted".
> 
> I'm sorry for bothering you with this, this is something I'm fairly new at and I don't quite understand everything.

Hi @Darkroll76 
I guess you are using Xiaomi or Realme device please cofirm.

Issue #786, opened 2019-09-05
Is any quickly way to pass utf8 text from computer to android ?


angelmic:
Hi

I need to do hundred times paste utf8 text to my android device. Is any way to speed up current way: to `Ctrl` + `Shift` + `v` to put text into android's clipboard then do paste.

Plandercozeron:
Try sending 'Paste' event with ADB (Android 7.0+)
`adb shell input keyevent 279`

or just send whole text with
`adb shell input "your_text"`

rom1v:
Oh, that's great a great trick! Thank you :+1:

```
adb shell input keyevent PASTE
```

I will probably use it to improve non-ascii support.

rom1v:
Ref #1426

rom1v:
Fixed by fc1dec027093ee8a2746a212db0c72a9d96d39a4.

Issue #612, opened 2019-06-23
Thank you for this project


barry-ran:
Amazing project. Thank you!
I reimplemented scrcpy with different technology stacks (Qt + Opengl) and added something:
- an interface for interaction.
- connecting multiple devices to the same application

Of course I want to share it. [QtScrcpy](https://github.com/barry-ran/QtScrcpy)
![win](https://raw.githubusercontent.com/barry-ran/QtScrcpy/master/screenshot/win.png)

MrTeeXD:
Now this is a something I want to see in scrcpy Plz rom1v. Well done

MrTeeXD:
@rom1v Can we have this implemented into scrcpy please, Its such a great idea.

rom1v:
Cool UI :+1: Great job!

This is something missing in _scrcpy_, especially for users not comfortable with the command-line.

---

You "converted" every "class" into in C++/Qt (the signal/slots mechanism in indeed very appropriate), but now you if you want to keep it up to date with scrcpy, you will have to reimplement the changes introduced by every commit.

I would like (for a long time) to separate the "core" and the UI, to provide something like a _libscrcpy_. This would allow to implement different UI (I want to keep SDL for a lightweight version, but Qt is better if we want widgets), or even implement custom apps which could use the core API to receive video frames and control the device (for automation or whatever).

The core would remain in C, but a C++ app could use it directly (and use some RAII wrappers for convenience). That way, you could just write the UI part, and keep up to date with scrcpy easily.

I will try to think more about such an API.

arley1708:
Is it available in english sir?

barry-ran:
> Cool UI 👍 Great job!
> 
> This is something missing in _scrcpy_, especially for users not comfortable with the command-line.
> 
> You "converted" every "class" into in C++/Qt (the signal/slots mechanism in indeed very appropriate), but now you if you want to keep it up to date with scrcpy, you will have to reimplement the changes introduced by every commit.
> 
> I would like (for a long time) to separate the "core" and the UI, to provide something like a _libscrcpy_. This would allow to implement different UI (I want to keep SDL for a lightweight version, but Qt is better if we want widgets), or even implement custom apps which could use the core API to receive video frames and control the device (for automation or whatever).
> 
> The core would remain in C, but a C++ app could use it directly (and use some RAII wrappers for convenience). That way, you could just write the UI part, and keep up to date with scrcpy easily.
> 
> I will try to think more about such an API.


[...]

Issue #37, opened 2018-03-11
Non english keyboard layout


elig0n:
I have hebrew keyboard layout on both my computer and my android keyboard that is in use but it wouldn't let me pass the letters.  Any chance to fix this?

rom1v:
> Any change to fix this?

[Probably not](https://blog.rom1v.com/2018/03/introducing-scrcpy/#handling-accented-characters).

Except if someone find a technical solution ;)

MiracleXYZ:
I used to have the same issue but it disappeared through some "practice":
1. Turn to AOSP and type some numbers like `1234` (that works fine).
2. Type some letters. If it doesn't work, press `shift` key several times. (You may say a bunch of letters "crowding in".
3. Turn to GBoard or other keyboards. It will work fine if you are lucky :)

wuairc:
One possible technical solution is to write a customized input method manager. [Vysor](https://chrome.google.com/webstore/detail/vysor/gidgenkbbabolejbgbpnhbimgjbffefm)(an scrcpy alternative) does this.
The input method manager can be installed and enabled by adb, without user interaction.

AndroidDeveloperLB:
+1 on this. Wanted to write in Hebrew too, but failed.
Are you sure it's impossible to use adb to pass unicode keys ?

vshymanskyy:
One of the possible options could be to create a custom keyboard, like `NULL keyboard` that I described here: https://github.com/Genymobile/scrcpy/issues/181, but with an ability to Input and Copy/Paste unicode. Probably not the easiest option, but at least something if there's not system API for such things.

AndroidDeveloperLB:
I would prefer without this workaround...

giorgiobeggiora:
+1

giorgiobeggiora:
I strongly hope this will be fixed soon

livin2:
https://remoboard.app/
example of this app may be help. 
pass unicode via an wireless server independent from adb server.

AndroidDeveloperLB:
@livin2 This means I have to change my default keyboard, which means I might have less features than what I want to have.
If an app is required, I would even install one that requires root, but having less features and having to switch keyboards as an alternative... This is just a hassle...

Is there maybe a way to do it using "monkeyrunner "? Someone wrote about this here:
https://stackoverflow.com/a/20615725/878126

livin2:
[...]

PR #1426, opened 2020-05-25
Inject UTF-8 text

rom1v:
Since Android 7, we could use the PASTE keycode to inject UTF-8 text, instead of injecting every ASCII char sequentially. (ref: https://github.com/Genymobile/scrcpy/issues/786#issuecomment-528308353)

So I just implemented it.

If the device runs Android >= 7:
 - backup the clipboard
 - set the text to inject into the clipboard
 - inject the PASTE key code
 - restore the clipboard

It is very fast and allows to inject UTF-8 text, which is awesome for non-ASCII languages like Chinese.

Please test and report any problem.

Thank you for your feedbacks.

---

Here are binaries so that you can test even if you don't know how to build:

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1426/1/scrcpy.exe)
  _SHA256: 4481d854630aaba9f31e4f3a19b46c4843c1918da03e1a486089bae3a3ebb09a_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1426/1/scrcpy-server)
  _SHA256: 64cd4d4277040430c48d35677ad8472a6315337651fb9cfc5f62cf5abf2021b7_

_(on Windows, replace both files in the v1.13 release; on other platforms, use `scrcpy-server` and follow [§Prebuilt server](https://github.com/Genymobile/scrcpy/blob/master/BUILD.md#prebuilt-server))_

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Controller.java
@@ -1,10 +1,7 @@
 package com.genymobile.scrcpy;
 
-import com.genymobile.scrcpy.wrappers.InputManager;
-
 import android.os.SystemClock;
 import android.view.InputDevice;
-import android.view.InputEvent;
 import android.view.KeyCharacterMap;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
@@ -50,7 +47,7 @@ public class Controller {
     public void control() throws IOException {
         // on start, power on the device
         if (!device.isScreenOn()) {
-            injectKeycode(KeyEvent.KEYCODE_POWER);
+            device.injectKeycode(KeyEvent.KEYCODE_POWER);
 
             // dirty hack
             // After POWER is injected, the device is powered on asynchronously.
@@ -133,7 +130,7 @@ public class Controller {
     }
 
     private boolean injectKeycode(int action, int keycode, int metaState) {
-        return injectKeyEvent(action, keycode, 0, metaState);
+        return device.injectKeyEvent(action, keycode, 0, metaState);
     }
 
     private boolean injectChar(char c) {
@@ -144,7 +141,7 @@ public class Controller {
             return false;
         }
         for (KeyEvent event : events) {
-            if (!injectEvent(event)) {
+            if (!device.injectEvent(event)) {
                 return false;
             }
         }
@@ -200,7 +197,7 @@ public class Controller {
         MotionEvent event = MotionEvent
                 .obtain(lastTouchDown, now, action, pointerCount, pointerProperties, pointerCoords, 0, buttons, 1f, 1f, DEVICE_ID_VIRTUAL, 0,
                         InputDevice.SOURCE_TOUCHSCREEN, 0);
-        return injectEvent(event);
+        return device.injectEvent(event);
     }
 
     private boolean injectScroll(Position position, int hScroll, int vScroll) {
@@ -223,26 +220,11 @@ public class Controller {
         MotionEvent event = MotionEvent
                 .obtain(lastTouchDown, now, MotionEvent.ACTION_SCROLL, 1, pointerProperties, pointerCoords, 0, 0, 1f, 1f, DEVICE_ID_VIRTUAL, 0,
                         InputDevice.SOURCE_TOUCHSCREEN, 0);
-        return injectEvent(event);
-    }
-
-    private boolean injectKeyEvent(int action, int keyCode, int repeat, int metaState) {
-        long now = SystemClock.uptimeMillis();
-        KeyEvent event = new KeyEvent(now, now, action, keyCode, repeat, metaState, KeyCharacterMap.VIRTUAL_KEYBOARD, 0, 0,
-                InputDevice.SOURCE_KEYBOARD);
-        return injectEvent(event);
-    }
-
-    private boolean injectKeycode(int keyCode) {
-        return injectKeyEvent(KeyEvent.ACTION_DOWN, keyCode, 0, 0) && injectKeyEvent(KeyEvent.ACTION_UP, keyCode, 0, 0);
-    }
-
-    private boolean injectEvent(InputEvent event) {
-        return device.injectInputEvent(event, InputManager.INJECT_INPUT_EVENT_MODE_ASYNC);
+        return device.injectEvent(event);
     }
 
     private boolean pressBackOrTurnScreenOn() {
         int keycode = device.isScreenOn() ? KeyEvent.KEYCODE_BACK : KeyEvent.KEYCODE_POWER;
-        return injectKeycode(keycode);
+        return device.injectKeycode(keycode);
     }
 }
--- a/server/src/main/java/com/genymobile/scrcpy/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Controller.java
@@ -149,6 +149,10 @@ public class Controller {
     }
 
     private int injectText(String text) {
+        if (device.injectTextPaste(text)) {
+            // The best method (fastest and UTF-8) worked!
+            return text.length();
+        }
         int successCount = 0;
         for (char c : text.toCharArray()) {
             if (!injectChar(c)) {
--- a/server/src/main/java/com/genymobile/scrcpy/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
@@ -10,8 +10,12 @@ import android.content.IOnPrimaryClipChangedListener;
 import android.graphics.Rect;
 import android.os.Build;
 import android.os.IBinder;
+import android.os.SystemClock;
 import android.view.IRotationWatcher;
+import android.view.InputDevice;
 import android.view.InputEvent;
+import android.view.KeyCharacterMap;
+import android.view.KeyEvent;
 
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -147,7 +151,7 @@ public final class Device {
         return supportsInputEvents;
     }
 
-    public boolean injectInputEvent(InputEvent inputEvent, int mode) {
+    public boolean injectEvent(InputEvent inputEvent, int mode) {
         if (!supportsInputEvents()) {
             throw new AssertionError("Could not inject input event if !supportsInputEvents()");
         }
@@ -157,6 +161,21 @@ public final class Device {
         }
 
         return serviceManager.getInputManager().injectInputEvent(inputEvent, mode);
+    }
+
+    public boolean injectEvent(InputEvent event) {
+        return injectEvent(event, InputManager.INJECT_INPUT_EVENT_MODE_ASYNC);
+    }
+
+    public boolean injectKeyEvent(int action, int keyCode, int repeat, int metaState) {
+        long now = SystemClock.uptimeMillis();
+        KeyEvent event = new KeyEvent(now, now, action, keyCode, repeat, metaState, KeyCharacterMap.VIRTUAL_KEYBOARD, 0, 0,
+                InputDevice.SOURCE_KEYBOARD);
+        return injectEvent(event);
+    }
+
+    public boolean injectKeycode(int keyCode) {
+        return injectKeyEvent(KeyEvent.ACTION_DOWN, keyCode, 0, 0) && injectKeyEvent(KeyEvent.ACTION_UP, keyCode, 0, 0);
     }
 
     public boolean isScreenOn() {
--- a/server/src/main/java/com/genymobile/scrcpy/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
@@ -186,6 +186,27 @@ public final class Device {
         return injectKeycode(keyCode, InputManager.INJECT_INPUT_EVENT_MODE_ASYNC);
     }
 
+    public boolean injectTextPaste(String text) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+            return false;
+        }
+
+        // On Android >= 7, we can inject UTF-8 text as follow:
+        // - set the clipboard
+        // - inject the PASTE key event
+        // - restore the clipboard
+
+        String clipboardBackup = getClipboardText();
+        isSettingClipboard.set(true);
+
+        rawSetClipboardText(text);
+        boolean ok = injectKeycode(KeyEvent.KEYCODE_PASTE, InputManager.INJECT_INPUT_EVENT_MODE_WAIT_FOR_RESULT);
+        rawSetClipboardText(clipboardBackup);
+
+        isSettingClipboard.set(false);
+        return ok;
+    }
+
     public boolean isScreenOn() {
         return serviceManager.getPowerManager().isScreenOn();
     }
@@ -214,9 +235,13 @@ public final class Device {
         return s.toString();
     }
 
+    private boolean rawSetClipboardText(String text) {
+        return serviceManager.getClipboardManager().setText(text);
+    }
+
     public boolean setClipboardText(String text) {
         isSettingClipboard.set(true);
-        boolean ok = serviceManager.getClipboardManager().setText(text);
+        boolean ok = rawSetClipboardText(text);
         isSettingClipboard.set(false);
         return ok;
     }
--- a/server/src/main/java/com/genymobile/scrcpy/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
@@ -167,15 +167,23 @@ public final class Device {
         return injectEvent(event, InputManager.INJECT_INPUT_EVENT_MODE_ASYNC);
     }
 
-    public boolean injectKeyEvent(int action, int keyCode, int repeat, int metaState) {
+    public boolean injectKeyEvent(int action, int keyCode, int repeat, int metaState, int mode) {
         long now = SystemClock.uptimeMillis();
         KeyEvent event = new KeyEvent(now, now, action, keyCode, repeat, metaState, KeyCharacterMap.VIRTUAL_KEYBOARD, 0, 0,
                 InputDevice.SOURCE_KEYBOARD);
-        return injectEvent(event);
+        return injectEvent(event, mode);
+    }
+
+    public boolean injectKeyEvent(int action, int keyCode, int repeat, int metaState) {
+        return injectKeyEvent(action, keyCode, repeat, metaState, InputManager.INJECT_INPUT_EVENT_MODE_ASYNC);
+    }
+
+    public boolean injectKeycode(int keyCode, int mode) {
+        return injectKeyEvent(KeyEvent.ACTION_DOWN, keyCode, 0, 0) && injectKeyEvent(KeyEvent.ACTION_UP, keyCode, 0, 0, mode);
     }
 
     public boolean injectKeycode(int keyCode) {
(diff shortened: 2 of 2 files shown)


AndroidDeveloperLB:
Where can I download it?

rom1v:
Here are binaries so that you can test even if you don't know how to build:

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1426/1/scrcpy.exe)
  _SHA256: 4481d854630aaba9f31e4f3a19b46c4843c1918da03e1a486089bae3a3ebb09a_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1426/1/scrcpy-server)
  _SHA256: 64cd4d4277040430c48d35677ad8472a6315337651fb9cfc5f62cf5abf2021b7_

_(on Windows, replace both files in the v1.13 release; on other platforms, use `scrcpy-server` and follow [§Prebuilt server](https://github.com/Genymobile/scrcpy/blob/master/BUILD.md#prebuilt-server))_

AndroidDeveloperLB:
Doesn't work well.
Each time I write a Hebrew letter, it also writes the English one .

Example: 
I write: "ש"
And I get: "שa"

Not only that, but for some reason it doesn't work on all places. In the create-contact screen, it didn't work for some of the EditTexts...

And for some reason I can't run it without the console using "scrcpy-noconsole.exe" . 

See here:
[2020-05-25_10-52-45.zip](https://github.com/Genymobile/scrcpy/files/4675938/2020-05-25_10-52-45.zip)

rom1v:
> Each time I write a Hebrew letter, it also writes the English one .
> 
> Example:
> I write: "ש"
> And I get: "שa"

What behavior do you get on v1.13?

What if you launch it with `scrcpy --prefer-text`?

> Not only that, but for some reason it doesn't work on all places. In the create-contact screen, it didn't work for some of the EditTexts...

Oh, that's bad :(

> And for some reason I can't run it without the console using "scrcpy-noconsole.exe" .

This is not a release, just a quick build so that you can test, I didn't post a new build of `scrcpy-noconsole.exe`.

AndroidDeveloperLB:
By "v1.13", you mean the official one here:
https://github.com/Genymobile/scrcpy/releases
?
If so, instead of Hebrew it writes in English. So for the above example, when I type "ש" it puts "a" .

Using `scrcpy --prefer-text` seems to handle the English writing, but I've noticed an issue that is caused by the method itself of copying-to-clipboard:
It puts the texts that are copied into the clipboard history of Gboard ... 

I still don't get how come the EditText issue exists. I can paste there...

rom1v:
> Not only that, but for some reason it doesn't work on all places. In the create-contact screen, it didn't work for some of the EditTexts...

> It puts the texts that are copied into the clipboard history of Gboard ...

OK, thank you for your feedbacks. This method could not be used when typing text then.

However, I could probably use it to paste the text immediately on `Ctrl`+`Shift`+`v` (if Android >= 7), in addition to clipboard synchronization.

AndroidDeveloperLB:
Can you please explain how could the EditText issue exist ? Pasting is allowed there...

Is it possible perhaps to create a (secured) app that uses root, to type characters via ScrCpy? 
One that will allow only ScrCpy to type characters?
Or is even if it's possible, it could be very dangerous, as other apps might exploit it ?

rom1v:
> Can you please explain how could the EditText issue exist ? Pasting is allowed there...

I don't know.

> Is it possible perhaps to create a (secured) app that uses root, to type characters via ScrCpy?
> One that will allow only ScrCpy to type characters?
> Or is even if it's possible, it could be very dangerous, as other apps might exploit it ?

Scrcpy should work without root. And indeed it looks dangerous.

AndroidDeveloperLB:
I'm talking only about the non-English typing. I don't think there is any other solution, sadly.

Someone proposed a keyboard, but that's changing the way we use the device and it doesn't seem like a stable solution.

I already requested Google to add unicode support:
https://issuetracker.google.com/issues/143375921

It doesn't even work for emulator:
https://issuetracker.google.com/issues/153619173

But I don't think it's going to be added anytime soon.

AndroidDeveloperLB:
I can think of a way to make a root option secured:

Each time you start ScrCpy via the PC, install&run a completely new companion APK that has a secret password that only the PC app knows, which will let communication only via this password.

In addition, the Android app will only be able to accept communication once for its entire life, and will be auto-removed when the communication is finished.

rom1v:
> Is it possible to use root to emulate unicode typing?

Probably writing to `/dev/input/...`.

But rooted devices is not the target of scrcpy. I don't plan to implement a root-specific solution.

AndroidDeveloperLB:
It's only a temporary solution, for a very specific problem : typing in unicode.
Only people who really want it would use it.

It's similar to a lot of apps out there, that offer basic functionality without root, and extra functionality (that can't be achieved without root) with root.

I hate it too, but I think it's the only solution for now.

AndroidDeveloperLB:
I can come up with other possible security measures if you wish.

AndroidDeveloperLB:
Would you like me to prepare the Android app? I can if you wish.

rom1v:
TBH, no, I don't want to add complexity in scrcpy for a root-only solution.

AndroidDeveloperLB:
Too bad. 
Can you think of another alternative solution for this, then?

rom1v:
I'm closing, instead I'll do later what I said here: https://github.com/Genymobile/scrcpy/pull/1426#issuecomment-633448242

AndroidDeveloperLB:
So unicode typing is lost :(

AndroidDeveloperLB:
Can you please consider starring the issues I wrote to Google, at least?

rom1v:
> I'm closing, instead I'll do later what I said here: https://github.com/Genymobile/scrcpy/pull/1426#issuecomment-633448242

Done: fc1dec027093ee8a2746a212db0c72a9d96d39a4

AndroidDeveloperLB:
If you will star them, I will say "Thank you" :)

quyleanh:
Hope Google can if this issue.

AndroidDeveloperLB:
@quyleanh Please consider starring. I believe that with enough attention, Google might consider adding it.

quyleanh:
@AndroidDeveloperLB done 😃

AndroidDeveloperLB:
@rom1v Found another mirroring app that does it, and as was written here, it uses a keyboard app on the device:
https://en.mirroid.io/

Sadly it's only for Windows OS, and it has some annoyances. But it works for this case.

Maybe you could do something similar? But make it smarter than them:
They replace the keyboard app while connected, and don't change it back to the app you used before (Gboard in this case) when you are done using it. I wonder if it's possible.

domenn:
Current version (1.16) appears not to take `--prefer-text` launch parameter into account. It behaves same as when launched without. So in 1.16 I cannot type slovenian letters (čžšČŽ...). In console I see `[server] WARN: Could not inject char u+010d` whether I run with or without the parameter. I tried the .exe files that you posted above in this issue discussions and it works. But they appear outdated (-v parameter outputs 1.13). Is this clipboard-transport functionality added to 1.16? If not I believe it should be added since it is a useful feature and it works. I understand the drawbacks (bad for games), but that's why we have a choice (run with or without the parameter).

Also, the .exe files in this post don't work when used without parameter for me. Console shows some output then it stops responding, screen doesn't display. I use Windows.

Should I open a new issue, for example focusing on `--prefer-text` not working? (This is what I understand tho. Maybe the parameter works, but is not actually intended to fix my usecase).

rom1v:
`--prefer-text` is not for supporting UTF-8 characters, but just to inject letters as "text" instead of "key events".

In general, only ASCII works correctly: https://github.com/Genymobile/scrcpy/blob/master/FAQ.md#special-characters-do-not-work
This is a limitation on Android about raw key injection (without an IME application).

domenn:
Thank you for explanation. I read FAQ already and posts regarding lack of UTF-8 support.

But with .exe files in this post non UTF-8 characters work. Could those changes be integrated in "mainline" under a different command line parameter for example? 

Also an interesting note, some non-ascii characters actually work by default (ß and ć for me)

rom1v:
These binaries were built at the time of the PR, based on v1.13, so you need the v1.13 of the `scrcpy-server`.

> Could those changes be integrated in "mainline" under a different command line parameter for example?

Probably not, it was a failed attempt.

AndroidDeveloperLB:
@rom1v Is there any hope in having this feature? Any idea ? Maybe using accessibility?
Could accessibility type text (including unicode) ?

quyleanh:
@rom1v I heard about Microsoft's Your Phone app can type and transfer unicode character. Could you take a look about that. I think there will be work around for this issue.

rom1v:
Basically, the only solution to get this working in scrcpy requires to install an app on the device and select an IME when scrcpy is running. Currently, scrcpy does not install any app (see #37).

> I heard about Microsoft's Your Phone app can type and transfer unicode character.

I don't know exactly how it works, but probably:
 - either they install an app on the device;
 - either the app is preinstalled (or system gets modified by the vendor, that would explain why they support only a limited set of devices) to expose a function to inject unicode characters.

AndroidDeveloperLB:
@rom1v 
" select an IME when scrcpy is running"
I've noticed "[mirroid](https://en.mirroid.io/)" alternative app does it for you. Sadly it doesn't restore it when it's done, though.

Can accessibility inject text, including unicode?

rom1v:
> Sadly it doesn't restore it when it's done, though.

scrcpy could restore it during the cleanup.

> I've noticed "mirroid" alternative app

https://mirroid.io/faq#shortcut

They use the very same shortcuts. :smile: 
Where is the desktop part of the app?

AndroidDeveloperLB:
the very same shortcuts.? 
What do you mean? 
As for desktop part, it's on the main website.
Here:
https://en.mirroid.io/

rom1v:
> As for desktop part, it's on the main website.

On that website, there is only a link "download for Android", which redirects to the play store :/

domenn:
I downloaded and tried mirrord in the past. I dislike it. It is strange that you can't find windows download link. Maybe it depends on what device you open. Here is direct link, hopefully:
https://res.mirroid.io/download/2020/08/18/MirroidInstaller_V1.5.0.3.exe

AndroidDeveloperLB:
@rom1v No. The website has a link to download for Windows. Maybe you changed the user-agent or something.
See:

[Zhhti7tIU5.zip](https://github.com/Genymobile/scrcpy/files/5164479/Zhhti7tIU5.zip)

And the download link is :

https://en.mirroid.io/en/download

And direct download link is:

https://res.mirroid.io/download/2020/08/18/MirroidInstaller_V1.5.0.3.exe

@domenn They got a UI with features that can be easily be used directly, which scrCpy doesn't have. 
However, scrCpy has many other things that this one doesn't and MirroId has some issues sometimes, so I use MirroId only when I want to type unicode.

rom1v:
OK, thank you for the links.

So Mirroid is completely based on scrcpy, without respecting the [license](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)) at all. In particular, scrcpy copyright is not mentioned anywhere.

`mirroid-server.jar` is just `scrcpy-server.jar` (even the class names are unchanged and package names is `com.genymobile.scrcpy`).

The client code is based on QtScrcpy (#612). The OpenGL shaders are the same as from (older versions) of https://github.com/barry-ran/QtScrcpy/blob/master/QtScrcpy/device/render/qyuvopenglwidget.cpp

Btw:
```
$ strings Mirroid.exe | grep scrcpy
inputscrcpy
qtscrcpy
com.genymobile.scrcpy.Server
```

They have an [user agreement](https://en.mirroid.io/useragreement):

> We may share your Personal Information with our affiliate and subsidiary companies, joint venture partners and third party service providers, contractors and agents, …

Interesting section:

> 11.1 The Software may use third party softwares or technologies (including without limitation, some open-source codes and/or public domain codes). The necessary licenses of use such third party softwares or technologies will be obtained legally by Mirroid Team.

No, they don't.

AndroidDeveloperLB:
@rom1v I see. But how did they break the rules of the license? They need to just mention that they use scrCpy?
Are they copycats?

quyleanh:
@rom1v so could you investigate the work around based on their implementation?
As your comment, I think people cannot trust mirroid.io at all. Thank you for comment anyway.

AndroidDeveloperLB:
@quyleanh Do you know perhaps of a different alternative to it, that does offer unicode?

quyleanh:
@AndroidDeveloperLB no, I don't. I have tested QtScrcpy but it doesn't support unicode. (of course, it's based on scrcpy). Hope @rom1v can find out the way.

AndroidDeveloperLB:
@rom1v I hope you sue them for violating.
But I also want to know how to avoid such an issue myself. I thought apache is quite free-to-use.

vip-liutao:
> 接时替换键盘应用程

mirroid真强大,是我需要的
一样不延迟并且解决了输入法问题，这个在scrcpy困扰了我很久
这两款软件的音频都可以采用
蓝牙连接电脑
win10应用商店下载
Bluetooth audio receiver
使用

gaussandhisgun:
can you add built linux binaries of this in here? my distro seems to have problems with meson as it provides only 0.45 and scrcpy needs 0.48 to be built

thedmd:
Version from top post works for me like a charm. Polish characters are handled well. Thank you!

heweisheng:
if check the message is utf8 char use paste?or not use ascii inject。maybe is a idea but i need check all char

AlexBurdu:
I addepted this a bit in [my fork](https://github.com/AlexBurdu/scrcpy/commit/4a7bce99c03bb4c3a41d0c4478d3bb85e0ddd4ea) and works great for my needs.
Thanks!

quyleanh:
@AlexBurdu could you add some brief info?

AlexBurdu:
So the implementation that you can see in [my fork](https://github.com/AlexBurdu/scrcpy/commit/4a7bce99c03bb4c3a41d0c4478d3bb85e0ddd4ea), which is a slightly modified version of the one in this PR works fine on Pixel phone. Unfortunately it does not work well on my Samsung Galaxy S20 FE.
I believe Samsung may have customized the way clipboard works on their rom, as when I'm sending characters through clipboard I see them in Gboard keyboard clipboard history, but I don't see them in OS clipboard or Samsung's clipboard.

I initially started poking at this to add [Romanian diacritics support](https://github.com/AlexBurdu/scrcpy/commit/4a7bce99c03bb4c3a41d0c4478d3bb85e0ddd4ea#diff-ef0095487caa8b924f050ab459bc5863fca6d33d6c13b0f0fd20587d1d939a0c), but it seems that support for them is missing in the Android OS `KeyCharacterMap` class, and no events are returned for `KeyCharacterMap.getEvents(chars)`
For example, [ACCENT_COMMA_BELOW](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/KeyCharacterMap.java;l=218;drc=master) is commented out in the source file. While [ACCENT_BREVE](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/KeyCharacterMap.java;l=200;drc=master) is there, `KeyCharacterMap.getEvents()` still returns null for `breve('A')` / `breve('a')`. I didn't dig further to see why that happens, I'm assuming that some mapping for it is also missing in this function that is [meant to be used for tests](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/KeyCharacterMap.java;l=609;drc=master).

AlexBurdu:
@rom1v would it be possible for scrcpy server to register an [IME](https://developer.android.com/guide/topics/text/creating-input-method.html#DefiningIME) that would take care of text input, or is not possible?
Is scrcpy-server seen by the OS as a regular application or like "something else" (what else)?

rom1v:
> Is scrcpy-server seen by the OS as a regular application or like "something else" (what else)?

No, it's not an Android app, just a Java executable accessing the Android APIs.

More details: https://blog.rom1v.com/2018/03/introducing-scrcpy/

About IME, also see #1751.

AlexBurdu:
Thanks, the hack in #1751 works great for my needs as a [fallback to key inject](https://github.com/AlexBurdu/scrcpy/commit/ee2f3faec6ad3c03938725169b83f792659ea84e)!

quyleanh:
@AlexBurdu could you please give me the Windows build for your latest implementation? I would like to test with my inject text scenario.
Thank you.

domenn:
@quyleanh Here. 
@AlexBurdu You did the hard work, I managed to build it for Windows and therefore I may as well share it with anyone else that's interested.
https://github.com/domenn/scrcpy/releases/tag/v1.19

AlexBurdu:
Thanks, I just adapted the work others have done in the linked prs. 
Please note that you need to install AdbKeyboard and select it as yout keyboard while you use scrcpy for it to work: https://github.com/senzhk/ADBKeyBoard

Also, i have a few changes done to the mouse buttons on my fork.

AlexBurdu:
For Linux users I use this trick to get the ADBkeyBoard set when I start scrcpy and then switch back to my previous one (GBoard) when I exit it.

```
export personal_phone=192.168.0.94:8888 && \
adb -s $personal_phone shell ime set com.android.adbkeyboard/.AdbIME && \
adb connect $personal_phone && \
scrcpy \
--turn-screen-off \
--stay-awake \
--window-x 2500 --window-y 0 \
--max-size 1024 \
-s $personal_phone; \
adb -s $personal_phone shell ime set com.google.android.inputmethod.latin/com.android.inputmethod.latin.LatinIME
```

gaussandhisgun:
> Thanks, I just adapted the work others have done in the linked PRs. Please note that you need to install [ADBKeyBoard](https://github.com/senzhk/ADBKeyBoard) and select it as your keyboard while you use scrcpy for it to work:
> 
> Also, i have a few changes done to the [mouse buttons bindings](https://github.com/AlexBurdu/scrcpy/commit/4820529887e0e97cf4598a3d082e4d5036a81e0a) on my fork.

since i for can not submit issues on your fork built by @domenn, i'll put it right here. for whatever reason, on non-english layout, when pressing a russian key, both english and russian chars appear at the same time. this might be my clipboard manager or kde connect freaking out but they seem to copy only russian chars. 
image

my devices are: redmi note 9s running pixelplusui on android 11 and a laptop running windows 10 21H1 x64, connection was done with usb, commandline is `scrcpy`.

AlexBurdu:
that sounds like a bug for https://github.com/senzhk/ADBKeyBoard, but I'm not sure.
I'm no longer using clipboard for this, but sending the chars through adb shell to ADBKeyBoard.

quyleanh:
@domenn @AlexBurdu thank you very much. I have tried and it work well with my case.
Input Vietnamese text has no problem now.

AlexBurdu:
We have diverged a lot from the initial PR discussion (sorry for that), but [in my fork](https://github.com/AlexBurdu/scrcpy/commit/df260fbc9cf279dc47da5ca3986fa16bdde2231d), I don't have any code left for injecting through keyboard

yilksd:
> that sounds like a bug for https://github.com/senzhk/ADBKeyBoard, but I'm not sure. I'm no longer using clipboard for this, but sending the chars through adb shell to ADBKeyBoard.

How to use your fork? I have just compiled your fork and launched it but I still cannot type Chinese character. I have the ADB keyboard app installed.

quyleanh:
I have compiled the Windows builds with @AlexBurdu 's ADBKeyBoard merged.
Feel free to use in this release.
https://github.com/quyleanh/scrcpy/releases/tag/v1.24

Issue #942, opened 2019-11-21
Request: be able to select text with Shift + Arrow Keys


KeronCyst:
Bluetooth keyboards can do this when connected to the device.

rom1v:
I implemented forwarding SHIFT, and injecting COPY on Ctrl+c (if Android >= 7, like I do for paste since [v1.14](https://github.com/Genymobile/scrcpy/releases/tag/v1.14)).

Here is a branch [`copy`](https://github.com/Genymobile/scrcpy/commits/copy).

Unfortunately, there is a bug on Android (at least on my ROM on One Plus 7 Pro).

To reproduce:
 - select any text and click on "copy"
 - click somewhere in the middle of a text in a textfield, it should show a dialog containing `Paste` and `Select all`
 - when this dialog is displayed, press `Shift`+`←` (left)

![dialog](https://user-images.githubusercontent.com/543275/83128983-74bcc900-a0dc-11ea-8762-84860aece9df.png)


On my device, it crash 100% reproducible:
```
E MessageQueue-JNI: Exception in MessageQueue callback: handleReceiveCallback
E MessageQueue-JNI: java.lang.IllegalArgumentException
E MessageQueue-JNI: 	at com.android.internal.util.Preconditions.checkArgument(Preconditions.java:35)
E MessageQueue-JNI: 	at android.widget.SelectionActionModeHelper$TextClassificationHelper.init(SelectionActionModeHelper.java:1062)
E MessageQueue-JNI: 	at android.widget.SelectionActionModeHelper.resetTextClassificationHelper(SelectionActionModeHelper.java:468)
E MessageQueue-JNI: 	at android.widget.SelectionActionModeHelper.resetTextClassificationHelper(SelectionActionModeHelper.java:476)
E MessageQueue-JNI: 	at android.widget.SelectionActionModeHelper.startSelectionActionModeAsync(SelectionActionModeHelper.java:123)
E MessageQueue-JNI: 	at android.widget.Editor.startSelectionActionModeAsync(Editor.java:2154)
E MessageQueue-JNI: 	at android.widget.Editor.refreshTextActionMode(Editor.java:2099)
E MessageQueue-JNI: 	at android.widget.TextView.spanChange(TextView.java:10719)
E MessageQueue-JNI: 	at android.widget.TextView$ChangeWatcher.onSpanChanged(TextView.java:13446)
[...]

PR #1465, opened 2020-06-01
Forward LCtrl and Shift to the device

### LINKED ISSUE #555: support passing in Home and Ctrl-x keys

rom1v:
This is a follow up of #1446.

To improve usability, the goal of these changes is to forward `Ctrl` and `Shift` keys to the device.

This allows for example to press `Ctrl`+`t` in Firefox to open a new tab or to press `Ctrl`+`c` or `Ctrl`+`d` in Termux. This also allows to select text using the keyboard (`Shift`+arrows, `Ctrl`+`a`…), and simplify copy-paste (`Ctrl`+`c`, `Ctrl`+`v`) and cut-paste (`Ctrl`+`x`, `Ctrl`+`v`).

Here is the list of the changes:
 - all shortcuts now use the `RCtrl` (right `Ctrl`) key instead of any `Ctrl` key, on all platforms
 - in addition, on macOS, `Cmd` can be used (for shortcuts which are not already captured by the system) as before
 - `LCtrl` (left `Ctrl`) and `Shift`(both left and right) are forwarded to the device
 - to copy a text, `Ctrl`+`c` works, because it is forwarded to the device, which, in a text field, copy the text into the clipboard… and the clipboard is now automatically (since v1.13) synchronized with the computer keyboard.
 - for convenience, and as special cases, `RCtrl`+`c` (`Cmd`+`c`) and `RCtrl`+`x` (`Cmd`+`x`) are forwarded to the device as `Ctrl`+`c` and `Ctrl`+`x`
 - pressing `LCtrl`+`v` synchronizes the computer clipboard to the device before sending the key event, therefore pasting works seamlessly (and the old `Ctrl`+`Shift`+`v` shortcut has been removed)
 - `RCtrl`+`v` injects the computer clipboard content as a sequence of text events (this is still useful sometimes)
 - to support "cut" `RCtrl`+`x`, I changed the "resize to fit" (the shortcut which removes the black borders) from `RCtrl`+`x` (oldly `Ctrl`+`x`) to `RCtrl`+`w`
 - seamless copy-paste now also works before Android 7

This fixes #555 and #942.

Feedbacks welcome :)

## Binaries

For windows users, take both `scrcpy.exe` and `scrcpy-server`, and replace them in your scrcpy v1.14 release.

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1465/1/scrcpy.exe)
  _SHA256: a22661070827d47a08618bcaf8550688aa2c56b0053251ba45cebee5b9b45850_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1465/1/scrcpy-server)
  _SHA256: 62f3cb793c8398e0f55f5e98be438be0c238ddb94efbdbbf0f76b7522ee51264_


nightshade427:
This works great! Any way to get ALT key forwarded as well?

rom1v:
> Any way to get ALT key forwarded as well?

I tested it, but when I press `Alt`+`tab` on my computer, it presses it both on the device and the computer, which is unexpected (Debian with XFCE on X11).

`AltGr`+`tab` works on the device though… (but it seems weird to forward `AltGr` only to avoid conflicts, and it would be inconsistent with forwarding left control).

You could test on your computer, and share your feedback:

```diff
diff --git a/app/src/event_converter.c b/app/src/event_converter.c
index 82b0501..94b5966 100644
--- a/app/src/event_converter.c
+++ b/app/src/event_converter.c
@@ -95,6 +95,8 @@ convert_keycode(SDL_Keycode from, enum android_keycode *to, uint16_t mod,
         MAP(SDLK_LCTRL,        AKEYCODE_CTRL_LEFT);
         MAP(SDLK_LSHIFT,       AKEYCODE_SHIFT_LEFT);
         MAP(SDLK_RSHIFT,       AKEYCODE_SHIFT_RIGHT);
+        MAP(SDLK_LALT,         AKEYCODE_ALT_LEFT);
+        MAP(SDLK_RALT,         AKEYCODE_ALT_RIGHT);
     }
 
     if (!(mod & (KMOD_NUM | KMOD_SHIFT))) {
diff --git a/app/src/input_manager.c b/app/src/input_manager.c
index 3e932a5..cbfaa74 100644
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -270,7 +270,6 @@ input_manager_process_key(struct input_manager *im,
 
     bool lctrl = event->keysym.mod & KMOD_LCTRL;
     bool rctrl = event->keysym.mod & KMOD_RCTRL;
-    bool alt = event->keysym.mod & (KMOD_LALT | KMOD_RALT);
     bool meta = event->keysym.mod & (KMOD_LGUI | KMOD_RGUI);
     bool shift = event->keysym.mod & (KMOD_LSHIFT | KMOD_RSHIFT);
 
@@ -284,11 +283,6 @@ input_manager_process_key(struct input_manager *im,
     }
 #endif
 
-    if (alt) {
-        // No shortcuts involve Alt, and it is not forwarded to the device
-        return;
-    }
-
     struct controller *controller = im->controller;
 
     SDL_Keycode keycode = event->keysym.sym;
```

nightshade427:
I use editors on Android and they use alt key combos for switching tabs, etc.

I would be happy to try out the above code but I don't have any tooling on my win10 machine to build project. Can you post binaries like above possibly?

rom1v:
Here are binaries with `Alt` and `AltGr` forwarding enabled:

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1465/withalt/scrcpy.exe)
  _SHA256: 1d56fcf25658574ec84dbbd4ff26b1e03a9309e36b8f33d4627ff14f6514f67e_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1465/withalt/scrcpy-server)
  _SHA256: 62f3cb793c8398e0f55f5e98be438be0c238ddb94efbdbbf0f76b7522ee51264_

The behavior depends on the device ROM. For example, on OxygenOS, pressing/releasing `Alt` opens the emoji keyboard and `AltGr` the normal keyboard. Long-pressing `Alt` then pressing tab triggers app switching on the device (in addition to app switching on the computer, which is annoying). `AltGr`+`tab` triggers app switching on the device only.

A compromise could be to forward keys when the Alt modifier is set, without sending the `Alt` events themselves (but it would be inconsistent with `Ctrl` and `Shift`):

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1465/withaltmod/scrcpy.exe)
  _SHA256: 9c69941eb692aeb3977e7b79e6180d0f1691f1a8123e1baa94f063ad6e3ac667_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1465/withaltmod/scrcpy-server) (it's the same)
  _SHA256: 62f3cb793c8398e0f55f5e98be438be0c238ddb94efbdbbf0f76b7522ee51264_

nightshade427:
The first binary above is sending alt to device, but it wont let me send a key combo (eg. alt+t) it sees the alt button being pressed but doesnt see the 't' while holding down alt. 

The second binary doesnt seem to register any alt keys at all.

rom1v:
Probably because your system/window manager captures and prevents it to be passed to SDL/scrcpy.

What about `AltGr`+`t`? (or `AltGr`+`tab`)

rom1v:
> it sees the alt button being pressed

(What do you use to see that?)

nightshade427:
I use a combo of micro editors raw input mode (to look at keystrokes) or online keyboard testers

rom1v:
So it confirms that Alt should not be forwarded IMO.

nightshade427:
I wonder if they're is a way to do it like virtual machine software does (eg. Vmware), where you press a key combo (eg. Ctrl+Alt+h) and all keys are sent to device until that combo is pressed again?

lilydjwg:
It works nice for me, except for one small case:

I can't paste from computer clipboard to Android termux app. Ctrl-v syncs the clipboard but it doesn't paste text in a terminal. So could you add a shortcut that only syncs the clipboard to Android? There are also apps that act against clipboard content on open too.

rom1v:
> I can't paste from computer clipboard to Android termux app. Ctrl-v syncs the clipboard but it does other things than paste text in a terminal. So could you add a shortcut that only syncs the clipboard to Android?

Indeed, Termux does not support pasting. But there is always the shortcut to inject the clipboard content as a sequence of text events, which is now `RCtrl`+`v`.

rom1v:
> There are also apps that act against clipboard content on open too.

What do you mean?

lilydjwg:
> Indeed, Termux does not support pasting.

It does support pasting, by using long-press menus.

> But there still is the shortcut to inject the clipboard content as a sequence of text events, which is now RCtrl+v. Could you confirm it works for you?

This only works with ASCII. If my text has Chinese characters I get this:

```
[server] WARN: Could not inject char u+5220
```

>> There are also apps that act against clipboard content on open too.
>
> What do you mean?

Some apps will analyze the clipboard content on open, and if it's some predefined pattern, the app acts accordingly. E.g. The popular shopping app Taobao will check the clipboard for some kind of goods id and open the page for it. So if you get such text (e.g. from a chat app) you'll need to put it into the clipboard and then open Taobao. There is no place for the user to paste.

rom1v:
Thank you for the feedback.

> It does support pasting, by using long-press menus.

Does this version breaks something compared to v1.14 related to copy-paste?

IIUC, I should keep a shortcut to synchronize clipboard without pasting?

> Some apps will analyze the clipboard content on open, and if it's some predefined pattern, the app acts accordingly. E.g. The popular shopping Chinese app Taobao will check the clipboard for some kind of goods id and open the page for it. So if you get such text (e.g. from a chat app) you'll need to put it into the clipboard and then open Taobao. There is no place for the user to paste.

Wow, it opens without any action from the user?

How did you paste in such case in v1.14? Does this PR breaks something regarding this behavior?

lilydjwg:
> IIUC, I should keep a shortcut to synchronize clipboard without pasting?

Yes, so the steps will be "sync, long press and click", instead of "ctrl-v, ctrl-v, backspace, long press and click". But anyway, I'll just ssh into termux when my computer is near and functional. I don't have an actual need for this.

> Wow, it opens without any action from the user?

The user opens the app. That's it.

> How did you paste in such case in v1.14? Does this PR breaks something regarding this behavior?

I didn't test. It's a possible use case. I didn't use keyboard input before this patch because I didn't feel like previous key handling and attached the keyboard directly to my phone when needed.

This patch works pretty well for me. I'd use scrcpy more because of this.

rom1v:
>> IIUC, I should keep a shortcut to synchronize clipboard without pasting?
> Yes, so the steps will be "sync, long press and click", instead of "ctrl-v, ctrl-v, backspace, long press and click".

I restored `RCtrl`+`Shift`+`v` to just synchronize the computer clipboard to the device clipboard (branch [`clipboard.3`](https://github.com/Genymobile/scrcpy/commits/clipboard.3)).

If there is no major problem on this branch, I will merge it into `dev` (and probably publish a new release after that).

Here are binaries:

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/pr1465/2/scrcpy.exe)
  _SHA256: cabfc66d9d13690bb22be0ea6d0fa9d090aa6460a45f53c01de2e61f6943d8a4_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/pr1465/2/scrcpy-server)
  _SHA256: 15533c3d4a4a12a20e2afe9cd7dfb8efecd3f990227ac8262bb79998c1fc3af5_

lilydjwg:
This branch works well for me, thanks!

rom1v:
Thank you :)

Could someone test on macOS and provide feedback, please?

lsdy0806:
Hello！Not very friendly for windows notebooks without Rctl.

lsdy0806:
Hello！Not very friendly for windows notebooks without Rctrl.

lsdy0806:
> 您好！对于没有Rctrl的Windows笔记本不是很友好。

I hope you can change the Rctrl key and use Alt or other keys instead of Rctrl!

lsdy0806:
This branch is not very friendly to me, Several of my laptops do not have Rctrl, I hope you can use Alt or other keys instead of Rctrl.

rom1v:
@lsdy0806 Yes, thank you for your feedback, it's like some [apple keyboard](https://github.com/Genymobile/scrcpy/pull/1465#issuecomment-642184032).

I will propose something else when I have time.

lsdy0806:
> @ lsdy0806是的，谢谢您的反馈，就像一些[苹果键盘一样](https://github.com/Genymobile/scrcpy/pull/1465#issuecomment-642184032)。
> 
> 有空的时候我会提出其他建议。

Thank you very much for your reply, Scrcpy has brought great convenience to my work and life. If the future version uses the combination of Rctrl+keys, you will not be able to use these key combination functions on my computer, because my windows computer lacks the Rctrl key. I hope you can use the Alt key instead of the Rctrl key.

rom1v:
> pressing `LCtrl+v` synchronizes the computer clipboard to the device before sending the key event, therefore pasting works seamlessly

In fact, there is a problem here: copying a password on the computer and pasting it to the device puts the password in the device clipboard (which any app can read/listen). :scream:

lsdy0806:
> > pressing `LCtrl+v` synchronizes the computer clipboard to the device before sending the key event, therefore pasting works seamlessly
> 
> In fact, there is a problem here: copying a password on the computer and pasting it to the device puts the password in the device clipboard (which any app can read/listen). 😱

This is not a problem, Many projection software already supports clipboard bidirectional synchronization.

rom1v:
The difference is that when you *paste* from computer to device , the content of the paste is copied into the device clipboard. That may be unexpected. (This does not happen when injecting the content char by char as a sequence of key events.)

Maybe I will just warn in the README.

brunoais:
I think we could try solving this problem by looking at what others do.
The most popular remote desktop and remote control tools I see, handle this problem by using some sort of toggle system.
Most popular methods:

1. Toggle options (with configurable shortcut press)
1.1 E.g. TeamViewer uses (or used to use) Scroll lock as shortcut.
1.2. E.g. AnyDesk uses a manual toggle option with a shortcut key combo
2. Full screen
2.1 E.g. Reminna & windows remote desktop only send dubious target shortcuts to target machine if the target machine is in full screen, otherwise, they are sent to the top machine.
3. 2-step key combination
3.1 E.g. VirtualBox defaults to the VM and uses a configurable key to press so that the next key combo applies outside the VM.

I think we should discuss which one to use for scrcpy. I do think the option 2 is the worst of all because the format of the phone image is a lot different from the format of a PC screen leaving huge black bezels and a huge feel of wasted space.
What do you think?

rom1v:
I plan to forward Ctrl and Shift, and use a configurable key for scrcpy shortcuts (probably Alt by default).

It could be changed:
```
# use RCtrl for shortcuts
scrcpy --shortcut-mod=rctrl
# use either Alt or RCtrl
scrcpy --shortcut-mod=lalt,rctrl
# use LCtrl+Alt
scrcpy --shortcut-mod=lctrl+lalt
# use LCtrl+Alt or Cmd
scrcpy --shortcut-mod=lctrl+lalt,lcmd,rcmd
```

I started to work on that locally, but I didn't have much time recently.

lsdy0806:
> I plan to forward Ctrl and Shift, and use a configurable key for scrcpy shortcuts (probably Alt by default).
> 
> It could be changed:
> 
> ```
> # use RCtrl for shortcuts
> scrcpy --shortcut-mod=rctrl
> # use either Alt or RCtrl
> scrcpy --shortcut-mod=lalt,rctrl
> # use LCtrl+Alt
> scrcpy --shortcut-mod=lctrl+lalt
> # use LCtrl+Alt or Cmd
> scrcpy --shortcut-mod=lctrl+lalt,lcmd,rcmd
> ```
> 
> I started to work on that locally, but I didn't have much time recently.

This is really good, solves the problem of no Rctrl, and can configure shortcut key combinations according to personal habits, thank you very much for your work!

brunoais:
> I plan to forward Ctrl and Shift, and use a configurable key for scrcpy shortcuts (probably Alt by default).

That sound good to me. In my case, as an FXCE desktop user, `Alt` is the default key to use to manipulate windows for all kinds and purposes. With configuration, it can be the default (majority uses windows+mac) and then I can just reassign it to something else thanks to that feature.
Thank you very much.

Maybe the `--shortcut-mod` could be quickly pressed to toggle whether the next key combinations go to scrcpy/host or go to the phone. In my case, I use both `Ctrl`, both `Alt`, `menu`, etc... considerably so in **VirtualBox** I use **Insert** as the **"host key"** (Although it required editing XML files). I think I would do the same in scrcpy.

Thank you!

rom1v:
> That sound good to me. In my case, as an Xfce desktop user, `Alt` is the default key to use to manipulate windows for all kinds and purposes.

I am also a Xfce user, I use `Alt`+clicks for moving or resizing windows, I don't see any problems with using `Alt` for scrcpy shortcuts.

What Xfce shortcuts involving `Alt` do you use?

brunoais:
I went to check the config files myself and I actually set all shortcuts myself to make use of the alt, I don't use the default ones (unlike what I though I was doing... It's been years.)
But still: 
<kbd>Alt</kbd>+<kbd>↑</kbd>: Restore/Maximize (just like Windows has it, but with <kbd>Alt</kbd>)
<kbd>Alt</kbd>+<kbd>↓</kbd>: Restore/Minimize (just like Windows has it, but with <kbd>Alt</kbd>)
<kbd>Alt</kbd>+<kbd>M</kbd>: Minimize all windows
<kbd>Alt</kbd>+<kbd>D</kbd>: Peek at desktop (just like Windows has it, but with <kbd>Alt</kbd>)
<kbd>Alt</kbd>+<kbd>Ctrl</kbd>+<kbd>Tab</kbd>: Cycle windows of the same program group
<kbd>Alt</kbd>+<kbd>A</kbd>: Toggles window in all workspaces
<kbd>Alt</kbd>+<kbd>Ctrl</kbd>+Arrow: Switch to workspace in that virtual direction
<kbd>Alt</kbd>+<kbd>F{1..9}</kbd>: Switch to workspace of number and take the selected window with you
<kbd>Alt</kbd>+<kbd>{1..9}</kbd>: Switch to workspace of number and don't take the selected window with you
<kbd>Alt</kbd>+<kbd>Insert</kbd>: Insert adjacent workspace
<kbd>Alt</kbd>+<kbd>Del</kbd>: Delete workspace and go to previous workspace

However, I'm working on moving many of these shortcuts to <kbd>Super</kbd> because the window icon on the key (as usual, this was manufactured to be a windows PC) makes more sense to me than an <kbd>Alt</kbd> and too many programs are starting to include <kbd>Alt</kbd>-based shortcuts in their defaults.  I'm still convincing myself to relearn and transition.

If you're curious, in keyboard shortcuts, I moved all default shortcuts to start apps to <kbd>Alt</kbd>+<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F{1..9}</kbd> although I'm deprecating them for myself and just use something like the `wisker-menu` instead.

rom1v:
Here is a new proposal: #1598

Issue #274, opened 2018-09-24
It's not a bug,but i want it to work better


singleNeuron:
Thanks for you work, but could you please show the FPS on the screen instead of stdout.
My English is poor, please don't mind.

rom1v:
Yes, this would be better, but requires some work.

Firstly, the dependency [SDL_ttf](https://www.libsdl.org/projects/SDL_ttf/) must be added to be able to draw text. This would allow to write an [OSD](https://en.wikipedia.org/wiki/On-screen_display) component, which could be used to print various messages.

Then, we need to refresh (compute+render) the FPS between frames. Frames are sent only when the device surface content changes, so there may be seconds or minutes without any screen refresh. The FPS should be updated in that case.

_By the way (just a side note), the FPS is just a hint to get an idea of the performance, but it may be misleading: if you watch a 25 FPS fullscreen video, then FPS will be 25, no more, which is perfect. On the other hand, if the encoding/decoding/rendering is not fast enough, the FPS can be 55 while the screen renders at 60 fps, in that case the framerate is too low. Just to say that the raw FPS is not meaningful in itself._

singleNeuron:
Could it be easier by showing it on window's title bar, it seems to be like a good idea.

AndroidDeveloperLB:
Can you please allow to show it in either the title bar or the screen?

PR #1484, opened 2020-06-07
Show text after file pushed

npes87184:
Hi,

In this series, I add a text framework. User can use this framework to show text on screen.
After it, I use this framework to show file pushed.

[imgur](https://i.imgur.com/I1VDleM.png)

Note:

Because the path of font file is difficult to specify in linux based system. I embed open-sans ttf to prevent loading font file failed.

Thanks.


app/src/screen.c
```diff
@@ -454,6 +466,90 @@ update_texture(struct screen *screen, const AVFrame *frame) {
     }
 }
 
+bool
+screen_render_text(struct screen *screen) {
```
rom1v, line 470: This opens the font and draws the glyphs on every frame :/

IMO, it should create the texture and render the text the first time, then cache it (to be invalidated on resize though).

app/src/screen.c
```diff
@@ -454,6 +466,90 @@ update_texture(struct screen *screen, const AVFrame *frame) {
...
+    struct render_text_request req;
+
+    // we check this flag outside lock, it may cause that we don't show text immediately.
```
rom1v, line 473: (but it's racy, you should use an atomic instead)

app/src/screen.c
```diff
@@ -454,6 +466,90 @@ update_texture(struct screen *screen, const AVFrame *frame) {
...
+    dstrect.w = width;
+    dstrect.h = height;
+
+    SDL_RenderCopy(screen->renderer, text_texture, NULL, &dstrect);
+
+    screen->curr_text.time -= 1;
+    if (screen->curr_text.time <= 0) {
+        SDL_free(screen->curr_text.text);
+        screen->curr_text.text = NULL;
+        screen->curr_text.time = 0;
```
rom1v, line 536: The time is handled as the number of frames. But the framerate is (very) variable: it may be as low as 0 fps if the device screen is constant to 90fps (for example) if it changes a lot.

The "toast" delay should be based on real time instead.

rom1v (commented):
Thank you for working on this. I agree that some OSD would be better to report status/actions to the user (like FPS count https://github.com/Genymobile/scrcpy/issues/274#issuecomment-423934442).

> Because the path of font file is difficult to specify in linux based system. I embed open-sans ttf to prevent loading font file failed.

I'm not fan of embedding the font into the source code (I know that I did that for the [icon](https://github.com/Genymobile/scrcpy/blob/master/app/src/icon.xpm), but it's ASCII-art readable, not an array of hexadecimal values). I'd prefer a separate font-file. By the way, unless we want a specific font (I think we don't), I think it should even use some default sans-serif font available on the system.

_I added some inline comments on the PR._

Also, from a user point of view, I think that it is not very clear that the text is added by scrcpy (that it is not part of the device screen). I'm not sure how to improve that. Maybe one reason is that Android already displays notifications as toasts at the same location.

Events like "file pushed" could maybe printed in the bottom in some (overlay) "status bar" (a bit like in Firefox when we hover a link). Or as notification bubbles in the top-right (a bit like the result of `notify-send mytext` on Linux, ideally, they would also appear/disappear gracefully). (These are just things I have in mind, but I'm very bad at UI design :wink:)

npes87184:
> Thank you for working on this. I agree that some OSD would be better to report status/actions to the user (like FPS count [#274 (comment)](https://github.com/Genymobile/scrcpy/issues/274#issuecomment-423934442)).
> 
> > Because the path of font file is difficult to specify in linux based system. I embed open-sans ttf to prevent loading font file failed.
> 
> I'm not fan of embedding the font into the source code (I know that I did that for the [icon](https://github.com/Genymobile/scrcpy/blob/master/app/src/icon.xpm), but it's ASCII-art readable, not an array of hexadecimal values). I'd prefer a separate font-file. By the way, unless we want a specific font (I think we don't), I think it should even use some default sans-serif font available on the system.

Yes, I also think that embed some binary in code is really ugly. However, as far as I know, there are not an easy and reliable way to get usable font cross platform...

> 
> _I added some inline comments on the PR._
> 
> Also, from a user point of view, I think that it is not very clear that the text is added by scrcpy (that it is not part of the device screen). I'm not sure how to improve that. Maybe one reason is that Android already displays notifications as toasts at the same location.
> 
> Events like "file pushed" could maybe printed in the bottom in some (overlay) "status bar" (a bit like in Firefox when we hover a link). Or as notification bubbles in the top-right (a bit like the result of `notify-send mytext` on Linux, ideally, they would also appear/disappear gracefully). (These are just things I have in mind, but I'm very bad at UI design 😉)

Maybe style like notify-send is a good design?
(I'm very bad at UI design too :/)

app/src/screen.c
npes87184: Agree.

npes87184: What should happen if time is up but the screen is constant? In correct implementation, it will remain on screen until screen changed.

app/src/screen.c
npes87184: Yes, it will race. But the effect is just that it will delay notice that it needs to show text.

This trick added only because I notice the performance has some degrade. Maybe if we don't create texture every frame, we don't need this.

app/src/screen.c
npes87184: Agree.

app/src/screen.c
rom1v: IMO, the popup/taost delay should be independant of the device video. It should fade out after a constant delay (whether the device screen is constant or not).

npes87184:
I have pushed new version, it includes:

1. Toast will fade out after a constant delay no matter the state of the screen.
2. Cache test_texture.
3. Prevent racing

Some missing parts:

1. UI/UX
2. Embedding font

BTW, windows lacks some dll now, I will fix it later.

npes87184:
@rom1v 

I find that in windows, to make this feature working, we need a lot of dll and it is annoy to use current method to manage a lot of dlls.
Do you have another good idea about this problem?

Thank you.

PR #1543, opened 2020-06-25
Linguistic improvement

Skaldebane:


rom1v:
Thanks :+1:

Merged as 8c27f59aa571504e2e038dc5386e2be6897cec32 (I changed the subject line to imperative mood, I hope it's correct in English :wink:).

Skaldebane:
@rom1v it would be better to replace "linguistic" with "grammar" 😁
"Linguistic" here is an adjective (description just like "big", "small", "tired" and so on) so it's incorrect to use it after a verb 😄
And thanks for this amazing software! I depend on it daily as my phone got broken nowadays ❤️❤️

rom1v:
> it would be better to replace "linguistic" with "grammar"

Too late, it's committed :D Sorry :)

PR #1576, opened 2020-07-08
1572/turn on screen when exit

### LINKED ISSUE #1572: Turn the screen on if the connection with scrcpy is dropped abruptly
Sometimes I disconnect my phone from the PC and the screen is left off.
I may forget that the phone is actually not locked and not even asleep and I leave the phone like that possibly for too long.
Also, my wifi is not very stable and I'm enough of a security freak but I also love the convenience so I care to press the "OK" every time I connect to the phone in debug mode.

By adding this behavior, I can control my phone immediately when a disconnect happens without having to press power twice (or 3 times, if the phone is already asleep):

In sum, reduce these steps:
1. Wake up (but screen doesn't) (Only if asleep)
2. Sleep + lock
3. Wake up; screen wake up
4. Fill in the pattern
5. (confirm fingerprint) Press OK

To:
1. (confirm fingerprint)  Press OK

Extra:
I intent to implement this myself. It can also allow me to familiarize with these workings of android.
This issue is to confirm if I have green light for such feature in scrcpy

brunoais:
This was more straightforward than I expected.
Does this seem to be correctly done?

closes #1572

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
@@ -76,7 +76,7 @@ public final class CleanUp {
             }
             if (scheduleScreenOn) {
                 Ln.i("Restoring \"displayPowerMode\"");
-                new Device(new Options()).setScreenPowerMode(Device.POWER_MODE_NORMAL);
+                Device.setScreenPowerMode(Device.POWER_MODE_NORMAL);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


rom1v:
> Does this seem to be correctly done?

Yes, thank you :+1: It works great.

I did the following changed:
 - squash your 2 commits
 - rebase on current `dev`
 - rename `scheduleScreenOn` to `restoreNormalPowerMode` (there is no scheduling here)
 - moved `if (restoreNormalPowerMode)` outside the other if-block (there is no need to create a `ServiceManager` for that option)

Please review: [`pr1576`](https://github.com/Genymobile/scrcpy/commit/pr1576) :wink:

brunoais:
LGTM :+1:

rom1v:
Thank you. Merged into `dev`: 30714aba34e8b447159beaebeaf4eeab7bd33907 :+1:

PR #1577, opened 2020-07-08
1573/Turn the screen back off if it's known to turn on by an action in scrcpy

### LINKED ISSUE #1573: When the screen is off, if the screen turns on from a command in the PC, order it to turn back off
Sometimes I get distracted by my phone when the screen turns back on and I'm working on it.
I have attention problems from flashing lights and moving things which atract me too much and the phone screen turning on and me having to act because of that is part of an issue to me.

1. If these are true:
1.1 `Ctrl+o` or `--screen-off` (command line option) is more recent than `Ctrl+Shift+o`
1.2. A command is sent to the phone, through `scrcpy` that is known to turn the screen on.
1.3. The screen tries to turn on in the following X (maybe 5) seconds.
1.3.1. 1.1. is still true
---> Turn the screen back off ASAP

I tried to use an automation tool that detects right clicks and `Ctrl+p` on the desktop app but it works like a terrible hack. Sometimes the command goes too early, other times the second command is mixed together with something else I'm trying to do.
In those cases, I have already ended up giving orders for things I didn't mean to.

Extra:
I intent to implement this myself. It can also allow me to familiarize with these workings of android.
This issue is to confirm if I have green light for such feature in scrcpy

brunoais:
On my phone (cheap phone running Android 6), this hack works.
I was unable to find a decent way to achieve this otherwise.

What do you think? Any better ideas?

closes #1573

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,7 +439,8 @@ scrcpy -S
 
 Or by pressing `Ctrl`+`o` at any time.
 
-To turn it back on, press `Ctrl`+`Shift`+`o` (or `POWER`, `Ctrl`+`p`).
+To turn it back on, press `Ctrl`+`Shift`+`o` (or the phone's physical `POWER` button).
+Otherwise, scrcpy will do effort (read: should work) to keep the screen off.
 
 It can be useful to also prevent the device to sleep:
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -566,6 +566,7 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
         {"show-touches",           no_argument,       NULL, 't'},
         {"stay-awake",             no_argument,       NULL, 'w'},
         {"turn-screen-off",        no_argument,       NULL, 'S'},
+        {"keep-screen-off",        no_argument,       NULL, 'S'},
         {"verbosity",              required_argument, NULL, 'V'},
         {"version",                no_argument,       NULL, 'v'},
         {"window-title",           required_argument, NULL, OPT_WINDOW_TITLE},
@@ -649,6 +650,7 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
                 opts->serial = optarg;
                 break;
             case 'S':
+                opts->keep_screen_off = opts->turn_screen_off;
                 opts->turn_screen_off = true;
                 break;
             case 't':
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -311,6 +311,7 @@ scrcpy(const struct scrcpy_options *options) {
         .lock_video_orientation = options->lock_video_orientation,
         .control = options->control,
         .display_id = options->display_id,
+        .keep_screen_off = options->keep_screen_off,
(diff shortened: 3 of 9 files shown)


xJigs:
i mean im using adb now, but i want to go back to wired

brunoais:
@rom1v Any comments or sth?

rom1v:
@brunoais I didn't have the time to test yet.

However, as I said in https://github.com/Genymobile/scrcpy/issues/1573#issuecomment-654886965, I still think this is too "hacky" and may cause problems. For example if 2 scrcpy are started for the same device (this is already a problem for cleanup, but here it's only a one-time problem), or it does not behave the same way if we press power physically or via Ctrl+p, and probably other problems…

brunoais:
> @brunoais Sorry I didn't have the time to test yet.
Ah OK. I didn't know whether you didn't have time or it was de-prioritized or forgotten.

You are right. The behavior is different between acting directly on the device and on scrcpy. Howeverr if there are more than one instance of scrcpy connected to the same device, I don't see what kind of problem this may really create.
However, I do understand the inconsistency this can create.

What about we get this feature opt-in by running it behind a command line argument, with (hopefully) proper explanation about it? Would that be a good compromise?

brunoais:
I made this feature opt-in.
Even though it's execution is still not so good, maybe with some decent docs about it, it it works acceptably well.

rom1v:
I quickly tested before my holidays, it worked quite well. I might consider merging with this behavior. But I will do more tests before. But probably not before few weeks.

brunoais:
Sure. Thank you for keeping me updated.
I'll wait and let me know how to update the README to be more clear to users. I'll try to get some ideas myself.

rom1v:
I acknowledge that it is convenient especially on right-click to turn the screen on (keeping the physical device screen off), even if `--stay-awake` (while plugged in) already provided a (partial) solution.

I took some of your commits and squashed them (I didn't make it optional, so I didn't include `Create keep_screen_off command line option`): 2a7f3d70d71730b7a79a1b0ce626a5ea772c5cbe

 - I replaced magic numbers by `KeyEvent` constants (like `KeyEvent.KEYCODE_POWER`).
 - I moved the test `screenStayOff` to the callers
 - I also moved the calls to `scheduleScreenOff()` to the injection methods (as a side-effect, this avoids to call it on right-click when it actually injects BACK).

I also suggest to replace `Timer` by `Executor`: https://stackoverflow.com/questions/409932/java-timer-vs-executorservice
as follow: e31b13b0b2a851ad421869e2290eed4feae39545

The branch is [`pr1577`](https://github.com/Genymobile/scrcpy/commits/pr1577). Are you ok with these changes?

_Also, I will probably rename "screen off" to "power mode off" (for example `schedulePowerModeOff()`), to avoid confusion with the existing "turn screen on" (in particular, `pressBackOrTurnScreenOn()`, when it turns the screen on, schedules "screen off", which is confusing). _

I don't know whether I will merge it yet. More opinions are welcome :)

brunoais:
@rom1v I approve all the changes you did.

rom1v:
Merged cf9d4497

PR #1623, opened 2020-07-27
Add no-repeat cli option

xeropresence:
Still need documentation, not sure what you would want to put in there.

This modifies event_loop to pass down the option struct instead of passing the individual options as it seemed silly to expand the function to add a third param.

Not sure of the point of 

https://github.com/xeropresence/scrcpy/commit/434ffbdbc1395de4a8382345fcc19d1c745a3d14#diff-4dd0bcc81fa9404d5dd8572e7682d3c5R245

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -531,6 +531,7 @@ guess_record_format(const char *filename) {
 #define OPT_CODEC_OPTIONS          1018
 #define OPT_FORCE_ADB_FORWARD      1019
 #define OPT_DISABLE_SCREENSAVER    1020
+#define OPT_NO_REPEAT              1021
 
 bool
 scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
@@ -553,6 +554,7 @@ scrcpy_parse_args(struct scrcpy_cli_args *args, int argc, char *argv[]) {
         {"no-control",             no_argument,       NULL, 'n'},
         {"no-display",             no_argument,       NULL, 'N'},
         {"no-mipmaps",             no_argument,       NULL, OPT_NO_MIPMAPS},
+        {"no-repeat",              no_argument,       NULL, OPT_NO_REPEAT},
         {"port",                   required_argument, NULL, 'p'},
         {"prefer-text",            no_argument,       NULL, OPT_PREFER_TEXT},
         {"push-target",            required_argument, NULL, OPT_PUSH_TARGET},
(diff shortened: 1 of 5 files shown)


besteducationpage:
Try community Forum

besteducationpage:
www.besteducationpae.com

Thanks for your comment
We will answer your question very soon

rom1v:
Thank you. :+1:

> This modifies event_loop to pass down the option struct instead of passing the individual options as it seemed silly to expand the function to add a third param.

I moved the `control` flag to the initialization (the same way as it worked for `prefer_text`: 74079ea5e4ce0282a5d078cfa80fc68d2d9da16e
Then I split your commit to first pass the whole `scrcpy_options` in `scrcpy.c`: 65d06a36634287426206b459d2f290142ea1bf46
then implement the option: https://github.com/Genymobile/scrcpy/commit/94269ccaad38b35fe2ae0e20dd1b81f97068e5c0

(I renamed to `--no-key-repeat`)

> Still need documentation, not sure what you would want to put in there.

I added documentation in that last commit.

The branch is [`pr1623`](https://github.com/Genymobile/scrcpy/commits/pr1623). Could you review and test, please?

---

> Not sure of the point of https://github.com/xeropresence/scrcpy/commit/434ffbdbc1395de4a8382345fcc19d1c745a3d14#diff-4dd0bcc81fa9404d5dd8572e7682d3c5R245

if the `#ifdef CONTINUOUS_RESIZING_WORKAROUND` block was not included, then `display` is not used, and it generates a warning:

```
../app/src/scrcpy.c: In function ‘event_loop’:
../app/src/scrcpy.c:248:17: warning: unused parameter ‘display’ [-Wunused-parameter]
  248 | event_loop(bool display, bool control) {
      |                 ^
```

xeropresence:
Seems to work fine.

 I see you inverted the naming scheme of the option.

https://github.com/Genymobile/scrcpy/pull/1623/commits/434ffbdbc1395de4a8382345fcc19d1c745a3d14#diff-1e7437254ef4ba8f538e75a3bd53846fR414

> options->ignore_key_repeat

vs 

https://github.com/Genymobile/scrcpy/commit/94269ccaad38b35fe2ae0e20dd1b81f97068e5c0#diff-1e7437254ef4ba8f538e75a3bd53846fR465

>!im->forward_key_repeat

I think ignore_key_repeat is easier to understand at first glance rather then !im->forward_key_repeat, but other then that it all looks good.

xeropresence:
With display no longer being a variable then that should no longer be needed as well right?

rom1v:
> I think ignore_key_repeat is easier to understand at first glance rather then !im->forward_key_repeat

The idea is to use boolean flags with a "positive" meaning. For example `--no-display`, `--no-control`, `--no-mipmaps` are stored as booleans `display`, `control` and `mipmaps` (to avoid conditions like `if (!no_display)`). For consistency, `--no-key-repeat` could be stored in a bool `key_repeat` (but I added `forward` to make it clear that we don't forward them even if they are received, and `--no-forward-key-repeat` is too long).

Here, the condition in the block is negative because it's an [early return](https://medium.com/better-programming/are-early-returns-any-good-eed4b4d03866) (an early return is often used in the negative form):

```c
if (im->forward_key_repeat) {
    // forward event
}
```

vs

```c
if (!im->forward_key_repeat) {
    return;
}
// forward_event
```

`ignore_key_repeat` is not bad , but I think `forward_key_repeat` is more consistent.

> With display no longer being a variable then that should no longer be needed as well right?

It's removed: https://github.com/Genymobile/scrcpy/commit/65d06a36634287426206b459d2f290142ea1bf46#diff-4dd0bcc81fa9404d5dd8572e7682d3c5L248

rom1v:
Merged 84f1d9e3

PR #1626, opened 2020-07-27
Traditional Chinese translation for README

RaenonX:
- Added Traditional Chinese translation for README.md

PR #1644, opened 2020-08-07
Traditional Chinese translation for README

RaenonX:
Closes #1626


RaenonX:
Duplicated with #1626

PR #1670, opened 2020-08-13
Send wakeup keycode alongside trying to turn screen on when restoring

brunoais:


server/src/main/java/com/genymobile/scrcpy/CleanUp.java
```diff
@@ -78,7 +80,10 @@ public static void main(String... args) {
 
         if (restoreNormalPowerMode) {
             Ln.i("Restoring normal power mode");
-            Device.setScreenPowerMode(Device.POWER_MODE_NORMAL);
+            Device device = new Device(new Options());
+            if(device.isScreenOn()){
+                Device.setScreenPowerMode(Device.POWER_MODE_NORMAL);
```
brunoais, line 85: @rom1v Is it better this way? It works too with my phone.

rom1v:
> LGTM.

:+1: Merged into `dev`.

> Doesn't work on my phone. I need this API to turn it back on, otherwise, the double-tab turns off and locks and then double-tab turns it back on.

Sorry, I don't get it.

With the following steps:
 - `scrcpy --turn-screen-off` (physical screen off, mirror screen on)
 - <kbd>Alt</kbd>+<kbd>p</kbd> (physical screen off, mirror screen off)
 - close scrcpy
 - double-tap on the physical screen

Without this change, it stays black. With this change, it works. (Initially, I thought it would be the contrary)

> True, that's by design.

:+1:

Btw, double-tap with a double-click in scrcpy works to turn the screen off, but not to turn it back on.

brunoais:
> With the following steps:
> 
>     * `scrcpy --turn-screen-off` (physical screen off, mirror screen on)
>     * Alt+p (physical screen off, mirror screen off)
>     * close scrcpy
>     * double-tap on the physical screen
>
> Without this change, it stays black. With this change, it works. (Initially, I thought it would be the contrary)


**Without this change:**
On my phone, that 3rd step would turn the screen and display on; ready to unlock. Then the double-tap would turn the screen off.
Closing scrcpy turns on the screen and the display.   

Seems like in phones like yours, only the screen turns on but not the display (no image but light on). However, the OS changes mode to "awaken" (or whatever name). So when you double-tap it, it turns off both screen and display.

**With this change:**
I guess we get consistent behavior because now my phone behaves as you describe in the quote above.  
If the display is on but the screen if off, the screen is turned back on.
If both display and screen are off, they stay off.

Makes sense?

PR #1701, opened 2020-08-24
Turn on device screen when scrcpy window become active

excitoon:


rom1v:
Thank you for the proposal.

But isn't it a bit overkill to wake up the device every time the window gets focus? A simple right-click does the same thing.

excitoon:
> Thank you for the proposal.
> 
> But isn't it a bit overkill to wake up the device every time the window gets focus? A simple right-click does the same thing.

Well, it would pretty user-friendly, I think. Why would we ever need a window with a black screen? I mean, in that scenario where we frequently use the device through `scrcpy`.

I shall solve few problems here:
1) `SDL_ActiveEvent` is not present in SDL 2.0;
2) check that real screen does not turn on in `-S`;

rom1v:
> Why would we ever need a window with a black screen?

But IMO having the window focused or not should not change the behavior on the device. Moreover, if you focus the window and waits without touching the device, the window will have focus and the device screen will still be black.

So for that use case, I think `--stay-awake` is a better solution. And otherwise, right-click to power on the screen is straightforward enough.

excitoon:
My point is that as a user, I don't need something black on my screen, because that would be waste of screen space. If I focus the window, then it's for a purpose, therefore I need to communicate with that window.

`-w` is a good thing as well, but it does not work well for me because it interfere with activity state - device would think that I am active, even though I'm not.

excitoon:
I would also propose swipe sequence to send to the device in order to unlock it after turning it on.

excitoon:
I shall test my changes for a while. I suspect there is a bug.

excitoon:
I've fixed it. @rom1v may I ask one thing, I have minor thing when I switch to `scrcpy` application with Alt+Tab keystroke, after I release Tab it is being sent to the device unfortunately (and make some actions there sometimes). How would we avoid that?

rom1v:
For reasons explained in https://github.com/Genymobile/scrcpy/pull/1701#issuecomment-679085488, I'm closing this PR.

> I have minor thing when I switch to scrcpy application with Alt+Tab keystroke, after I release Tab it is being sent to the device unfortunately (and make some actions there sometimes). How would we avoid that?

The event is generated by SDL, I already encounter such problems. I guess it's a SDL bug.

excitoon:
Can you fix it in SDL?

rom1v:
SDL is a dependency of scrcpy. I'm not familiar with the codebase and don't have time to invest in it.

PR #1716, opened 2020-08-27
Fix server build script by using more portable and safer bash and rely on PATH when no ANDROID_HOME

ljmf00:
See the commit messages.

### DIFF
--- a/server/build_without_gradle.sh
+++ b/server/build_without_gradle.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # This script generates the scrcpy binary "manually" (without gradle).
 #
--- a/server/build_without_gradle.sh
+++ b/server/build_without_gradle.sh
@@ -17,14 +17,17 @@ SCRCPY_VERSION_NAME=1.16
 PLATFORM=${ANDROID_PLATFORM:-29}
 BUILD_TOOLS=${ANDROID_BUILD_TOOLS:-29.0.2}
 
-if [ -z ${ANDROID_HOME+x} ]; then
+if [ "$ANDROID_HOME" != "" ]; then
     AIDL="$ANDROID_HOME/build-tools/$BUILD_TOOLS/aidl"
     DX="$ANDROID_HOME/build-tools/$BUILD_TOOLS/dx"
+    ANDROID_PLATFORMS_PATH="$ANDROID_HOME/platforms"
 else
     [ -x "$(which dx)" ] && DX="dx" \
         || (echo "Please provide \$ANDROID_HOME or set \$PATH to include 'dx' executable"; exit 1)
     [ -x "$(which aidl)" ] && AIDL="aidl" \
         || (echo "Please provide \$ANDROID_HOME or set \$PATH to include 'aidl' executable"; exit 1)
+    [ "$ANDROID_PLATFORMS_PATH" == "" ] && \
+        (echo "Please provide \$ANDROID_HOME or set \$ANDROID_PLATFORMS_PATH"; exit 1)
 fi
 
 BUILD_DIR="$(realpath ${BUILD_DIR:-build_manual})"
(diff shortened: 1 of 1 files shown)


server/build_without_gradle.sh
```diff
@@ -9,14 +9,24 @@
 #
 #     BUILD_DIR=my_build_dir ./build_without_gradle.sh
 
-set -e
+set -eo pipefail
```
rom1v, line 12: Isn't it too violent?

```bash
set -eo pipefail
result=$(echo a | grep b)
```

server/build_without_gradle.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
rom1v, line 1: Ref #426

OK, I don't know which one is good practice (`/bin/bash` seems less risky to me, but why not).

server/build_without_gradle.sh
```diff
@@ -9,14 +9,24 @@
...
+if [ -z ${ANDROID_HOME+x} ]; then
+    AIDL="$ANDROID_HOME/build-tools/$BUILD_TOOLS/aidl"
+    DX="$ANDROID_HOME/build-tools/$BUILD_TOOLS/dx"
+else
+    [ -x "$(which dx)" ] && DX="dx" \
+        || (echo "Please provide \$ANDROID_HOME or set \$PATH to include 'dx' executable"; exit 1)
+    [ -x "$(which aidl)" ] && AIDL="aidl" \
+        || (echo "Please provide \$ANDROID_HOME or set \$PATH to include 'aidl' executable"; exit 1)
+fi
+
```
rom1v, line 29: For info, why do you need this? Where do your `dx` and `aidl` come from?

(That way, the `dx` and `aidl` version are not necessarily from the expected platform.)

ljmf00: > For info, why do you need this? Where do your dx and aidl come from?

The problem I'm trying to solve here is building `scrcpy` completely from scratch with `android-tools` compiled from source. On Arch Linux `scrcpy` is built using a prebuilt `scrcpy-server`. 

Because AOSP source has a huge and monolithic build system and because there's no proper tagging, Arch Linux, to build everything from source, need to do this: https://github.com/archlinux/svntogit-community/blob/packages/android-tools/trunk/generate_build.rb#L3 .

> (That way, the dx and aidl version are not necessarily from the expected platform.)

You're right. In fact, I didn't covered this, if you depend on a specific `$PLATFORM` and `$BUILD_TOOLS` versions, then, this is a problem.

server/build_without_gradle.sh
ljmf00: It depends on what behavior you want here. You might want something like this:`echo a | grep b || true`, instead.

ljmf00:
I added 3e0dfb1 patch to complete 1cf5918 .

server/build_without_gradle.sh
ljmf00: It depends of the portability you want to have here. I do think it's a good practice for 2 major reasons:

1. All Linux distributions I'm aware of are compatible with the approach I described in the patch.

2. Some BSD-based distributions like FreeBSD doesn't support your approach. See here: https://forums.freebsd.org/threads/bash-not-found.31849/ .

The `env` binary simply search for the executable in the provided PATH environment variable, so, here, you rely on it instead of providing the full path.

I didn't noticed the previous patch you referred, so if you're good with this, I can rebase and drop it.

rom1v:
I merged the "shebang" patch (02a882a0a2fbe40504889bf753aa055bfc3c4fdf) in to `dev`, as well as #426 and 1c44dc2259f7111f8432fcf1ab8509c68f48d9a0.

ljmf00:
> I merged the "shebang" patch ([02a882a](https://github.com/Genymobile/scrcpy/commit/02a882a0a2fbe40504889bf753aa055bfc3c4fdf)) in to `dev`, as well as #426 and [1c44dc2](https://github.com/Genymobile/scrcpy/commit/1c44dc2259f7111f8432fcf1ab8509c68f48d9a0).

That said, I'm closing this.

Issue #1421, opened 2020-05-23
Scrcpy via ssh directly to android device (no computer+usb in between)


kaefert:
Hi there!

Thanks for this wonderful piece of software! It works absolutely amazing via USB, perfect quality, super low latency, simply amazing for controlling my phone through my computer! Much better than any device emulator I've seen! :)

Now I also have an android media player remote somewhere that I can connect to directly via ssh (installed and started through [Termux](https://wiki.termux.com/wiki/Main_Page)) which I'd like to remote control using scrcpy

---------------------

 - [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
See above and below.

**Describe the solution you'd like**
Connect directly to android device via ssh without computer with usb-connection in between. I can't use wireless adb without ssh, because the internet is in between and only the ssh port is forwarded through the NATing router.

**Describe alternatives you've considered**
Running `tsudo screencap -p see.png ; tsudo chmod 666 see.png` on the device and then using sftp to look at the image.
When I need to click something I use `su -c input tap $x $y`. That method is very slow and annoying for manual control though.

**Additional context**
The android media player is used to drive a public display of sorts, and most of the time I try to avoid going there for maintance and instead use ssh+sftp.

All I could find relating to scrcpy + ssh seems to talk about the scenario where you have a remote computer that's connected to the android device via usb and the local computer connects to the remote one via ssh and uses port forwards through the ssh tunnel to access the adb server on that computer to talk to the android device.

There's also some information how to use wireless adb for Scrcpy but I couldn't find anything on how to do it through ssh without a computer with adb server on the remote locaton.

kaefert:
So I did some more research and found the following related threads:
https://github.com/Genymobile/scrcpy/issues/5
https://github.com/Genymobile/scrcpy/issues/144
https://github.com/Genymobile/scrcpy/issues/242
https://github.com/Genymobile/scrcpy/issues/289

rom1v:
The _scrcpy_ client necessarily connects to the device via `adb`. In particular, it executes the _scrcpy server_ via `adb shell` to get shell permissions (to capture the video without asking and to inject events).

[...]

PR #1741, opened 2020-09-06
Added `use-ssh` option

excitoon:
Progress on #1421

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -651,6 +651,7 @@ guess_record_format(const char *filename) {
 #define OPT_DISABLE_SCREENSAVER    1020
 #define OPT_SHORTCUT_MOD           1021
 #define OPT_NO_KEY_REPEAT          1022
+#define OPT_USE_SSH                1024
 
 bool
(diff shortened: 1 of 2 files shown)


excitoon:
Things to be done:
[ ] reverse / forward connections;
[ ] copying server binary.

excitoon:
At this point there are (1) only reverse connections implemented, (2) forwarding is not ideally stable, but generally it works.

excitoon:
About (2), I would test this build for a while. Also, it will be needed to have our own NDK-based tool instead of `socat`.

excitoon:
In addition to redirection of sockets this tool also makes sure that server is killed after it receives EOF in STDIN (when client disconnects).

excitoon:
In current version there is no requirement to turn on "USB debugging (security settings)", assuming you run as root.

excitoon:
I found that forward mode is working much more stable than reverse one. @rom1v do you know if application can miss connections to it?

excitoon:
Fixed that in 02fc377e9520a1b42fda7f29b475b3dd8364ffc5

e-t-l:
Any progress on this? It has the potential to be a major enhancement to SCRCPY.

PR #1745, opened 2020-09-07
Improved stability of making connections in reverse mode

excitoon:


rom1v (commented):
Hi,

Sorry, I don't see why it would have any impact on "connection stability".

excitoon:
Ah, I see, it really won't help as you make `connect` one by one. But it may help in my case (though it is still pretty rare case).

PR #1751, opened 2020-09-10
Using ADBKeyboard for injecting text

excitoon:


server/src/main/java/com/genymobile/scrcpy/Controller.java
```diff
@@ -145,18 +147,28 @@ private boolean injectKeycode(int action, int keycode, int repeat, int metaState
...
-        if (events == null) {
-            return false;
-        }
-        for (KeyEvent event : events) {
-            if (!device.injectEvent(event)) {
+        if (options.useADBKeyboard()) {
+            // Process latin keys the same way in order to provide same reaction speed.
+            try {
+                Process process = Runtime.getRuntime().exec("am broadcast -a ADB_INPUT_CHARS --eia chars " + String.valueOf((int) c));
+                return process.waitFor() == 0;
```
rom1v, line 154: You should send the broadcast programmatically (to avoid execute a new process and add latency for every letter).

I guess this involves retrieving an `IActivityManager` by calling [`ActivityManager.getService()`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/ActivityManager.java#4070) (by introspection) (or [`ActivityManagerNative.getService()`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/ActivityManagerNative.java#50) on older versions), and calling [`broadcastIntent()`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/IActivityManager.aidl#120).

app/src/cli.c
```diff
@@ -651,6 +651,7 @@ guess_record_format(const char *filename) {
 #define OPT_DISABLE_SCREENSAVER    1020
 #define OPT_SHORTCUT_MOD           1021
 #define OPT_NO_KEY_REPEAT          1022
+#define OPT_USE_ADB_KEYBOARD       1026
```
rom1v, line 654: Some numbers don't exist :trollface:

rom1v (commented):
Thank you for the PR :+1:

Ideally, I would like the IME code to be included in scrcpy-server: https://github.com/Genymobile/scrcpy/issues/37#issuecomment-681676085

Moreover, this would also allow a socket connection between the server and the IME (to minimize latency).

However, until this is implemented (I have zero time to work on this currently), I think calling ADBkeyboard is an acceptable solution.

If this option is passed, I think it should also save the current IME, set the ADBKeyboard IME, and it should restore the previous one on cleanup.

server/src/main/java/com/genymobile/scrcpy/Controller.java
excitoon: Is that method stable enough? I only have A9 devices and there is something made about activities in A10. I'm not Android developer though.

app/src/cli.c
excitoon: > Thank you for the PR +1
> 
> Ideally, I would like the IME code to be included in scrcpy-server: [#37 (comment)](https://github.com/Genymobile/scrcpy/issues/37#issuecomment-681676085)
> 
> Moreover, this would also allow a socket connection between the server and the IME (to minimize latency).
> 
> However, until this is implemented (I have zero time to work on this currently), I think calling ADBkeyboard is an acceptable solution.
> 
> If this option is passed, I think it should also save the current IME, set the ADBKeyboard IME, and it should restore the previous one on cleanup.

All of these is right. :+1:

Artur202030:
https://github.com/Lurker00/scrcpy/releases/tag/v1.10L0
The new features are:
-10-fingers multitouch support.
-Any language input support with an aid of ADBKeyBoard (included).
...

rom1v:
Hi,

I just implemented what I suggested [here](https://github.com/Genymobile/scrcpy/pull/1751#discussion_r486393847), and now it works great (without latency).

You could test my work-in-progress branch [`adbkeyboard`](https://github.com/Genymobile/scrcpy/commits/adbkeyboard).

```bash
scrcpy --use-adb-keyboard
```

What remains to do:
 - automatically select the ADB keyboard on start
 - automatically restore the initial keyboard on stop (even if that won't work if we start 2 instances of scrcpy with ADBKeyboard, since the second one will consider that the initial keyboard was ADBKeyboard).
 - some cleanup

excitoon:
Hey.

You can just never restore keyboard to ADBKeyboard, that shall solve the
problem.

Kind regards,
Vladimir.

rom1v:
A future next step could be to include the IME in scrcpy (as a real android app), to avoid the requirement to install a separate app.

I'm wondering if it's still possible to create a real/installable android app without gradle (using build-tools directly). The reason is that the official debian package uses [`build_without_gradle.sh`](https://github.com/Genymobile/scrcpy/blob/master/server/build_without_gradle.sh) because the full Android SDK is not available in Debian official repositories.

Artur202030:
hi,
both English and Russian letters are printed simultaneously.

rom1v:
@Artur202030 Try with:

```
scrcpy --use-adb-keyboard --prefer-text
```

rom1v:
An issue with using the existing `ADBKeyBoard` is the absence of permissions.

As a consequence, once enabled, any installed app could inject any text or events (anything managed by [`AdbIME`](https://github.com/senzhk/ADBKeyBoard/blob/master/keyboardservice/src/main/java/com/android/adbkeyboard/AdbIME.java)) as they like. This is a security issue IMO.

To avoid the problem, the IME should declare a permission, and with `shell` rights we can send a broadcast with this permission. In short, [the equivalent of](https://stackoverflow.com/a/55824554/1987178) (not tested):
```
adb shell am broadcast -a com.com.myexample.MY_ACTION --receiver-permission com.myexample.permission.MY_PERMISSION
```

Note that this may only work for broadcasts, not services or activities. For these cases (or if the solution above does not work), we could [require some shell permission](https://stackoverflow.com/a/35194227/1987178). But it will cause issues for XiaoMI devices: https://github.com/Genymobile/gnirehtet/issues/5 (and _gnirehtet_ may not be started by broadcast receivers https://github.com/Genymobile/gnirehtet/commit/f8681809a28a0073f791f0d9eae65da02ae73c3d).

Artur202030:
> @ Artur202030 Попробуйте:
> 
> ```
> scrcpy --use-adb-keyboard --prefer-text
> ```
so normal but 1 sec delay

rom1v:
> so normal but 1 sec delay

Which branch?

Artur202030:
i tested your code and **excitoon**. Mixed up files. 
Sorry.

rom1v:
Could you confirm that there is no delay on my branch (https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724268986)?

Artur202030:
> Could you confirm that there is no delay on my branch ([#1751 (comment)](https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724268986))?

Your code works without delay.
Thanks.

quyleanh:
@excitoon could you please take a look at this PR again?

quyleanh:
@rom1v I have built my self `[adbkeyboard](https://github.com/Genymobile/scrcpy/commits/adbkeyboard) and used `scrcpy --use-adb-keyboard`, but I still cannot enter Vietnamese character.

Is there any way to report log?

quyleanh:
@rom1v could you please take a look at this PR?

rom1v:
@quyleanh I implemented broadcasts so that it works without latency (https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724268986), but the absence of permissions of AdbKeyboard is a problem (https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724532949). Also, scrcpy should enable/disable the keyboard on start/stop.

I consider adding a feature to "install" the scrcpy server instead. That way, it could provide its own IME: https://github.com/Genymobile/scrcpy/issues/1880#issuecomment-725655830

> but I still cannot enter Vietnamese character.

It should allow to inject any UTF-8 chars (try to [copy paste](https://github.com/Genymobile/scrcpy/#copy-paste) a vietnamese character with <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>v</kbd> to inject characters).

quyleanh:
@rom1v thank you for your kind response.

> It should allow to inject any UTF-8 chars (try to [copy paste](https://github.com/Genymobile/scrcpy/#copy-paste) a vietnamese character with <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>v</kbd> to inject characters).

I know that copy paste feature. But I would like to inject UTF-8 characters directly from keyboard.

> @quyleanh I implemented broadcasts so that it works without latency ([#1751 (comment)](https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724268986)), but the absence of permissions of AdbKeyboard is a problem ([#1751 (comment)](https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724532949)). Also, scrcpy should enable/disable the keyboard on start/stop.
> 
> I consider adding a feature to "install" the scrcpy server instead. That way, it could provide its own IME: [#1880 (comment)](https://github.com/Genymobile/scrcpy/issues/1880#issuecomment-725655830)

So in summary, we will not have inject UTF-8 character directly from keyboard soon? As I understand, there are a lot of problem needed to do.

rom1v:
> I know that copy paste feature. But I would like to inject UTF-8 characters directly from keyboard.

Yes, I know, but since you said that Vietnamese characters were not injected, I wanted to check where it failed.

There are two copy-paste:
 - once using the clipboard, where any byte stream work (including UTF-8)
 - once injecting the characters, which acts as if every character was pressed from a keyboard (injected char by char), currently it only supports ASCII (+ some other accented chars)

If you use the second method with the ADBKeyboard, I expect that Vietnamese characters work. Is it the case?

quyleanh:
> If you use the second method with the ADBKeyboard, I expect that Vietnamese characters work. Is it the case?

Yes, I use the second method. There are only some Vietnamese characters work.
For example, if I type `ô` (unicode hex is `U+00F4`), it is fine. But if I type `ồ` (unicode hex is `U+1ED3`), the warning appreases.
`[server] WARN: Could not inject char u+1ed3`

You can refer to [this summary table](http://vietunicode.sourceforge.net/charset/).

rom1v:
> [server] WARN: Could not inject char u+1ed3

You're not running my branch `adbkeyboard` with `scrcpy --use-adb-keyboard`: https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724268986

quyleanh:
> You're not running my branch `adbkeyboard` with `scrcpy --use-adb-keyboard`: [#1751 (comment)](https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724268986)

I build your branch and run it now, but I forgot to use `scrcpy --use-adb-keyboard` syntax.
I checked again with `scrcpy --use-adb-keyboard`, but I cannot enter any Vietnamese characters now.
How can I provide log since there is no output or warning on terminal?

rom1v:
Try with AdbKeyboard alone: https://github.com/senzhk/ADBKeyBoard (without scrcpy).

quyleanh:
Thank you for your suggestion but I want to use scrcpy for both mirroring screen and injecting text.
There is no way after all?

rom1v:
Yes, I got it, but try with AdbKeyboard alone to check if Vietnamese characters work. If they don't, then it's expected that the `adbkeyboard` branch can't do it either. If it works, then there is a problem somewhere else.

quyleanh:
Thank you for your detail instruction. The [senzhk/ADBKeyBoard](https://github.com/senzhk/ADBKeyBoard) works well. This is my test:
```shell
adb shell am broadcast -a ADB_INPUT_TEXT --es msg 'Hôm nay trời rất là đẹp. Ồ ngoài kia có chú ngựa ô. Ổ gà. Ở nhà không đi học. Hỏi han. Hello?'
Broadcasting: Intent { act=ADB_INPUT_TEXT flg=0x400000 (has extras) }
Broadcast completed: result=0
```
Is there any other suggestion?

rom1v:
Test with ints:

```
adb shell am broadcast -a ADB_INPUT_CHARS --eia chars 7891
```
(7891 is 0x1ed3)

quyleanh:
It's fine. The result is `ồ`.

rom1v:
OK so in theory it should work with scrcpy on `adbkeyboard`. I don't know why it does not, but will not investigate for now, since in the end scrcpy should have its own IME (https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-744586560).

quyleanh:
Actually I have to use an external IME to type and inject Vietnamese on Windows. However other apps are OK with it.
In the previous test, all Vietnamese characters are injected from terminal since I already paste it in `adb` command. I don't know is there any different between injecting directly from `adb` command and injecting from other IME.

By the way, I'm glad to hear that you are working on a work around. Could you please push it in this repo, so I can try it? Or if it needs to test, just let me know.

Helaer:
@rom1v Using ADBKeyboard to input text is a good solution, which can solve the problem that Samsung and other devices cannot paste text, Why is there no progress?

yilksd:
I'm using ubuntu 20.04 + fcitx, and I have found 2 problems when using adbkeyboard:
1.adbkeyboard won't stop working when scrcpy window loses focus and I type in a QT application , so the characters also appear in scrcpy . The characters won't appear in scrcpy when I type in Firefox.
2.It seems there is a limitation in string's length. If I type a long sentence , only the first several characters can be injected. There is no such a limitation in adbkeyboard's ADB_INPUT_CHARS feature because I can inject such a long sentence using `am broadcast -a ADB_INPUT_CHARS --eia chars 'xxx' `, so maybe the limitation is in another place.

dieskim:
@rom1v - I was wondering if there is any update on "scrcpy having its its own IME" ? We tried scrcpy on our current stack and even with "--keyboard=uhid" we are still having clipboard paste issue on Samsung devices etc. Switching keyboard inputs on the device is also needed when switching between languages. In the past we have used Vysor with its on IME and that has been a good solution for different language inputs and clipboard management. 

We are happy to put in work to make this happen if thats something that could be helpful?

Thanks for all the work you do.

rom1v:
> I was wondering if there is any update on "scrcpy having its its own IME" ?

No, since most of the need is now covered by UHID keyboard.

I understand that there are still issues that UHID keyboard does not solve, but I am not keen on adding a new subproject just for that.

Installing AdbKeyboard can also be a problem for security: any app on the device could inject text (https://github.com/Genymobile/scrcpy/pull/1751#issuecomment-724532949).

dieskim:
Yes I understand. Thanks for getting back to me.

I wonder if adding a cli option that would enable a “keyboard broadcaster”,
or something similar, from scrcpy would be something that you might be
willing to add to the core?

Looking at how adbkeyboard was implemented before in the adbkeyboard
branch, we could capture the keyboard inputs and broadcast those, we could
even add a secret key for security that scrcpy could broadcast with the
broadcast and then a custom IME app could listen for those broadcasts. Not
sure if we could add to adbkeyboard but making something similar should be
possible and most likely not too hard.

Let me know what you think, we could look into it more. If it’s something
you would be willing to ado into the core if done correctly we could most
likely put in the work to get it done.

We have looked into using scrcpy in our development and testing stack
before but different language inputs and clipboard management has always
been an issue for us.

master255:
Fix compatibility with different languages for 2 years? Are you idiots?

Issue #1753, opened 2020-09-11
FFmpeg windows builds will be removed


rom1v:
I developed scrcpy for Linux, where the dependencies are just retrieved via `apt`.

But scrcpy had to work on Windows too. I'm bad at build systems, and have a very limited knowledge of Windows, so I initially just recommended to build from something which looks like a Linux environment (MSYS2). For the first public release (v1.0), I even just took the DLL installed from the MSYS environment.

But it was too cumbersome for me to create a release (I had to build from a Windows machine), so I wrote some script to cross-compile the binaries from Linux, and just downloaded prebuilt dependencies of FFmpeg and SDL (106b87a4d2459ae881ab89b987cc9347a843a4ea). This was a quick&dirty solution (which btw makes adding or upgrading dependencies very annoying). I'm not very proud of it, but at least it was good enough for solving my problem.

The FFmpeg binaries are downloaded from https://ffmpeg.zeranoe.com/builds/. This site will close soon:

> ffmpeg.zeranoe.com will close on Sep 18, 2020, and all builds will be removed. If you're using Zeranoe FFmpeg Builds in your product, please ensure that you have a hosting strategy in place by the 18th. Thank you to everyone who supported the project along the way!

so the release script of scrcpy will just stop to work correctly.

In the short term, I could just keep the latest versions and ship them. But it becomes urgent to build the dependencies along with scrcpy. And an option to build static binaries would be awesome (#1624).

https://linuxfr.org/users/roger21/journaux/ffmpeg-pour-windows-ca-va-couper

npes87184:
#1484 also suffered from dll hell. IMO, it can be solved in the same time?

Biswa96:
> I even just took the DLL installed from the MSYS environment.

MSYS2/MinGW FFmpeg builds are dynamically linked. So, you have to download all the dependencies which may not be the right solution here.

> The FFmpeg binaries are downloaded from https://ffmpeg.zeranoe.com/builds/. This site will close soon:

How those guys build it. Any script?

> so the release script of scrcpy will just stop to work correctly.

Neevvveeerrr. With CI, possibility is endless.

One lazy solution would be to create a package for MSYS2 which already provides FFmpeg and SDL2. But the package will not be a standalone like the current releases.

ouvtk:
Not sure if it will help much, but ffmpeg peeps already have a compilation guides: https://trac.ffmpeg.org/wiki/CompilationGuide
[...]

PR #1805, opened 2020-10-02
Adding new option for falling back to clipboard pasting on character injection failures.

### LINKED ISSUE #37: Non english keyboard layout

starrify:
So that non-ASCII characters may be entered via IMEs. Resolves #37.

Submitting the changes here for now for review. Further documentation work would be needed on [`README.md`](https://github.com/Genymobile/scrcpy/blob/56d237f152fc4d454a65fe594412b4bd9409fd64/README.md#input-control) and [`FAQ.md`](https://github.com/Genymobile/scrcpy/blob/56d237f152fc4d454a65fe594412b4bd9409fd64/FAQ.md#special-characters-do-not-work) either in this PR or as separate ones if this gets merged early.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -26,6 +26,11 @@ scrcpy_print_usage(const char *arg0) {
         "        Unit suffixes are supported: 'K' (x1000) and 'M' (x1000000).\n"
         "        Default is %d.\n"
         "\n"
+        "    --char-inject-fallback\n"
+        "        Fallback to the clipboard (copy and paste) when non-supported\n"
+        "        non-ASCII characters are typed.\n"
+        "        Would override the device clipboard content when it happens.\n"
+        "\n"
         "    --codec-options key[:type]=value[,...]\n"
         "        Set a list of comma-separated key:type=value options for the\n"
         "        device encoder.\n"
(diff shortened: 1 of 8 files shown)


rom1v:
Hi,

Thank you for the proposal.

Using this method to inject text has drawbacks: https://github.com/Genymobile/scrcpy/pull/1426#issuecomment-633448242
But behind a flag, and only as a fallback, it might be acceptable.

rom1v:
@npes87184 Any opinion about this?

npes87184:
I don't like this workaround. It seen to me that we will **mix** the concept between clipboard and text injection. IMO, we should find time to actually implement the IME into scrcpy to resolve the problem correctly.

BTW, @rom1v I guess you are doing static build for client now, right?

rom1v:
> From my point of view, instead of adding an option that will be removed eventually,

Does it work fine for you when you type chinese? One advantage is that it may work without installing an app.

Do you think that it could be enabled by default when the character may not be injected (so that there is no new option)?

> I guess you are doing static build for client now, right?

No, until #1753 is solved, I use the last version of FFmpeg built by zeranoe.

npes87184:
> > From my point of view, instead of adding an option that will be removed eventually,
> 
> Does it work fine for you when you type chinese? One advantage is that it may work without installing an app.
> 

I will give it a try recently (maybe tonight if possible). Yes, without needing to install an app is a benefit in this method and I haven't noticed this.

> Do you think that it could be enabled by default when the character may not be injected (so that there is no new option)?
> 

Because the benefit of not having to install apps, I change my mind, I think it is worth to add this workaround in this time. However, too many options is also a problem now. If possible, I stand for not adding new option for this method.

> > I guess you are doing static build for client now, right?
> 
> No, until #1753 is solved, I use the last version of FFmpeg built by zeranoe.

Ok, I will take a look in my free time for static build. BTW, don't give high expectation, I am really busy recently :P.

npes87184:
> > From my point of view, instead of adding an option that will be removed eventually,
> 
> Does it work fine for you when you type chinese? One advantage is that it may work without installing an app.
> 

It doesn't work for me. I ran scrcpy with `--char-inject-fallback`. And I cannot type chinese. Do I need to do something special? Like using some specific IME?

> Do you think that it could be enabled by default when the character may not be injected (so that there is no new option)?
> 
> > I guess you are doing static build for client now, right?
> 
> No, until #1753 is solved, I use the last version of FFmpeg built by zeranoe.

starrify:
> It doesn't work for me.

Hmm I'm sorry to know that. I also used an IME for Chinese for testing, and the environments are:
- `GNU/Linux x86_64 5.8.10-arch1-1`
- `Xorg 1.20.9`
- `Gnome-shell 3.36.6 / GDM 3.36.3`
- `ibus 1.5.22`

Could you share some info on your side so I may have a look?

quyleanh:
Is there any progress for this PR? @rom1v @npes87184

deltachonglee81 (changes requested):
Change the system

PR #1828, opened 2020-10-12
Changed type of got _picture to boolean

r4matero:


app/src/decoder.c
```diff
@@ -79,7 +83,7 @@ decoder_push(struct decoder *decoder, const AVPacket *packet) {
         return false;
     }
 #else
-    int got_picture;
+    bool got_picture;
```
rom1v, line 86: No, `avcodec_decode_video2` expects an `int *`, you can't pass a `bool *`.

app/src/decoder.c
```diff
@@ -15,6 +15,10 @@
 #include "util/buffer_util.h"
 #include "util/log.h"
 
+typedef enum bool {
+    false, true
+} bool;
```
rom1v, line 20: If you need a `bool`, include `<stdbool.h>`.

PR #1838, opened 2020-10-18
Use win64 ffmpeg built from BtbN

npes87184:
There are three changes in this PR.

1. In script `prepare-dep` the third parameter will be the output folder name now.
2. Stop building win32 version due to lack of ffmpeg libs
3. Use BtbN built ffmpeg (win64-lgpl-shared)

### DIFF
--- a/prebuilt-deps/Makefile
+++ b/prebuilt-deps/Makefile
@@ -20,14 +20,9 @@ prepare-ffmpeg-dev-win32:
 		ffmpeg-4.3.1-win32-dev
 
 prepare-ffmpeg-shared-win64:
-	@./prepare-dep https://ffmpeg.zeranoe.com/builds/win64/shared/ffmpeg-4.3.1-win64-shared.zip \
-		dd29b7f92f48dead4dd940492c7509138c0f99db445076d0a597007298a79940 \
+	@./prepare-dep https://github.com/BtbN/FFmpeg-Builds/releases/download/autobuild-2020-10-17-12-30/ffmpeg-n4.3.1-20-g8a2acdc6da-win64-lgpl-shared-4.3.zip \
+		db3a355a2569044ceb7b417aae0b4bed4a01b1ef0a762dd7a4610bd34d886203 \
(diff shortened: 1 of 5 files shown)


prebuilt-deps/Makefile
```diff
@@ -20,15 +20,10 @@ prepare-ffmpeg-dev-win32:
 		ffmpeg-4.3.1-win32-dev
 
 prepare-ffmpeg-shared-win64:
-	@./prepare-dep https://ffmpeg.zeranoe.com/builds/win64/shared/ffmpeg-4.3.1-win64-shared.zip \
-		dd29b7f92f48dead4dd940492c7509138c0f99db445076d0a597007298a79940 \
+	@./prepare-dep https://github.com/BtbN/FFmpeg-Builds/releases/download/autobuild-2020-10-17-12-30/ffmpeg-n4.3.1-20-g8a2acdc6da-win64-lgpl-shared-4.3.zip \
```
rom1v, line 23: One problem is that the releases disappear :/

nkh0472:
It's possible to fetch [the latest stable 64bit build of FFmpeg](https://www.gyan.dev/ffmpeg/builds/ffmpeg-release-full-shared.7z) and [the sha256sum of it](https://www.gyan.dev/ffmpeg/builds/sha256-release-full-shared) in https://www.gyan.dev/ffmpeg/builds/, more info about them can be seen in https://github.com/GyanD/codexffmpeg.

nkh0472:
Meanwhile, seems scripts in https://github.com/BtbN/FFmpeg-Builds could build win32 version.

https://github.com/BtbN/FFmpeg-Builds#targets-variants-and-addins

I'm not sure whether this can still be used at present. If so, compiling with GitHub server can save a lot of resources. If you only need a stable version, you can compile it once after it is released, and there is no need to build frequently. 

---
Since there is no need to build new versions frequently, maybe I can ask gyanD if he can provide a 32-bit version for the stable release.

rom1v:
#2952

Issue #1497, opened 2020-06-11
Use better resolution icon for app


OJ7:
Screenshot of CMD+Tab task switcher on Mac. Looks like the icon is very low resolution.

![image](https://user-images.githubusercontent.com/5333243/84422512-22ef6500-abeb-11ea-81ba-681792c57af9.png)

rainman0607:
[https://support.apple.com/guide/mac-help/create-custom-icons-files-folders-mac-mchlp2313/mac](url)

rom1v:
Please see #1987.

PR #1987, opened 2020-12-22
High resolution icon

rom1v:
There are two problems with the scrcpy icon:
 1. there is no scrcpy icon (it just uses a standard Android icon)
 2. it's low definition (48×48) (#1497)

This PR intends to fix the latter.

### Why was the icon in 48×48?

For the initial release of scrcpy, to set an icon quickly (and in a funny way), I used an [XPM](https://en.wikipedia.org/wiki/X_PixMap): [`icon.xpm`](https://github.com/Genymobile/scrcpy/blob/v1.0/app/src/icon.xpm).
I explained here: https://blog.rom1v.com/2018/03/introducing-scrcpy/#set-a-window-icon

I use XFCE, and the icons are never shown at a higher resolution, so it was not a problem for me. However, when the icon is displayed at a bigger size, the quality is awful: #1497.

### Icon format

To support a higher resolution icon, I had to use another format.

SDL expects a packed (non-planar) bitmap format, so I choose PNG (but other formats could be supported).

To avoid adding a new dependency, I implemented icon loading using FFmpeg (which is **not** the more straightforward way to just load an image!). For simplicity, I only implemented support for RGB-like input.

The icon is read from a file, either in `/usr/share/icons/hicolor/512x512/apps/scrcpy.png` or just `icon.png` for portable builds (i.e. Windows). Like `SCRCPY_SERVER_PATH`, an environment variable allows to customize the icon path: `SCRCPY_ICON_PATH`. The main purpose is to be able to execute `./run x` to run a build without installing.

### Debug mode

To easily distinguish debug builds, I used to change the icon color (because it was simple and funny): https://github.com/Genymobile/scrcpy/blob/431c9ee33b49f5ce7317b4c687e00bc436adbc02/app/src/tiny_xpm.c#L40-L43

To keep things simple with a real image, I will just abandon this debug-specific change.

### TODO

 - I'd like to add support for palette formats (`AV_PIX_FMT_PAL8`), in particular because it's the format produced by `optipng` if I optimize the current icon.
 - I'd like a real scrcpy icon :smile:


rom1v:
The downscaled 512×512 icon looks bad on Windows. But SDL does not expose a way to provide an icon set for each resolution :confused:

rom1v:
@varlesh Could you please set the logo on the [scrcpy wikipedia page](https://en.wikipedia.org/wiki/Scrcpy)?

When uploading an image, wikipedia requires to check the box _I am the author of this work_. :wink:

varlesh:
No problem,added
https://commons.wikimedia.org/wiki/File:Scrcpy_logo.svg

вс, 14 нояб. 2021 г. в 19:39, Romain Vimont ***@***.***>:

Issue #1975, opened 2020-12-15
scrcpy-noconsole compiled


douxiaoping:
I configured windows_noconsole to be true in meson_options. Ninja-cx compiled, but it still didn't work.
![image](https://user-images.githubusercontent.com/29052998/102163103-6f329f00-3ec5-11eb-9d6d-fb64d1b811ac.png)

douxiaoping:
#define WINDOWS_NOCONSOLE 
I defined this, and ninja-CX didn't work either

rom1v:
_(Please post console content as text instead of screenshot.)_

What do you mean "it still didn't work"? If you double click on the generated scrcpy.exe, it opens a new terminal?

> Directory already configured

(Here your command did not work, either `rm -rf x` before the command, or use `meson configure`.)

douxiaoping:
this is meson_options.txt
option('compile_app', type: 'boolean', value: true, description: 'Build the client')
option('compile_server', type: 'boolean', value: true, description: 'Build the server')
option('crossbuild_windows', type: 'boolean', value: false, description: 'Build for Windows from Linux')
option('windows_noconsole', type: 'boolean', value: true, description: 'Disable console on Windows (pass -mwindows flag)')
option('prebuilt_server', type: 'string', description: 'Path of the prebuilt server')
option('portable', type: 'boolean', value: false, description: 'Use scrcpy-server from the same directory as the scrcpy executable')
option('hidpi_support', type: 'boolean', value: true, description: 'Enable High DPI support')
option('server_debugger', type: 'boolean', value: false, description: 'Run a server debugger and wait for a client to be attached')
option('server_debugger_method', type: 'combo', choices: ['old', 'new'], value: 'new', description: 'Select the debugger method (Android < 9: "old", Android >= 9: "new")')

rm-rf X will still pop up the terminal interface after compilation
Is it configured incorrectly there

douxiaoping:
``belen@DESKTOP-S47BQP1 MINGW64 ~/scrcpy
$ meson configure
WARNING: The source directory instead of the build directory was specified.
WARNING: Only the default values for the project are printed, and all command line parameters are ignored.

Core properties:
  Source dir D:/msys64/home/belen/scrcpy

Main project options:

  Core options            Default Value                                                                                               Possible Values                                               Description
  ------------            -------------                                                                                               ---------------                                               -----------
  auto_features           auto                                                                                                        [enabled, disabled, auto]                                     Override value of all 'auto' features
[...]

PR #1988, opened 2020-12-22
Replace `scrcpy-noconsole.exe` by a simple wrapper script

rom1v:
In the Windows releases (until v1.16), there are two binaries:
 - `scrcpy.exe` (which pauses on error to let the user a chance to see any error message before the terminal closes)
 - `scrcpy-noconsole.exe` which does not open a terminal

They both contain the whole program, and complicate the build system.

For the next release, the idea is to have 3 files:
 - `scrcpy.exe` the main binary, with the same behavior as on other platforms (no pause on error)
 - `scrcpy-console.bat` a simple wrapper script to pause on error (already mreged 904d47057987fcd2bc2bb9494758a611fa6ab2bc)
 - `scrcpy-noconsole.vbs` a one-line VBS script to start `scrcpy.exe` without a console (this PR)

`scrcpy-noconsole.vbs` just contains:
```vbs
CreateObject("Wscript.Shell").Run "cmd /c scrcpy.exe", 0, false
```

Refs https://github.com/Genymobile/scrcpy/issues/1975#issuecomment-745314161

### DIFF
--- a/Makefile.CrossWindows
+++ b/Makefile.CrossWindows
@@ -11,8 +11,7 @@
 .PHONY: default clean \
 	build-server \
 	prepare-deps-win32 prepare-deps-win64 \
-	build-win32 build-win32-noconsole \
-	build-win64 build-win64-noconsole \
+	build-win32 build-win64 \
 	dist-win32 dist-win64 \
 	zip-win32 zip-win64 \
 	sums release
@@ -21,9 +20,7 @@ GRADLE ?= ./gradlew
 
 SERVER_BUILD_DIR := build-server
 WIN32_BUILD_DIR := build-win32
-WIN32_NOCONSOLE_BUILD_DIR := build-win32-noconsole
 WIN64_BUILD_DIR := build-win64
-WIN64_NOCONSOLE_BUILD_DIR := build-win64-noconsole
 
 DIST := dist
 WIN32_TARGET_DIR := scrcpy-win32
@@ -39,7 +36,7 @@ release: clean zip-win32 zip-win64 sums
 clean:
 	$(GRADLE) clean
 	rm -rf "$(SERVER_BUILD_DIR)" "$(WIN32_BUILD_DIR)" "$(WIN64_BUILD_DIR)" \
-		   "$(WIN32_NOCONSOLE_BUILD_DIR)" "$(WIN64_NOCONSOLE_BUILD_DIR)" "$(DIST)"
+		   "$(DIST)"
 
 build-server:
 	[ -d "$(SERVER_BUILD_DIR)" ] || ( mkdir "$(SERVER_BUILD_DIR)" && \
@@ -60,17 +57,6 @@ build-win32: prepare-deps-win32
 			-Dportable=true )
 	ninja -C "$(WIN32_BUILD_DIR)"
 
-build-win32-noconsole: prepare-deps-win32
-	[ -d "$(WIN32_NOCONSOLE_BUILD_DIR)" ] || ( mkdir "$(WIN32_NOCONSOLE_BUILD_DIR)" && \
-		meson "$(WIN32_NOCONSOLE_BUILD_DIR)" \
-			--cross-file cross_win32.txt \
-			--buildtype release --strip -Db_lto=true \
-			-Dcrossbuild_windows=true \
-			-Dcompile_server=false \
-			-Dwindows_noconsole=true \
-			-Dportable=true )
-	ninja -C "$(WIN32_NOCONSOLE_BUILD_DIR)"
-
 prepare-deps-win64:
 	-$(MAKE) -C prebuilt-deps prepare-win64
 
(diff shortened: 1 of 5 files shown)


Artur202030:
> They both contain the whole program, and complicate the build system.

For those who have no building problems, do not remove the option to create a scrcpy-noconsole.exe

rom1v:
I think there is no benefit to keep it. If it's not necessary, I prefer to remove it to simplify the Makefile and remove the ifdefs from the sources.

Alfakennyone:
> 
> 
> I think there is no benefit to keep it. If it's not necessary, I prefer to remove it to simplify the Makefile and remove the ifdefs from the sources.


There is, at least for me. 

I like to create a shortcut (to change the icon) and pin it to my taskbar.

jeansagi:
Sorry, but I think [#1988](https://github.com/Genymobile/scrcpy/pull/1988) is bad implemented.

I was happily, BTW, using scrcpy-noconsole.exe without any problem but, after vacation, I found it just simply does not work in 1.17.

I understand the desire to have just one executable and use wrappers if the same functionality is preserved but that is not the case.

When I used a console (cmd, bash) when I invoked scrcpy.exe the console gets locked until the scrcpy window closes or if you press Ctrl-C on the console window.
If you use scrcpy-noconsole.exe instead of scrcpy.exe the console doesn't get locked and you can keep using it and even close it and the scrcpy window remains visible.

Now If I use scrcpy-noconsole.exe it just does not work. Besides if I use scrcpy-noconsole.bat the console window still remains locked!! because it just calls scrcpy.exe which locks the console while the scrcpy window remains open. That is not the previous behavior.

Sure, I can use scrcpy-noconsole.vbs but I can use it only in cmd but not in bash by default and besides, you can't pass parameters to the vbs (for example --render-driver opengles2 or any other) which you can easily do with previous version of scrcpy-noconsole.exe.

So, 

The new implementation of scrcpy-noconsole, IMO, is simply wrong as it does not mimic the previous behavior.
And BTW I don't get how adding scrcpy-noconsole.exe how *much* complicates the build system but as long as the previous behavior isn't well implemented it is better to have the option.

Just my thoughts,


J.

rom1v:
> When I used a console (cmd, bash) when I invoked scrcpy.exe the console gets locked until the scrcpy window closes or if you press Ctrl-C on the console window.

To run in background (to get prompt immediately):
 - in cmd, execute `start /B scrcpy`
 - in bash, execute `scrcpy &`

You could also pass arguments.

> Besides if I use scrcpy-noconsole.bat the console window still remains locked!! because it just calls scrcpy.exe which locks the console while the scrcpy window remains open. That is not the previous behavior.

The goal was not to keep the previous behavior, the fact that it was executed in background was a side-effect.
Any command could be executed in background easily from a script.

> besides, you can't pass parameters to the vbs (for example --render-driver opengles2 or any other)

That's a real drawback, the vbs script should forward the arguments to scrcpy. PR welcome.

> I can use scrcpy-noconsole.vbs but I can use it only in cmd but not in bash by default

The only purpose of `scrcpy-noconsole` when it was introduced was to provide a way for windows users to double-click without opening a visible console. If it's called from a script, it's not necessary to use a "no console" version.

jeansagi:
> > When I used a console (cmd, bash) when I invoked scrcpy.exe the console gets locked until the scrcpy window closes or if you press Ctrl-C on the console window.
> 
> To run in background (to get prompt immediately):
> 
> * in cmd, execute `start /B scrcpy`
> * in bash, execute `scrcpy &`
> 
> You could also pass arguments.

Do you try this yourself? In my case I have the Windows 10 Pro 20H2, BTW, and that just don't work.
Surely after executing `start /B scrcpy` in cmd or `scrcpy &` in bash the console returns but get unusable. To effectively use the console, one must close scrcpy window.

So, both options aren't really a workaround. In fact, if I create a scrcpy.bat with `start /B scrcpy` to use the same command on cmd/bash (hate to use different commands for the same...) it does not work... the console gets stuck. I tried this on cmd and bash on their own programs and through windows terminal and always is the same: The console gets stuck.

> > Besides if I use scrcpy-noconsole.bat the console window still remains locked!! because it just calls scrcpy.exe which locks the console while the scrcpy window remains open. That is not the previous behavior.
> 
> The goal was not to keep the previous behavior, the fact that it was executed in background was a side-effect.
> Any command could be executed in background easily from a script.
>

Not so easy without getting the console stuck, as I said.
 
> > besides, you can't pass parameters to the vbs (for example --render-driver opengles2 or any other)
> 
> That's a real drawback, the vbs script should forward the arguments to scrcpy. PR welcome.
> 
> > I can use scrcpy-noconsole.vbs but I can use it only in cmd but not in bash by default
> 
> The only purpose of `scrcpy-noconsole` when it was introduced was to provide a way for windows users to double-click without opening a visible console. If it's called from a script, it's not necessary to use a "no console" version.

Maybe, that were their original purpose but the more than I think about it the more I believe that it should be a parameter to scrcpy, not a duplicate program compiled with a special flag.

That would be indeed a useful feature for me... and it seems to others as well.


J.

rom1v:
> Do you try this yourself? In my case I have the Windows 10 Pro 20H2, BTW, and that just don't work.

Yes, that works.

> Surely after executing start /B scrcpy in cmd or scrcpy & in bash the console returns but get unusable.

Because of the scrcpy output? Yes, if you don't want it, disable it (like any shell command):

```batch
start /B scrcpy >nul 2>&1
```
```bash
scrcpy >&/dev/null
```

> In fact, if I create a scrcpy.bat with start /B scrcpy to use the same command on cmd/bash (hate to use different commands for the same...) it does not work... the console gets stuck. I tried this on cmd and bash on their own programs and through windows terminal and always is the same: The console gets stuck.

Yes, you need the `-noconsole` version (from a batch script, call `scrcpy-noconsole`, you don't need `start /B`). As you noticed, there is a temporary issue that the current `scrcpy-noconsole` script does not forward its arguments, but this can be fixed. Then in practice there will be no difference between executing the old `scrcpy-noconsole.exe` and the new `scrcpy-noconsole.vbs`.

> the more than I think about it the more I believe that it should be a parameter to scrcpy, not a duplicate program compiled with a special flag.

Of course that would be better, but that's not possible (otherwise it would have been implemented that way): on Windows the fact that a console is opened or not is determined before executing the `main()` of the program, so the program itself could not decide to not start a new window. That's the reason why you can't execute a `.bat` file without displaying a console.

The wanted behavior is that of Linux:
 - if you run from a terminal, you get the output and you get the prompt back once scrcpy is terminated (unless you call it with `&`)
 - if you run it without a terminal, it still works without opening a new window

On Windows, the behavior of consoles is totally broken (IMO, I might be biased):
 - if you create a "Windows" application:
   - from a terminal, it behaves as expected
   - with a double-click from the file manager, it opens a terminal (could not be avoided programmatically)
 - if you create a "Console" application:
   - from a terminal, there is no output and it is started in background (wtf)
   - with a double-click from the file manager, it behaves as expected
 
(and that's not the only thing which is [broken](https://daviddeley.com/autohotkey/parameters/parameters.htm#WINPASS) with command-line execution)

PR #2040, opened 2021-01-16
Added Save Screenshot

Spudmn:
Pressing the MOD + Q key will save a screenshot the next time the screen is rendered.

It saves a PNG file with a date time stamp filename.

It uses lodepng to encode the file.  Thanks Lode Vandevenne (https://github.com/lvandeve/lodepng)

The save_screenshot function has been modified from user1902824 (https://stackoverflow.com/questions/34255820/save-sdl-texture-to-file)

Limitations.

If 2 screenshots are taken within the same second, the file name will be the same and the first file will be overwritten

Does not handle client side screen rotation.

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -20,6 +20,7 @@ src = [
     'src/stream.c',
     'src/tiny_xpm.c',
     'src/video_buffer.c',
+    'src/lodepng.c',
     'src/util/net.c',
     'src/util/process.c',
     'src/util/str_util.c'
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -456,6 +456,11 @@ input_manager_process_key(struct input_manager *im,
                     screen_switch_fullscreen(im->screen);
                 }
                 return;
+            case SDLK_q:
+                if (!shift && !repeat && down) {
+                   im->screen->save_screenshot = true;
+                }
+                return;
             case SDLK_w:
                 if (!shift && !repeat && down) {
                     screen_resize_to_fit(im->screen);
--- a/app/src/lodepng.c
+++ b/app/src/lodepng.c
@@ -1,0 +1,6464 @@
+/*
+LodePNG version 20201017
+
+Copyright (c) 2005-2020 Lode Vandevenne
+
+This software is provided 'as-is', without any express or implied
+warranty. In no event will the authors be held liable for any damages
+arising from the use of this software.
+
+Permission is granted to anyone to use this software for any purpose,
+including commercial applications, and to alter it and redistribute it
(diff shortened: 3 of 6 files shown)


rom1v:
Thank you for working on this. This is a feature often requested (#684), and it is not trivial.

> will save a screenshot the next time the screen is rendered.

Why not the current frame immediately? (if the screen content does not change, the next frame could be sent seconds or minutes later in theory)

I have several general remarks.

This PR adds a dependency (lodepng). Scrcpy already depends on FFmpeg, which can encode images, so I'd prefer to use FFmpeg. This is far less straightforward, but this avoids a separate image encoder (I have pending work for [decoding images](https://github.com/Genymobile/scrcpy/pull/1987) too). Btw, since the input picture is already compressed (it comes from a H.264 stream), it would be better to save it to JPG than PNG (lossless is unnecessary on a lossy input).

Also, encoding a picture takes time: it should not be done on the UI thread, but on a separate thread.

I think that it would be more straightforward to encode the content of the decoded frame (the rendering frame in `struct video_buffer`, which would need to be refcounted), rather than to retrieve the SDL texture content (moreover, it's in YUV, which is the native JPEG format). Even semantically, in theory we could want to capture a screenshot even if we don't display the frame on screen.

Regards

JASHIKO:
Also came here to add that instant screenshot would be amazing to add. It's probably the only feature missing from this amazing program at the moment.

vincespice:
If I press a key mod + Q, it doesn't take a screenshot. What could be the reason? My version: scrcpy v1.17 on Jan 2.

npes87184:
> If I press a key mod + Q, it doesn't take a screenshot. What could be the reason? My version: scrcpy v1.17 on Jan 2.

This PR has not been merged. You can build scrcpy based on this PR manually to include this feature.

vincespice:
Can you help me with how I can do it? I'd be grateful for.

Spudmn:
> Can you help me with how I can do it? I'd be grateful for.
You will need to build scrcpy from my PR.  
If you like you can build it from my fork here https://github.com/Spudmn/scrcpy/tree/Added_Save_Screenshot
Please note that I do not have time to maintain this fork.  It worked for what I needed, maybe it will work for you.

YerongLi:
> > If I press a key mod + Q, it doesn't take a screenshot. What could be the reason? My version: scrcpy v1.17 on Jan 2.
> 
> This PR has not been merged. You can build scrcpy based on this PR manually to include this feature.

How we can build with this PR?

hellishvictor:
Looking forward for this feature to be implemented.

PR #2052, opened 2021-01-20
Pass arguments through to scrcpy.exe

slingmint:


rom1v:
Thank you :+1:

However, it did not correctly handled arguments with spaces (like `--window-title "a b c"`), so I added quotes. And I simplified a bit.

Please review: https://github.com/Genymobile/scrcpy/commit/pr2052

slingmint (commented):
No worries. And, yes, I'm not sure how I forgot about adding quotes, but good catch. Looks good to me.

rom1v:
Oh but my version does not work in case the parsed arguments already contain `"`.

For example `--window-title="a b c"` does not work, while `--window-title "a b c"` works. (The argument parsing is very different from bash.)

rom1v:
I fail to pass a real double-quote (for example a window title `"abc"` with `"` included) as argument correctly:

```cmd
scrcpy           --window-title="\"a b c\""  REM works
scrcpy-noconsole --window-title="\"a b c\""  REM does not work

scrcpy           --window-title="""a b c"""  REM works
scrcpy-noconsole --window-title="""a b c"""  REM does not work
```

https://superuser.com/a/962805/310383

rom1v:
`--window-title "a b c"` works correctly, that's good enough :+1:

Merged into `dev`: dd453ad041b0dcf0486610a2215c2d86499d66a7

PR #2065, opened 2021-01-28
remain title bar when use --window-borderles so that we can move and close the window

liyinong6666:
Only remove left and right borders so that we can move and close the window

### DIFF
--- a/app/src/device.c
+++ b/app/src/device.c
@@ -11,7 +11,7 @@ device_read_info(socket_t device_socket, char *device_name, struct size *size) {
         LOGE("Could not retrieve device information");
         return false;
     }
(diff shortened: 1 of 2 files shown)


rom1v:
What's your OS/environment?

On Debian/XFCE, removing `SDL_WINDOW_RESIZABLE` just make the window non-resizable (so if you try to resize it moves the window), but has no effect on left and right borders display.

liyinong6666:
I'm on Windows 10.  
well, removing SDL_WINDOW_RESIZABLE is a easy way to solve my problem，but I didn't consider about cross-platform.I will try to think about a more general solution

PR #2185, opened 2021-03-09
meson: Do not use full path with mingw tools name.

Biswa96:
This helps to use mingw toolchains which are not in /usr/bin path.

### DIFF
--- a/cross_win32.txt
+++ b/cross_win32.txt
@@ -2,11 +2,11 @@
 
 [binaries]
 name = 'mingw'
-c = '/usr/bin/i686-w64-mingw32-gcc'
-cpp = '/usr/bin/i686-w64-mingw32-g++'
-ar = '/usr/bin/i686-w64-mingw32-ar'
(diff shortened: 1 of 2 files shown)


Biswa96:
I could also change the upstream branch. BTW, have you consider to use pkg-config or cmake to find cross-compiled sdl and ffmpeg dependencies instead of hardcoded path in app/meson.build?

rom1v:
> I could also change the upstream branch.

What do you mean?

> BTW, have you consider to use pkg-config […] to find cross-compiled sdl and ffmpeg dependencies instead of hardcoded path in app/meson.build

Oh yes, please. These hardcoded paths in meson are so hacky.

In the end, I would like something like [`contrib`](https://code.videolan.org/videolan/vlc/-/tree/master/contrib) in VLC, where each dependency is in its own folder, and we can `make install` them to somewhere (a first step could be to just use the prebuilts binary, but `make install` them to the "prefix" folder), and build scrcpy against them.

PR #2212, opened 2021-03-23
Disable scecure screen content on Android Q+

slavkoder:
Using `createDisplay(name, secure)` with `secure == true` goes against the security recommendations and posses risk for leaking sensitive user/app data. 

Starting on Android 12 this policy will be enforced and the call will crash for all non-secure displays with the following output:
```
[server] ERROR: Exception on thread Thread[main,5,main]
java.lang.AssertionError: java.lang.reflect.InvocationTargetException
	at com.genymobile.scrcpy.wrappers.SurfaceControl.setDisplaySurface(SurfaceControl.java:75)
	at com.genymobile.scrcpy.ScreenEncoder.setDisplaySurface(ScreenEncoder.java:243)
	at com.genymobile.scrcpy.ScreenEncoder.internalStreamScreen(ScreenEncoder.java:91)
	at com.genymobile.scrcpy.ScreenEncoder.streamScreen(ScreenEncoder.java:60)
	at com.genymobile.scrcpy.Server.scrcpy(Server.java:80)
	at com.genymobile.scrcpy.Server.main(Server.java:252)
	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:355)
Caused by: java.lang.reflect.InvocationTargetException
	at java.lang.reflect.Method.invoke(Native Method)
	at com.genymobile.scrcpy.wrappers.SurfaceControl.setDisplaySurface(SurfaceControl.java:73)
	... 7 more
Caused by: java.lang.IllegalArgumentException: displayToken must not be null
	at android.view.SurfaceControl$Transaction.setDisplaySurface(SurfaceControl.java:3111)
	at android.view.SurfaceControl.setDisplaySurface(SurfaceControl.java:2194)
	... 9 more
```

To prevent Scrcpy from crashing and to protect sensitive app data, this change is suggested for apps running Android Q and higher.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
+++ b/server/src/main/java/com/genymobile/scrcpy/ScreenEncoder.java
@@ -7,6 +7,7 @@ import android.media.MediaCodec;
 import android.media.MediaCodecInfo;
 import android.media.MediaCodecList;
 import android.media.MediaFormat;
+import android.os.Build;
 import android.os.IBinder;
 import android.view.Surface;
 
@@ -225,7 +226,7 @@ public class ScreenEncoder implements Device.RotationListener {
     }
 
     private static IBinder createDisplay() {
-        return SurfaceControl.createDisplay("scrcpy", true);
+        return SurfaceControl.createDisplay("scrcpy", Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q);
     }
 
     private static void configure(MediaCodec codec, MediaFormat format) {


slavkoder:
I just noticed that you already have this https://github.com/Genymobile/scrcpy/commit/ca3d807e7fe2cd795d0efe74034640d1594a1954

I would recommend that you extend this feature to pre-12 to help maintain security of the apps on the earlier versions.

rom1v:
Hi,

> Using `createDisplay(name, secure)` with `secure == true` goes against the security recommendations and posses risk for leaking sensitive user/app data.
> I would recommend that you extend this feature to pre-12 to help maintain security of the apps on the earlier versions.

IMO, the device belongs to the user. An app should not be able to prevent the user to take a screenshot by setting a flag.
Unless the user have control over the flag (like in Signal for example); that way, this provides security for the user (sensitive data is not leaked by keeping the flag enabled) without excessive restriction (users are still able to take a screenshot of their own device).

A scrcpy option `--secure-flag` (controlled by the user) would be totally ok.

Ref: 1fdde490fd2a0b89680a2b5da5e5274192398023
I also opened a bug report to restore the previous behavior (without much hope): https://issuetracker.google.com/issues/180777866

> I just noticed that you already have this ca3d807

Yes, I'll have no choice to do this if they keep that behavior in Android 12.

LBensman:
> I would recommend that you extend this feature to pre-12 to help maintain security of the apps on the earlier versions.

I'm not sure I'd agree with this.  The security from AOSP's perspective here is to prevent nefarious code from taking screenshots to steal sensitive info (like a banking app showing account number or something) - i.e. without user's awareness or consent.  And yes, in fact, I think AOSP allowing shell to have sufficient access always seemed to me like a bit of an oversight AOSP left until now.  But in case of scrcpy, it's showing the screen to that same user (that probably has the phone right next to them).

Having scrcpy receive secure screen here does not open (as far as I can tell) any new vectors not already available to the user using it.  So scrcpy enforcing AOSP's desire on lower releases achieves nothing - especially since anyone can take scrcpy code, flip `false` to `true` recompile, and voila!

> I also opened a bug report to restore the previous behavior (without much hope): https://issuetracker.google.com/issues/180777866

@rom1v, accessing that link gives me "Access denied".  I think only you can see it - I don't know if you can change visibility of the issue - it may be limited to author and AOSP internal users.

rom1v:
>  I think only you can see it

Yes, I think so.

Here is the bug report content:

> In Android 12 preview, shell permissions are not sufficient anymore to capture "protected" screen content. As a consequence, scrcpy (of which I'm the author) could not mirror some apps anymore like it did for Android <= 11. Note that it's just about the "secure" flag, the content with DRM was still not viewable.
>
> IMO, shell permissions are sufficiently privileged to allow capturing such content. Some apps like Silence or Signal allow to disable this "screen protection", but some apps don't provide such a settings, while it's still convenient to control them from the computer with screen mirroring.
>
> Could you bring back this ability with shell permission on Android 12?
>
> Ref: https://github.com/Genymobile/scrcpy/issues/2129
>
> Thank you,
> Regards
>
>     * Are you an Android developer?" Yes
>     * Is this a regression from Android 11 to 12? Yes
>     * What are the steps to reproduce the problem? Start scrcpy to mirror the device on the computer
>     * Issue Category: Framework (platform) / Runtime Permissions
>     * What was the expected result? For convenience, shell permissions should allow to mirror content having FLAG_SECURE (like it is the case since at least Android 5)
>     * What was the actual result? Could not create a secure display with shell permissions: E SurfaceFlinger: Only privileged processes can create a secure display
>     * Relevant logcat output: See https://github.com/Genymobile/scrcpy/issues/2129

The bug has been assigned, and its state has been set to:

> Blocked by -/1 180941209 [Details unavailable.]

I have no more details.

slavkoder:
Thanks for your comments and thoughts on the feature. Closing this PR in favor of existing fix https://github.com/Genymobile/scrcpy/commit/ca3d807e7fe2cd795d0efe74034640d1594a1954

PR #2223, opened 2021-03-29
Clarify the order of operations

rayfoss:
Mistakes I made, admittedly low on sleep:
- Used root dispite all warnings
- Ran non-prebuilt code... and couldn't undo.

Clarified by using headings and re-ordering to make prebuilt section stand-alone

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -6,7 +6,7 @@ You may want to build only the client: the server binary, which will be pushed
 to the Android device, does not depend on your system and architecture. In that
 case, use the [prebuilt server] (so you will not need Java or the Android SDK).
 
-[prebuilt server]: #prebuilt-server
+[prebuilt server]: #option-2-prebuilt-server
 
 ## Branches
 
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you.

I moved more prebuilt server-related doc and changed minor things, here is the result: fc5de88eaacfd3ef0af2939f19706d72f9bc427d

Is it ok for you?

Then I wrote a script to simplify installation (on `install_release` branch):
https://github.com/Genymobile/scrcpy/blob/install_release/BUILD.md#simple

I will merge both on `master` soon if everything is ok.

rayfoss:
Looks good! I was concerned about breaking the #Prebuilt-server link, but making people read from top to bottom is probably desireable.

rom1v:
> I was concerned about breaking the #Prebuilt-server link

Yes, me too, but Github-markdown does not allow to force an anchor name (without breaking other things) :confused: 

But anyway, with the "simple" method, few people would need to use the prebuilt server explicitly (they'll just use `install_release.sh` I guess).

PR #2258, opened 2021-04-16
Actions for 4th and 5th mouse buttons

brunoais:
For now, the 4th and 5th mouse buttons have erratic behavior from misinterpretation as being either 1st or 3rd buttons.

Nowadays, non-basic mice have, at least, 4 buttons, so, for me, makes most sense that those buttons are assigned. In this case, they are assigned based on my perceived difficulty in activating the feature using the mouse and how often used.

Given such, I decided that app switch to be the 4th button and notifications panel to be the 5th button.
These are what I believe to be the best use for these two. The 4th button being the app switch should be close to the no-brainer.
As for the notifications being the 5th, that one is more open for discussion. IMO, the major competitors being:

1. Notifications panel
2. Pinch-to-zoom
3. paste

No more buttons are assigned because SDL doesn't support them. 
For now, I would present this way, as a way to start a discussion

### DIFF
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -646,13 +646,17 @@ input_manager_process_mouse_button(struct input_manager *im,
     }
 
     bool down = event->type == SDL_MOUSEBUTTONDOWN;
-    if (!im->forward_all_clicks && down) {
+    if (!im->forward_all_clicks) {
         if (control && event->button == SDL_BUTTON_RIGHT) {
-            press_back_or_turn_screen_on(im->controller);
+            if (down) {
+                press_back_or_turn_screen_on(im->controller);
+            }
             return;
         }
         if (control && event->button == SDL_BUTTON_MIDDLE) {
-            action_home(im->controller, ACTION_DOWN | ACTION_UP);
+            if (down) {
+                action_home(im->controller, ACTION_DOWN | ACTION_UP);
+            }
             return;
         }
 
@@ -665,7 +669,9 @@ input_manager_process_mouse_button(struct input_manager *im,
             bool outside = x < r->x || x >= r->x + r->w
                         || y < r->y || y >= r->y + r->h;
             if (outside) {
-                screen_resize_to_fit(im->screen);
+                if (down) {
+                    screen_resize_to_fit(im->screen);
+                }
                 return;
             }
         }
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -647,6 +647,8 @@ input_manager_process_mouse_button(struct input_manager *im,
 
     bool down = event->type == SDL_MOUSEBUTTONDOWN;
     if (!im->forward_all_clicks) {
+        int action = down ? ACTION_DOWN : ACTION_UP;
+
         if (control && event->button == SDL_BUTTON_RIGHT) {
             if (down) {
                 press_back_or_turn_screen_on(im->controller);
@@ -654,9 +656,7 @@ input_manager_process_mouse_button(struct input_manager *im,
             return;
         }
         if (control && event->button == SDL_BUTTON_MIDDLE) {
-            if (down) {
-                action_home(im->controller, ACTION_DOWN | ACTION_UP);
-            }
+            action_home(im->controller, action);
             return;
         }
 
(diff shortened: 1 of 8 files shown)


rom1v:
:+1: It's better to bind them to something.

I made some changes on `dev` to properly handle DOWN and UP events for shortcuts: #2259
Could you review/test, please?

That way, you could properly propagate them in your `action_app_switch()`.

> The 4th button being the app switch should be close to the no-brainer.

Yes, I guess.

> As for the notifications being the 5th, that one is more open for discussion.

Why not. But it's maybe odd to bind "open navigation panel" but not "close navigation panel".
Maybe `MENU` (<kbd>MOD</kbd>+<kbd>m</kbd>), especially because it's the "unlock screen" shortcut?

I'd like other opinions on that (and my mouse has 3 buttons).

(In the future, I would like to implement some kind of shortcut customization. But it's not a priority.)

brunoais:
> +1 It's better to bind them to something.
> 
> I made some changes on `dev` to properly handle DOWN and UP events for shortcuts: #2259
> Could you review/test, please?

Sure. I will be able to in about 1h.
> 
> > As for the notifications being the 5th, that one is more open for discussion.
> 
> Why not. But it's maybe odd to bind "open navigation panel" but not "close navigation panel".
> Maybe `MENU` (MOD+m), especially because it's the "unlock screen" shortcut?

No need. Just right click (back button) and it closes. No need for a button to close when the back button does that already. I wanted to have a shortcut key to open towards the 2nd menu, when sliding the finger the 2nd time. Maybe I will end up implementing it as a finger slide... Unsure...
 
> I'd like other opinions on that (and my mouse has 3 buttons).

Mine has 7. Unexpected behaviors happen when I try them all.

> 
> (In the future, I would like to implement some kind of shortcut customization. But it's not a priority.)

Great. Want me to try that?

brunoais:
Your branch looks good. I rebased this one on the one you did.

app/src/input_manager.c
```diff
@@ -646,13 +658,24 @@ input_manager_process_mouse_button(struct input_manager *im,
...
-    if (!im->forward_all_clicks && down) {
+    if (!im->forward_all_clicks) {
+        int action = down ? ACTION_DOWN : ACTION_UP;
+
+        if (control && event->button == SDL_BUTTON_X1) {
+            action_app_switch(im->controller, action);
+            return;
+        }
+        if (control && event->button == SDL_BUTTON_X2) {
+            expand_notification_panel(im->controller);
```
rom1v, line 669: only when `action == ACTION_DOWN`

brunoais: Like so?

rom1v:
Merged into `dev`: b28adab86fecb4c4a617d790230100930a03f2b1

Thank you.

rom1v:
Oops, sorry, I changed X2 to RIGHT_CLICK to test locally, and I kept that in the commit. :facepalm: 

I just fixed and force-pushed `dev`: c4f3a768d5478cf27c8dd16aca409a45a670bc16

luke-jr:
The 4th button is typically interpreted as "Back" by regular PC applications...

Can we at least have a setting for that normalish behaviour?

brunoais:
> Can we at least have a setting for that normalish behaviour?

Because we already have that function assigned to the right click mouse button.  
If you prefer the 4th button to be a back button, feel free to make your own build of scrcpy that does that or use some other program that temporarily reassigns buttons.

luke-jr:
Wouldn't it make more sense to support standard behaviours, and have people who want non-standard behaviours reassign things? :/

Switching between right-click and back-click between applications has been a bit confusing since I started using scrcpy.

rom1v:
> Switching between right-click and back-click between applications has been a bit confusing since I started using scrcpy.

The reason is that BACK is very useful even if you have a 2-button mouse.

Yes, designing a "reassing buttons" feature could be useful, but it's not done.

brunoais:
> How can I customize the 4th and 5th mouse button clicks?

What do you want them to do? The process is different depending on what you want them to mean.

brunoais:
> I would like to set button 4 (the one that is used to go back) as the back button

In that case, you edit here:

https://github.com/Genymobile/scrcpy/pull/2258/files#diff-fd408d3adc5df68bc93e54b8de3830f0e2814ff8e07c581f41f60008f1e4775eR665

Replace that line's code with:
```
press_back_or_turn_screen_on(im->controller, ACTION_DOWN)
```

rom1v:
@123SONIC321 Please read https://github.com/Genymobile/scrcpy/pull/4877#issuecomment-2185307250

PR #2259, opened 2021-04-16
Improve mouse buttons events

rom1v:
Some mouse clicks DOWN are captured for shortcuts, but the matching UP
event was still forwarded to the device. Instead, capture both DOWN and UP for shortcuts, and do nothing on UP.

After than, forward middle-click and right-click shortcuts (respectively HOME and BACK-if-screen-is-off) respecting the DOWN and UP events from the mouse. This is consistent with the keyboard shortcuts.

Read individual commits for more details.


rom1v:
Merged.

PR #2260, opened 2021-04-17
Ctrl+n+n for settings panel

brunoais:
Show the settings panel if pressing `CTRL`+`n` and then, without releasing CTRL, press `n` again

Still not done/updated:
README: 

- [x] New shortcut
- [x] Rename what `CTRL`+`SHIFT`+`n` does

Closes: #2264

PR #2264, opened 2021-04-18
Press 2x extra mouse button to open settings

brunoais:
Requires #2260


rom1v:
Merged into `dev`.

rom1v:
A minor inconsistency: with the keyboard shortcut, the settings panel is expanded regardless of the delay between the first <kbd>MOD</kbd>+<kbd>n</kbd> and the second. With the mouse, it only works with a double-click. Maybe it should work with a longer delay too? What do you think?

brunoais:
Sounds good to me. That feature acting consistent between both ways is better.

rom1v:
OK, I noticed this when testing before releasing v1.18. But it's minor, so I still released. It can be fixed later :wink:

rom1v:
> Maybe it should work with a longer delay too?

(Or maybe the keyboard shortcut should only work for a short delay (like a double click), I don't know which is better.)

brunoais:
After thinking for a while. I think both should have the reset after a short delay.
Any ideas on how to achieve that? I don't know SDL API well enough for that.

PR #2265, opened 2021-04-18
Intents on start & exit

brunoais:
Send intents when scrcpy starts and when scrcpy exits

Still missing (will be added if feature accepted)

- [x] Mention feature in README
- [x] Create command line option to choose whether to send the intents (defaulting to off)

Extra: video showing a visual use:

https://www.mediafire.com/file/ktcxs0exclvoqha/intent_demo.mp4/file

(I was unable to attach the file in this post)


brunoais:
@rom1v Anything that can be discussed about this already?

server/src/main/java/com/genymobile/scrcpy/Intents.java
```diff
@@ -0,0 +1,42 @@
...
+        for (String e = iter.previous(); iter.hasPrevious(); e = iter.previous()) {
+            if(e == null){
+                // If there's no value, remove the entry
+                iter.remove();
+                iter.previous();
+                iter.remove();
+                iter.previous();
+                iter.remove();
+            }
+        }
```
brunoais, line 33: I thought it's easier to read if just adding all and then remove the unused ones than to just add the ones provided.
Doing the other way is fine too

rom1v:
Sorry, I have no much time to work on scrcpy presently.

I'm ok with sending a new intent on start and close, but IMO:
 - it should be enabled only with an explicit parameter
 - the implementation should send the intent directly (without executing a new program `am broadcast …`, but calling the functions used by `am` directly); see this old WIP commit: 160ae02b93ed028e8d881aed5893c49d8cf0a068

Regards

brunoais:
> Sorry, I have no much time to work on scrcpy presently.

Sad but understood. Must be really bad that can't even comment. Thanks for taking the time for that comment.

> it should be enabled only with an explicit parameter

Updated the top comment with that in mind.

>     the implementation should send the intent directly see this old WIP commit: [160ae02](https://github.com/Genymobile/scrcpy/commit/160ae02b93ed028e8d881aed5893c49d8cf0a068)

Got it! I'll keep that in mind. Thank you for that commit. It should be enough for what I'm aiming at. I didn't know what could be done.

rom1v:
> Must be really bad that can't even comment.

I still can (and often) answer to comments/issues.

About this one, I didn't answer immediately because in my TODO list I wanted to improve the drag&drop behavior by sending an intent after "adb push" a file (the equivalent of `adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Download` for example), so I initially wanted to implement it quickly, then answer with `sendBroadcast()` already available in the codebase. But I didn't have time to implement it, and I forgot to comment.

brunoais:
Ty for clarification. Those things happen.

brunoais:
> > Must be really bad that can't even comment.
> 
> I still can (and often) answer to comments/issues.
> 
> About this one, I didn't answer immediately because in my TODO list I wanted to improve the drag&drop behavior by sending an intent after "adb push" a file (the equivalent of `adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Download` for example), so I initially wanted to implement it quickly, then answer with `sendBroadcast()` already available in the codebase. But I didn't have time to implement it, and I forgot to comment.

Want me to finish that implementation?

rom1v:
> Want me to finish that implementation?

Thank you for the proposal, but I started implementing it few days ago (i just pushed the WIP: [`broadcast`](https://github.com/Genymobile/scrcpy/commits/broadcast)), but I discovered that [`ACTION_MEDIA_SCANNER_SCAN_FILE`](https://developer.android.com/reference/android/content/Intent?hl=en#ACTION_MEDIA_SCANNER_SCAN_FILE) is now deprecated, so I finally abandoned the idea. :disappointed:

brunoais:
It's deprecated but lots of people use older phones. It can still make sense to go forward with it.

brunoais:
@rom1v 
I think I finished this feature.
There was 1 detail I had problems deciding how to implement. Which is the "scrcpy" prefix. I decided to end up with a field and a method in the `Intents` enum but I'm not fully sure of it.

server/src/main/java/com/genymobile/scrcpy/Server.java
```diff
@@ -188,6 +219,9 @@ private static Options createOptions(String... args) {
         boolean powerOffScreenOnClose = Boolean.parseBoolean(args[15]);
         options.setPowerOffScreenOnClose(powerOffScreenOnClose);
 
+        EnumSet<Intents> broadcastIntents = Intents.fromBitSet(BitSet.valueOf(new long[]{Long.parseLong(args[16])}));
```
brunoais, line 222: This is a roundabout way of translating but it is the way that seemed to be most direct without implementing manually

rom1v:
Hi,

Thank you for your work.

IMO, this feature should be kept simple:
 - two intents only: STARTED and STOPPED (no CLEANED, which is an implementation detail)
 - STOPPED would be sent by the CleanUp process only (so that it is sent even on adb disconnection)
 - a simple option `--send-broadcasts` without additional parameters: either the user wants the broadcasts, either they don't, IMO there is no need for enabling individual broadcasts
 - no intent extras: currently, AFAIK, they provide no more information than the broadcast action itself (but they might be added later to provide scrcpy info, like capture resolution?)

Regards

brunoais:
Here's my counter-opinions:

1. Two intents only where STOPPED is sent when CLEANED is currently sent. <- Nothing against this one
2. When I did it like that, I had into account needs such as not announcing when the connection happens (possible app that learned about it and is now listening) but still wanting to be announced when the disconnection happens. One kind of app I was thinking is a game that checks for scrcpy and blocks running if scrcpy is detected as connected but the user still wants to be able to pause the game automatically in case a disconnection happens. Do you think it would be bad to maintain such code?
3. The intent extras are mostly useful when using automation tools, instead of custom code. For custom code, they don't add anything because you can do pretty much anything. With automation tools such as Tasker, having such variables helps having less programming steps and profiles+tasks easier to maintain, specially as scrcpy evolves.

rom1v:
> When I did it like that, I had into account needs such as not announcing when the connection happens (possible app that learned about it and is now listening) but still wanting to be announced when the disconnection happens.

The STARTED and STOPPED broadcasts could be sent, but the app would only react to STOPPED.

> The intent extras are mostly useful when using automation tools, instead of custom code.

I don't know Tasker, but I guess that the automation tools could probably react to broadcasts without duplicate information in the extras. I don't want custom hacks specific to Tasker (which is not even open source).

brunoais:
> The STARTED and STOPPED broadcasts could be sent, but the app would only react to STOPPED.

In this case, it would be to hide scrcpy. It's more about apps you don't control than apps you do control.

> I don't know Tasker, but I guess that the automation tools could probably react to broadcasts without duplicate information in the extras. I don't want custom hacks specific to Tasker (which is not even open source).

It's not a feature specific for Tasker, I only mentioned it as an example, but fair enough.

rom1v:
> In this case, it would be to hide scrcpy. It's more about apps you don't control than apps you do control.

I'm not sure to understand the problem.

IIRC, in your initial version, broadcasts were sent unconditionally. The purpose of the cli option was to disable broadcasts by default because they are not used in 99.9% of users. But if a user needs them, why configuring which broadcasts are sent by scrcpy would help?

Scrcpy could just send both STARTED and STOPPED, and apps (or Tasker or whatever) could just do actions on reception of one of them.

brunoais:
> IIRC, in your initial version, broadcasts were sent unconditionally.

That's because I wanted to confirm implementation direction. Such as:

1. Whether OK to do such kind of feature
2. If doing on the server is sth you accept
3. If done on the server, if it should work by calling the shell or if I'm missing sth.

As I mention in the title, it was code as a base for discussion ;).
I already had the two checkmarks you see on the top post since the very beginning. I guess I should have mentioned that fact more explicitly.

If I were to do all the cli API by then and you didn't want to accept intent sending, I'd wasted my time on the C code when it would be just thrown away.

> Scrcpy could just send both STARTED and STOPPED, and apps (or Tasker or whatever) could just do actions on reception of one of them.

I'll try to explain with different words. About selecting which intents to send, my idea is more directed towards apps (most commonly, games but also some bank apps) that try to detect everything and then refuse to work when some parameter is different than expected. If you still insist on a "no" then I can remove the selection part with no hard feelings...

brunoais:
Given what you mentioned and also not continuing the discussion, I decided my reasons are weak and I'll remove the ability to choose which intents to use completely

rom1v:
Given the specific needs related to the intent format and their intent extra, maybe a more generic approach would be preferable:

```
scrcpy --exec-on-start='am broadcast -a ...;settings ...' --exec-on-exit='am broadcast -a ...;settings ...'
```

What do you think?

brunoais:
Even though that approach is flexible, it can lead to any command being run and then will come the need to accommodate for it by what different people want, etc, etc...
It's flexible, good willed but I think, in the long run, it's worse than just removing the ability to choose which intents.
In other words: IMO, it makes sense for scrcpy to announce what it's doing but it doesn't make sense for scrcpy to execute arbitrary code. Even if, technically, it's not creating vulnerabilities.

brunoais:
Now I just need to test if all is good

DanEdens:
This is a really good thread. Cheers all.

brunoais:
Oh... More conflicts to fix... Maybe next weekend... I'll put in my calendar.

brunoais:
A hook script is more acceptable...
However, how can we properly tell the user that the stop is not guaranteed to run?
The only real way to make sure a "stop" script is run is by having something running by an app installed on the phone which checks for when scrcpy closes and then ensures the cleanup is actually done.
I'd like to provide that but I understand you don't want it.

rom1v:
> The only real way to make sure a "stop" script is run is by having something running by an app installed on the phone which checks for when scrcpy closes and then ensures the cleanup is actually done.

There is. See `CleanUp.java`

brunoais:
> There is. See `CleanUp.java`

That one works most of the time... Not even nearly every time.
For me, it works about 90% of the time (Do note: way better than the expected ~40% without it, though!).
However, it's still flawed. Sometimes I have to reset the android system settings myself because `CleanUp.java` didn't do it.
There's also a very easy way to force the cleanup code not to run.

rom1v:
> For me, it works about 90% of the time

If this doesn't work, my understanding is that it's a bug on the device which kills the child process after few hundreds of milliseconds. On a Nexus 5, i have this problem. On a OnePlus 7 Pro, it works 100℅ reliably (even if I had `sleep(5 seconds)`), for example when I unplug the device.

brunoais:
@rom1v The easiest way for me to reproduce is to just turn off debug mode automatically.
In my case, for some reason, just leaving it there and wait, my phone (for some reason) decides that debug mode has been on for too long and turns it off. I can't control that. I went to the battery saver app (the one from the manufacturer) and I can't find any way to control it.

rom1v:
> The easiest way for me to reproduce is to just turn off debug mode automatically.

It works as expected for me.

For example:

```
# enable show touches
scrcpy -t
```

When I turn USB debugging off, show touches is correctly disabled.

brunoais:
In my phone, if I turn debugging off, everything running as adb is immediately killed (as in `kill -9`). I don't know if it's manufacturer, android, both... neither.... I don't know

brunoais:
Replaced by #3191

Issue #2252, opened 2021-04-13
--window-borderless not working on macOS, scrcpy version 1.17


frankleonrose:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: macOS 10.15.7 (Catalina) and 10.14.6 (Mojave)
 - scrcpy version: 1.17_1
 - installation method: brew
 - device model: Galaxy Note 9, others
 - Android version: 10

**Describe the bug**
Passing `--window-borderless` to version 1.16 correctly showed a window with no title bar. 
Since update to 1.17 the title bar appears despite `--window-borderless`.

As an experiment, I tried building with an additional line in `screen_init_rendering`:
```
    SDL_SetWindowBordered(screen->window, !window_borderless);
```
This change resulted in a restored control over the window border.

rom1v:
Please post screenshots of both v1.16 and v1.17. I guess it's due to some SDL change.

frankleonrose:
This is v1.17_1 showing the command to launch adjacent to the problematic window on macOS Catalina.

![image](https://user-images.githubusercontent.com/1261725/115308542-e0ff4b80-a138-11eb-89ab-c04b38c9ea7e.png)

I don't have 1.17 or 1.16_4 installed anymore and cannot easily roll back the version (looks like I'd have to create a tap to extract into...dunno). Basically, the screenshot would have *no* title bar showing.

lilithkrakatoa:
Is reproducible

OS: macOS Big Sur 11.2.3
scrcpy version: 1.17
Installation method: brew
Device: Essential Products PH-1 (Android 10)

![Xnip2021-04-20_03-34-10](https://user-images.githubusercontent.com/82754016/115356117-5266ea80-a189-11eb-8baa-034ba2be6c40.jpg)

rom1v:
Thank you for your screenshots.

> As an experiment, I tried building with an additional line in […]

Could you please try building v1.16 or even earlier versions and test?

You mentioned that it worked in v1.16 but not in v1.17, but I suspect that it may be related to some independent SDL upgrade. This would help to determine the cause.

Thank you.

frankleonrose:
Ok, got the screenshot. And the error remains in 1.16.
[...]

PR #2276, opened 2021-04-22
Ensure the return value from get_server_path is freeable

frankleonrose:
While testing in support of https://github.com/Genymobile/scrcpy/issues/2252 I ran into this crash bug which looks like this:

```
% x/app/scrcpy
2021-04-22 12:50:31.596 scrcpy[82192:39073792] INFO: scrcpy 1.16 <https://github.com/Genymobile/scrcpy>
2021-04-22 12:50:31.596 scrcpy[82192:39073792] ERROR: Could not get executable path, using scrcpy-server from current directory
scrcpy(82192,0x11122fdc0) malloc: *** error for object 0x10a439c2c: pointer being freed was not allocated
scrcpy(82192,0x11122fdc0) malloc: *** set a breakpoint in malloc_error_break to debug
scrcpy-server: 1 file pushed. 4.5 MB/s (33622 bytes in 0.007s)
zsh: abort      x/app/scrcpy --window-borderless
```

The fix is simply that in the case where `get_server_path` returns just `SERVER_FILENAME`, it must be duplicated, because the caller expects to be able to free the returned value. If the duplication fails, we return NULL which indicates failure, so no further error handling is necessary, unless we wanted to log something about that extremely rare situation.

### DIFF
--- a/app/src/server.c
+++ b/app/src/server.c
@@ -61,7 +61,7 @@ get_server_path(void) {
         LOGE("Could not get executable path, "
              "using " SERVER_FILENAME " from current directory");
         // not found, use current directory
-        return SERVER_FILENAME;
+        return SDL_strdup(SERVER_FILENAME);
     }
     char *dir = dirname(executable_path);
     size_t dirlen = strlen(dir);
@@ -73,7 +73,7 @@ get_server_path(void) {
         LOGE("Could not alloc server path string, "
              "using " SERVER_FILENAME " from current directory");
         SDL_free(executable_path);
-        return SERVER_FILENAME;
+        return SDL_strdup(SERVER_FILENAME);
     }
 
     memcpy(server_path, dir, dirlen);


rom1v:
Yes, thank you :+1: LGTM.

However, I'm surprised you execute this code. Did you explicitly enabled `-Dportable=true` in your `meson` command?

rom1v:
I rebased on `dev` (so I also replaced `SDL_strdup()` by `strdup()`, because of changes on that branch, and merged: aaf7875d923e492047e66ca88585079704abf5ad

frankleonrose:
Yes, I did specify portable because I was testing the 1.16 version and had the compatible server local.

Thanks for taking the fix. Sorry I hadn't targeted dev.

rom1v:
> I did specify portable because I was testing the 1.16 version and had the compatible server local.

OK. Just for info, instead, you could just use `-Dprebuilt_server=...` in your meson command, or set `SCRCPY_SERVER_PATH` at runtime.

Issue #2256, opened 2021-04-15
OK, I gave up on scrcpy. Not computer illiterate but was expected easier install on Mac. 


itoldusoandso:
MacOS 10.13
Have done as instructed.. CLT, Homebrew and scrcpy.  The install process was sluggish, took 16 hours and decided to post on XDA to check sanity of what I am doing. 

Sorry guys, this just didn't work out for me. Total fail ;(

Now two more hours later, the install process stayed hanging and still showing "make"... So now after 18 hours I am pulling the plug want to quit. I want to delete scrcpy and put my machine back where it was before start. However, it gives me error when I type in the command:

bash-3.2$ brew uninstall scrcpy
Error: No available formula or cask with the name "scrcpy".
bash-3.2$ 

So what do I do now. 

See link to XDA post with picture.   I don't see how to attach picture here.

https://forum.xda-developers.com/t/installing-scrcpy-now-already-more-than-16-hours-comman-line-tools-homebrew-is-this-normal.4262295/

PembaTamang:
were you able to fix it ?

itoldusoandso:
Unfortunately not and I am still waiting on instructions on how to get rid of the software. I don't know what to do at this point.

rom1v:
> MacOS 10.13

Probably related: #2059

In particular: https://github.com/Genymobile/scrcpy/issues/2059#issuecomment-767032047

> macs that can't update their OS beyond 10.13 need to be warned that homebrew doesn't know or care if they try to update scrcpy it will install incompatible ffmpeg dependencies.

rom1v:
As an alternative, you could try [MacPorts](https://ports.macports.org/) on macOS (I don't know, I have no macOS).

itoldusoandso:
Well, I looked at MacPorts but it seems to me it's the page doesn't seem to be safe for toddlers. I need the instructions to be for a 5-years old. It's Babylon language for me as far as this is concerned.

People mention the issue starting after 1.6_2?  What are your thoughts if I try same steps but I try install 1.6_2 version. Can I do that somehow?

But first I guess I need to clean the mess on my machine. I have CLT, Homebrew and scrcpy installed. So I guess I need to start removing scrcpy but I don't know how, what and where. I don't want to start deleting the stuff in home directory as I said I it's not like regular apps so stuff seems to be spread around like birds feed.

harens:
[...]

Issue #2183, opened 2021-03-09
macos not install


zhou-zc:
log：

brew install scrcpy
Updating Homebrew...
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
Error: No available formula or cask with the name "scrcpy".
==> Searching for a previously deleted formula (in the last month)...
Error: No previously deleted formula found.
==> Searching taps on GitHub...
Error: No formulae found in taps.

macos version
 mac catalina 10.15.7

rom1v:
This is a homebrew issue.

Issue #2059, opened 2021-01-24
Error: scrcpy: no bottle available!


BardsDirge:
i was trying to update from 1.16 with homebrew and it seems i got hung up at 
`==> Installing scrcpy dependency: rust`
i waited over an hour. my macbook was pushing 100ªc the whole time.
i tried
`brew install scrcpy`
it told me i had 1.16 and, as usual, it told me to change the command to upgrade. i did and it stoped at rust again.
now for the first time ever, homebrew is telling me 
`Error: scrcpy: no bottle available! `

i'm disabled and use scrcpy to access my phone. i'm really screwed right now without it. please advise.

**Environment**
 - OS:  macOS 10.13.6
 - scrcpy version: 1.16_2
 - installation method: brew
 - device model: LG Tribute Dynasty SP200
 - Android version: 7.1.2

BardsDirge:
`==> Installing ffmpeg dependency: rust
==> ./configure --prefix=/usr/local/Cellar/rust/1.49.0_1 --release-channel=stable
==> make`

rom1v:
Hi.

You should report to homebrew, this is totally independant of scrcpy.

BardsDirge:
i'm not the only user that needs to know 1.16_2 is the end of the line.

would it really disrupt the ecosystem if you allowed discussing homebrew _**as it pertains to scrcpy?**_ the entirety of your mac users instructions are "use homebrew", even for compiling it yourself.

macs that can't update their OS beyond 10.13 need to be warned that homebrew doesn't know or care if they try to update scrcpy it will install incompatible ffmpeg dependencies. _**people need to know this**_.

i spent all day sorting out what went wrong and then manually replacing all my old scrcpy components in /usr/local/ from a backup.

rom1v:
> would it really disrupt the ecosystem if you allowed discussing homebrew

You could discuss it, but this is not the better place to get help, since this is a homebrew problem, you have more chances to get answers there I guess. I have no idea what is the cause of your issue.

Btw, I'm surprised that ffmpeg depends on Rust by default (maybe for rav1e).

BardsDirge:
i've spent a long time herding ducklings on message boards. i know you  need to be firm where you draw lines. this really isn't a thing homebrew cares about though. you're either running "big sur" or you're off the radar.

that's why i hope somehow people will be made aware that scrcpy's instructions only apply to OS 10.14+. 

[...]

Issue #1414, opened 2020-05-21
MacOS Segmentation fault: 11


wudelu:
/usr/local/Cellar/scrcpy/1.8/share/scr...shed. 0.6 MB/s (19850 bytes in 0.030s)
2020-05-21 18:31:56.356 scrcpy[30780:1970626] INFO: Initial texture: 1080x2160
Segmentation fault: 11
JokerWudeMacBook-Pro:~ jokerwu$ scrcpy -v
scrcpy 1.8

dependencies:
 - SDL 2.0.9
 - libavcodec 58.35.100
 - libavformat 58.20.100
 - libavutil 56.22.100

rom1v:
Which device? Which Android version?

Try with the latest scrcpy version (1.13).

Climax777:
I get the same. Using prebuild server 1.13 and scrcpy built in MacPorts:
scrcpy 1.13

dependencies:
 - SDL 2.0.12
 - libavcodec 58.65.102
 - libavformat 58.35.101
 - libavutil 56.38.100

Running lldb on scrcpy, it seems that adb push fails for some reason

```
(lldb) target create "scrcpy"
Current executable set to 'scrcpy' (x86_64).
(lldb) run
Process 25502 launched: '/usr/local/bin/scrcpy' (x86_64)
2020-05-21 22:45:39.568061+0200 scrcpy[25502:253291] INFO: scrcpy 1.13 <https://github.com/Genymobile/scrcpy>
/usr/local/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 11.1 MB/s (27694 bytes in 0.002s)
2020-05-21 22:45:39.645469+0200 scrcpy[25502:253291] ERROR: "adb push" exited unexpectedly
Process 25502 exited with status = 1 (0x00000001)
(lldb)
```

Android Debug Bridge version 1.0.41
Version 30.0.1-6435776

rom1v:
```
adb shell ls -l /data/local/tmp
```

Climax777:
```
-rw-rw-rw- root     root      4322329 1970-01-01 02:00 apk1590094170308.apk
-rwxr-xr-x root     root        17880 2018-09-11 03:44 getport
-rw-rw-rw- root     root           55 2020-05-18 17:00 mobile_info.properties
drwxrwxrwx shell    shell             2020-05-18 16:49 tc
-rw-rw-rw- root     root            7 1970-01-01 02:00 vysor.pwd
```

[...]

PR #2299, opened 2021-05-06
Add MacPorts Documentation

harens:
It goes without saying that I'm very thankful to the maintainers that put their time into keeping this awesome project running. Thank you for all your work :)

For all those that have various issues with Homebrew: https://github.com/Genymobile/scrcpy/issues/2256 https://github.com/Genymobile/scrcpy/issues/1414 https://github.com/Genymobile/scrcpy/issues/2059 https://github.com/Genymobile/scrcpy/issues/2183 (the list goes on...)

Document the installation instructions for MacPorts, and note that setting up adb is not required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,15 @@ You can also [build the app manually][BUILD].
 
 ### macOS
 
-The application is available in [Homebrew]. Just install it:
+The application is available in [MacPorts], which sets up adb for you:
+
+```bash
+sudo port install scrcpy
+```
+
+[MacPorts]: https://www.macports.org/
+
+It's also available in [Homebrew]. Just install it:
 
 [Homebrew]: https://brew.sh/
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you :+1:

I edited the commit to just add MacPorts after Homebrew: https://github.com/Genymobile/scrcpy/commits/pr2299

I personally don't have preference between the two (I don't use macOS), but for now it seems Homebrew is better known, and it was documented first, so I prefer to keep it that way :wink:

harens:
That's more than understandable. Thanks for editing it @rom1v.

rom1v:
Merged into master.

Coool:
[scrcpy](https://github.com/Coool/scrcpy-macOS-Binaries) for macOS

PR #2332, opened 2021-05-21
Holding Mod+p gives Power Options. Updated in Readme

sidhant-khamankar:


brunoais:
This is more of a consequence of how Android works and not something that Scrcpy does.
If you hold down the power button on an android phone, that is what will happen.

PR #2371, opened 2021-06-08
Log input event details to verbose log

intgr:
I wanted to automate a few actions using 'adb shell input' commands, but
it requires pixel coordinates. Logging input event details from scrcpy is
helpful for achieving that.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -184,7 +184,7 @@ scrcpy_print_usage(const char *arg0) {
         "\n"
 #endif
         "    -V, --verbosity value\n"
-        "        Set the log level (debug, info, warn or error).\n"
+        "        Set the log level (verbose, debug, info, warn or error).\n"
 #ifndef NDEBUG
         "        Default is debug.\n"
 #else
@@ -505,6 +505,11 @@ parse_display_id(const char *s, uint32_t *display_id) {
 
 static bool
 parse_log_level(const char *s, enum sc_log_level *log_level) {
+    if (!strcmp(s, "verbose")) {
+        *log_level = SC_LOG_LEVEL_VERBOSE;
+        return true;
+    }
+
     if (!strcmp(s, "debug")) {
         *log_level = SC_LOG_LEVEL_DEBUG;
         return true;
--- a/app/src/control_msg.c
+++ b/app/src/control_msg.c
@@ -170,7 +170,12 @@ control_msg_log(const struct control_msg *msg) {
                          (long) msg->inject_touch_event.buttons);
             } else {
                 // numeric pointer id
-                LOG_CMSG("touch [id=%" PRIu64 "] %-4s position=%" PRIi32 ",%"
+#ifndef __WIN32
+# define PRIu64_ PRIu64
+#else
+# define PRIu64_ "I64u"  // Windows...
+#endif
+                LOG_CMSG("touch [id=%" PRIu64_ "] %-4s position=%" PRIi32 ",%"
                              PRIi32 " pressure=%g buttons=%06lx",
                          id,
                          MOTIONEVENT_ACTION_LABEL(action),
--- a/app/src/control_msg.c
+++ b/app/src/control_msg.c
@@ -1,12 +1,59 @@
 #include "control_msg.h"
 
 #include <assert.h>
+#include <inttypes.h>
 #include <stdlib.h>
 #include <string.h>
 
 #include "util/buffer_util.h"
 #include "util/log.h"
 #include "util/str_util.h"
+
+/**
+ * Map an enum value to a string based on an array, without crashing on an
+ * out-of-bounds index.
+ */
+#define ENUM_TO_LABEL(labels, value) \
+    ((size_t) (value) < ARRAY_LEN(labels) ? labels[value] : "???")
+
+#define KEYEVENT_ACTION_LABEL(value) \
+    ENUM_TO_LABEL(android_keyevent_action_labels, value)
+
+#define MOTIONEVENT_ACTION_LABEL(value) \
+    ENUM_TO_LABEL(android_motionevent_action_labels, value)
+
+#define SCREEN_POWER_MODE_LABEL(value) \
+    ENUM_TO_LABEL(screen_power_mode_labels, value)
+
+static const char *const android_keyevent_action_labels[] = {
+    "down",
+    "up",
+    "multi",
+};
+
+static const char *const android_motionevent_action_labels[] = {
+    "down",
+    "up",
+    "move",
+    "cancel",
+    "outside",
+    "ponter-down",
+    "pointer-up",
+    "hover-move",
+    "scroll",
+    "hover-enter"
+    "hover-exit",
+    "btn-press",
+    "btn-release",
+};
+
+static const char *const screen_power_mode_labels[] = {
+    "off",
+    "doze",
+    "normal",
+    "doze-suspend",
+    "suspend",
+};
 
 static void
 write_position(uint8_t *buf, const struct position *position) {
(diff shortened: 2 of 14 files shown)


intgr:
PS: Thanks a lot for developing and maintainig scrcpy. :cupid:

rom1v:

I think that the log should not be in the "convert" method (wihch is a kind of util method), but in the caller.

Input events can be very verbose, they should be enabled in verbose mode only (`LOGV`) IMO.

For consistency, verbose logs could also be added for other input events (keyboard, touches). (This could be done separately later though.)

intgr:
:+1: I will try to make changes later today.

intgr:
Fixed your notes and added logging for finger events. Didn't implement keyboard event logging yet.

rom1v:
After reading your `LOGV` calls just before a `control_push_msg()`, I'm wondering if we shouldn't log all control messages (with [all its values](https://github.com/Genymobile/scrcpy/blob/d3d955f67b92d08bbee21bebce40fa072eea38f6/app/src/control_msg.h#L42-L74)).

For example, by adding a function `control_msg_log(const struct control_msg *msg)` called at the beginning of `control_push_msg()`. What do you think?

intgr:
Makes sense to me.

Do you think I should log `AMOTION_EVENT_ACTION_MOVE` events too or just up/down? For my use case, I would find logging all events a bit noisy.

For example, if I want to script an `adb shell input swipe` command, I need the start and end coordinates but not the intermediate ones. But I think it isn't a big deal.

Can you imagine any use cases, where logging intermediate events would be useful?

rom1v:
I would prefer to log all events sent to the device without distinction. If you want only some of them, it's easy to grep :)

intgr:
@rom1v I pushed a partial draft, if you happen to be around, it would be nice to hear if I'm going in the right direction with this.

I also noticed only now (embarrassing!) that scrcpy does not currently support a `SC_LOG_LEVEL_VERBOSE` log level, I suppose I need to add support for that as well?

app/src/android/input.h
```diff
@@ -91,6 +91,8 @@ enum android_input_event_type {
 
 /**
  * Key event actions.
+ *
+ * Also keep `android_keyevent_action_descriptions` updated.
```
rom1v, line 95: This file is copied from AOSP: [`include/android/input.h`](https://android.googlesource.com/platform/frameworks/native/+/refs/heads/master/include/android/input.h). Please do not modify it.

app/src/controller.c
```diff
@@ -4,6 +4,34 @@
...
 #include "util/log.h"
 
+/**
+ * Map an "enum" value from array to a string, without crashing on an
+ * out-of-bounds value.
+ */
+#define ENUM_TO_DESCRIPTION(descriptions, value) \
+    ((size_t) (value) < (sizeof(descriptions)/sizeof(descriptions[0])) ? descriptions[value] : "???")
+
+static const char *android_keyevent_action_descriptions[] = {
```
rom1v, line 14:
```c
static const char *const android_keyevent_action_descriptions[] = { … }
```

app/src/controller.c
```diff
@@ -4,6 +4,34 @@
 
 #include "util/log.h"
 
+/**
+ * Map an "enum" value from array to a string, without crashing on an
+ * out-of-bounds value.
+ */
+#define ENUM_TO_DESCRIPTION(descriptions, value) \
+    ((size_t) (value) < (sizeof(descriptions)/sizeof(descriptions[0])) ? descriptions[value] : "???")
```
rom1v, line 12: There is a macro `ARRAY_SIZE()`.

app/src/controller.c
```diff
@@ -45,9 +73,67 @@ controller_destroy(struct controller *controller) {
     receiver_destroy(&controller->receiver);
 }
 
+static void
+control_msg_log(const struct control_msg *msg) {
+    switch(msg->type) {
+        case CONTROL_MSG_TYPE_INJECT_KEYCODE:
+            LOGV("Event: key %s code=%u repeat=%u meta=%06x",
```
rom1v, line 80: The actual type of an enum is implementation defined ([specs](https://port70.net/~nsz/c/c11/n1570.html#6.7.2.2p4)), so the value should be cast to a concrete integer type (`int` for example, with `"%d"` as format),

The `repeat` value is `uint32_t`, so the format must be `PRIu32`.

_Idem for other values._

```c
LOGV("Event: key %s code=%d repeat=%" PRIu32 " meta=%06lx",
    ENUM_TO_DESCRIPTION(android_keyevent_action_descriptions, msg->inject_keycode.action),
    (int) msg->inject_keycode.keycode,
    msg->inject_keycode.repeat,
    (long) msg->inject_keycode.metastate); // in theory, int may be as small as 16bits, so not sufficient
```
_(keep your own indentation, I just wrote it from the github interface.)_

rom1v (commented):
> I pushed a partial draft, if you happen to be around, it would be nice to hear if I'm going in the right direction with this.

Yes, looks good :+1: Thank you.

I added few remarks inline.

> I also noticed only now (embarrassing!) that scrcpy does not currently support a `SC_LOG_LEVEL_VERBOSE` log level

Oops.

> I suppose I need to add support for that as well?

Yes please, as a separate commit :)

intgr:
Thanks! My C has been getting a bit rusty. Pushed an updated version.

* I added `--verbosity=verbose` option in a separate commit
* I went through all format string arguments and should have correct types now.
* Moved logging code to `control_msg.c`
* Added remaining event types

I was trying to avoid building the Java part. Took me a while to realize that I need to downgrade, seems like it doesn't work with JDK 16 (got some weird error "module java.base does not "opens java.io" to unnamed module").

app/src/control_msg.c
```diff
@@ -93,6 +128,74 @@ control_msg_serialize(const struct control_msg *msg, unsigned char *buf) {
...
+                 (int) msg->inject_keycode.keycode,
+                 msg->inject_keycode.repeat,
+                 (long) msg->inject_keycode.metastate);
+            break;
+        case CONTROL_MSG_TYPE_INJECT_TEXT:
+            LOGV("Input: text \"%s\"", msg->inject_text.text);
+            break;
+        case CONTROL_MSG_TYPE_INJECT_TOUCH_EVENT: {
+            size_t action = msg->inject_touch_event.action & AMOTION_EVENT_ACTION_MASK;
+            LOGV("Input: touch %-4s position=%" PRIi32 "x%" PRIi32 " pressure=%g buttons=%06lx",
```
intgr, line 147: I omitted `pointer_id` from this message because it didn't didn't seem very useful.

server/src/main/java/com/genymobile/scrcpy/Server.java
```diff
@@ -251,6 +251,7 @@ public void uncaughtException(Thread t, Throwable e) {
         Options options = createOptions(args);
 
         Ln.initLogLevel(options.getLogLevel());
+        Ln.v("Using verbose log level");
```
intgr, line 254: Should I drop this log and leave the `v()` method unused?

app/src/control_msg.c
intgr, line 147: The `%-4s` makes output more pleasantly aligned
```
VERBOSE: Input: touch down position=286x786 pressure=1 buttons=000001
VERBOSE: Input: touch move position=287x786 pressure=1 buttons=000001
VERBOSE: Input: touch move position=288x786 pressure=1 buttons=000001
VERBOSE: Input: touch up   position=288x786 pressure=0 buttons=000001
```

app/src/control_msg.c
```diff
@@ -8,6 +8,41 @@
 #include "util/log.h"
 #include "util/str_util.h"
 
+/**
+ * Map an enum value to a string based on an array, without crashing on an
+ * out-of-bounds value.
+ */
+#define ENUM_TO_LABEL(labels, value) \
```
intgr, line 15: Renamed "description" -> "label" everywhere.

rom1v: :+1: (or even `ENUM_TO_STRING`?)

server/src/main/java/com/genymobile/scrcpy/Server.java
rom1v: I think you should drop it.

app/src/control_msg.c
rom1v: I think you should log it, it allows to distinguish several simultaneous fingers on touch events.

app/src/cli.c
```diff
@@ -184,7 +184,7 @@ scrcpy_print_usage(const char *arg0) {
         "\n"
 #endif
         "    -V, --verbosity value\n"
-        "        Set the log level (debug, info, warn or error).\n"
+        "        Set the log level (verbose, debug, info, warn or error).\n"
```
rom1v, line 187: :+1:

Also in the manpage:

```diff
diff --git a/app/scrcpy.1 b/app/scrcpy.1
index 7b6f1faa..62dc9677 100644
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -193,7 +193,7 @@ It requires to lock the video orientation (see --lock-video-orientation).
 
 .TP
 .BI "\-V, \-\-verbosity " value
-Set the log level ("debug", "info", "warn" or "error").
+Set the log level ("verbose", "debug", "info", "warn" or "error").
 
 Default is "info" for release builds, "debug" for debug builds.
 
```

rom1v:
Hi,

I allowed myself to work on this based on your branch. I did the following changes:
 - introduce a few more macros to simplify
 - log pointer id (either `mouse`, `vfinger` or a numeric id)

The logs highlighted that motion events were sent even if left-click was not pressed (but for example right-click or 4th button), so I fixed it: 2b7f19d2699ed1a16342605a72740a3fdc377460 :+1:

Here is a branch: [`logv`](https://github.com/Genymobile/scrcpy/commits/logv). Please review/test :wink:

I plan to do few more things before merging:
 - move all log level stuff to log.h/log.c
 - expose a function to know the current log level
 - do not call `control_msg_log()` if the current log level is not verbose.

rom1v:
> I plan to do few more things before merging:
> - move all log level stuff to log.h/log.c
> - expose a function to know the current log level
> - do not call `control_msg_log()` if the current log level is not verbose.

Done. I updated the `logv` branch.

app/src/control_msg.c
intgr: Yeah I slightly prefer "label" too :)

intgr:
> I discovered, thanks to the logs, that motion events were sent even if left-click was not pressed

Thanks, cool to hear that there are other use cases/benefits of this work :)

intgr:
How do we proceed? Should I pull your changes to my branch? I'll do that for now.

rom1v:
> How do we proceed? Should I pull your changes to my branch?

Not necessarily, just tell me if it's ok or if some thkngs should be changed :) I'll merge manually.

app/src/controller.c
```diff
@@ -48,6 +48,10 @@ controller_destroy(struct controller *controller) {
 bool
 controller_push_msg(struct controller *controller,
                       const struct control_msg *msg) {
+    if (sc_get_log_level() == SC_LOG_LEVEL_VERBOSE) {
```
intgr, line 51: I'd use `(sc_get_log_level() <= SC_LOG_LEVEL_VERBOSE)` even if we don't expect any lower log levels.

intgr:
Other than that minor comment, this looks good to me.

> Not necessarily, just tell me if it's ok or if some thkngs should be changed :) I'll merge manually.

OK, I already pushed it, I'm sure you can sort it out :)

app/src/control_msg.c
```diff
@@ -93,6 +140,94 @@ control_msg_serialize(const struct control_msg *msg, unsigned char *buf) {
...
+            LOG_CMSG("text \"%s\"", msg->inject_text.text);
+            break;
+        case CONTROL_MSG_TYPE_INJECT_TOUCH_EVENT: {
+            int action = msg->inject_touch_event.action
+                       & AMOTION_EVENT_ACTION_MASK;
+            uint64_t id = msg->inject_touch_event.pointer_id;
+            if (id == POINTER_ID_MOUSE || id == POINTER_ID_VIRTUAL_FINGER) {
+                // string pointer id
+                LOG_CMSG("touch [id=%s] %-4s position=%" PRIi32 ",%" PRIi32
+                             " pressure=%f buttons=%06lx",
```
intgr, line 164: I used `%g` for pressure so it hides the fractional part when there are no digits. I guess you changed this intentionally?

`pressure=0.000000` vs `pressure=0`

app/src/controller.c
rom1v: :+1:

app/src/control_msg.c
rom1v: Oh, yes, I forgot to discuss this change, the reason was that `%g` prints exponential notation when necessary, and I think we don't want that here:

```c
printf("%f %g\n", 0.0000001f, 0.0000001f);
```

prints:

```
0.000000 1e-07
```

But yes, `%g` is probably better in practice. I will change back to `%g`.

I also changed `position=500x400` to `position=500,400`, since `500x400` suggests a dimensions rather than a position.

rom1v: Pushed to `logv`.

rom1v:
> OK, I already pushed it, I'm sure you can sort it out :)

OK, it's indeed better that you pushed it, it's more convenient for discussions :wink:

app/src/control_msg.c
intgr: Agreed, the `,` separator makes sense.

intgr:
Pushed again. Seems fine to me now :+1:

rom1v:
> Seems fine to me now +1

Thanks :+1:

Then let's merge into `dev` :tada:

PR #2383, opened 2021-06-13
Request scan media on push

rom1v:
After a file is pushed (by drag&drop), send a broadcast to request the media scanner to scan the "push target" directory.
    
In practice, this allows some apps to detect the new file immediately.

---

Refs https://github.com/Genymobile/scrcpy/pull/2265#issuecomment-860182404 cc @brunoais 

In the end, I'm not convinced by the feature, for two reasons:
 - [`ACTION_MEDIA_SCANNER_SCAN_FILE`](https://developer.android.com/reference/android/content/Intent?hl=en#ACTION_MEDIA_SCANNER_SCAN_FILE) is now deprecated
 - in practice, I tested some apps, either they refresh immediately on file push (with or without the change), either they don't (with or without the change)

In the past, it was useful, but here even on a test device with Android 6, it does not seem very useful (maybe I should test more).

### DIFF
--- a/app/src/control_msg.c
+++ b/app/src/control_msg.c
@@ -80,6 +80,13 @@ control_msg_serialize(const struct control_msg *msg, unsigned char *buf) {
         case CONTROL_MSG_TYPE_SET_SCREEN_POWER_MODE:
             buf[1] = msg->set_screen_power_mode.mode;
             return 2;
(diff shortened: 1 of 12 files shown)


brunoais:
Unfortunately, I can't say much about it because I don't know that mechanism of Android. If you are not convinced. Certainly it won't be me who does :).

BTW, I'm now working on the command line options for the intents

PR #2384, opened 2021-06-13
Push to /sdcard/Download/ by default

rom1v:
Change the default push target from `/sdcard/` to `/sdcard/Download/`.

Pushing to the root of `/sdcard/ is not very convenient, many apps do not expose its content directly.

It can still be changed by `--push-target`.

---

I would like to change the default push target. What do you think? (cc @brunoais)

### DIFF
--- a/README.md
+++ b/README.md
@@ -709,15 +709,15 @@ There is no visual feedback, a log is printed to the console.
 
 #### Push file to device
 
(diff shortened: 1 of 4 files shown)


brunoais:
I would certainly do that.

rom1v:
Merged into `dev`.

PR #2008, opened 2021-01-03
Build by CMake

sman2013:
With this branch, we can debug with Clion.

Issue #1358, opened 2020-05-05
Pass server arguments over socket


rom1v:
I tried to improve readability/robustness of configuration passed from the client ot the server: 870ced088ef469829722678f71a05412db34e88d (branch `name_params`)

But on Windows, it failed:

```
ERROR: Command too long (255 chars)
ERROR: Failed to execute: [adb], [shell], [CLASSPATH=/data/local/tmp/scrcpy-server.jar], [app_process], [/], [com.genymobile.scrcpy.Server], [1.13], [max_size=0], [bit_rate=8000000], [max_fps=0], [lock_video_orientation=-1], [tunnel_forward=false], [crop=], [send_frame_meta=true], [control=true], [display_id=0], [show_touches=false], [stay_awake=false] 
```

I think many of these key=value pairs should be passed over the socket (not `tunnel_forward` because the socket creation depends on it).

Or maybe you have a better idea?

Biswa96:
Do you mean using some kind of JSON, XML or INI file?

berniegp:
I got redirected here by this Google search: "scrcpy Command too long (255 chars)"

My problem is that I set this
```
ADB="C:\Program Files\Unity\Hub\Editor\2020.2.1f1\Editor\Data\PlaybackEngines\AndroidPlayer\SDK\platform-tools\adb.exe"
```

and got this:
```
C:\Builds>scrcpy --max-size 720 --bit-rate 1M --max-fps 30 --show-touches --always-on-top
INFO: scrcpy 1.17 <https://github.com/Genymobile/scrcpy>
C:\ProgramData\chocolatey\lib\scrcpy\tools\scrcpy-server: 1 file pushed. 2.3 MB/s (34930 bytes in 0.014s)
ERROR: Command too long (255 chars)
ERROR: Failed to execute: [C:\Program Files\Unity\Hub\Editor\2020.2.1f1\Editor\Data\PlaybackEngines\AndroidPlayer\SDK\platform-tools\adb.exe], [shell], [CLASSPATH=/data/local/tmp/scrcpy-server.jar], [app_process], [/], [com.genymobile.scrcpy.Server], [1.17], [info], [720], [1000000], [30], [-1], [false], [-], [true], [true], [0], [true], [false], [-], [-]
```

I get this error even if I enable long path support in Windows 10 via
```
HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled
```

[...]

PR #2397, opened 2021-06-17
Initial work done on making it work natively on all platforms/buildsystems 

Wirtos:
Tried to make it less platform-dependent, the only bug so far is pts assertion at line 45, stream.c if you try to close the window. Not sure what causes it. By integrating vcpkg and cmake manifest adding more libraries will be more trivial too. A review would be nice. 
All changes narrow down to removing gcc function macros by turning them into comma operator ones, removing C11 stdatmic and using SDL atomics instead, also removed VLAs and used size_t instead of ssize_t since there's no real need to use ssize_t at all.


app/src/sys/win/command.c
```diff
@@ -27,7 +32,7 @@ cmd_execute(const char *const argv[], HANDLE *handle) {
     memset(&si, 0, sizeof(si));
     si.cb = sizeof(si);
 
-    char cmd[256];
+    char cmd[512];
```
rom1v, line 35: Oh, thank you: https://github.com/Genymobile/scrcpy/issues/1358#issuecomment-862989748

rom1v (commented):
Hi,

Thank you for your work.

> By integrating vcpkg and cmake manifest adding more libraries will be more trivial too.

However, I will not merge another build system, we will not maintain several build systems (https://github.com/Genymobile/scrcpy/pull/2008#issuecomment-753622027).

> removing C11 stdatmic and using SDL atomics instead

SDL atomics usage had been removed by 54ccccd8839f16cc289f98160882b909ea7aa3c0. In the end, I would like to make a "core" independant of SDL (and use SDL only for the UI).
The scrcpy target is C11.

> used `size_t` instead of `ssize_t` since there's no real need to use `ssize_t` at all.

We use `ssize_t` in VLC, which is compiled for many platforms without problems. ([OK](https://mailman.videolan.org/pipermail/vlc-devel/2020-October/139507.html), we don't support MSVC, but this specific `ssize_t` issue could be fixed by a typedef)

(Also, with `size_t` raw comparisons with -1 will warn with `-Wsign-compare`, and `< 0` are invalid.)

> also removed VLAs

I'm ok to remove VLAs.

> the only bug so far is pts assertion at line 45, stream.c if you try to close the window. Not sure what causes it.

```diff
diff --git a/app/src/stream.c b/app/src/stream.c
index d1b8b9f3..cdf682c1 100644
--- a/app/src/stream.c
+++ b/app/src/stream.c
@@ -38,7 +38,7 @@ stream_recv_packet(struct stream *stream, AVPacket *packet) {
 
     uint64_t pts = buffer_read64be(header);
     uint32_t len = buffer_read32be(&header[8]);
-    assert(pts == NO_PTS || (pts & 0x8000000000000000) == 0);
+    assert(pts == NO_PTS || (pts & 0x8000000000000000UL) == 0);
     assert(len);
 
     if (av_new_packet(packet, len)) {
```

(EDIT: I'm ok with your `>> 63` too :+1:)

?

Wirtos:
> The scrcpy target is C11.
Well, MSVC has a roadmap on supporting C11 threading, but knowing them it's not going to happen anytime soon.
The solution is to create adapter functions + typedefs which would make other atomics/theads a drop-in replacement. 

> We use `ssize_t` in VLC, which is compiled for many platforms without problems. ([OK](https://mailman.videolan.org/pipermail/vlc-devel/2020-October/139507.html), we don't support MSVC, but this specific `ssize_t` issue could be fixed by a typedef)
I don't see why non-portable ssize_t might be in use here since you can't replicate its behaviour with standard language tools. (size_t) -1 fits just fine and follows the standard. 
> (Also, with `size_t` raw comparisons with -1 will warn with `-Wsign-compare`, and `< 0` are invalid.)
Indeed, probably should add an explicit cast.

> > the only bug so far is pts assertion at line 45, stream.c if you try to close the window. Not sure what causes it.
> ?
When closing the window assertion happens with MSB set for some reason. Some kind of race condition?

> However, I will not merge another build system, we will not maintain several build systems (#2008 (comment)).
Completely up to you, but imo it's way cleaner plus allows to integrate nicely with source-based dependency manager like vcpkg without a ton of platform-specific code to get ffmpeg, for example. A lot of projects use cmake and it allows to use make/visual studio generators instead of ninja. If needed, I might replace meson with cmake if needed.

rom1v:
> > The scrcpy target is C11.
>
> Well, MSVC has a roadmap on supporting C11 threading, but knowing them it's not going to happen anytime soon.

I understand your wish to use MSVC, but supporting exotic (and non-open source) compilers which do not support C11 is not a goal for scrcpy (yes, I'm :trollface:ing a bit). It brings additional complexity that I would like to avoid.

We can make some small changes to fix issues with MSVC, but not prevent using C features that MSVC do not provide. MSVC support is not a goal (I'm sorry).

I try to only use standard things when possible, but if a convenient feature is available in both `gcc` and `clang`, then I might use it (I'm less strict than VLC on that point). This can be discussed though: for example, if the only blocking point to support MSVC was the usage of "compound statements" in macros, then I would adapt the code.

> When closing the window assertion happens with MSB set for some reason. Some kind of race condition?

Because if you change the type from `ssize_t` to `size_t`, then `-1` is not `< HEADER_SIZE` anymore:
https://github.com/Genymobile/scrcpy/pull/2397/files#diff-10a4372c63b8fcd8c07fc76d3f5e8ac626be9a4eef1811c6031400aa1d1f7b08L40

> imo it's way cleaner plus allows to integrate nicely with source-based dependency manager like vcpkg without a ton of platform-specific code to get ffmpeg

I agree that building for Windows with the dependencies is a PITA. But for other systems, it's just:

```
dependencies = [
    dependency('libavformat'),
    dependency('libavcodec'),
    dependency('libavutil'),
    dependency('sdl2'),
]
```

If there is an open source package manager which provides/builds FFmpeg and SDL dependencies easily for Windows (but uses the dependencies from the system for other platforms), compatible with Meson, then why not? My own additional constraint: the Windows release must be cross-compiled from Linux, the whole release must be buildable from Linux.

Also see discussions on #1753.

> A lot of projects use cmake and it allows to use make/visual studio generators instead of ninja. If needed, I might replace meson with cmake completely.

No, thanks. I definitely want to keep Meson over CMake. Sorry.

Wirtos:
>It brings additional complexity that I would like to avoid.

Fair. What about making conditional `__STDC_NO_ATOMICS__` along with `__STDC_VERSION__` macro replacement with SDL counter-variants then? it's an optional C11 feature, so might as well define SDL fallback. 

> Because if you change the type from `ssize_t` to `size_t`, then `-1` is not `< HEADER_SIZE` anymore:
> https://github.com/Genymobile/scrcpy/pull/2397/files#diff-10a4372c63b8fcd8c07fc76d3f5e8ac626be9a4eef1811c6031400aa1d1f7b08L40

Whoops, a little bit of an oversight on my part, shouldn't have used automatic type replacement 👀

> No, thanks. I definitely want to keep Meson over CMake. Sorry.

Roger that. It still should work just fine for mason even with the given changes code-wise. 
So what changes should I actually keep/remove?

rom1v:
> What about making conditional `__STDC_NO_ATOMICS__` along with `__STDC_VERSION__` macro replacement with SDL counter-variants then? it's an optional C11 feature, so might as well define SDL fallback.

That would still be a SDL dependency. One of my goal (in the future) is to have a "scrcpy-core" which does not depend on SDL at all. Moreover, the SDL version is very limited, and does not support the memory order parameter, so it's not just a function name replacement.

> So what should I actually keep/remove?

I'm interested in the VLA removals, and a change [to grow the Windows cmd buffer](https://github.com/Genymobile/scrcpy/pull/2397/files) (even to 8192, the max length on Windows, but with a `malloc()` instead of stack-allocated).

If you could, please open a PR for each (otherwise, please tell me).
Also, you should base your changes on `dev`, not `master`.

Thank you :wink:

Wirtos:
k, will do

PR #2405, opened 2021-06-19
Vla/static buffer removal

Wirtos:


rom1v:
Thank you. But since it's just local, IMO it's overkill to compute the length then alloc, you could just alloc 8192 directly (and it is freed immediately).

Wirtos:
> 
> 
> Thank you. But since it's just local, IMO it's overkill to compute the length then alloc, you could just alloc 8192 directly (and it is freed immediately).

Thought about it too. What about the length of additional args? Should I account for them too?

rom1v:
Thank you very much.

I made the following changes:
 - remove the reverted commits
 - rename `cmd` to `argv` in a separate commit (for diff clarity)
 - remove `CMD_MAX` from `process.h` and used separate local constants for windows implementation and error message
 - simplify the logic in `sys/win/process.c`

I pushed to branch [`novla`](https://github.com/Genymobile/scrcpy/commits/novla). Please review the result to tell me if this is ok for you.

Regards

Wirtos:
Seems fine, not entirely sure why len here https://github.com/Genymobile/scrcpy/blob/b24be9e2c2527b4667568551d93102151b4fc16d/app/src/adb.c#L84 differs from here https://github.com/Genymobile/scrcpy/blob/b24be9e2c2527b4667568551d93102151b4fc16d/app/src/sys/win/process.c#L9
since they're both being fed the same argv?

rom1v:
> Seems fine, not entirely sure why len here […]

Oh, my inline reviews were not committed on github (still pending). The reason is explained here: https://github.com/Genymobile/scrcpy/pull/2405#discussion_r654840475
I guess 512 would be sufficient, but it's arbitrary (In practice, it will not make any difference.)

rom1v:
Merged. :+1:

PR #2406, opened 2021-06-19
remove compound statements

Wirtos:


app/src/util/queue.h
```diff
@@ -2,8 +2,6 @@
 #ifndef QUEUE_H
 #define QUEUE_H
 
-#include "common.h"
```
rom1v, line 5: `common.h` is willingly included in all source files, at least to always correctly load defines from `config.h` and `compat.h`. This is a convention I introduced to avoid this kind of error: 068253a3a299a60d78c7c1b0e760f4bbe3b75bc1

rom1v (commented):
TBH, I'm a reluctant to remove compound statements from scrcpy. It's more readable and convenient, and is supported by gcc and clang. I know that it's not supported by MSVC, but anyway there are other problems with MSVC, so it does not bring benefits to remove them.

For the little story, I implemented the generic vector ([`vlc_vector`](https://code.videolan.org/videolan/vlc/-/commit/983c43f05928032a14f201c506d6b9c51d0c5145)) in VLC few years ago, and was annoyed not to be able to use compound statements (and also `typeof`).

So after that I was pleased to be able use this feature in scrcpy :)

Wirtos:
As for template data structures, things like https://github.com/rxi/vec or https://github.com/Wirtos/cmap worked just fine for me without the need for any compiler extensions, but since I've got my fork building for any general compiler/platform I don't want too invasive.

rom1v:
> As for template data structures, things like https://github.com/rxi/vec

Thank you for the reference. After a quick look, the impl does not seem correct in term of pointer aliasing (a `T**` could not be cast to a `char **`). Cf [discussion](https://mailman.videolan.org/pipermail/vlc-devel/2018-August/121078.html).

Wirtos:
> (in theory, a `T**` should not be cast to a `char **`). 

Indeed, my local version uses void** and iirc this exact impl has a bug in insert macro, which always returns success, even if underlying function fails

rom1v:
> Indeed, my local version uses void**

But even with `void **` I think it is a strict aliasing violation: it aliases a `T **` with a `void **`.

It would be great to produce a real sample to demonstrate the issue, but here is a sample with a simpler aliasing (`struct A*` vs `struct B*`):

```c
#include <stdio.h>

struct A { int x; };
struct B { int x; };

void f(struct A *a, struct B *b) {
    a->x = 2;
    b->x = 3;
    printf("%i\n", a->x);
}

int main() {
    struct A a = { .x = 1 };
    f(&a, (struct B *) &a);
    return 0;
}
```

```
$ gcc a.c && ./a.out
3
$ gcc -O2 a.c && ./a.out
2
```

rom1v:
I can adapt the sample to highlight the strict aliasing violation with a `char **`:

```c
#include <stdio.h>

static int v0 = 100;
static char v1 = 42;

void f(int **a, char **b) {
    *a = &v0;
    *b = &v1;
    printf("%i\n", **a);
}

int main() {
    int data = 5;
    int *a = &data;
    f(&a, (char **) &a);
    return 0;
}
```
```
$ gcc a.c && ./a.out
42
$ gcc -O2 a.c && ./a.out
100
```

However, I cannot expose the issue with a `void **` (which does not mean that it would be correct).

Wirtos:
> However, I cannot expose the issue with a `void **` (which does not mean that it would be correct).

Iirc it's UB, valid impl would require an intermediate void * variable, having a void ** cast simply silents the warning in compilers (which is still invalid)

PR #2464, opened 2021-07-06
Implement (optional) buffering

### LINKED ISSUE #2417: Increasing Latency
Is there a way to increase latency? Weird question but mainly because I dont need low latency, but instead smooth gameplay with minimal lag spikes. Since I will be streaming, I can always adjust audio to sync in OBS afterwards.

Thanks.

rom1v:
To minimize latency, scrcpy always displays a frame as soon as it is available, without waiting. This design decision is (on purpose) at the cost of jitter: the delay between frames is not constant.

In an ideal world, every frame produced by the device would be displayed some constant time later on the device (for simplicity, the schema assumes a constant framerate):

nojitter

However, in reality, the delay is not constant; the encoding time, transfer time and decoding time may vary from frame to frame:

jitter

In practice, this is not a problem for many real-time use cases. It is slightly noticeable when playing a video on the device (or anything involving regular movement) while mirroring.

On recording (`--record`), scrcpy captures the timestamps on the device and write them in the resulting file, so that the playback is correct without jitter.

However, some real-time use cases might benefit from adding a small latency to compensate for jitter too. For example, few tens of seconds of latency for live-streaming are not important, but jitter is noticeable.

To support this, the principle is to delay the frames to compensate for jitter (notice how the blue circles have the same spacing as the green circles):

buffering

This PR implements optional buffering:

```bash
scrcpy --display-buffer=50  # add 50 ms buffering for display
scrcpy --v4l2-buffer=500    # add 500 ms buffering for v4l2 sink
```

Display buffering and [v4l2](https://github.com/Genymobile/scrcpy#v4l2loopback) are independent, so that it is possible for example to control it with minimal delay but stream it with a larger delay.

In practice, here is a graph showing the difference in smoothness (after a small initialization phase) with a buffer of 50ms on a wireless connection:

capture_graph3
([graph_old_impl](https://user-images.githubusercontent.com/543275/124667050-29641500-deaf-11eb-8796-6e2403c2bceb.png))

Fixes #2417

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,24 @@ For example, you could capture the video within [OBS].
 [OBS]: https://obsproject.com/fr
 
 
+#### Buffering
+
+It is possible to add buffering. This increases latency but reduces jitter (see
+#2464).
+
+The option is available for display buffering:
+
+```bash
+scrcpy --display-buffer=50  # add 50 ms buffering for display
+```
+
+and V4L2 sink:
+
+```bash
+scrcpy --v4l2-buffer=500    # add 500 ms buffering for v4l2 sink
+```
+
+
 ### Connection
 
 #### Wireless
--- a/app/meson.build
+++ b/app/meson.build
@@ -10,6 +10,7 @@ src = [
     'src/event_converter.c',
     'src/file_handler.c',
     'src/fps_counter.c',
+    'src/frame_buffer.c',
     'src/input_manager.c',
     'src/opengl.c',
     'src/receiver.c',
--- a/app/meson.build
+++ b/app/meson.build
@@ -25,6 +25,7 @@ src = [
     'src/util/process.c',
     'src/util/str_util.c',
     'src/util/thread.c',
+    'src/util/tick.c',
 ]
 
 if host_machine.system() == 'windows'
--- a/app/meson.build
+++ b/app/meson.build
@@ -2,6 +2,7 @@ src = [
     'src/main.c',
     'src/adb.c',
     'src/cli.c',
+    'src/clock.c',
     'src/compat.c',
     'src/control_msg.c',
     'src/controller.c',
--- a/app/meson.build
+++ b/app/meson.build
@@ -168,6 +168,10 @@ if get_option('buildtype') == 'debug'
             'src/cli.c',
             'src/util/str_util.c',
         ]],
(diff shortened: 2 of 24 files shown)


rom1v:
I improved the smoothness on start, reworked a bit and fixed some bugs.

orogenic:
Public note to self. I am curious to look more into v4l2loopback performance limitations. I noticed v4l2loopback appears less stable - as well possibly lower - framerate compared to the scrcpy window, with no buffering in both cases. I became curious is there some performance issue with scrcpy's v4l2loopback implementation, or with v4l2loopback itself, that can account for this discrepancy I am noticing with no buffering in both cases?

Also, irrelevant, curious about the v4l2loopback-ctl set-fps command and how it can interact with scrcpy --max-fps option.

rom1v:
> I noticed v4l2loopback appears less stable - as well possibly lower - framerate compared to the scrcpy window

Which player do you use to play the v4l2 stream?

orogenic:
obs, discord, ffplay. they all give similar results. since the above note started so unscientifically, I feel like apologizing... there are a lot of configuration options I am experimenting with... I may be blaming v4l2loopback for something unrelated... I was fiddling more this morning and started removing options one-by-one from my main script I use --v4l2-sink with... first I removed the --no-display option and compared the video players to the scrcpy window, and the same stability issue was present in both... that's the first thing I should have tried, sorry. then I removed the --turn-screen-off option and the stability of both the window and v4l2loopback stream seemed to return to normal. this is still just my perception, but whatever I'm noticing may have to do with the --turn-screen-off option, not anything to do with v4l2loopback. apologies.

shersty:
> On recording (--record), scrcpy captures the timestamps on the device and write them in the resulting file, so that the playback is correct without jitter.

hello, does this mean that when recording a video, scrcpy will get the timestamp corresponding to the frame on the device (mobile phone), and then encode the video according to this timestamp? How can I get this exact timestamp?

rom1v:
> does this mean that when recording a video, scrcpy will get the timestamp corresponding to the frame on the device (mobile phone)

Yes.

> How can I get this exact timestamp?

Grep `pts` in this file: https://github.com/Genymobile/scrcpy/blob/master/app/src/recorder.c

PR #2490, opened 2021-07-15
fix gentoo ebuild name

pascalbrax:
A tiny fix in the gentoo's ebuild name syntax.


rom1v:
I think it is correct: contrary to the others which points directly to a package, the link for gentoo points to a directory of packages (with different versions), hence `scrcpy/` instead of `scrcpy`. The prefix of the repo (`app-mobilephone`) is irrelevant.

PR #2519, opened 2021-07-26
add plugin system

gamegrd:
###  You can use `PostMessage` to operate the device on the windows platform

`````python
 
import win32api,win32gui,win32con

hld = win32gui.FindWindow("SDL_app", None)
if hld > 0:
    SDL_FINGERDOWN      = 0x700
    WM_PLUGIN_BASE     = win32con.WM_USER
    x = int( (55 / 270) * (1000*1000) )
    y = int( (515 / 555) * (1000*1000) )
    win32gui.PostMessage(hld, WM_PLUGIN_BASE + SDL_FINGERDOWN ,x, y)
    win32gui.PostMessage(hld, WM_PLUGIN_BASE + SDL_FINGERDOWN +1  ,x, y)

 
`````

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,36 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ gamegrd ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: BSFishy/meson-build@v1.0.3
+
+      # Runs a single command using the runners shell
+      - name: Meson Build
+        run: |
+          meson x --prefix /usr --buildtype release --strip -Db_lto=true -Dc_link_args=-lcrypto --cross-file cross.txt meson x --prefix /usr --buildtype debug --strip -Db_lto=true -Dc_link_args=-lcrypto
+          
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
+        run: |
+          echo Add other actions to build,
+          echo test, and deploy your project.
--- a/app/meson.build
+++ b/app/meson.build
@@ -2,6 +2,7 @@ src = [
     'src/main.c',
     'src/adb.c',
     'src/cli.c',
+    'src/plugin.c',
     'src/compat.c',
     'src/control_msg.c',
     'src/controller.c',
--- a/app/src/plugin.c
+++ b/app/src/plugin.c
@@ -1,0 +1,49 @@
+#include "plugin.h"
+#include "scrcpy.h"
+#include <stdio.h>
+#include <string.h>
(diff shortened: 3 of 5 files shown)


Helaer:
@rom1v  Great, exactly what most people need!

brunoais:
@gamegrd IMO, if you add a plugin system, you add for all builds, not just windows builds.

Bethk85 (approved):

svaningelgem:
Hello @gamegrd, I took your code and improved it a bit:
https://github.com/svaningelgem/scrcpy/tree/feature/interactions_via_code

The way I improved on it:
- you can click
- you can type
- you can take a screenshot (BMP only because SDL doesn't support anything else)

This way you could interact fully with a scrcpy in an automated way. (take a screenshot, find a certain button, click it, type something etc etc)


@brunoais : I completely agree with your statement that this should support more than just Windows. However, the only thing I could find to support sending signals on Linux is the `kill` command. Could you give me some pointers on how to achieve this on Linux & Mac? I would love to give it a try to add this.

brunoais:
@svaningelgem In linux realm, there's no message system by default. There's a communications daemon which is common but not necessarily available. It's common enough that if it is required for the task it's acceptable.
It's called `dbus`.
I think the API is decently explained here: http://www.matthew.ath.cx/misc/dbus
It matches with what I can remember right now.

I have no idea about Mac realm. However, you can ditch dbus completely and use a client-server relationship (my kind of solution).

You can expose a [fifo/named pipe](https://www.man7.org/linux/man-pages/man7/fifo.7.html) to receive bytes. That would be represented as a file to write to in UNIX (linux, mac, openBSD, etc...) and a pipe location in windows.
I think for win32, it's as explained here: https://docs.microsoft.com/en-us/windows/win32/ipc/pipe-names

That would be the same style that many multiplatform programs work, such as mysql.

As for the formatting, for example, to split messages, a simple non-blocking read that checks if the last byte is a certain special byte would suffice while escaping that byte is done by sending it twice. That is a very simple algorithm also used for physical hardware communication.
As for the encoding, you can go with a mimcry of `ControlMessageReader.java` kind of formatting or you can come up with something more complete.

Is this enough for you?

svaningelgem:
Thanks for the lengthy explanation!

I'm wondering if I would be allowed to introduce new dependencies? I quickly glanced and there is Boost.Interprocess, and there are protobuffers that could help. Additionally there is signal/mmap in Python (search term: python ipc).

The simplest way would be to start a tcp server on localhost, but I don't know if that's allowed either? There are enough one-file cross-platform server implementation I can include if the license is correct.

What are your thoughts on the additional inclusions or is a tcp server a more preferable approach?

brunoais:
> Thanks for the lengthy explanation @brunoais !
> 
> I'm wondering if I would be allowed to introduce new dependencies? I quickly glanced and there is Boost.Interprocess, and there are protobuffers that could help. Additionally there is signal/mmap in Python (search term: python ipc).

I don't know if you can add more dependencies... @rom1v is the only one that can answer to that. I think there's no need.

I don't know Boost, so I can't comment. Protobuffer is not only an overkill, it's also a mess to deal with. The error messages are mediocre and fails easily (API not so good).

> The simplest way would be to start a tcp server on localhost, but I don't know if that's allowed either? 

I think that is fine. Then publicize the port of that server in a file in some default location. In unix, it would be in `/run`.

> There are enough one-file cross-platform server implementation I can include if the license is correct.

File cross-platform implementation? Then it's not tcp? If there's an implementation of a server and its license is compatible with scrcpy's license, it should be ok.

> What are your thoughts on the additional inclusions or is a tcp server a more preferable approach?

I'd prefer either a pipe or tcp.

> [ A very nice http library is: https://github.com/yhirose/cpp-httplib.git, MIT matches Apache, so that's fine -- any other options that you know ? ]

`scrcpy` is written in C, not C++. So that library is not a good option.

PR #2582, opened 2021-08-24
Upgrade SDL (2.0.14) Adb (31.0.3) for Windows

nkh0472:
They are the latest versions.
ADB 31.0.2 --> 31.0.3
SDL2 2.0.14 --> 2.0.16

---
Maybe there is some better solution to fetch the latest deps:

For ADB, we can use `curl -L -O https://dl.google.com/android/repository/platform-tools-latest-windows.zip`, this only affectS `./prebuilt-deps/Makefile`.

For SDL2, I'm not sure how to do this, but at least we can use some kind of variate, so we only need to change the dictionary name in one file (like in `./prebuilt-deps/Makefile`), instead of in 4 files (including `./prebuilt-deps/Makefile`, `./cross_win32.txt`, `./cross_win64.txt`, `./release.mk`).

Additionally, the situation about FFmpeg is more complicated, I'll open a new PR to describe it.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -76,7 +76,7 @@ The server uses 3 threads:
  - the **main** thread, encoding and streaming the video to the client;
  - the **controller** thread, listening for _control messages_ (typically,
    keyboard and mouse events) from the client;
- - the **receiver** thread (managed by the controller), sending _device messges_
+ - the **receiver** thread (managed by the controller), sending _device messages_
    to the clients (currently, it is only used to send the device clipboard
    content).
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -153,6 +153,26 @@ You may also need to configure the [scaling behavior]:
 [scaling behavior]: https://github.com/Genymobile/scrcpy/issues/40#issuecomment-424466723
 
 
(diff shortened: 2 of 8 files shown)


rom1v:
> They are the latest versions.
> ADB 31.0.2 --> 31.0.3
> SDL2 2.0.14 --> 2.0.16

Thank you.

I use to upgrade them just before a release (because in practice this is when it matters).

I may accept bumping versions at other times though. However, could you please rebase properly on `dev`, and create one commit for the adb upgrade and one for the SDL upgrade? (if the SDL upgrade is split in several commits, it will break on `git bisect`)

Example of previous upgrades:
 - ff7baad7093f4574eb04e82cebd7c1829310b663
 - d039a7a39a80940c3eca226ce3d7d0555157fa8c

> Maybe there is some better solution to fetch the latest deps:

I want to reference a specific version, so that a release from an old version can be re-created with the same dependencies (ok, it's only a best-medium effort, because I don't archive all the downloaded binaries). And bump the versions only after testing them.

nkh0472:
Done in #2588 and #2589.

PR #2595, opened 2021-08-26
Changes to turn scrcpy into a library and hook into video frames.

TeofilisMartisius:
You asked me to switch to "dev" branch. I have now done so.

This is continuation of work started on: https://github.com/Genymobile/scrcpy/pull/2591


TeofilisMartisius:
Ok, just in case you want to see how this is used, you can take a look here:

https://github.com/TeofilisMartisius/scrcpy-java

Issue #279, opened 2018-09-30
Send keyboard events as physical keyboards instead 欧风


AlynxZhou:

rom1v:
See https://github.com/Genymobile/scrcpy/issues/87#issuecomment-378775565 about OTG.

lasek101:
Hello
First thanks for Your work.
Now to my problem.
I use polish programmer phisical keyboard. It is normal qwerty keyboard with polish national characters obtaining by pressing right alt with specific letter. 
I have compiled, on linux box, client and server fom dev and rawalpha branch but can't get any polish diacritical letter. Pressing ctrl+k does nothing. 
What I'm doing wrong?

rom1v:
@lasek101 Unfortunately, I have no solution to generate such keys on the device.
See https://github.com/Genymobile/scrcpy/issues/193 and https://github.com/Genymobile/scrcpy/pull/194.

AlynxZhou:
@rom1v what about letting computer dealing with input method and just pass input string to phone?

npes87184:
Hi, @AlynxZhou 

Does the problem become to be able to input (Chinese) words from computer?
If the answer is yes, it may be a simple workaround to cover the problem which is some character cannot be prefectly handled. However, there are still some problems to deal.
For example, pass input string to where? clipboard or pass directly to the edittext? 

If we pass the string to clipboard, there will be inconvenience to use. We need to paste again and again.
On the other hand, if we want to directly pass to input field. As I know, there are many restricts. For instance, how about webview?

Thanks.

lasek101:
Maybe there is a hope. 
If I connect physical keyboard in OTG mode I can choose layout in Android (settings-languages and input methods - physical keyboard - then I choose keyboard I've connected and choose correct layout). After that I can get polish diacritical letters by pressing right alt and certain letter. 
Mayby there is a way to tell android that keyboard used by scrcpy is like OTG one.

rom1v:
@AlynxZhou
> what about letting computer dealing with input method and just pass input string to phone?

This is what it does (except for letters and space since v1.4, see #87).

The problem is that the device does not support injecting characters other than ASCII. See [handle text inputs](https://blog.rom1v.com/2018/03/introducing-scrcpy/#handle-text-input).

> If I connect physical keyboard in OTG mode

See https://github.com/Genymobile/scrcpy/issues/87#issuecomment-378775565.

> After that I can get polish diacritical letters by pressing right alt and certain letter.

[...]

PR #2268, opened 2021-04-19
Add v4l2loopback support

rom1v:
This is a follow up for v4l2loopback support (#2232 #2233) implemented by @martinellimarco.

I refactored scrcpy to support this feature properly, then I rebased @martinellimarco's work to use the refactored codebase.

The main purpose is to make the v4l2 component consume frames directly from the decoder, in order to avoid decoding the video stream twice if mirroring display is enabled:

```
                      screen/display
                    /
            decoder
          /         \
         /            v4l2_sink
  stream
         \
           recorder

          \--------/  \-------------/
         packet sinks   frame sinks
```

Here is a summary of the changes:
 - Remove `--render-expired-frames` option (see 8bae1f6b7fc7a82ea2d04c76e21fd127617c97d6 for details)
 - Remove compat with old FFmpeg decodeing and codec params API (deprecated since 2016)
 - Make video_buffer not consume the input frame (to make possible to send it to several sinks)
 - Add two abstractions (traits):
   - `sc_packet_sink`: a component which receives packets. It is implemented by the `decoder` and the `recorder`, and the `stream` now push its packets to the registered packet sinks (without depending on the concrete sink types)
   - `sc_frame_sink`: a component which receives frames. It is implemented by the `screen` (and now `v4l2`), and the `decoder` now pushes its frames to the registered frame sinks (without depending on the concrete sink types)
 - Add a new mode for `--lock-video-orientation`: `initial`. This locks the video in the orientation the device happens to be on start and avoids to pass an explicit value. This mode is set by default when v4l2 sink is enabled.
 - Rework the v4l2 sink component to implement `sc_frame_sink`, and be registered to the `decoder`.
 - Use a `video_buffer` to always send the very last frame, like for the display (it may skip frames, it's on purpose, is it a problem?)
 - Add some documentation in the README

Here is a usage summary:

```bash
sudo apt install v4l2loopback-dkms   # install v4l2loopback
sudo modprobe v4l2loopback           # create a video device
scrcpy --v4l2-sink /dev/video1       # mirror + v4l2
scrcpy --v4l2-sink /dev/video1 -N    # v4l2 only
ffplay -i /dev/video1                # test
```

The video stream could also be captured from [OBS](https://obsproject.com/).

Thank you for your review/feedbacks.

PR #2632, opened 2021-09-15
Add HID keyboard simulation via USB HID over AoAv2

AlynxZhou:
Add HID keyboard simulation via USB HID over AoAv2, which provides better input experience if connected with USB cable.

Currently this needs `--serial` to provide USB serial, we may update it to fetch serial via adb automatically.

A new option `--input-mode` is added, by default is `auto`, try HID first and fallback to old input injection, user can set `hid` or `inject` manually, it will fail if user's selection not work.

Please add your comments if you have any advice and I'll try to update it.

Fixes <https://github.com/Genymobile/scrcpy/issues/279>.


AlynxZhou:
Oh sorry I forgot Windows, I'll do it after I have my dinner.

rom1v:
Thank you for your work :+1:

I will try to review this week-end (this might take some time, it's a big change/PR :wink:).

AlynxZhou:
I've built libusb for windows, however it seems cannot open and compare serial for my phone, maybe some driver is needed but I don't know, windows is really a mess...

AlynxZhou:
Good news: I build windows version successfully with libusb dependencies.

Bad news: libusb unable to open Android phone on windows, I've searched a lot and only find a relevant page: <https://libusb-devel.narkive.com/vENuKzdR/getting-the-serial-number-of-the-camera-fails>, seems that windows has some limitation of accessing multi end point device, and libusb needs to visit end point 0 directly which windows does not allow.

So I think we have nothing to do with it, it will fallback to inject mode on windows, so just leave it here should be fine, and Linux users need this feature most (so they can use apps that do not have Linux version but Android version).

Please take your time to review it and I am waiting for advice! Thanks!

rom1v:
> Bad news: libusb unable to open Android phone on windows, I've searched a lot and only find a relevant page: https://libusb-devel.narkive.com/vENuKzdR/getting-the-serial-number-of-the-camera-fails, seems that windows has some limitation of accessing multi end point device, and libusb needs to visit end point 0 directly which windows does not allow.

OK, I never managed to use AOA on Windows either. It is totally ok to implement it for Linux only (at least as a first step). Moreover, there is already another feature (V4L2) available only on Linux. I suggest you keep your commit `Add Windows build support for libusb` locally, it might be useful later, but drop it from this PR for now.

Also, I have a little request (I didn't review yet, but looked at commit titles): could you `rebase -i` to clean up some history. For example, `Move HID keyboard setup code into functions`: this code could have been written directly "into functions" I guess. Or `Make sdl_keymod_to_hid_modifiers() more readable` (make it more readable directly). And other spurious changes to make clean commits… Thank you :)

AlynxZhou:
> > Bad news: libusb unable to open Android phone on windows, I've searched a lot and only find a relevant page: https://libusb-devel.narkive.com/vENuKzdR/getting-the-serial-number-of-the-camera-fails, seems that windows has some limitation of accessing multi end point device, and libusb needs to visit end point 0 directly which windows does not allow.
> 
> OK, I never managed to use AOA on Windows either. It is totally ok to implement it for Linux only (at least as a first step). Moreover, there is already another feature (V4L2) available only on Linux. I suggest you keep your commit `Add Windows build support for libusb` locally, it might be useful later, but drop it from this PR for now.
> 
> Also, I have a little request (I didn't review yet, but looked at commit titles): could you `rebase -i` to clean up some history. For example, `Move HID keyboard setup code into functions`: this code could have been written directly "into functions" I guess. Or `Make sdl_keymod_to_hid_modifiers() more readable` (make it more readable directly). And other spurious changes to make clean commits… Thank you :)

Will do it tomorrow, it's midnight 3:00 and I am gonna sleep after fighting against windows for whole night 😹

AlynxZhou:
Squashed into 2 commits.

I think it's better to keep the windows build commit here, since the code itself is correct -- it's not our problem but libusb's problem, once it works correctly (maybe hard though), our code will work.

Plus if I dropped build support for libusb on Windows, I need to add a lot of `#ifdef _WIN32` (or something similiar) to build release on Windows (since no libusb is here and `aoa_hid` `hid_keyboard` needs it when compiling so the only way will be disable them, too). So I think keep it will be better than drop it.

AlynxZhou:
😄 I have to rebase it again because I found that I mis-spell AOAv2 into AoAv2.

chldbwnstm:
Can we please simulate mouse click as well?

AlynxZhou:
I tried locally, but the position syncing is bad, and position is needed
for detect what has been clicked. And it should not be better then current
experience.

rom1v:
@AlynxZhou I don't know if you received my e-mail this week-end (probably went into your spams, like often with gmail recipients). (If you missed it, you didn't miss a lot, I said I had no time yet.)

Anyway, here is some quick feedback.


### Keyboard

I did a quick test on my devices, when I use HID (either in French or English locale) from my laptop (a French AZERTY keyboard), it injects keys as if my keyboard was QWERTY (so numbers, accented characters, some letters… are incorrect).

I'm not sure how to solve this problem if we inject scancodes. Any idea?


### Input mode selection

> > (Maybe a good idea is like usbaudio, if only one device with adb, use it, others ask for serial?)
>
> I'd really prefer to be explicit. Otherwise, people will never know immediately which method is used and why/why not.

After reflection, I confirm that I would really prefer the mode to be explicit. It greatly impacts how the injection works, so it should not just be the consequence of some heuristic/fallback.

> > If it's explicitly selected, `scrcpy` should fail IMO, not fallback.
>
> Then what should be the default option?

IMO, the default should be the Android text injection: it's the most basic, but works the same way everywhere:
 - Linux, Windows and macOS
 - over USB and wirelessly / TCP/IP

HID injection is better, but only over USB and only on Linux, and it might cause significant differences (hopefully for the better, but it can also impact things like the AZERTY/QWERTY issue I encounter).

So IMO it should be enabled explicitly with `scrcpy --input-method=hid`.


### Auto device selection

For now, as we discussed, HID can only be enabled if a serial is explicitly passed.

I'm not asking you to do anything about it, I'm just listing some ideas/solutions.

We could use `adb get-serialno` and match by serial (as already said earlier, like in [usbaudio](https://github.com/rom1v/usbaudio/blob/6f46c43b9a04249f5de61255b834dc7ebd56421c/src/aoa.c#L146)):

```console
$ adb get-serialno
05f5e60a0ae518e5
```

Or with `adb-devpath` to find the `vid:pid`:

```console
$ adb get-devpath
usb:1-4
$ cat /sys/bus/usb/devices/1-4/id{Vendor,Product}
18d1
4ee2
```

Filtering by `adb` interfaces could also help: https://github.com/rom1v/usbaudio/blob/6f46c43b9a04249f5de61255b834dc7ebd56421c/src/aoa.c#L85-L117

Please don't do anything about it, I must rebase my old work to read the output of a process: https://github.com/Genymobile/scrcpy/issues/279#issuecomment-917369145

AlynxZhou:
> @AlynxZhou I don't know if you received my e-mail this week-end (probably went into your spams, like often with gmail recipients). (If you missed it, you didn't miss a lot, I said I had no time yet.)
> 

Gmail puts your email into junk mail directly, sad.

> Anyway, here is some quick feedback.
> ### Keyboard
> 
> I did a quick test on my devices, when I use HID (either in French or English locale) from my laptop (a French AZERTY keyboard), it injects keys as if my keyboard was QWERTY (so numbers, accented characters, some letters… are incorrect).
> 

Hmm, I only use QWERTY keyboard so I did not test this, however, I have some thoughts about it. I suggest you run it in verbose mode, and see if the key in logs is the same as what you typed (I think it should be the same).

Then I suggest you to try to change keyboard layout in your Android IME, for example I am using Gboard, I can add QWERTY and AZERTY in Alphabet, maybe this is the problem, you have to set it manually because HID has no info about keyboard layout.

Screenshot from 2021-09-21 01-34-55

I personally only have on tweak of my keyboard that I swapped Control and Caps Lock with udev rules, and it works fine on Android (because this has been handled before SDL). I think the problem is that by default Android IME uses QWERTY as keyboard layout. But I cannot test it, I need your help.

> I'm not sure how to solve this problem if we inject scancodes. Any idea?

I think this is not problem of those codes, they just forward scancodes...

> ### Input mode selection
> 
> > > (Maybe a good idea is like usbaudio, if only one device with adb, use it, others ask for serial?)
> > 
> > 
> > I'd really prefer to be explicit. Otherwise, people will never know immediately which method is used and why/why not.
> 
> After reflection, I confirm that I would really prefer the mode to be explicit. It greatly impacts how the injection works, so it should not just be the consequence of some heuristic/fallback.
> 
> > > If it's explicitly selected, `scrcpy` should fail IMO, not fallback.
> > 
> > 
> > Then what should be the default option?
> 
> IMO, the default should be the Android text injection: it's the most basic, but works the same way everywhere:
> 
>     * Linux, Windows and macOS
> 
>     * over USB and wirelessly / TCP/IP
> 
> 
> HID injection is better, but only over USB and only on Linux, and it might cause significant differences (hopefully for the better, but it can also impact things like the AZERTY/QWERTY issue I encounter).
> 
> So IMO it should be enabled explicitly with `scrcpy --input-method=hid`.

I prefer the current logic, but if you think the old way is better for default, it's OK, should be easily changed in one commit I think.

> ### Auto device selection
> 
> For now, as we discussed, HID can only be enabled if a serial is explicitly passed.
> 
> I'm not asking you to do anything about it, I'm just listing some ideas/solutions.
> 
> We could use `adb get-serialno` and match by serial (as already said earlier, like in [usbaudio](https://github.com/rom1v/usbaudio/blob/6f46c43b9a04249f5de61255b834dc7ebd56421c/src/aoa.c#L146)):
> 
> ```
> $ adb get-serialno
> 05f5e60a0ae518e5
> ```
> 
> Or with `adb-devpath` to find the `vid:pid`:
> 
> ```
> $ adb get-devpath
> usb:1-4
> $ cat /sys/bus/usb/devices/1-4/id{Vendor,Product}
> 18d1
> 4ee2
> ```
> 
> Filtering by `adb` interfaces could also help: https://github.com/rom1v/usbaudio/blob/6f46c43b9a04249f5de61255b834dc7ebd56421c/src/aoa.c#L85-L117
> 
> Please don't do anything about it, I must rebase my old work to read the output of a process: [#279 (comment)](https://github.com/Genymobile/scrcpy/issues/279#issuecomment-917369145)

I did nothing about auto selection, the only thing that might be related is that I added a `adb_getserialno` function (which cannot return serial because we cannot read output of process), and I never use it, I forget to remove it because it's harmless. It could be removed with a single commit.

It's mid night 2:00 and my brain is not so clear, my words might be confusing, please add comment if you have any questions about my reply, I'll reply after waking up 😄.

AlynxZhou:
Actually I am not sure what a AZERTY keyboard is, when you press A on your keyboard, does it send scancode for A? Or send scancode Q (because the key is Q on QWERTY) and your system convert it into keycode A? If it sends scancode A, all things should works fine, if it sends scancode Q, Android is responsible for convert it into keycode A like what your system do. But anyway this should have nothing related to our code, it just forward scancode.

Please have a test for setting your IME to AZERTY and see if it works fine.

rom1v:
> when you press A on your keyboard, does it send scancode for A? 

It generates `SDL_SCANCODE_Q` for physical key code (`event->keysym.scancode`) and `SDLK_a` for virtual key code (`event->keysym.sym`).

Changing the device keyboard layout seems to have no impact.

rom1v:
> Maybe the layout is part of some USB descriptor?

https://www.usb.org/document-library/device-class-definition-hid-111

`bCountryCode` p22-23

AlynxZhou:
> > Maybe the layout is part of some USB descriptor?
> 
> https://www.usb.org/document-library/device-class-definition-hid-111
> 
> `bCountryCode` p22-23

However AOA does not use HID descriptor...

AlynxZhou:
> > when you press A on your keyboard, does it send scancode for A?
> 
> It generates `SDL_SCANCODE_Q` for physical key code (`event->keysym.scancode`) and `SDLK_a` for virtual key code (`event->keysym.sym`).
> 

Hmm, if you got SDL_SCANCODE_Q, I think send Q to Android should be the correct behavior, because your keyboard is sending Q to your system.

> Changing the device keyboard layout seems to have no impact.
> 

What's your Android IME? Could you please tell me what you did to change it? Thanks.

> Maybe the layout is part of some USB descriptor? (I didn't read the specs)

For AOA we only have Report Descriptor in documents.

I guess your keyboard is a normal QWERTY keyboard and just swap keycaps? The layout is converted by system.

Another way is that you use an OTG cable to connect your keyboard directly... my code should works the same as connecting keyboard to Android physically.

AlynxZhou:
OK, I find where to change the physical keyboard in Android, seems not the IME, for my Galaxy S9+ is Settings -> General management -> Language and input -> Physical Keyboard -> Gboard - Multilingual typing (maybe change other IME if you don't use Gboard).

After choose "French, Azerty style", when I press Q on my QWERTY keyboard, A is shown on screen. You may check this.

Those settings only appear when Android detected a physical keyboard (OTG or AOA).

rom1v:
Indeed :+1: (I don't use GBoard, but OpenBoard, but it seems independent anyway)

key

AlynxZhou:
Should be in document, I am not sure whether Android can change layout automatically when you plugged in your keyboard via OTG, but seems we need to set it manually for once in AOA.

AlynxZhou:
See new commits about remove adb_get_serialno and use legacy way as default input mode, if you still have any questions, please add your comment.

app/src/hid_keyboard.c
```diff
@@ -0,0 +1,347 @@
...
+    0x95, 0x05,
+    // Output (Data, Variable, Absolute): LED report
+    0x91, 0x02,
+
+    // Report Size (3)
+    0x75, 0x03,
+    // Report Count (1)
+    0x95, 0x01,
+    // Output (Constant): LED report padding
+    0x91, 0x01,
```
rom1v, line 81: Are these 18 bytes about LEDs necessary in the case of scrcpy?

They appear in [hid1_11.pdf](https://www.usb.org/sites/default/files/hid1_11.pdf) n Appendinx B §`B.1 Protocol 1 (Keyboard)`, but it's not clear to me if it is just an example or if it must contain these items.

The Appendix B precises:

> Devices may append additional data to these boot reports, but the first 8 bytes of keyboard reports and the first 3 bytes of mouse reports must conform to the format defined by the Boot Report descriptor in order for the data to be correctly interpreted by the BIOS.

So I think we could remove these bytes safely. At least it works on my device.

Note: On the other hand, if we start mirroring with <kbd>CAPS LOCK</kbd> (or <kbd>NUM LOCK</kbd>) on, then the state on the device is reversed. Maybe we could synchronize states? However, it could be a bit tricky, since several scrcpy windows mirroring different devices might not be in the same state, so it would require to synchronize on window focus, etc. I'm not sure it is worth it. (I'm just thinking out loud.)

app/src/hid_keyboard.c
```diff
@@ -0,0 +1,347 @@
...
+    0x05, 0x01,
+    // Usage (Keyboard)
+    0x09, 0x06,
+
+    // Collection (Application)
+    0xA1, 0x01,
+    // Report ID (1)
+    0x85, 0x01,
+
+    // Usage Page (Keyboard)
```
rom1v, line 39:
```
Usage Page (Key Codes)
```
(I think)

app/src/hid_keyboard.h
```diff
@@ -0,0 +1,89 @@
...
+#define HID_KEYBOARD_EVENT_SIZE (3 + HID_KEYBOARD_MAX_KEYS)
+#define HID_KEYBOARD_REPORT_ID 0x01
+#define HID_KEYBOARD_RESERVED 0x00
+#define HID_KEYBOARD_ERROR_ROLL_OVER 0x01
+#define HID_MEDIA_EVENT_SIZE 2
+#define HID_MEDIA_REPORT_ID 0x02
+#define HID_MEDIA_KEY_INDEX (HID_REPORT_ID_INDEX + 1)
+/**
+ * Media keys handle as mask so we define them here.
+ * Currently not used because desktop environment catches them before window.
```
rom1v, line 40: I think that we should not forward media keys anyway (which are "global shortcuts"): if I want to change volume or mute my desktop, I don't want to also mute or change volume on the device as a consequence.

So it's a good thing that the desktop environment catches them. I think we could remove these defines.

app/src/hid_keyboard.c
```diff
@@ -0,0 +1,347 @@
...
+    // Logical Minimum (0)
+    0x15, 0x00,
+    // Logical Maximum (1)
+    0x25, 0x01,
+    // Report Size (1)
+    0x75, 0x01,
+    // Report Count (8)
+    0x95, 0x08,
+    // Input (Data, Array)
+    0x81, 0x02,
```
rom1v, line 140: The media keys should not be forwarded (see my other comment), so this could be removed too.

(IIUC, this also removes the need for handling `Report ID`s.)

app/src/hid_keyboard.c
```diff
@@ -0,0 +1,347 @@
...
+    0x95, 0x08,
+    // Input (Data, Variable, Absolute): Modifier byte
+    0x81, 0x02,
+
+    // Report Size (8)
+    0x75, 0x08,
+    // Report Count (1)
+    0x95, 0x01,
+    // Input (Constant): Reserved byte
+    0x81, 0x01,
```
rom1v, line 61: Is it necessary (it works without)?

rom1v (commented):
I added some comments inline during a first-pass review, after I read the USB/HID spec.

app/src/hid_keyboard.h
rom1v: Oh, unless they are generated from the scrcpy shortcuts (e.g. Alt + up could trigger the HID media volume up). :+1:

app/src/hid_keyboard.c
rom1v: Oh, unless they are generated from the scrcpy shortcuts (e.g. Alt + up could trigger the HID media volume up). :+1:

app/src/hid_keyboard.c
AlynxZhou: > Are these 18 bytes about LEDs necessary in the case of scrcpy?

I suggest to keep them because they are harmless, we never use them, but every keyboards have LEDs, we'd better behave more like a real keyboard.

Remove those LEDs should work, but maybe some systems require them, I am not sure. So I suggest to keep those harmless definations.

Nearly every examples on Internet have those lines, so if some one wants to modify those code, they can compare those with examples line by line if we keep them.

> Note: On the other hand, if we start mirroring with CAPS LOCK (or NUM LOCK) on, then the state on the device is reversed. Maybe we could synchronize states? However, it could be a bit tricky, since several scrcpy windows mirroring different devices might not be in the same state, so it would require to synchronize on window focus, etc. I'm not sure it is worth it. (I'm just thinking out loud.)

I think user is responsible for sync those state, if we do sync in code, a lot of hack is needed (those LEDs are control by system so you need permission to modify keyboard state), grabbing focus may be not allowed on some platform (Wayland I think), so the simplest way is let user handle it to start scrcpy with CAPS LOCK off.

app/src/hid_keyboard.c
AlynxZhou: Should be Key Codes.

app/src/hid_keyboard.c
AlynxZhou: Could be removed (except we want to implement shortcuts with them).

I kept them just in case.

(Removing the Report ID byte might change a lot code, so I suggest not to remove Report ID, it should be fine I think.)

app/src/hid_keyboard.c
AlynxZhou: Every examples on Internet have those reserved byte, so we'd better keep it.

app/src/hid_keyboard.c
AlynxZhou: Another way to sync Caps Locks may be that when we start scrcpy, we detect the state of those locks, and if they are on, we send HID events for them before handling user actions. 😄(However I am not sure to do it, maybe we can implement it future.)

AlynxZhou:
I've pushed new commits according to the review comments. please check them.

rom1v:
Hi,

I had some time today to work on this PR/feature, so here is a small follow up.

Firstly, I reviewed a bit in more details, the work is very good :+1: Thank you.

I took your branch and did some changes locally (I removed _report id_ as evoked [here](https://github.com/Genymobile/scrcpy/pull/2632#discussion_r716006053), I fixed minor issues and code style, etc.), I will share when it's ready (so please don't modify your PR, this would conflict).

Then, as a preparation step, similar to the refactors I did for integrating the V4L2 feature (#2268), I extracted keyboard and mouse processors to "traits" (interfaces), so that we can provide an alternative implementation without intrusive changes: [`input_traits`](https://github.com/Genymobile/scrcpy/commits/input_traits) (this is not merged into `dev` yet, I need more testing/time to review my own code).

I will rebase/adapt your PR onto this branch, to provide an HID key processor implementation.

I'll then retrieve the adb serial even if it is not provided explicitly.

---

Then, one complicated issue I need to deal with is the clipboard synchronization on <kbd>Ctrl</kbd>+<kbd>v</kbd> (that you noticed):
https://github.com/AlynxZhou/scrcpy/blob/2aa94b8e020cf5ed37894946e1a918f41488168b/app/src/input_manager.c#L597

When pressing <kbd>Ctrl</kbd>+<kbd>v</kbd>, currently, a control message with the current computer clipboard is sent to the device just before sending the control message for the <kbd>Ctrl</kbd>+<kbd>v</kbd> keys. As a consequence, when injecting the keys, the clipboard is already synchronized, so <kbd>Ctrl</kbd>+<kbd>v</kbd> pastes the correct content.

But this hack works only because the clipboard synchronization is guaranteed to be done before the <kbd>Ctrl</kbd>+<kbd>v</kbd> injection (the control messages are serialized), which is not the case if keys are sent via HID. One quick solution is to disable the feature if HID is enabled (what you did), but I'd like to keep it working.

---

I'll keep you updated when I make some progress (sorry, it's quite slow because I can work on it only on my free time).

AlynxZhou:
Just notify me if there is anything I need to do.

> But this hack works only because the clipboard synchronization is guaranteed to be done before the Ctrl+v injection (the control messages are serialized), which is not the case if keys are sent via HID. One quick solution is to disable the feature if HID is enabled (what you did), but I'd like to keep it working.

I guess this is because control message and HID event are in different threads?

rom1v:
Here is an update after a Sunday afternoon of work:

 - I finished to rebase/adapt your PR on the "input traits" refactor
 - I changed the option `--input-mode mode` to `-K/--keyboard-hid` (without argument), for simplicity. I kept `keyboard` in the name because in the future there might be `--mouse-hid`.
 - The default remains the input injection method. However, if `-K/--keyboard-hid` is requested, if it fails, then it fallbacks to the default input injection method (instead of failing). The reason is to be able to use the same command/shortcut when the device is connected via USB or via TCP/IP, without changing the arguments.
 - I rebased my function to execute a command with redirection, to read the `stdout` of a child process, both on Linux/POSIX and Windows.
 - I exposed `adb_get_serialno()`, and use it to retrieve the serial if it is not explicitly provided by the user (so that `scrcpy -K` is sufficient to use HID over AOA, providing a serial is not necessary).
 - I added ifdefs to compile AOA HID only on supported platforms (Linux). I don't know if it can work on macOS yet.

Here is my current WIP branch: [`hid.10`](https://github.com/Genymobile/scrcpy/commits/hid.10).

What I will do before merging:
 - find a solution for the <kbd>Ctrl</kbd>+<kbd>v</kbd> issue
 - adapt README and your commit message to the recent changes

> I guess this is because control message and HID event are in different threads?

Because currently there is no guarantee that the HID event is injected after the paste is executed on the device (using the same thread for control and HID events would not solve the problem, it would only guarantee that the control message is sent before the HID event, but not that the execution is complete).

rom1v:
For Windows: https://github.com/libusb/libusb/wiki/Windows#How_to_use_libusb_on_Windows

AlynxZhou:
> For Windows: https://github.com/libusb/libusb/wiki/Windows#How_to_use_libusb_on_Windows

I tried all of this page and still not work.

AlynxZhou:
> synchronize CAPS_LOCK and NUM_LOCK on start (?)

Should be easy, we can send a caps lock/num lock key event on start if they are on.

krishtoautomate:
Hi, will this otg keyboard work for ios as well?

rom1v:
No.

castillofrancodamian:
How do I change the keyboard language? I already tried changing the language of the physical keyboard on Android but it still doesn't work.

Channpyae:
> How do I change the keyboard language? I already tried changing the language of the physical keyboard on Android but it still doesn't work.

Hi, what phone do you use? Go to settings in phone. Type "physical" in search bar. Click on physical keyboard settings. You'll see keyboard shortcuts settings. Reply to me if not working.

castillofrancodamian:
@Channpyae It's a Nokia 2.3. I did exactly that and it's still with the English keyboard (I guess).

Channpyae:
> @Channpyae It's a Nokia 2.3. I did exactly that and it's still with the English keyboard (I guess).

I'm facing the same issue, now. I tried multiple methods like installing physical keyboard layouts on both my system and phone(Android 12, Samsung) but doesn't work. It seems scrcpy don't take unicode characters other than English; based on terminal outputs. Sorry, man. I can't help.

rom1v:
What do you test exactly? scrcpy in otg mode or the current PR about uhid?

(physical keyboard has no impact on the default injection mode)

castillofrancodamian:
@rom1v I use `scrcpy --otg --hid-keyboard --hid-mouse` .

PR #2686, opened 2021-10-06
Update Readme.pt-br.md

PedroMiguelCarraro:


PedroMiguelCarraro:
Updated with this fixes, now looks like its fine

PR #2783, opened 2021-11-14
Inherit only specific handles on Windows

### LINKED ISSUE #2779: V1.20 version can only open one device window when multiple devices are connected.

rom1v:
To be able to communicate with a child process via stdin, stdout and
stderr, the CreateProcess() parameter bInheritHandles must be set to
TRUE. But this causes *all* handles to be inherited, including sockets.

As a result, the server socket is inherited by the process running adb
to execute the server on the device, so it may not be closed properly,
causing other instances of scrcpy to fail.

To fix the issue, use an extended API to explicitly set the HANDLE to
inherit:
 - <https://stackoverflow.com/a/28185363/1987178>
 - <https://devblogs.microsoft.com/oldnewthing/20111216-00/?p=8873>

Fixes #2779

---

Replace these binaries in your v1.20 release:
 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/2779/5/scrcpy.exe) _sha256:33a57452cbd58c689a30497c28e005111553e2ab21f6ddfee270767c29b5cceb_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/2779/5/scrcpy-server) _sha256:b20aee4951f99b060c4a44000ba94de973f9604758ef62beb253b371aad3df34_

### DIFF
--- a/app/src/sys/win/process.c
+++ b/app/src/sys/win/process.c
@@ -1,4 +1,9 @@
+// <https://devblogs.microsoft.com/oldnewthing/20111216-00/?p=8873>
+#define _WIN32_WINNT 0x0600 // For extended process API
+
 #include "util/process.h"
+
+#include <processthreadsapi.h>
 
 #include <assert.h>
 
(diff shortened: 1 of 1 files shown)


rom1v:
Hmm, that fixes #2779, but then calling `adb_get_serialno()` (which we never do on Windows) does not work as expected. Will investigate before merging.

rom1v:
Merged into `dev`: 9cb14b51663d61c6f0b41da3c5699de418d633aa

PR #2790, opened 2021-11-15
Fix wrong package name for libusb dev in BUILD.md

AlexBurdu:
Fix wrong package to install for Ubuntu/Debian

Without it I was getting:
```
Run-time dependency libusb-1.0 found: NO (tried pkgconfig and cmake)

app/meson.build:88:8: ERROR: Dependency "libusb-1.0" not found, tried pkgconfig and cmake
```


rom1v:
Thank you, merged: 02ae0db6cd14a06899e6f8e92bfe107210ec51a3

Issue #2801, opened 2021-11-17
Ability to specify adb and scrcpy server Host/Port's


clienthax:
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
I require a way to specify connection host and port for both adb and the scrcpy connection (assuming forward connection).
I have multiple android devices connected to one parent host, and each usb is isolated inside a docker container (similar to https://github.com/sorccu/docker-adb)
This results in both the adb port and the scrcpy server port being exposed to different ports on the host system.

**Describe the solution you'd like**

- Options for ADB Host/Port
- Options for scrcpy server Host/Port

eg `scrcpy --adbHost 192.168.1.50 --adbPort 9999 --scrcpyHost 192.168.1.50 --scrcpyPort 10000 --force-adb-forward`

**Describe alternatives you've considered**

- Using soccat to forward the ports locally from the remote and using --force-adb-forward, however this is fairly clunky.
- Using the ADB env variable to specify the host and port via "ADB=adb -H host -P port", however this doesn't allow for scrcpy to function after installing the server.

**Additional context**
The client is not located on the same system as the devices, ssh forwarding is also not a good option due to these being isolated in their own containers.

rom1v:
> eg `scrcpy --adbHost 192.168.1.50 --adbPort 9999 --scrcpyHost 192.168.1.50 --scrcpyPort 10000 --force-adb-forward`

Isn't the host necessary the same for adb and scrcpy?

I would say yes, because the `adb forward`/`adb reverse` command executed by scrcpy opens a tunnel on the adb host, so the scrcpy host is necessarily the same, isn't it?

`--scrcpyPort` is basically the existing `--port` option IIUC.

> Using the ADB env variable to specify the host and port via "ADB=adb -H host -P port"

Or even `ADB_SERVER_SOCKET`: https://github.com/Genymobile/scrcpy/issues/144#issuecomment-387765711

The remaining problem is the hardcoded [localhost](https://github.com/Genymobile/scrcpy/blob/02ae0db6cd14a06899e6f8e92bfe107210ec51a3/app/src/server.c#L362) here I guess.

clienthax:
[...]

PR #2807, opened 2021-11-18
Add remote host/port options for use when forwarding.

RipleyTom:
Add 2 new options:
--tunnel-host to set the final remote IP address when using --force-adb-forward.
--tunnel-port to set the final remote port  when using --force-adb-forward.

Note that those are different from the forwarding port given through --port(as adb will still forward on that port).
I tried to keep the changes relatively minimal(ie if you don't use the options it will behave as it used to).

I also added a .clang-format file as every C/C++ project should have one.
Another notable change is I forced _WIN32_WINNT and WINVER to 0x0600(from mingw default of 0x0502) through meson to have access to a few more posix inet functions for windows.

Implement https://github.com/Genymobile/scrcpy/issues/2801.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,4 @@
+UseTab: Never
+TabWidth: 4
+IndentWidth: 4
+AlwaysBreakAfterReturnType: AllDefinitions
--- a/app/meson.build
+++ b/app/meson.build
@@ -137,6 +137,12 @@ foreach f : check_functions
   endif
 endforeach
 
+if host_machine.system() == 'windows' or get_option('crossbuild_windows')
+    # Set the windows version higher than mingw default, needed for inet_pton on windows
+    add_global_arguments('-D_WIN32_WINNT=0x0600', language : 'c')
+	add_global_arguments('-DWINVER=0x0600', language : 'c')
+endif
+
 # the version, updated on release
 conf.set_quoted('SCRCPY_VERSION', meson.project_version())
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -13,6 +13,14 @@
 #include "util/strbuf.h"
 #include "util/term.h"
 
+#ifdef __WINDOWS__
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#else
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#endif
+
 #define STR_IMPL_(x) #x
 #define STR(x) STR_IMPL_(x)
 
@@ -46,6 +54,8 @@
 #define OPT_V4L2_SINK              1027
 #define OPT_DISPLAY_BUFFER         1028
 #define OPT_V4L2_BUFFER            1029
+#define OPT_TUNNEL_HOST            1030
+#define OPT_TUNNEL_PORT            1031
 
 struct sc_option {
     char shortopt;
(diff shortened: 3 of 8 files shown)


rom1v:
Thank you very much. Works as expected :+1:

I applied the following changes:
 - extract util method to parse an IPv4, so that `cli.c` do not have to handle that (and split it into a separate commit)
 - accept --tunnel-port=0, because we should be able to pass the default value ("do not force a port")
 - replace remaining `scrcpy_ip` and `scrcpy_port` to `tunnel_host` and `tunnel_port` in the code
 - move `_WIN32_WINNT` and `WINVER` definitions from meson to source code (see below)
 - document in the manpage (same content as in the --help)
 - document in the README with an example (separate commit)
 - automatically enable `--force-adb-forward` if tunnel host or port is provided (separate commit)
 - fix minor code style issues

The resulting branch is [pr2807](https://github.com/Genymobile/scrcpy/commits/pr2807). Please review and test :)

Could you please provide a real e-mail address (your commit author is currently `RipleyTom <RipleyTom@users.noreply.github.com>`)?

> Another notable change is I forced _WIN32_WINNT and WINVER to 0x0600(from mingw default of 0x0502) through meson to have access to a few more posix inet functions for windows.

I did a similar thing recently, but I [defined](https://github.com/Genymobile/scrcpy/blob/2d6a96776c9e4afeba74e98f70dd98d29e226b0d/app/src/sys/win/process.c#L2) the variable directly in the source code, not via meson. It's more local and does not complexify the build system.

> I also added a .clang-format file as every C/C++ project should have one.

This is off-topic. And it would require more work (configure more options in the `.clang-format` file and make all the source code match the rules).

RipleyTom:
I'm using that email specifically for privacy(see https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address).

_WIN32_WINNT and WINVER should be defined at the project level, not in some files and no in others. Windows API version used should be consistent among the whole project.

I don't really care about the .clang-format but it was a terrible experience trying to contribute to the project without one(I just set enough options for it to format the functions correctly).

rom1v:
> I'm using that email specifically for privacy(see https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address).

OK.

> _WIN32_WINNT and WINVER should be defined at the project level, not in some files and not in others. Windows API version used should be consistent among the whole project.

Here is what the [GNU documentation](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html) says:

> You should define these macros by using '#define’ preprocessor directives at the top of your source code files. These directives must come before any #include of a system header file. It is best to make them the very first thing in the file, preceded only by comments. You could also use the ‘-D’ option to GCC, but it’s better if you make the source files indicate their own meaning in a self-contained way.

Is it different for Windows?

> I don't really care about the .clang-format but it was a terrible experience trying to contribute to the project without one(I just set enough options for it to format the functions correctly).

What's your editor?

RipleyTom:
> > _WIN32_WINNT and WINVER should be defined at the project level, not in some files and not in others. Windows API version used should be consistent among the whole project.
> 
> Here is what the [GNU documentation](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html) says:
> 
> > You should define these macros by using '#define’ preprocessor directives at the top of your source code files. These directives must come before any #include of a system header file. It is best to make them the very first thing in the file, preceded only by comments. You could also use the ‘-D’ option to GCC, but it’s better if you make the source files indicate their own meaning in a self-contained way.
> 
> Is it different for Windows?
The goal of _WIN32_WINNT/WINVER is to indicate a minimum API version(and thus windows version) necessary for the program to run, win32 api is backward compatible(sometimes to a fault) so there is no drawback to setting the minimum project wide(unlike the cited page which has an explanation about incompatible standards).
This project is quite small so it's fine if you want to use #define macros each time you want to use a more advanced version of the API but on a bigger project it'd lead to issues with redefinitions if you use different values in headers or the like, just not a good idea.

> 
> > I don't really care about the .clang-format but it was a terrible experience trying to contribute to the project without one(I just set enough options for it to format the functions correctly).
> 
> What's your editor?
You know what's the point of .clang-format? Not having to worry about which editor people are using ;).

rom1v:
Hmm, in fact, the way these feature test macros are currently defined is not very good:
https://github.com/Genymobile/scrcpy/blob/57fb08e443272a9d346b036fb421cdf0ccff3074/app/src/compat.h#L4-L9

In VLC, in fact we define them in `config.h` (via autootols). But I submitted #2812 with `add_global_arguments()`. Please review :wink:

rom1v:
Merged into `dev`.

 - 52e5181c841580b582e44dbdbdf08df533637d07
 - 7bdbde7363836b79864f876f6a4cc107785a7147
 - 68eaee5bb086383e601ac64e9ff0aa65830599e3
 - 44721ed98235b262fc77d686ec5020c72a8dcb42

PR #1960, opened 2020-12-09
fix build error for macOS.

mqiezi:
Fix error below when build for macOS version.
---------------
```bash
~/git/GitHub/scrcpy(master ✗) ninja -Cx
ninja: Entering directory `x'
[0/3] Generating scrcpy-server with a custom command

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.3/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 2s
25 actionable tasks: 1 executed, 24 up-to-date
[1/3] Compiling C object app/scrcpy.p/src_sys_unix_command.c.o
FAILED: app/scrcpy.p/src_sys_unix_command.c.o
cc -Iapp/scrcpy.p -Iapp -I../app -I../app/src -I/usr/local/Cellar/ffmpeg/4.3.1_4/include -I/usr/local/include/SDL2 -I/usr/local/opt/llvm/include -I/usr/local/opt/openssl/include -I/usr/local/opt/flex/include -flto -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -Wextra -std=c11 -O3 -D_THREAD_SAFE -MD -MQ app/scrcpy.p/src_sys_unix_command.c.o -MF app/scrcpy.p/src_sys_unix_command.c.o.d -o app/scrcpy.p/src_sys_unix_command.c.o -c ../app/src/sys/unix/command.c
In file included from ../app/src/sys/unix/command.c:24:
In file included from ../app/src/util/log.h:4:
In file included from /usr/local/include/SDL2/SDL_log.h:40:
/usr/local/include/SDL2/SDL_stdinc.h:426:5: error: implicit declaration of function 'memset_pattern4' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    memset_pattern4(dst, &val, dwords * 4);
    ^
../app/src/sys/unix/command.c:31:12: error: implicit declaration of function 'strdup' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    path = strdup(path);
           ^
../app/src/sys/unix/command.c:31:12: note: did you mean 'strcmp'?
/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/string.h:77:6: note: 'strcmp' declared here
int      strcmp(const char *__s1, const char *__s2);
         ^
[...]

PR #2812, opened 2021-11-20
Configure feature test macros in meson

rom1v:
Use ~~add_global_arguments() like in PR #2807~~ `add_project_arguments()`.

However, the [documentation](https://mesonbuild.com/Adding-arguments.html) says:

> Compilation tests that are run as part of your project configuration do not use these flags. The reason for that is that you may need to run a test compile with and without a given flag to determine your build setup. For this reason tests do not use these global arguments.

We could like these definitions to be set in tests too. But in practice, they are correctly set on my machine :thinking: 

An alternative could be to add the flags (still from Meson) to the `config.h` generated file. I don't know what is better. Any idea?

@mqiezi Could you confirm that this does not cause a regression for #1960?

cc @RipleyTom

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -44,11 +44,19 @@ if host_machine.system() == 'windows'
         'src/sys/win/file.c',
         'src/sys/win/process.c',
     ]
+    add_project_arguments('-D_WIN32_WINNT=0x0600', language: 'c')
+    add_project_arguments('-DWINVER=0x0600', language: 'c')
 else
     src += [
         'src/sys/unix/file.c',
         'src/sys/unix/process.c',
     ]
+    add_project_arguments('-D_POSIX_C_SOURCE=200809L', language: 'c')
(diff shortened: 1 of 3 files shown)


RipleyTom:
Yes add_project_arguments seems to be the appropriate one, sorry I never edited a meson build before.
Shouldn't the _WIN32_WINNT and WINVER also be defined when crosscompiling?

rom1v:
> Shouldn't the _WIN32_WINNT and WINVER also be defined when crosscompiling?

`host_machine == "windows"` when [cross-compiling](https://mesonbuild.com/Cross-compilation.html):

> The tl/dr summary is the following: if you are doing regular cross compilation, you only care about `build_machine` and `host_machine`. Just ignore `target_machine` altogether and you will be correct 99% of the time.

rom1v:
Superseded by #2813.

PR #2813, opened 2021-11-20
Configure feature test macros in meson (via config.h)

rom1v:
This is an alternative to #2812.

cc @RipleyTom


rom1v:
Merged into `dev`.

PR #2870, opened 2021-12-11
Fix adb server hang

yan12125:
Since commit 04267085441d6fcd05eff7df0118708f7622e237, the server is run
in a dedicated thread. For SDL, many signals, including SIGINT and
SIGTERM, are masked for new threads. As a result, if the adb server is
not already running, adb commands invoked by scrcpy will start an adb
server that ignores those signals and cannot be terminated at system
shutdown.

### DIFF
--- a/app/src/sys/unix/process.c
+++ b/app/src/sys/unix/process.c
@@ -119,6 +119,13 @@ sc_process_execute_p(const char *const argv[], sc_pid *pid, unsigned flags,
 
         close(internal[0]);
         enum sc_process_result err;
+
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you :+1:

It seems it also fixes #2873 (that I discovered by investigating this issue/PR, but I thought it was a different problem).

rom1v:
Rebase onto `dev` and merged: d0496719080ccf9ee00fccf6cb0765e31765529f

yan12125:
Thanks! Glad that this change helps in more cases :)

PR #2887, opened 2021-12-17
Request: To be able to resize windows in borderless with the mouse by grabbing one corner (keeping aspect ratio) Dev

manolito62:
To be able to resize windows in borderless with the mouse by grabbing one corner (keeping aspect ratio)


rom1v:
This is not a pull request.

PR #2945, opened 2022-01-14
Reference USB driver for Google devices for Windows.

johnjohndoe:


rom1v:
This link is provided in the first paragraph of the "Install OEM USB drivers":

> To connect and debug with any of the Google Nexus devices using Windows, you need to install the [Google USB driver](https://developer.android.com/studio/run/win-usb).

But ok, we can probably add it anyway.

rom1v:
Merged 479abc8c7723d2d05b7c5027528acd91946a328c

PR #2962, opened 2022-01-23
Updated README.sp to 1.21

Secreto31126:
Some anchors still require to be updated to match with the translated titles, I will change them later.

Btw, I thought Github had an option to mark a pull request as a draft, but I couldn't find it. It would have been useful here :(


rom1v:
Gracias :+1:
Merged: 86158130051d450a449a2e7bb20b0fcef1b62e80

PR #2970, opened 2022-01-26
Add a shorcut "open a terminal here" on Windows

rom1v:
On Windows, the file explorer does not provide any "open a terminal here" shortcut.

Add a bat file which just runs `cmd` to easily open a terminal directly in the scrcpy directory (so there is no need to `cd C:\path\
…`).


AndroidDeveloperLB:
I don't understand how it's related to ScrCpy. What was added?

rom1v:
> I don't understand how it's related to ScrCpy.

It's for convenience to get easily a terminal in the correct directory on Windows, to execute `adb` or `scrcpy` with arguments.

> What was added?

If you look at the change, you can see a new file `open_a_terminal_here.bat`.

AndroidDeveloperLB:
@rom1v Can you please demonstrate a scenario of when it's useful, and how to use it?
Suppose I open ScrCpy. Now I mirror the device on the PC. Now what?

rom1v:
> Can you please demonstrate a scenario of when it's useful, and how to use it?

Double click on `open_a_terminal_here.bat`, and in the terminal write for example:
```
scrcpy --record file.mp4
```

Now your session is recorded.

> Suppose I open ScrCpy. Now I mirror the device on the PC.

The terminal is used to open scrcpy with arguments. Don't open scrcpy before the terminal.

RiggiG:
For future reference, holding Shift adds "open X here" to the Windows Explorer context menu for Terminal, PowerShell, etc.

AndroidDeveloperLB:
@RiggiG OK but how is it related to ScrCpy? And why is it essential?
BTW, on Windows 11, when I right click (without Shift), I can already see "Open in Windows terminal". Maybe I've installed something for that though

rom1v:
@RiggiG

> For future reference, holding Shift adds "open X here" to the Windows Explorer context menu for Terminal, PowerShell, etc.

Oh, thank you, I didn't know that.

Deishelon:
@rom1v another fast way to open terminal in current dir, type "cmd" or "powershell" in windows explorer path bar.

PR #2971, opened 2022-01-26
Add server options to send raw H.264 without using the client

rom1v:
This is useful to use the `scrcpy-server` alone (without the client) just to get the raw H.264 stream:

```bash
adb push scrcpy-server /data/local/tmp/scrcpy-server.jar
adb forward tcp:1234 localabstract:scrcpy
adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar \
    app_process / com.genymobile.scrcpy.Server 1.21 \
    raw_video_stream=true tunnel_forward=true control=false
```
    
As soon as a client connects via TCP to localhost:1234, it will receive the raw H.264 stream.

For example, it can be played with VLC (but with a lot of delay due to buffering):

```bash
vlc --demux h264 tcp://localhost:1234
```

Refs https://github.com/Genymobile/scrcpy/pull/1419#issuecomment-1013964650

---

Here is a binary:

 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/2971/1/scrcpy-server) _sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855_

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/Server.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Server.java
@@ -253,6 +253,13 @@ public final class Server {
                     boolean sendDummyByte = Boolean.parseBoolean(value);
                     options.setSendDummyByte(sendDummyByte);
                     break;
+                case "raw_video_stream":
+                    boolean rawVideoStream = Boolean.parseBoolean(value);
+                    if (rawVideoStream) {
+                        options.setSendDeviceMeta(false);
+                        options.setSendFrameMeta(false);
+                        options.setSendDummyByte(false);
+                    }
                 default:
                     Ln.w("Unknown server option: " + key);
                     break;


rom1v:
GitHub has lost the commits in the PR after a rebase. Here they are:

 - 6b21f4ae13e0644632d2bef4f14ff590b1fa5e18
 - 3ba32c2a0d57b708980df121caad84252f720699
 - 45a5e560dfeabed8b32482caad857424ff472518
 - 2faf9715bebe720ed75ce98f5f8c2b2b1532880a

elf-world:
Can windows support this?

rom1v:
Yes, it's platform-independent.

elf-world:
> 

But VLC didn't respond when I used it on windows.

rom1v:
The H.264 has no fixed framerate: a new frame is produced whenever the device screen surface "changes".

To play it with VLC, see https://code.videolan.org/rom1v/vlc/-/merge_requests/20

elf-world:
C:\>adb -s 192.168.31.18:1472 shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 1.22 raw_video_stream=true tunnel_forward=true control=false encoder=OMX.google.h264.encoder
[server] WARN: Unknown server option: encoder
[server] INFO: Device: vivo V2055A (Android 12)
[server] DEBUG: Using encoder: 'OMX.qcom.video.encoder.avc'
[server] DEBUG: Screen streaming stopped
能否修改解码器？

Abunwer:
@rom1v thank u for this, but I'm trying to get the server.jar file, I think I have to update the gradle file to build a jar file, I tried but I couldn't find the right way to do it? I'm new to gradle, do you have any idea how I can get the server.jar file by building the source code?

nnnpa31:
Does it apply to the latest version 1.25?
I pull the latest release server file and execute the following commands in sequence.
```shell
adb push . \scrcpy-server /data/local/tmp/scrcpy-server
adb forward tcp:1234 localabstract:scrcpy
adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 1.25 raw_video_stream=true tunnel_forward =true control=false
```
But it returns Aborted and exits.

nnnpa31:
> You put an extra space between `.` and `\`.

Thanks, this is really a low mistake due to my oversight. That's why I deleted my reply. Thank you very much for your reply.

JJLo1:
after running the last command on adb it returns "Aborted" after about 2 seconds
im running the command in 1 line: 
`adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process com.genymobile.scrcpy.Server 1.22 raw_video_stream=true tunnel_forward=true control=false`

my scrcpy-server is the one from the current scrcpy-win64-v2.1.zip
any idea whats the problem?

rom1v:
Read https://github.com/Genymobile/scrcpy/blob/master/doc/develop.md#protocol

JJLo1:
ah, ok i broke it by accidently deleting the forward slash between app_process and com.genymobile...
ty for quick answer though

ouinouin:
@rom1v  , I tried the commandlines , on my side i couldnt get the adb commandline to work, this gives me : `aborted` as a result (the phone is able to stream the camera otherwise.. the concept of scrcpy streaming directly some h264 is promising and opens a lot of possibilities of recycling old phones as cameras over tcpip (looking at you frigate and go2rtc), but for this to happen, we would need something easy to just enable the stream while the phone is starting. would it be easy to have a dead simple user interface on the apk/jar scrcpy-server sideloaded ? (unsure if the scrcpy is a jar file or an APK..).

PR #2916, opened 2022-01-01
Add HID mouse support

rom1v:
Similar to the `--hid-keyboard` (or `-K`) added in scrcpy v1.20 (#2632), add a new option `--hid-mouse` (or `-M`).

Basically, you can run `scrcpy` with both HID keyboard and mouse with:

```bash
scrcpy -KM
```

The mouse is "captured": the mouse pointer disappears from the computer and appears on the Android device.

Special captures keys, either <kbd>Alt</kbd> or <kbd>Super</kbd>, toggle (disable or enable) the mouse capture. Use one of them to give the control of the mouse back to the computer.

These capture keys do not conflict with shortcuts, since a shortcut is always a combination of the <kbd>MOD</kbd> key and some other key, while the capture key triggers an action only if it is pressed and released alone.

Like HID keyboard, it works only when the device is connected via USB, and is only supported on Linux.

Tests and feedback welcome.

cc @AlynxZhou (author of #2632)
cc @JonasLoeffelholz (author of #2653)
cc @chldbwnstm (https://github.com/Genymobile/scrcpy/pull/2632#issuecomment-951942015)

---

This PR is based on previous refactors I already pushed to `dev` branch to support the feature properly.

When run with `-KM --forward-all-clicks`, on Firefox/Android, mouse button 4 works as expected (previous page), but mouse button 5 does not (I'd expect to go to the next page). I don't know if I did something wrong or if it is not expected to go to the next page.

PR #2974, opened 2022-01-26
Add OTG mode (HID only)

rom1v:
Add an option `--otg` to run scrcpy with only physical keyboard (#2632) and mouse (#2916) simulation, without mirroring and without requiring adb (USB debugging), as if the computer keyboard and mouse were plugged directly to the device via an OTG cable.

```bash
scrcpy --otg
# Pass the serial if several USB devices are available
scrcpy --otg -s 0123456789abcdef
```

This opens a window (with just the scrcpy logo, there is no mirroring), which when focused forwards the keyboard and mouse to the device:

otg

The mouse capture works the same way as in "normal" mode (as explained in #2916):
 - when the mouse is "captured", the mouse pointer disappears from the computer and appears on the Android device
 - a click in the window captures the mouse
 - special captures keys, either <kbd>Alt</kbd> or <kbd>Super</kbd>, toggle (disable or enable) the mouse capture (use one of them to give the control of the mouse back to the computer)

Like `--hid-keyboard` and `--hid-mouse`, `--otg` only works if the device is connected by USB, and is currently only supported on Linux.


lucas-bortoli:
This is awesome! Nice work!

BachoSeven:
@rom1v Would it be possible to also add mirroring to this OTG mode which doesn't require USB debugging? This would be especially useful for broken screens :'D 

thanks in advance

rom1v:
@BachoSeven This is not possible without installing an app on the device and authorizing capture, so that would not work with a broken screen.

vielavlrs:
I tried installing scrcpy v1.22 on Ubuntu 20.04 but the terminal shows my scrcpy is still v1.12? How do I run v1.22 in Ubuntu 20.04?

rom1v:
> I tried installing scrcpy v1.22 on Ubuntu 20.04 but the terminal shows my scrcpy is still v1.12? How do I run v1.22 in Ubuntu 20.04?

Uninstall the one from the distribution (`apt remove scrcpy`), and [follow these steps](https://github.com/Genymobile/scrcpy/blob/master/BUILD.md#simple) to install the latest release.

PR #3005, opened 2022-02-06
Improve device selection

rom1v:
Currently, if several devices are connected, `scrcpy` fails with:

```console
$ scrcpy
scrcpy 1.22 <https://github.com/Genymobile/scrcpy>
error: more than one device/emulator
ERROR: "adb get-serialno" returned with value 1
ERROR: Could not get device serial
```

This is not very user-friendly: the user must call `adb devices` manually to get the serial, then call `scrcpy -s xxxxxxxxxx` explicitly.

This PR brings features to improve device selection.

If several devices are connected, scrcpy prints the list of devices:

```console
$ scrcpy
…
ERROR: Multiple (3) ADB devices:
ERROR:     -->   (usb)  fedcba09                        device  GM1913
ERROR:     -->   (usb)  01234567890abcdef         unauthorized  
ERROR:     --> (tcpip)  192.168.1.1:5555                device  Nexus_5
ERROR: Select a device via -s (--serial), -d (--select-usb) or -e (--select-tcpip)
```

So it is possible to immediately call scrcpy again with `-s`.

In addition, two new options allow to select a USB or TCP/IP device when there is only one connected:
 - `-d` (or `--select-usb`) uses a USB device (like `adb -d`)
 - `-e` (or `--select-tcpip`) uses a TCP/IP device (like `adb -e`)

For example, to run `scrcpy` with the single TCP/IP device:

```console
$ scrcpy -e
…
DEBUG: ADB device found:
DEBUG:           (usb)  fedcba09                        device  GM1913
DEBUG:           (usb)  05f5e60a0ae518e5          unauthorized  
DEBUG:     --> (tcpip)  192.168.1.1:5555                device  Nexus_5
```

To run `scrcpy` with a single USB device:

```console
$ scrcpy -d
…
ERROR: Multiple (2) ADB devices over USB:
ERROR:     -->   (usb)  fedcba09                        device  GM1913
ERROR:     -->   (usb)  01234567890abcdef         unauthorized  
ERROR:         (tcpip)  192.168.1.1:5555                device  Nexus_5
ERROR: Select a device via -s (--serial), -d (--select-usb) or -e (--select-tcpip)
```

Since there are several devices connected over USB in this example, it fails, with a detailed error message.

Notice that the list of devices is printed at `DEBUG` level if it works (so it is not printed in release mode unless `-Vdebug` is passed), whereas it is printed at `ERROR` level if it fails.
(Maybe we should print it at `INFO` level when everything works, so that it is always printed?)

The first commits implement the same behavior for listing USB devices for HID/OTG mode (introduced in [v1.22](https://github.com/Genymobile/scrcpy/releases/tag/v1.22):

```console
$ scrcpy --otg
…
ERROR: Multiple (2) USB devices:
ERROR:     --> fedcba09           (xxxx:xxxx)  OnePlus SM8150-MTP
ERROR:     --> 01234567890abcdef  (xxxx:xxxx)  LGE Nexus 5
ERROR: Select a device via -s (--serial)
```

---

Do not conflate `--select-tcpip` with the `--tcpip` option introduced in [v1.21](https://github.com/Genymobile/scrcpy/releases/tag/v1.21).
 - `--tcpip` (without argument):
   1. connects to the selected device (which could be selected via `-s`, even `-U` or `-T`)
   2. enables TCP/IP mode (execute `adb tcpip 5555`)
   3. retrieves the device IP (by parsing `adb shell ip route`)
   4. connects to the device (`adb connect IP:5555`)
   5. starts mirroring
 - `--select-tcpip` just selects a device already connected (i.e. listed in `adb devices`) via TCP/IP

_In hindsight, maybe `--tcpip` should have been called `--connect-tcpip` or similar to avoid confusion, but it's too late._

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,7 +422,7 @@ scrcpy -b2M -m800  # short version
 
 #### Multi-devices
 
-If several devices are listed in `adb devices`, you must specify the _serial_:
+If several devices are listed in `adb devices`, you can specify the _serial_:
 
 ```bash
 scrcpy --serial 0123456789abcdef
@@ -434,6 +434,19 @@ If the device is connected over TCP/IP:
 ```bash
 scrcpy --serial 192.168.0.1:5555
 scrcpy -s 192.168.0.1:5555  # short version
+```
+
+If only one device is connected via either USB or TCP/IP, it is possible to
+select it automatically:
+
+```bash
+# Select the only device connected via USB
+scrcpy -d             # like adb -d
+scrcpy --select-usb   # long version
+
+# Select the only device connected via TCP/IP
+scrcpy -e             # like adb -e
+scrcpy --select-tcpip # long version
 ```
 
 You can start several instances of _scrcpy_ for several devices.
--- a/app/meson.build
+++ b/app/meson.build
@@ -1,6 +1,7 @@
 src = [
     'src/main.c',
     'src/adb/adb.c',
+    'src/adb/adb_device.c',
     'src/adb/adb_parser.c',
     'src/adb/adb_tunnel.c',
     'src/cli.c',
@@ -221,6 +222,7 @@ if get_option('buildtype') == 'debug'
     tests = [
         ['test_adb_parser', [
             'tests/test_adb_parser.c',
+            'src/adb/adb_device.c',
             'src/adb/adb_parser.c',
             'src/util/str.c',
             'src/util/strbuf.c',
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -44,6 +44,12 @@ The values are expressed in the device natural orientation (typically, portrait 
 value is computed on the cropped size.
 
 .TP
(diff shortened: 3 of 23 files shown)


erawhctim:
> maybe -U should be replaced by -d and -T replaced by -e to match what adb provides

This sounds ideal 👌 matches the muscle memory from using `adb` standalone

rom1v:
> This sounds ideal :ok_hand: matches the muscle memory from using `adb` standalone

Thank you for your feedback.

On the other hand, these `adb` options were intended to select device (`-d`) VS emulator (`-e`), and their semantics have shifted to mean usb VS tcpip. I'm not sure using these legacy letters is a good idea (in practice, scrcpy never mirrors an emulator, and everything is a device).

rom1v:
I changed to use `-d` and `-e`.

Issue #2930, opened 2022-01-08
Add bash completion script


rom1v:
scrcpy have a lot of options. It would be convenient if they were auto-completed on the command line.

For example:

```bash
scrcpy --loc
# press <tab><tab>
scrcpy --lock-video-orientation
```

https://iridakos.com/programming/2018/03/01/bash-programmable-completion-tutorial

On debian, the existing completion scripts are stored in `/usr/share/bash-completion/completions`.

PR #3012, opened 2022-02-10
add shell completions for all versions

hltdev8642:
Created completion files by hand, paraphrasing description/value fields to make them a suitable length to be displayed in terminal  (singleline). 

Not sure exactly where these would be put within the repo, so I made a completions directory for the time being to place the files. 

Hope this may be a helpful contribution, as I always wanted a completion file but couldn't find one or autogenerate it properly (so decided to make it by hand). 

Best,
Jared (hltdev)


rom1v:
Hi, thank you, there was even a feature request for that: https://github.com/Genymobile/scrcpy/issues/2930 :wink:

Remarks:
 - the 2 files looks almost identical, could they be "merged" into one?
 - the format seems different from the files in `/usr/share/bash-completion/completions` (which btw don't provide a text description for each command).

@yangfl is there a specific format required by Debian?

yangfl:
AFAIK there's no such specification. But seems these completions are for zsh and, of course we can ship them into `/usr/share/zsh/vendor-completions`, this seems a bit less useful as Debian's default shell is bash.

(Disclaimer: I use zsh too.)

hltdev8642:
> * the 2 files look almost identical, could they be "merged" into one?

Correct. They are identical except for `_scrcpy.exe` vs `_scrcpy` , which would just determine which word would trigger completion in the terminal (I have scrcpy aliased to scrcpy.exe on WSL anyways, which results in `scrcpy [tab]` and `scrcpy.exe [tab]` both triggering the completion anyways. 

So basically yes, you could potentially merge them but it would only trigger for either the `scrcpy.exe` or `scrcpy` command. FYI I generated the completion using the windows version (assuming that all the distributions have the same set of commands).

> * the format seems different from the files in `/usr/share/bash-completion/completions` (which btw don't provide a text description for each command).

as @yangfl  said it is a zsh completion 😉 . 
I can create a bash one too, shouldn't be too difficult. Could do that in the next few days and add it to the pull request 😄 👍

rom1v:
> you could potentially merge them but it would only trigger for either the scrcpy.exe or scrcpy command.

> This is the format that you will use most often for the first line, but you can also use the same file for completing several different functions if you want. See [here](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Autoloaded-files) for more details.

https://github.com/zsh-users/zsh-completions/blob/3fb84ee9cc290adb38ac02c629f7fc026acfd2ff/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command

(after a quick look, it's not clear to me how to, but it looks possible)

> I can create a bash one too

That would be great :+1:

Thank you :)

hltdev8642:
@rom1v, 

I have made a few other completions that do something similar to what you mentioned above, so I should be able to rig something up no problem 👍 (I used a combo of switch statements and functions, that should work in this scenario as well).

Also I worked on the bash completion and made one that works correctly, although it only completes the flags w/o descriptions. Not sure if generating a description per flag is even possible, but unless someone here knows I'll just look into it.

If it is fine without the descriptions, then I'll just merge the two zsh ones and push that and the bash one asap :-)

I'll report back soon!
Jared

rom1v:
I squashed your commits into a single one, then applied the following changes:
 - install to `$PREFIX/share/zsh/site-functions` (it does not work if I install the script to `/usr/local/share/zsh/vendor-completions`)
 - add completion for new options implemented recently (`-d`, `-e`, `--no-cleanup`, `--print-fps`)
 - remove `'*:filename:_files'` (scrcpy never accepts an input filename)
 - remove `.` at the end of descriptions for consistency
 - remove execution flag for `_scrcpy` (`chmod -x`) (the completions scripts are not "executable")
 - add missing values for `--lock-video-orientation`
 - document possible values using the same pattern everywhere: `(val1, val2, val3…)`
 - fix some typos

The resulting commit is here: [`zsh_completion`](https://github.com/Genymobile/scrcpy/commit/zsh_completion)

Please review :)

rom1v:
(FYI, I started working on a bash completion script yesterday, I will continue tonight or tomorrow.)

rom1v:
I fixed the completion of parameters: [`zsh_completion.2`](https://github.com/Genymobile/scrcpy/commit/zsh_completion.2)

For example:
```
% scrcpy --lock-video-orientation=
Completing orientation
0         1         2         3         initial   unlocked
```

rom1v:
> I started working on a bash completion, which works but only completes the flags (w/o any description fields). Not too sure if having descriptions is even possible/needed for a bash completion anyways (so unless someone here knows I can look into that).
> I can push that file as well if it will be good enough in that state, otherwise I will work on figuring out how to add the descriptions to the bash completion (and the zsh one should be all set now).

Oh, sorry, I missed the part where you actually started working on it (I thought you just planned to).

> (FYI, I started working on a bash completion script yesterday, I will continue tonight or tomorrow.)

I finalized my script yesterday evening, I planned to open a PR after this one is merged, but I can share it here: [`bash_completion`](https://github.com/Genymobile/scrcpy/commits/bash_completion)

Could you share a branch of your current bash script, please? Maybe there are possible improvements on both. Thank you.

hltdev8642:
> I fixed the completion of parameters: [`zsh_completion.2`](https://github.com/Genymobile/scrcpy/commits/zsh_completion.2)
> 
> For example:
> 
> ```
> % scrcpy --lock-video-orientation=
> Completing orientation
> 0         1         2         3         initial   unlocked
> ```

looks great, nice job! 👍 😄 👍 

> Could you share a branch of your current bash script, please? Maybe there are possible improvements on both. Thank you.

Sure, I just finished it, pretty basic but does work nonetheless! 

Note that I don't have a ton of experience with branch-related tasks using git, so _hopefully_ I did this correctly...
(if not, definitely let me know , as I am more then happy to learn something new👍) 

[\[bash-completion\]](https://github.com/hltdev8642/scrcpy/blob/bash_completion/completions/bash-completion/_scrcpy)

PR #3016, opened 2022-02-12
I have macOS High Sierra cant install Scrcpy on it please help

Manish-Bandi:
below is the error:

Error: scrcpy: Failed to download resource "theora" Download failed: Homebrew-installed `curl` is not installed for: https://downloads.xiph.org/releases/theora/libtheora-1.1.1.tar.bz2


rom1v:
1. it's an homebrew issue
2. You opened a PR instead of an issue

PR #3020, opened 2022-02-13
Add --no-cleanup option

### LINKED ISSUE #1764: Add posibilty to keep screen off after disconection

On version 1.15 was added that after disconection with option -S (--turn-screen-off) the screen will turn on again , but as for many this was annoying for me was a very usefull feature , so ¿can an extra option be added so that it turn screen off and dont turn on after disconection? 

on code seems quite posible as for now its hardcoded to always turn it on again (**true**)

`CleanUp.configure(mustDisableShowTouchesOnCleanUp, restoreStayOn, true);`

rom1v:
It might be useful not to cleanup on exit (to leave the screen turned off for example).

Fixes #1764

### DIFF
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -147,6 +147,12 @@ It may only work over USB, and is currently only supported on Linux.
 Also see \fB\-\-hid\-keyboard\fR.
 
 .TP
+.B \-\-no\-cleanup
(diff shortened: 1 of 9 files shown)


bbarcee:
I too would like this option - going to fix my issue!

PR #3029, opened 2022-02-17
Dynamically allocate ADB devices

DarkZeros:
This fixes connection to farm servers  with many more than 16 devices.

### DIFF
--- a/app/src/adb/adb.c
+++ b/app/src/adb/adb.c
@@ -519,9 +519,7 @@ bool
 sc_adb_select_device(struct sc_intr *intr,
                      const struct sc_adb_device_selector *selector,
                      unsigned flags, struct sc_adb_device *out_device) {
-    struct sc_adb_device devices[16];
-    ssize_t count =
-        sc_adb_list_devices(intr, flags, devices, ARRAY_LEN(devices));
+    ssize_t count = sc_adb_list_devices(intr, flags, NULL, 4096);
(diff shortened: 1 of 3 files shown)


rom1v:

> This fixes connection to farm servers with many more than 16 devices.

Out of curiosity, how many in practice?

If there are more than 50~100 devices, it will also reach this limit: https://github.com/Genymobile/scrcpy/blob/85edba20e79cdbde5295696ecc4950f584819899/app/src/adb/adb.c#L410-L417

`sc_usb_list_devices()` must also be adapted.

DarkZeros:
> Thank you for your contribution.
> 
> > This fixes connection to farm servers with many more than 16 devices.
> 
> Out of curiosity, how many in practice?
> 
> If there are more than 50~100 devices, it will also reach this limit:
> 
> https://github.com/Genymobile/scrcpy/blob/85edba20e79cdbde5295696ecc4950f584819899/app/src/adb/adb.c#L410-L417
> 
> `sc_usb_list_devices()` must also be adapted.

Right now the server has 20 devices. But in practice, as many as the USB protocol allows.
We use them as a way to "share" devices between users. Just discovered today that 4 were out of reach for scrcpy so I quickly drafted this change.

Indeed it may be useful to update the USB part as well, but I couldn't test that feature.

rom1v:
Superseded by #3035. Please test ;)

PR #3035, opened 2022-02-18
Dynamically allocate adb/usb devices using a vector, to avoid a hardcoded limit

rom1v:
The maximum number of adb devices (and usb devices for HID/OTG) was hardcoded to 16, which is too low.

The first commit (8d0694494697b940daf3d9cbff1fb3359462adc1) adds a `sc_vector` data structure and API. It is adapted from [`vlc_vector`](https://code.videolan.org/videolan/vlc/-/blob/0857947abaed9c89810cd96353aaa1b7e6ba3b0d/include/vlc_vector.h) that I initially [wrote while implementing the VLC playlist](https://blog.rom1v.com/2019/05/a-new-core-playlist-for-vlc-4).

It is then used for listing usb devices and adb devices to remove any hardcoded limitation (the vector grows dynamically as necessary).

There is still a limitation in the parsing of `adb devices -l`, which is implemented for simplicity in a single pass on the whole output. The output was limited to 4096 bytes (which represents ~40 devices). I changed to 65536 bytes, it should be sufficient in practice (~650 devices listed by `adb devices` should never happen).

This replaces #3029.

@DarkZeros I added you as co-author for the adb devices listing. Please test this PR and confirm that it works for your use case.

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -282,6 +282,9 @@ if get_option('buildtype') == 'debug'
             'src/util/str.c',
             'src/util/strbuf.c',
         ]],
+        ['test_vector', [
+            'tests/test_vector.c',
+        ]],
     ]
 
     foreach t : tests
--- a/app/src/adb/adb.c
+++ b/app/src/adb/adb.c
@@ -5,6 +5,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "adb_device.h"
 #include "adb_parser.h"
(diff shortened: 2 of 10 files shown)


DarkZeros:
I tested and managed to connect to the 17-20 devices with this patch, seems working from my side!
I didn't thoroughly review though.

rom1v:
Thank you for your tests :+1: Merged.

PR #3163, opened 2022-04-02
support udt timeout codec option

HEZI0427:
Non-negative timeout had been specified in the call to dequeueOutputBuffer,
 indicates that the call timed out, this will fix idr req not return frame bug


HEZI0427:
@

PR #3191, opened 2022-04-16
Actions on start exit

brunoais:
Allow running arbitrary scripts on start and exit of scrcpy.

Do note the exit script is not guaranteed to run depending on the phone you use and the way scrcpy exits.
See: https://github.com/Genymobile/scrcpy/pull/2265#issuecomment-977617087

Replaces #2265

### DIFF
--- a/app/src/server.c
+++ b/app/src/server.c
@@ -93,6 +93,7 @@ sc_server_params_copy(struct sc_server_params *dst,
     COPY(crop);
     COPY(codec_options);
     COPY(encoder_name);
+    COPY(hook_script);
     COPY(tcpip_dst);
 #undef COPY
 
@@ -232,6 +233,23 @@ execute_server(struct sc_server *server,
     }
     if (params->encoder_name) {
         ADD_PARAM("encoder_name=%s", params->encoder_name);
+    }
+    if (params->hook_script) {
+        char* requoted_hook;
+        int64_t replace_result = sc_str_find_replace(params->hook_script, "'", "'\"'\"'", &requoted_hook);
+        switch(replace_result){
+            case -2:
+                LOG_OOM();
+                break;
+            case -1:
+            case 0:
+                ADD_PARAM("hook_script='%s'", params->hook_script);
+                break;
+            default:
+                ADD_PARAM("hook_script='%s'", requoted_hook);
+                free(requoted_hook);
+        }
+
     }
     if (params->power_off_on_close) {
         ADD_PARAM("power_off_on_close=true");
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -56,6 +56,7 @@
 #define OPT_OTG                    1036
 #define OPT_NO_CLEANUP             1037
 #define OPT_PRINT_FPS              1038
+#define OPT_HOOK_SCRIPT            1039
 
 struct sc_option {
     char shortopt;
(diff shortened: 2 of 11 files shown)


app/src/server.c
```diff
@@ -233,6 +234,23 @@ execute_server(struct sc_server *server,
     if (params->encoder_name) {
         ADD_PARAM("encoder_name=%s", params->encoder_name);
     }
+    if (params->hook_script) {
+        char* requoted_hook;
+        int64_t replace_result = sc_str_find_replace(params->hook_script, "'", "'\"'\"'", &requoted_hook);
```
brunoais, line 239: Unlike what would be expected, adb shell just concatenates the arguments as strings same
as if they were just separated by spaces. Sending the commands in one argument
or sending as multiple arguments doesn't preduce any different outcome.
Sad but that's how adb works.

try typing this in your shell:
`adb shell 'am broadcast'`
vs
`adb shell am broadcast`

If they both work, means there's no reasonable way to deliver the arguments as-is, even when using `execvp` (which scrcpy is using).
Sadly, they have to be escaped.

Whether the escape be done here or before the calling stack to call adb shell is up for debate. However, no doubt it's needed and no doubt the behavior adb has is improper when intending to provide arguments as-is

brunoais: In that case, should I add sending the script to `control_msg.c` by queueing a message using `sc_controller_push_msg()` sometime in the beginning of the program execution while also implementing a new type at `sc_control_msg_serialize()` for that new feature (and then adding the counter-part code in the server)?

rom1v: I don't know what to do. I have mixed feelings about this feature.

> should I add sending the script to `control_msg.c` by queueing a message using `sc_controller_push_msg()`

This would require to start the `CleanUp` process only after this message is received :/

brunoais: > I don't know what to do. I have mixed feelings about this feature.

What about sending much of the first data through stdin? It can be encoded in some way and then just sent to stdin when `adb shell` is run.
With that, we can also deal with the issue of limited size of args, the argv inconsistency, etc... and just send all arguments through a stream in stdin as soon as the stdin stream exists. Maybe just encoding as base64 is enough and then ending with maybe the '\0'

> This would require to start the CleanUp process only after this message is received :/
Only if a script exists but yes

rom1v: > What about sending much of the first data through stdin?

IIRC, `adb shell` does not forward stdin to the process it runs (contrary to ssh).

brunoais: It works for me:
``` 
$ echo "love me" | adb shell 'cat -'
love me
love me
```
It is weird that `adb shell` first outputs the program it receives.
However, it is shown in the manual:
image

rom1v: Indeed. Even with an old adb version (tested with 36).

However, the behavior is slightly different depending on the adb server on the device:
 - on my Nexus 5, the input is "echoed" to the output, and stdin is never closed (even if it is closed on the desktop side)
 - on my OnePlus 7 Pro, the input is not echoed, and stdin is correctly closed

brunoais: > on my Nexus 5, the input is "echoed" to the output, and stdin is never closed (even if it is closed on the desktop side)

Which android version?

On my version:

1. the echo happens
2. stdin is not closed but it can be closed by the running program.
3. Terminating the running program correctly terminates `adb shell`

Do you think this can be tried?

brunoais: @rom1v Any answers you can give before I try again? It's been a few weeks

brunoais: @rom1v May I proceed to try to implement what I proposed? So far I can tell it appears to work.

rom1v: I have mixed feelings about this. IMO this feature is secondary, it adds complexity and requires to change how the parameters are passed, and apparently it does not behave the same way on all adb/devices. So TBH I'm not very keen it on for the moment :/

brunoais: Alright. In that case this and the alternative way that was already shot down are shot down.
I'll see if I can find an alternative way to process this.

Maybe it can be sent as a `ControlMessage` message like most features are done. Possibly in a similar way that the clipboard message is sent.
However, it would be convenient to increase the limit to `1«19` 512KB. That way a script can be ~400KB which would allow a lot of flexibility.

I think that the downside of using the `ControlMessage` API is acceptable with that upside of stability between versions.

What do you think?

brunoais: @rom1v ?

PR #3210, opened 2022-04-23
Add option to prevent power on on start

### LINKED ISSUE #3148: Option to power-off at start
- [X] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

I have a device permanently living next to my computer, and I use the systemd integration from #3101, so basically I always have scrcpy running. Unless I power-off (Alt-P), scrcpy will occupy the CPU all time (#3112). I am looking for an option that automatically hits the power button after connection, similarly to how `--turn-screen-off` just turns the device screen off. This would allow auto-connection, but without involving the CPU too much until I need to interact.

Thanks!

rom1v:
By default, on start, the device is powered on. To prevent this behavior, add a new option `--no-power-on`.

Fixes #3148

---

Are you ok with the option name (`--no-power-on`)?
There is already an option `--power-off-on-close`, so we could use `--no-power-on-on-start`, but I think I prefer `--no-power-on`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -721,6 +721,16 @@ To turn the device screen off when closing scrcpy:
 scrcpy --power-off-on-close
 ```
 
+#### Power on on start
+
+By default, on start, the device is powered on.
+
+To prevent this behavior:
+
+```bash
+scrcpy --no-power-on
+```
+
 
 #### Show touches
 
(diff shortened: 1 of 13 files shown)


brunoais:
I'm fine with this but wouldn't this be achievable with https://github.com/Genymobile/scrcpy/pull/3191?

PR #3217, opened 2022-04-25
Improve German translation

Sea-n:
Change `Wifi` to `Wi-Fi` as English and other translations.


rom1v:
Merged: 85512b14670db9b41606ec9841b8755a86633536

PR #3252, opened 2022-05-09
Add build support for Windows on ARM64 (based on llvm-mingw)

ZIXT233:
I tried to cross compile scrcpy for win on arm64 using llvm-mingw, which seems to be working properly and significantly improves video performance in win on arm64.

What I did:
- Prebuilt libusb,ffmpeg,SDL2 for win on arm64, which was placed in my responsitories. (Removed libusb build flags "-mwin32" and "--add-stdcall-alias" to support llvm)
- Added scripts to get prebuilt-deps.
- Added win on arm64 build option in release.mk,  which should be manually specified as llvm-mingw is a special toolchain. In addtion, I removed the linker option "-Wl,--allow-shlib-undefined" unsupported by llvm in build.ninja after meson completed. (There seems to be no way to prevent meson from passing this option so far, see:  https://github.com/mesonbuild/meson/issues/7506 )
- Modified app/build.ninja to select libusb dll depending on host_machine.cpu_family().

Known issue:
- Reported error "couldn't find image decorder" and "couldn't load icon" when starting. 
36F3D3382058CE4FCE21E4AEB551C50B

I'm not sure where to put the prebuilt package to ensure its maintenance .
Let me know if there are any problems.

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -145,13 +145,21 @@ else
     prebuilt_libusb_root = meson.get_cross_property('prebuilt_libusb_root')
     libusb_bin_dir = meson.current_source_dir() + '/prebuilt-deps/data/' + prebuilt_libusb
     libusb_include_dir = 'prebuilt-deps/data/' + prebuilt_libusb_root + '/include'
-
-    libusb = declare_dependency(
-        dependencies: [
-            cc.find_library('msys-usb-1.0', dirs: libusb_bin_dir),
+    if host_machine.cpu_family() == 'x86'
+      libusb = declare_dependency(
+	dependencies: [
+	  cc.find_library('msys-usb-1.0', dirs: libusb_bin_dir),
         ],
         include_directories: include_directories(libusb_include_dir)
-    )
+      )
+    elif host_machine.cpu_family() == 'aarch64'
+      libusb = declare_dependency(
+	dependencies: [
+	  cc.find_library('libusb-1.0', dirs: libusb_bin_dir),
+        ],
+        include_directories: include_directories(libusb_include_dir)
+      )
+    endif
 
     dependencies = [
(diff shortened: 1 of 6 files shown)


rom1v:
Thank you for your work.

The "prebuilt" mechanism of scrcpy is quite rudimentary (it downloads prebuilt releases). This can be a limitation (we could not patch dependencies as necessary), but in practice it "works".

However, this is acceptable only if the prebuilt binaries are "official" (e.g. linked from the official website of the dependency). It could not depend on binaries from random repositories. I understand that official binaries do not exist for ARM64 (so you did build them by yourself), but it is a problem to depend on `https://github.com/ZIXT233/`. :disappointed:

ZIXT233:
It is feasible to build dependencies from the official code. Maybe I can modify the script to support it？ @rom1v

rom1v:
@ZIXT233 Please read the discussion in #1753 first :)

It would be great to (optionally) build (from Linux) the dependencies (SDL, FFmpeg, libusb, adb), but still be able to use either the official prebuilt or the existing package already installed, if any. After that, yes, we could add the scripts/options to cross-build for Windows-arm64.

ZIXT233:
So are you considering adding optional compilation deps dynamic build scripts/options for more common situation?

rom1v:
Separate scripts to build the dependencies when requested (similar to the contribs in vlc), so that the resulting binaries could be used to build scrcpy, would be great, yes.

ZIXT233:
I'll try it. It may take some time to design.

rom1v:
I started a skeleton: #3283

tormodvolden:
We'd like to fix this properly in libusb: https://github.com/libusb/libusb/issues/1192

PR #3283, opened 2022-05-21
[WIP] Build dependencies from sources

rom1v:
For Windows releases, scrcpy use prebuilt dependencies downloaded from "official" sources.
However, the vendors don't always provide builds for all the architectures, and we can't apply patches when necessary (to fix a bug until the next release for example). Also see #1753.

For Linux, we use the dependencies from the package manager.
But it might be interesting to build the dependencies to provide a "portable" version (either a static build or an archive with all .so included and RPATH set).

This MR aims to provide scripts (one for each dependency: ffmpeg, sdl, libusb, adb(?)), intended to be called from the release script (from Linux to build for Linux and Windows (macOS is more complicated from Linux)).

For now, only the script to build ffmpeg for linux is provided:

```bash
export HOST=x86_64-linux-gnu
export HOST_SYSTEM=linux  # for convenience, could be deduced from $HOST later
cd app/deps/src
./ffmpeg.sh
```

This downloads the FFmpeg sources, then build locally for the target platform, and install them into `app/deps/target-x86_64-linux-gnu`. The structure looks like this (only the relevant parts of the tree are shown here):

```
app/deps
├── data
│   ├── ffmpeg-5.0.1-ef2efae259ce80a240de48ec85ecb062cecca26e4352ffb3fda562c21a93007b
│   └── ffmpeg-5.0.1.tar.xz
├── src
│   ├── ffmpeg.sh
│   └── init_deps
└── target-x86_64-linux-gnu
    ├── include
    │   ├── libavcodec
    │   ├── libavdevice
    │   ├── libavformat
    │   └── libavutil
    ├── lib
    │   ├── libavcodec.so -> libavcodec.so.59.18.100
    │   ├── libavcodec.so.59 -> libavcodec.so.59.18.100
    │   ├── libavcodec.so.59.18.100
    │   ├── libavdevice.so -> libavdevice.so.59.4.100
    │   ├── libavdevice.so.59 -> libavdevice.so.59.4.100
    │   ├── libavdevice.so.59.4.100
    │   ├── libavformat.so -> libavformat.so.59.16.100
    │   ├── libavformat.so.59 -> libavformat.so.59.16.100
    │   ├── libavformat.so.59.16.100
    │   ├── libavutil.so -> libavutil.so.57.17.100
    │   ├── libavutil.so.57 -> libavutil.so.57.17.100
    │   ├── libavutil.so.57.17.100
    │   └── pkgconfig
    │       ├── libavcodec.pc
    │       ├── libavdevice.pc
    │       ├── libavformat.pc
    │       └── libavutil.pc
    └── …
```

The correct target directory should then be used by meson to build (not done yet).

The scripts (for now ffmpeg only) must be adapted to cross-compile with mingw when the host system is Windows (not done yet).

### DIFF
--- a/app/deps/.gitignore
+++ b/app/deps/.gitignore
@@ -1,0 +1,2 @@
+/data
+/target-*
--- a/app/deps/src/build-ffmpeg.sh
+++ b/app/deps/src/build-ffmpeg.sh
@@ -29,6 +29,7 @@ cd "build-$HOST"
 
 params=(
     --prefix="$INSTALL_DIR"
+    --arch="$ARCH"
     --disable-autodetect
     --disable-programs
     --disable-everything
@@ -44,10 +45,19 @@ params=(
     --enable-muxer=mp4
     --enable-muxer=matroska
 )
-if [[ "$HOST_SYSTEM" == 'linux' ]]
-then
-    params+=(--enable-libv4l2)
-fi
+
+case "$HOST_SYSTEM" in
+    linux)
+        params+=(--enable-libv4l2)
+        ;;
+    windows)
+        params+=(--target-os=mingw32)
+        params+=(--cross-prefix="$HOST-")
+        ;;
+    *)
+        fail "Unsupported platform: $HOST"
+        ;;
+esac
 
 ../configure "${params[@]}"
 make -j $NJOBS
--- a/app/deps/src/build-ffmpeg.sh
+++ b/app/deps/src/build-ffmpeg.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+. init_deps
+
+VERSION=5.0.1
+FILENAME=ffmpeg-$VERSION.tar.xz
+
+URL=http://ffmpeg.org/releases/ffmpeg-$VERSION.tar.xz
+SHA256SUM=ef2efae259ce80a240de48ec85ecb062cecca26e4352ffb3fda562c21a93007b
+
+DEP_DIR="$DATA_DIR/ffmpeg-$VERSION-$SHA256SUM"
+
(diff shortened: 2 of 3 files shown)


eli-schwartz:
There is also https://gitlab.freedesktop.org/gstreamer/meson-ports/ffmpeg/ which the gstreamer project is maintaining to build ffmpeg using meson.

You could add this as a meson subproject, and if Meson cannot detect a suitable version of ffmpeg it would then build ffmpeg automatically using the same build state (as it is part of the same unified build).

rom1v:
> There is also https://gitlab.freedesktop.org/gstreamer/meson-ports/ffmpeg/ which the gstreamer project is maintaining to build ffmpeg using meson.

Ah cool. But it does not seem maintained / up to date.

Unless all scrcpy dependencies use meson, we need a "generic" system, so I prefer using the official FFmpeg sources/build system directly.

eli-schwartz:
> Unless all scrcpy dependencies use meson,

That's an emphatic "maybe".

There's an sdl2 port in the WrapDB available as `meson wrap install sdl2` which is up to date and being actively maintained by someone using it for another project.

Libusb is pretty simple to add support for, dunno about adb.

Ffmpeg is of course complex enough that updating it is a major hassle and done infrequently AFAIK, and may not necessarily be the ideal situation. Though adding a wrap file for it would not actually make a difference unless meson would otherwise fail with "dependency not found", so you may feel like adding it alongside the script in this PR? You could then choose to use the script or have meson build the wrap.

rom1v:
See 0fc62bfcd63fdccba759c1421ca7e4b5a2278cb9.

PR #3351, opened 2022-06-27
Desktop entry file for Linux app launchers #295

yan12125:
See: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

* Address review comments

Seems #296 is stale, so I created a replacement with fixes. Differences from the previous pull request are in the second commit.

### DIFF
--- a/app/data/scrcpy-console.desktop
+++ b/app/data/scrcpy-console.desktop
@@ -1,0 +1,13 @@
+[Desktop Entry]
+Name=scrcpy (console)
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+# For some users, the PATH or ADB environment variables are set from the shell
+# startup file, like .bashrc or .zshrc… Run an interactive shell to get
+# environment correctly initialized.
+Exec=/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'
+Icon=scrcpy
+Terminal=true
+Type=Application
+Categories=Utility;RemoteAccess;
+StartupNotify=false
--- a/app/data/scrcpy.desktop
+++ b/app/data/scrcpy.desktop
@@ -2,7 +2,10 @@
 Name=scrcpy
 GenericName=Android Remote Control
 Comment=Display and control your Android device
-Exec=scrcpy
+# For some users, the PATH or ADB environment variables are set from the shell
+# startup file, like .bashrc or .zshrc… Run an interactive shell to get
+# environment correctly initialized.
+Exec=/bin/sh -c '"$SHELL" -i -c scrcpy'
 Icon=scrcpy
 Terminal=false
 Type=Application
--- a/app/data/scrcpy.desktop
+++ b/app/data/scrcpy.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Name=scrcpy
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+Exec=scrcpy
+Icon=scrcpy
+Terminal=false
+Type=Application
+Categories=Utility;RemoteAccess;
+StartupNotify=false
--- a/app/meson.build
+++ b/app/meson.build
@@ -235,6 +235,13 @@ install_data('data/zsh-completion/_scrcpy',
 install_data('data/bash-completion/scrcpy',
              install_dir: join_paths(datadir, 'bash-completion/completions'))
 
+# Desktop entry file for application launchers
+if host_machine.system() == 'linux'
+    # Install a launcher (ex: /usr/local/share/applications/scrcpy.desktop)
+    install_data('data/scrcpy.desktop',
+                 install_dir: join_paths(datadir, 'applications'))
+endif
+
 
 ### TESTS
 
--- a/app/meson.build
+++ b/app/meson.build
@@ -240,6 +240,8 @@ if host_machine.system() == 'linux'
     # Install a launcher (ex: /usr/local/share/applications/scrcpy.desktop)
     install_data('data/scrcpy.desktop',
                  install_dir: join_paths(datadir, 'applications'))
+    install_data('data/scrcpy-console.desktop',
+                 install_dir: join_paths(datadir, 'applications'))
 endif
 
 
--- a/app/meson.build
+++ b/app/meson.build
@@ -223,14 +223,17 @@ executable('scrcpy', src,
            install: true,
            c_args: [])
 
+# <https://mesonbuild.com/Builtin-options.html#directories>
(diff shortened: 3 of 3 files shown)


yan12125:
Awesome improvements! Just one minor suggestion -

> For some users, `adb` is not in default $PATH but the one configured in .bashrc/ or .zshrc/...

"or" indeed makes it clearer, but a slash after .bashrc is somewhat unusual. Probably better with

> For some users, `adb` is not in default $PATH but the one configured in `.bashrc`, `.zshrc`, or other shell startup scripts.

(Term "shell startup scripts" is just the first one out of my head. Different people use different terms for those files :D)

`scrcpy.desktop` works as usual. For `scrcpy-console.desktop`, the "complex" Exec= line hits some other software. I'm using QTerminal on LXQt, where https://github.com/lxqt/qterminal/pull/961 is needed to correctly parse that line. That's a QTerminal bug, so keeping the current `scrcpy-console.desktop` is fine.

rom1v:
> (Term "shell startup scripts" is just the first one out of my head. Different people use different terms for those files :D)

In `man bash`:

> ```
>        ~/.bashrc
>               The individual per-interactive-shell startup file
> ```

So "shell startup script" looks good to me.

> For some users, adb is not in default $PATH but the one configured in .bashrc, .zshrc, or other shell startup scripts.

I suggest:

```diff
+# For some users, the PATH or ADB environment variables are set from the shell
+# startup file, like .bashrc or .zshrc… Run an interactive shell to get
+# environment correctly initialized.
```

What do you think? See [`pr3351.2`](https://github.com/Genymobile/scrcpy/commits/pr3351.2).

> For `scrcpy-console.desktop`, the "complex" Exec= line hits some other software. I'm using QTerminal on LXQt, where https://github.com/lxqt/qterminal/pull/961 is needed to correctly parse that line.

:disappointed: 

> That's a QTerminal bug, so keeping the current `scrcpy-console`.desktop is fine.

OK

yan12125:
Looks great, thank you very much for the efforts!

rom1v:
@asnelling as the author of #296, any remark?

app/scrcpy.desktop
```diff
@@ -0,0 +1,12 @@
+[Desktop Entry]
+Name=scrcpy
+GenericName=Android Remote Control
+Comment=Display and control your Android device
+# For some users, `adb` is not in default $PATH but the one configured in .bashrc/.zshrc/...
+# Run an interactive shell to get the same path used in terminals.
+Exec=/bin/sh -c '"$SHELL" -i -c scrcpy'
```
asnelling, line 7: For system-wide installations, Exec should use the system-wide `adb` installed at, e.g.: `/usr/bin/adb` and leave customization concerns up to the user in e.g.: `~/.local/applications/scrcpy.desktop`
:+1: good to merge

asnelling:
IMO, `Exec=` should let the system resolve `scrcpy` and _not_ invoke a shell. Users customizations belong in `~/.local/share/applications/scrcpy.desktop` . It seems the convention (at least in Ubuntu) is `Exec=scrcpy` or `Exec=/usr/bin/scrcpy`, which you can see with:

```Shell
$ grep '^Exec=' /usr/share/applications/*.desktop
```

More importantly, merge it :+1:

yan12125:
> For system-wide installations, Exec should use the system-wide `adb` installed at, e.g.: `/usr/bin/adb`

It's up to whether rom1v wants to support `$PATH` or `$ADB` configured in .bashrc/.zshrc/... or not. I'm fine with either.

> It seems the convention (at least in Ubuntu) is `Exec=scrcpy` or `Exec=/usr/bin/scrcpy`

It's also possible to extract those "complex" Exec= lines to standalone scripts like `/usr/bin/scrcpy-console.sh` and `/usr/bin/scrcpy-noconsole.sh`, and use them in Exec=.

rom1v:
> IMO, `Exec=` should let the system resolve `scrcpy` and _not_ invoke a shell. 

So, for the non-console version:

```
Exec=/usr/bin/scrcpy
```

For the console version:

```
Exec=/usr/bin/scrcpy || bash -i -c 'read -p "Press any key to quit..."'
```

?

Where should the user define `ADB` or `PATH` so that it is used for the launcher in that case?

@yangfl any opinion about the content of `Exec=`?

yangfl:
I think you can't use the shell syntax in `Exec=`

yan12125:
>> `Exec=/usr/bin/scrcpy || bash -i -c 'read -p "Press any key to quit..."'`
>
> I think you can't use the shell syntax in `Exec=`.

Yeah, `||` is passed as an argument and failed. Here is a working example with waiting and no interactive shell:

```
Exec=/bin/bash -c 'scrcpy || read -p "Press any key to quit..."'
```

> `"$SHELL" -i -c scrcpy`
> Strange, but acceptable as long as the software author satisfies with it.

I checked and found that qterminal and similar programs [expand environment variables](https://github.com/lxqt/qtermwidget/blob/1.1.0/lib/ShellCommand.cpp#L117). I'm not sure if other terminal emulators do that or not.

> Where should the user define `ADB` or `PATH` so that it is used for the launcher in that case?

That depends on how an Xorg or Wayland session is started. Apparently there is no universal way - see https://wiki.archlinux.org/title/Environment_variables#Graphical_environment

rom1v:
> Yeah, || is passed as an argument and failed. Here is a working example with waiting and no interactive shell:
>
> ```
> Exec=/bin/bash -c 'scrcpy || read -p "Press any key to quit..."'
> ```

OK. But then the scrcpy-console version will load environment variables from .bashrc (whatever the default shell), while the non-console version will not.

> `"$SHELL" -i -c scrcpy`
>
> Strange, but acceptable as long as the software author satisfies with it.

OK, so maybe I will keep [`pr3351.2`](https://github.com/Genymobile/scrcpy/commits/pr3351.2).

Thank you.

yan12125:
> But then the scrcpy-console version will load environment variables from .bashrc (whatever the default shell), while the non-console version will not.

Yeah that's an issue. How about `/bin/bash --norc --noprofile ...`?

Viileeu72:
> See: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html
> 
> * Address review comments
> 
> Seems #296 is stale, so I created a replacement with fixes. Differences from the previous pull request are in the second commit.

yan12125:
> How about `/bin/bash --norc --noprofile` ...?

Assuming no objection for this approach, I imported pr3351.2 branch, added proposed bash arguments and force-updated this PR.

rom1v:
Merged into `dev`.

yan12125:
Thanks!

PR #3378, opened 2022-07-12
Optimized icons filesize.

y04:
PNG image is visually identical. SVG image optimized for more smaller size and faster rendering, visually is the same.


rom1v:
Which tools did you use to optimize the files? Both `pngcrush` and `optipng` say that the png is already optimized.

jinu-jacob:
The app is really helpful broh. Bring it to the store.

y04:
> Which tools did you use to optimize the files? Both `pngcrush` and `optipng` say that the png is already optimized.

`Pngcrush` + `OptiPNG`+ `ECT` + `AdvanceCOMP` + `DeflOpt` + `defluff` + `pingo` + `pngwolf` + `TruePNG` + `ZopfliPNG` and 1 day of IDAT chunk brute-forcing to find the best compression.

Issue #3137, opened 2022-03-22
When multiple devices are present, let the user pick one.


oscarb:
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
_I'm always frustrated when I_ run `scrcpy` only to realize I had multiple devices/emulators connected and then need to copy the serial of one to connect to the device of my choice. 

**Describe the solution you'd like**
Instead of yielding an error, output the list of devices with each device prefixed by a number and then have  scrcpy` await user input for which device to use:
```
> scrcpy 
Multiple (3) ADB devices:
1) (usb)  QZ125F6WA1                      device  XQ_BQ52
2) (usb)  emulator-5554                   device  sdk_google_atv_x86
3) (usb)  emulator-5555                   device  sdk_google_atv_x86
Which device would you like to use (0 to exit)? 2
```
The copy could perhaps be a bit better.

**Describe alternatives you've considered**
An alternate solution would perhaps be if I wrote my own script which wraps the scrcpy tool with logic based on the scrcpy output, but that could likely break at any time.

**Additional context**
N/A

rom1v:
> Instead of yielding an error, output the list of devices with each device prefixed by a number and then have scrcpy` await user input for which device to use:

I already considered that, but scrcpy could not rely on user input, the user may start it without a console (that's quite common). And even when a console is available, it may be started from a script which does not expect it to wait.

rom1v:
#3005 added `-d` and `-e` to select USB and TCP/IP devices. But contrary to `adb`, an emulator is considered as an USB device. This s a bug, I'll fix it in the next release.

In your specific case, that would allow to call `scrcpy -d` to select the real device.

rom1v:
> But contrary to `adb`, an emulator is considered as an USB device. This s a bug, I'll fix it in the next release.

Fixed by #3140.

oscarb:
> > Instead of yielding an error, output the list of devices with each device prefixed by a number and then have scrcpy` await user input for which device to use:
> 
> I already considered that, but scrcpy could not rely on user input, the user may start it without a console (that's quite common). And even when a console is available, it may be started from a script which does not expect it to wait.

Thanks for taking the time to reply, and that you already had given it some thought :) Now that I understand better the reasons for not implementing a feature like this, feel free to close this.

[...]

Issue #1965, opened 2020-12-10
Provide an option to select a device, when multiple devices are connected


Kikju:
Often I have connected multiple emulators and phones to my computer. I have to open separate terminal, list the devices via `adb devices` and then copy each serial number and run scrcpy with `-s` argument. Or often I just launch scrcpy without arguments, and I get an error `adb: error: failed to get feature set: more than one device/emulator`.
It would be helpful, to provide an option to choose a specific device and continue instead of error "more than one device/emulator".

It could looks like this:
```
Multiple devices detected. To continue, please type the order of device you want to launch:
1. emulator-5554
2. 0123456789abcdef
```

Then I could just type "2" to continue and open scrcpy with selected device.

 - [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

PonyPC:
I agree with adding this feature, this also occurred while enabled wifi adb connection in one device.
But there are several applications that provide a user friendly UI such as QtScrcpy you can try: https://github.com/barry-ran/QtScrcpy
It is enough for developing android apk though some bugs in listing devices on my desktop.
I prefer this software AnLink https://anl.ink which seems use same technology from scrcpy, it is more convenient than QtScrcpy but not open source software, so choose your favorite.
Wish scrcpy could add an UI to choose devices.

erawhctim:
Would love to see something like this implemented. Another option (that mimics `adb` behavior) would be to add `-d` and `-e` flags to indicate a device vs. emulator?

rom1v:
Oh, I forgot this feature request.

I implemented such a feature this week-end: please review #3005.

rom1v:
Fixed by #3005.

Issue #485, opened 2019-04-01
What causes the "more than one divice/emulator" and how do I fix it?


MrMondayHonday:
Hi, the newbie here again.

For some reason I can no longer launch scrcpy.
after typing the usual  adb tcpip and followed by the adb connect  plus scrcpy I get:

adb: error: connect failed: more than one device/emulator 
ERROR: adb push" returned with value 1

So, what causes this  and how can I fix it?

npes87184:
The problem is you have more than one devices connected to your PC.

Using `adb devices` first.

And start `scrcpy` by `scrcpy -s [TARGET DEVICE'S SERIAL]`.

MrMondayHonday:
Thank you, I will investigate yhis.
But I have only one device connected to my ubuntu laptop. and that is my samsung phone.
However, I have noticed that it shows up as two distinct drives in my file manager, one called samsung android and the other mtp I believe. But this is not something new if related, and it worked before.

zhurai:
Just as a general FYI, if you have emulators running on your computer, those are also considered "other devices" by adb

MrMondayHonday:
A restart of computer and phone fixed it. 
I did not have any emulators running as far as I know.

But thanks!

rom1v:
If you were connected both via USB and via `adb connect`, you have 2 devices connected in `adb devices`.

jemckaleman:
> The problem is you have more than one devices connected to your PC.
> 
> Using `adb devices` first.
> 
> And start `scrcpy` by `scrcpy -s [TARGET DEVICE'S SERIAL]`.

An excellent contribution, good work bredrin

kaiphu:
I have the same issue cause I am I have multiple emulators running while I'm trying to use scrcpy just for my usb device. (scrcpy does work while the emulators are off, but I want to use it while they are on.) I have tried [scrcpy -s ****** and scrcpy usb] however it hits me with the "error: device unauthorized." When I [adb devices] while my phone is plugged in and have the emulators running only my phone shows up in devices. (unauthorized) Yes phone usb debugging is on. How can I get everything working at the same time?

Medievaldragon:
> 
> 
> Hi, the newbie here again.
> 
> For some reason I can no longer launch scrcpy. after typing the usual adb tcpip and followed by the adb connect plus scrcpy I get:
> 
> adb: error: connect failed: more than one device/emulator ERROR: adb push" returned with value 1
> 
[...]

PR #3418, opened 2022-08-05
Update scrcpy-console.bat

MrLeopard:
easy to use when multiple devices connected. 
requirement : https://github.com/Genymobile/scrcpy/issues/3137

### DIFF
--- a/doc/camera.md
+++ b/doc/camera.md
@@ -18,6 +18,17 @@ scrcpy --video-source=display --audio-source=mic    # force display AND micropho
 scrcpy --video-source=camera --audio-source=output  # force camera AND device audio output
 ```
 
+Audio can be disabled:
+
+```bash
+# audio not captured at all
+scrcpy --video-source=camera --no-audio
+scrcpy --video-source=camera --no-audio --record=file.mp4
+
+# audio captured and recorded, but not played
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you for your proposal.

If we really want this interaction with the user, it should not be in `scrcpy-console.bat` (which is only a script for Windows users to double-click without closing the terminal on error).

But scrcpy, by default, should not be interactive either (there might be no console, or it may be started from scripts, which are not expected to wait). An `--interactive` command line option could enable it, but then, it's useless (people do not explicitly pass the flag).

Since the command gives the list of users with their serial/ip, ready to be copy-pasted, I'm not sure it's worth it.

Ideally, I'd like scrcpy to be exposed as a library, and develop an alternative UI, which would expose the list of devices. But that requires a lot of work, that I cannot do on my free time.

MrLeopard:
Thanks for your replay.

When multiple devices connected, they have to use command line to run the programme, and I saw some guys also troubled by this design--#3137,#3005,#1965,#485

I read your opinion :

> "_but scrcpy could not rely on user input_"

So I thought how about just changing the script file scrcpy-console.bat, then I wrote this bat file, it just help people set ANDROID_HOME environment variable, scrcpy core not changed, but fresh user not need to see multiple devices error.

Thanks for your reading.

PR #3449, opened 2022-08-16
Remove deprecated meson.source_root()

rom1v:
This method is deprecated since Meson 0.56.0:
<https://mesonbuild.com/Release-notes-for-0-56-0.html#mesonbuild_root-and-mesonsource_root-are-deprecated>

We could replace it with meson.project_source_root(), but this would
make Meson 0.56 or above mandatory. Since the path in always computed
from the server/ directory, just add '..' to reference the root project
directory.

Refs c456e3826470753bdaefcffc93e9ae1bf25b12f7

cc @npes87184 as the author of c456e3826470753bdaefcffc93e9ae1bf25b12f7.

### DIFF
--- a/server/meson.build
+++ b/server/meson.build
@@ -13,8 +13,8 @@ if prebuilt_server == ''
                   install_dir: 'share/scrcpy')
 else
     if not prebuilt_server.startswith('/')
-        # relative path needs some trick
-        prebuilt_server = meson.source_root() + '/' + prebuilt_server
+        # prebuilt server path is relative to the root scrcpy directory
+        prebuilt_server = '../' + prebuilt_server
     endif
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> This method is deprecated since Meson 0.56.0:

Since you support versions older than that, it's fine to just not switch away from the deprecated function.

It's also possible to define `source_root = meson.current_source_dir()` inside the top-level meson.build, and refer to that source_root variable later on.

eli-schwartz:
Ah right. This is being passed as `input:` to a custom target, which is always calculated relative to the current context directory. It would be different if it was passed as a string argument to e.g. a run_target, but custom_target does make sure to handle it as a file.

So using ../ is fine here, it never needed to be a source_root() path. Never mind.

PR #3511, opened 2022-09-28
Update scrcpy-windows.manifest

Aninooo:


Aninooo:
none

rom1v:
Nope. https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process

Issue #1880, opened 2020-11-03
Avoid rewrites in tmp


lollita:
If I have understood correctly, scrcpy, every time it runs, it writes scrcpy-server.odex and scrcpy-server.vdex in /data/local/tmp.
Do it is possible to avoid all these rewrites, and keep the file by overwriting it if there is a new one?

rom1v:
See #1722

lollita:
I think it's unnecessary wear and tear on the memory card if avoidable.

Biswa96:
Maybe make an apk package for the scrcpy-server?

lollita:
I use
`chmod -w /data/local/tmp/oat/arm64/scrcpy-server.*`
Until the new version.

npes87184:
If user has concern about this, the best way is create an apk and install it. However, it may need to do something to make server as apk. In short time, maybe we can remain the server in device and calculate the check sum to ensure the server is expected.

rom1v:
For information, I started to work on a new option (probably `--install`) which would install the server (if the same version is not already installed) instead of pushing it (the same way I do for [gnirehtet](https://github.com/Genymobile/gnirehtet)). As a side effect, this will solve this issue.

The real goal is to install a real Android app. This will pave the way to:
 - provide an IME (to inject non-ASCII text)
 - forward audio (a proper integration of what [sndcpy](https://github.com/rom1v/sndcpy) does)
 - optionally mirror camera instead of the screen (which requires a user authorization from an app)
 - …

npes87184:
> For information, I started to work on a new option (probably `--install`) which would install the server (if the same version is not already installed) instead of pushing it (the same way I do for [gnirehtet](https://github.com/Genymobile/gnirehtet)). As a side effect, this will solve this issue.
> 
> The goal is to install a real Android app. This will pave the way to:
> 
> * provide an IME (to inject non-ASCII text)
> * forward audio (a proper integration of what [sndcpy](https://github.com/rom1v/sndcpy) does)
> * optionally mirror camera instead of the screen (which requires a user authorization from an app)
> * …

I am excited about this, things like #1484 can also be solved by provding apk. For example, we can add a notification or toast directly on phone.

quyleanh:
@rom1v how about progress of scrcpy's IME?

rom1v:
Unfortunately, no. For now, I'm working on #40 to add asynchronous swscale, and I'm doing more refactors (so that in the end I will be able to split the core from the UI).

[...]

Issue #1722, opened 2020-08-30
Keep server apk installed instead of pushing each time


psychowood:
Hi,
I tried searching for a similar feature request without success, hopefully it's not a duplicate

**Is your feature request related to a problem? Please describe.**
I start & stop scrcpy very frequently, and I was wondering if this could lead to useless flash memory degradation on my android device. 
From my understanding, the file is pushed to [/data/local/tmp/scrcpy-server.jar](https://github.com/Genymobile/scrcpy/blob/521f2fe994019065e938aa1a54b56b4f10a4ac4a/app/src/server.c#L22)

**Describe the solution you'd like**
Option to persist the server instead of pushing and deleting it each time. Maybe comparing the shasum with something like

`adb shell sha2sum  /data/local/tmp/scrcpy-server.jar`

before pushing.

**Describe alternatives you've considered**
Always pushing to a tmpfs to avoid writing on persistent storage.

rom1v:
> From my understanding, the file is pushed to /data/local/tmp/scrcpy-server.jar

Correct.

The reason why it is always push is that:
 - it's very fast
 - it avoids a mechanism to check if the installed version is up-to-date
 - checking would actually take as much time as just pushing it (and check+push takes more time)

I didn't thing about flash memory degradation though. But a priori, I would say that the write frequency is very low (it's not like a log file which could be written several times per second).

I'm not against an option to avoid it.

psychowood:
What about the alternative, pushing to ramfs/tmpfs? I don't know Android well enough, is there a standard/default path that can be used in alternative?

rom1v:
```
$ adb shell mount | grep tmpfs
tmpfs /dev tmpfs rw,seclabel,nosuid,relatime,mode=755 0 0
none /sys/fs/cgroup tmpfs rw,seclabel,relatime,mode=750,gid=1000 0 0
tmpfs /mnt tmpfs rw,seclabel,relatime,mode=755,gid=1000 0 0
tmpfs /storage tmpfs rw,seclabel,relatime,mode=755,gid=1000 0 0
```

`shell` do not have permission to write to these directories.

rom1v:
https://github.com/Genymobile/scrcpy/issues/1880#issuecomment-725655830

rom1v:
Please check #3517.

PR #3517, opened 2022-10-02
[Draft] Add option to install the server as an APK

rom1v:
To get shell permissions, the server must execute a [java main()](https://blog.rom1v.com/2018/03/introducing-scrcpy/#run-a-java-main-on-android) executable from `adb shell`.

The jar/apk containing the executable can be either pushed to any location (`/data/local/tmp`), or installed on the device.

Since the [beginning of the project](https://blog.rom1v.com/2018/03/introducing-scrcpy/#quick-installation), scrcpy just pushed the server to `/data/local/tmp`, because it's faster and simpler.

However, scrcpy might benefit from a companion Android app for supporting new features:
 - provide an IME (to inject non-ASCII text);
 - forward audio (a proper integration of what [sndcpy](https://github.com/rom1v/sndcpy) does);
 - optionally mirror camera instead of the screen (which requires a user authorization from an app)…

Instead of building a separate Android app to manage (install/reinstall), this PR makes possible to install the server directly as an APK (and execute the server from there):

```
scrcpy --install
scrcpy --reinstall
```

This paves the way to easily integrate Android component and start activities or services embedded in the server.

As a consequence, the server is now named `scrcpy-server.apk` (which is a drawback IMO, since people will want to install it manually instead of letting the client manage it) and is apk-signed so that it can be installed.

Refs https://github.com/Genymobile/scrcpy/issues/1722
Refs https://github.com/Genymobile/scrcpy/issues/1880#issuecomment-725655830

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -188,7 +188,7 @@ conf.set_quoted('SCRCPY_VERSION', meson.project_version())
 # the prefix used during configuration (meson --prefix=PREFIX)
 conf.set_quoted('PREFIX', get_option('prefix'))
 
-# build a "portable" version (with scrcpy-server accessible from the same
+# build a "portable" version (with scrcpy-server.apk accessible from the same
 # directory as the executable)
 conf.set('PORTABLE', get_option('portable'))
 
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -111,6 +111,10 @@ However, the option is only available when the HID keyboard is enabled (or a phy
 Also see \fB\-\-hid\-mouse\fR.
 
 .TP
+.B \-\-install
+Install the server (via "adb install") rather than pushing it to /data/local/tmp (via "adb push").
+
+.TP
 .B \-\-legacy\-paste
 Inject computer clipboard text as a sequence of key events on Ctrl+v (like MOD+Shift+v).
 
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -247,6 +247,10 @@ option if set, or by the file extension (.mp4 or .mkv).
 Force recording format (either mp4 or mkv).
 
 .TP
+.B \-\-reinstall
+Reinstall the server (via "adb install"), even if the correct version is already installed. Implies \fB\-\-install\fR.
+
+.TP
 .BI "\-\-render\-driver " name
 Request SDL to use the given render driver (this is just a hint).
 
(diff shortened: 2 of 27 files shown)


rom1v:
@yangfl This might impact Debian packaging: with this PR, the APK is signed. I updated `build_without_gradle.sh` do perform these tasks manually (without gradle). Please let me know if it's ok for you (but it's not urgent).

yangfl:
It's ok, since we can ask users to download and install apk from android side, or just sign it with a random key.

rom1v:
I am a bit worried about security implications though.

If someone manage to make you install a malicious app having package `com.genymobile.scrcpy`, then the next time the scrcpy client is executed, it will execute the java "executable" embedded in this package **with shell permissions**. This is a form of privilege escalation.

Maybe the server (executed by `adb shell`) and the companion APK to install should be a separate APK in the end. That would prevent this problem (no code from the APK could be executed with shell permissions).

When the app is pushed to `/data/local/tmp`, this is ok because no app can write to `/data/local/tmp/` (to replace the server just after it is pushed but before it is executed).

yangfl:
I think we can check signature of preinstalled apk and stop (or at least warn users and let them choose) running of unknown apk.

Maybe checking hash of apk file is a better choice, but I can't find how to do it...

quyleanh:
Sorry for asking but is there any progress of this PR?

rom1v:
> Sorry for asking but is there any progress of this PR?

Due to https://github.com/Genymobile/scrcpy/pull/3517#issuecomment-1265000612, I think it would be better to use a companion app instead. This PR is a good base but it is not intended to be merged as is.

quyleanh:
Thank you for explaining.
It's sad when an other approach of injecting non-ASCII text goes nowhere...

rom1v:
> It's sad when an other approach of injecting non-ASCII text goes nowhere...

[HID keyboard](https://github.com/Genymobile/scrcpy#physical-keyboard-simulation-hid) (over USB) shouldnwork pretty well.

quyleanh:
> [HID keyboard](https://github.com/Genymobile/scrcpy#physical-keyboard-simulation-hid) (over USB) should work pretty well.

Actually this feature doesn't fit my need. I would like a faster switch for mouse and keyboard.
Sometimes it also cannot detect my keyboard hardware.

rom1v:
Now, there is audio, camera and HID without the need for an app. Closing.

Issue #3522, opened 2022-10-07
Add bash completion script to scrcpy installation (script included here)


animaone:
- [ x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
Related to the completion of device id for when you are using multiple devices.

**Describe the solution you'd like**
When typing scrcpy -s + TAB_KEY there is no bash completion for the device id. There is already a bash script for adb. What I've done is copying the adb script and made some modifications to it.

**Describe alternatives you've considered**
The current alternative is copying devices ids manually or typing adb -s + TAB_KEY, then changing the adb command to scrcpy

**Additional context**
I've attached the script. Just EXTRACT IT and copy it to /etc/bash_completion.d/ and enjoy!
[scrcpy.tar.gz](https://github.com/Genymobile/scrcpy/files/9729338/scrcpy.tar.gz)
It would be very nice if this script(or a similar/cleaner version) could be installed automatically with the scrcpy installation.

rom1v:
scrcpy already has [bash completion](https://github.com/Genymobile/scrcpy/blob/master/app/data/bash-completion/scrcpy) and [zsh completion](https://github.com/Genymobile/scrcpy/blob/master/app/data/zsh-completion/_scrcpy) scripts.

I agree that they should be improved so that the `-s` option is completed by the available devices exposed by `adb devices`.

animaone:
Thanks. I've made a pull request with the modification in the default bash_completion script.
https://github.com/Genymobile/scrcpy/pull/3523

PR #3523, opened 2022-10-07
Add -s auto-complete option

animaone:
Add -s auto-complete option.

Commented at the following issue: https://github.com/Genymobile/scrcpy/issues/3522

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -95,7 +95,7 @@ _scrcpy() {
             ;;
         -s|--serial)
             # Use 'adb devices' to list serial numbers.
-            COMPREPLY=($(compgen -W "$(adb devices | awk '/(device|recovery|sideload)$/{print $1}')" -- ${cur} ) )
+            COMPREPLY=($(compgen -W "$(adb devices | awk '/device$/{print $1}')" -- ${cur} ) )
             return
             ;;
         -b|--bitrate \
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -93,6 +93,11 @@ _scrcpy() {
             COMPREPLY=($(compgen -W 'verbose debug info warn error' -- "$cur"))
             return
             ;;
+        -s|--serial)
+            # Use 'adb devices' to list serial numbers.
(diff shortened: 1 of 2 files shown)


app/data/bash-completion/scrcpy
```diff
@@ -93,6 +93,11 @@ _scrcpy() {
             COMPREPLY=($(compgen -W 'verbose debug info warn error' -- "$cur"))
             return
             ;;
+        -s|--serial)
+            # Use 'adb devices' to list serial numbers.
+            COMPREPLY=($(compgen -W "$(adb devices | awk '/(device|recovery|sideload)$/{print $1}')" -- ${cur} ) )
```
rom1v, line 98: I think that scrcpy may not work in recovery or sideload state, so maybe `adb devices |awk '/device$/{print $1}'` is sufficient.

rom1v (commented):
Awesome, thank you. :+1:

Could you also add it to the zsh completion script?

app/data/bash-completion/scrcpy
```diff
@@ -95,7 +95,7 @@ _scrcpy() {
             ;;
         -s|--serial)
             # Use 'adb devices' to list serial numbers.
-            COMPREPLY=($(compgen -W "$(adb devices | awk '/(device|recovery|sideload)$/{print $1}')" -- ${cur} ) )
+            COMPREPLY=($(compgen -W "$(adb devices | awk '/device$/{print $1}')" -- ${cur} ) )
```
rom1v, line 98: Also, if the `ADB` environment variable is set, this adb should be used instead.

```bash
"${ADB:-adb}" devices | awk …
```

animaone:
Ok. I've added a similar solution to the zsh completion. :)

rom1v:
Thank you :+1:

I rebased your commits onto `dev`, and made the following changes:
 - handle the case where the `ADB` environment variable is set
 - slightly adapt the `awk` command: `awk '$2 == "device" {print $1}'`
 - use the same `awk` command for zsh completion

I pushed to branch [`pr3523`](https://github.com/Genymobile/scrcpy/commits/pr3523). Could you please check?
If that's ok, I'll merge into `dev`.

animaone:
Checked. Many thanks +1    :)

rom1v:
Thanks :+1:

Merged into `dev`. :rocket:

Issue #3458, opened 2022-08-19
When you drag a window under windows, the screen will stop playing


cnuicn:
windows 下拖动窗口的时候，画面会停止播放
When you drag a window under windows, the screen will stop playing

rom1v:
During a resize or a move?

This is probably a SDL issue, but scrcpy uses a workaround:

https://github.com/Genymobile/scrcpy/blob/ed84e18b1ae3e51d368f8c7bc88ba4db088e6855/app/src/screen.c#L310-L320

So both resizing and moving should not freeze the mirroring.

Which scrcpy version do you use? Which windows version?

cnuicn:
用鼠标拖动和移动窗口的时候，画面都会冻结
When you drag and move the window with the mouse, the screen will freeze
scrcpy1.24 x64     windows11 x64

rom1v:
I have no access to a Windows machine right now. Could someone on Windows could reproduce/confirm?

rom1v:
Thank you. Now that I have access to a Windows machine, I can reproduce too, even while resizing.

I thought the workaround for this problem in macOS during resize was also effective for Windows.

It looks like a SDL issue, but I have no workaround for this, I can't make the screen refresh even if I call the code to render from the resize/move callback :/

rom1v:
It's a bug in SDL. SDL3 might fix the issue, when it's released.

Migrating to Qt is not a plan currently.

yume-chan:
According to https://github.com/libsdl-org/SDL/issues/1059#issuecomment-1842148648 and https://github.com/libsdl-org/SDL/commit/509c70c6982b6927f5a8d4fb32f9319cbaf0c2ef, it should listen to the `SDL_WINDOWEVENT_EXPOSED` event.

But since normal event loop is paused, `SC_EVENT_NEW_FRAME` events are not handled, it will keep rendering the same frame.

Something like this will work, but I ignored `screen->fb->mutex`:

```diff
diff --git a/app/src/screen.c b/app/src/screen.c
index 1d694f12..b5e32130 100644
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -230,6 +230,9 @@ sc_screen_render_novideo(struct sc_screen *screen) {
 #endif
 
 #ifdef CONTINUOUS_RESIZING_WORKAROUND
+static bool
+sc_screen_update_frame(struct sc_screen *screen);
+
 // On Windows and MacOS, resizing blocks the event loop, so resizing events are
[...]

PR #3547, opened 2022-10-23
Support wchar_t in argv in windows

### LINKED ISSUE #2932: Garbled characters appear when setting Chinese title
- [ ] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [ ] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Windows
 - scrcpy version: 1.21
 - installation method: Windows release
 - device model: Pixel 2 XL
 - Android version: 11

**Describe the bug**
Garbled characters appear when setting Chinese title

微信截图_20220110093928
微信截图_20220110094428

PS D:\soft\scrcpy-win64-v1.21>  .\scrcpy.exe -s 806KPSL1950254 --window-title 中文标题hello
scrcpy 1.21 <https://github.com/Genymobile/scrcpy>
D:\soft\scrcpy-win64-v1.21\scrcpy-server: 1 file pushed, 0 skipped. 63.2 MB/s (40067 bytes in 0.001s)
[server] INFO: Device: Google Pixel 2 XL (Android 11)
INFO: Renderer: direct3d
INFO: Initial texture: 1440x2880

npes87184:
This MR will fix #2932.

There are various approaches that can be used to address this issue, such as the usage of `wmain` with link flag `-municode`. Only the applied method, however, was successful.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -4,6 +4,10 @@
 #include <stdbool.h>
 #include <unistd.h>
 #include <libavformat/avformat.h>
+#ifdef _WIN32
+#include <windows.h>
+#include "util/str.h"
+#endif
 #ifdef HAVE_V4L2
 # include <libavdevice/avdevice.h>
 #endif
@@ -18,8 +22,8 @@
 #include "version.h"
 
 int
-main(int argc, char *argv[]) {
-#ifdef __WINDOWS__
+main_scrcpy(int argc, char *argv[]) {
+#ifdef _WIN32
     // disable buffering, we want logs immediately
     // even line buffering (setvbuf() with mode _IOLBF) is not sufficient
     setbuf(stdout, NULL);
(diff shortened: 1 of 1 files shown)


rom1v:
Awesome, thank you :+1:

I will test soon.

app/src/main.c
```diff
@@ -20,6 +24,17 @@
 int
 main(int argc, char *argv[]) {
 #ifdef __WINDOWS__
+    int wargc;
+    wchar_t **wargv = CommandLineToArgvW(GetCommandLineW(), &wargc);
+
+    if (argc != wargc) {
```
rom1v, line 30: Is it possible? Shouldn't it be an assertion?

npes87184: From my point of view, it wouldn't. However, it doesn't indicate it in the official document.

rom1v:
I think this is not correct to modify `argv` passed to main (it might be non-writable memory IIRC).

I added error handling, and to keep things "separate", I moved the code to a main "wrapper". That way, the "main" `main` (`main_scrcpy`) always receives `argc`/`argv` in UTF-8 as expected. This implementation does not even require `argc == wargc`, so I removed the check.

Here is a branch: [`pr3547`](https://github.com/Genymobile/scrcpy/commits/pr3547).

Please review and test if that still works (currently, I could not test on Windows).

npes87184:
I tested and it worked well.

However, I found that the elements in array `argv_utf8` forgotten to free.
I slightly added related logic, please check on it.

```c++
  for (int i = 0; i < wargc; ++i) {
      free(argv_utf8[i]);
  }
```

Also in the windows `main`, the `argc` does not use and it will cause compiler complains.

```
../app/src/main.c: In function ‘main’:
../app/src/main.c:89:10: warning: unused parameter ‘argc’ [-Wunused-parameter]
   89 | main(int argc, char *argv[]) {
      |      ~~~~^~~~
```

Additionally, I saw that you changed `__WINDOWS__` to `_WIN32` . I discovered that the code base still contains some of `__WINDOWS__`. Do you want them all taken out? I can change all of them if it is intentional.

rom1v:
> However, I found that the elements in array `argv_utf8` forgotten to free.

Good catch :+1:

> Also in the windows main, the `argc` does not use and it will cause compiler complains.

OK, I added:
```c
    (void) argc;
    (void) argv;
```

(yes, also `argv`, which was used mistakenly instead of `argv_utf8` in the first version)

Here is a branch: [`pr3547.2`](https://github.com/Genymobile/scrcpy/commits/pr3547.2).

> Additionally, I saw that you changed `__WINDOWS__` to `_WIN32`

Yes, `__WINDOWS__` is a [SDL define](https://github.com/LuaDist/sdl/blob/master/include/SDL_platform.h). Often, we can just use the "native" one `_WIN32`. Except maybe in `icon.c` which also uses `__APPLE__` (but the windows ifdef can probably be completely removed here, since the workaround seems not to work on Windows anyway #3458).

npes87184:
LGTM in pr3547.2 branch.

rom1v:
:+1: merged in `dev`: 48bb6f2ea858a02e05900d9828c0089166de17df

Issue #3497, opened 2022-09-13
Can no longer paste from Windows 10 to Android 13 (QPR Beta 1)


Icarus3000:
Device: Pixel 6 Pro
Android OS: 13 QPR Beta 1
Windows 10
Scrcpy 1.24

Terminal output when I press CTRL-V in scrcpy window:

```
[server] ERROR: Could not invoke method
[server] ERROR: Could not invoke method
java.lang.NoSuchMethodException: android.content.IClipboard$Stub$Proxy.getPrimaryClip [class java.lang.String, int]
        at java.lang.Class.getMethod(Class.java:2103)
        at java.lang.Class.getMethod(Class.java:1724)
        at com.genymobile.scrcpy.wrappers.ClipboardManager.getGetPrimaryClipMethod(ClipboardManager.java:28)
        at com.genymobile.scrcpy.wrappers.ClipboardManager.getText(ClipboardManager.java:63)
        at com.genymobile.scrcpy.Device.getClipboardText(Device.java:251)
        at com.genymobile.scrcpy.Device.setClipboardText(Device.java:264)
        at com.genymobile.scrcpy.Controller.setClipboard(Controller.java:300)
        at com.genymobile.scrcpy.Controller.handleEvent(Controller.java:124)
        at com.genymobile.scrcpy.Controller.control(Controller.java:75)
        at com.genymobile.scrcpy.Server$3.run(Server.java:133)
        at java.lang.Thread.run(Thread.java:1012)
java.lang.NoSuchMethodException: android.content.IClipboard$Stub$Proxy.setPrimaryClip [class android.content.ClipData, class java.lang.String, int]
        at java.lang.Class.getMethod(Class.java:2103)
        at java.lang.Class.getMethod(Class.java:1724)
        at com.genymobile.scrcpy.wrappers.ClipboardManager.getSetPrimaryClipMethod(ClipboardManager.java:39)
        at com.genymobile.scrcpy.wrappers.ClipboardManager.setText(ClipboardManager.java:77)
        at com.genymobile.scrcpy.Device.setClipboardText(Device.java:274)
        at com.genymobile.scrcpy.Controller.setClipboard(Controller.java:300)
        at com.genymobile.scrcpy.Controller.handleEvent(Controller.java:124)
        at com.genymobile.scrcpy.Controller.control(Controller.java:75)
        at com.genymobile.scrcpy.Server$3.run(Server.java:133)
        at java.lang.Thread.run(Thread.java:1012)
```

[...]

PR #3579, opened 2022-11-13
Force mouse source when --forward-all-clicks

### LINKED ISSUE #3568: mouse right click ACTION_UP is lost

Existing issues:
#2711, similar to my experience with right-click
#3070, related, if this is implemented right, the hoover events can be delivered

**Environment**
 - OS: arch x86_64
 - scrcpy version: 1.24, rebuild server with extra event logging
 - installation method: pacman/yay
 - device model, os: 
  google pixel 6 pro, grapheneOS (android13)
  asus nexus7, android 6
  samsung SM-T510, android 11

Primary work on pixel, other devices used to make sure problem is not os/device specific

**Describe the bug**
```
scrcpy --max-fps=30  --forward-all-clicks --verbosity=verbose --display=2
```
When using scrcpy with aFreeRDP, the right and middle mouse button don't work as expected.
I don't use the hid (AOA). In case of pixel, mouse pointer appears on the primary display. In case of samsung tab, it shows up on the extra display. I have not found a way to control it.

**Investigation**
Initially I assumed that the problem is in the aFreeRDP and started by dumping all motion events delivered to dispatchTouchEvent and dispatchGenericMotionEvent.
The first impression: ACTION_UP is missing for right and middle click. The event is not delivered. But perhaps it is the app itself.
So, I took a physical mouse and captured events. Than captured events when delivered from scrcpy. 

Right click events from physical mouse as seen by the application, the attributes which are always the same are in const section.
```
dispatchTouchEvent:         ACTION_DOWN,           actionButton=0, id[0]=0, x[0]=0.0, y[0]=2257.9397,                buttonState=BUTTON_SECONDARY, eventTime=11857929, ventId=507666466 }
dispatchGenericMotionEvent: ACTION_BUTTON_PRESS,   actionButton=BUTTON_SECONDARY, id[0]=0, x[0]=0.0, y[0]=2257.9397, buttonState=BUTTON_SECONDARY, eventTime=11857929, ventId=608438892 }
dispatchGenericMotionEvent: ACTION_BUTTON_RELEASE, actionButton=BUTTON_SECONDARY, id[0]=0, x[0]=0.0, y[0]=2257.9397, buttonState=0,                eventTime=11858273, ventId=612919901 }
dispatchTouchEvent:         ACTION_UP,             actionButton=0, id[0]=0, x[0]=0.0, y[0]=2257.9397,                buttonState=0,                eventTime=11858273, ventId=76737027 }
dispatchGenericMotionEvent: ACTION_HOVER_MOVE,     actionButton=0, id[0]=0, x[0]=0.0, y[0]=2257.9397,                buttonState=0,                eventTime=11858273, ventId=655777164 }

const:
toolType[0]=TOOL_TYPE_MOUSE,
classification=NONE, metaState=0, flags=0x0, edgeFlags=0x0, pointerCount=1, historySize=0,
downTime=11857929, deviceId=6, source=0x2002, displayId=0,

```

Right click events when delivered from scrcpy, the attributes which are always the same are in const section.
In addition log contains scrcpy events just before injection point (Controller.java, injectTouch)
```
scrcpy  injectEvent: ACTION_DOWN, id[0]=0, x[0]=710.0, y[0]=1668.0, buttonState=BUTTON_SECONDARY, eventTime=14278106, deviceId=0,  source=0x2002, displayId=0, eventId=-1002372252 }
FreeRDP dispatchTou: ACTION_DOWN, id[0]=0, x[0]=710.0, y[0]=1668.0, buttonState=BUTTON_SECONDARY, eventTime=14278106, deviceId=-1, source=0x2002, displayId=0, eventId=-1002372252 }
scrcpy  injectEvent: ACTION_UP,   id[0]=0, x[0]=710.0, y[0]=1668.0, buttonState=0,                eventTime=14278149, deviceId=0,  source=0x1002, displayId=0, eventId=-687213207 }

const:
toolType[0]=TOOL_TYPE_FINGER,
actionButton=0,
classification=NONE, metaState=0, flags=0x0, edgeFlags=0x0, pointerCount=1, historySize=0,
downTime=14278106,
```

The highlights are:
- missing touch ACTION_UP in the aFreeRDP, but can see the injected touch event in server
- missing motion events (generic motion events) which is probably not the real problem, but would be nice for consistency. Not sure about the performance, the hoover is send on each move.
- inconsistent deviceId, probably not a problem
- toolType is FINGER instead of MOUSE, again not a deal breaker
- wrong source (FINGER instead of MOUSE) in ACTION_UP event injected by the scrcpy

When I hard-code source to 0x2002 in injectTouch, the ACTION_UP is delivered to the application.
But this is just to prove why android is swallowing the event.

rom1v:
Right click and middle click require the source device to be a mouse, not a touchscreen). Therefore, the source device was changed only when a button other than the primary button was pressed (see adc547fa6e8e6167cd9633a97d98de6665b8c23a).

However, this led to inconsistencies between the `ACTION_DOWN` when a secondary button is pressed (with a mouse as source device) and the matching `ACTION_UP` when the secondary button is released (with a touchscreen as source device, because then there is no button pressed).

To avoid the problem in all cases, force a mouse as source device when --forward-all-clicks is set. As a consequence, pinch-to-zoom using a virtual finger is disabled in this case.

Concretely, in --forward-all-clicks mode:
 - device source is set to `InputDevice.SOURCE_MOUSE`;
 - motion event `toolType` is set to `MotionEvent.TOOL_TYPE_MOUSE`;
 - pinch-to-zoom simulation is disabled.

Otherwise:
 - left-click mouse event simulates a first virtual finger;
 - pinch-to-zoom simulation uses a second virtual finger.

For all (virtual or not) finger events:
 - device source is set to `InputDevice.SOURCE_TOUCHSCREEN`;
 - motion event `toolType` is set to `MotionEvent.TOOL_TYPE_FINGER`.

Fixes #3568

TODO:
 - [X] test that nothing is broken on a computer with a touchscreen with `--forward-all-clicks`
 - [X] test that nothing is broken on a computer with a touchscreen without `--forward-all-clicks`


rom1v:
Given [this remark](https://github.com/Genymobile/scrcpy/issues/3568#issuecomment-1320384378) (we could use 2 mice for pinch-to-zoom), I adapted the implementation to also support pinch-to-zoom if `--forward-all-clicks` is enabled.

cc @paweljasinski

paweljasinski:
Here are the test results with --forward-all-clicks done on pixel 6 pro with stock android 13.

1. play store - unable to scroll up/down with mouse as a touch, but roller works - no worry 
2. aFreeRDP - pinch works
3. termux - pinch works
4. RD Client (microsoft) - the app does not implement pinch, in addition pressing ctrl triggers conflicting functionality
5. aRDP Free - pinch emulation doesn't work, pinch works with real touch

For 2,3,4 and 5 the right click works.

rom1v:
And for all of them, pinch work without `--forward-all-clicks` on this branch?

paweljasinski:
> And for all of them, pinch work without --forward-all-clicks on this branch?

1. yes
2. yes
3. yes*
4. same problems
5. yes

* you have to carefully pick the location where you initiate pinch. If the emulated finger gets over the gui keyboard the app interprets it as dragging over the keyboard. This is consistent with real finger touch.

rom1v:
Thank you for your feedback.

So it seems to me that this PR is acceptable:
 - it guarantees that primary and secondary clicks are consistent when `--forward-all-clicks` is set
 - in addition, pinch-to-zoom is not disabled even in `--forward-all-clicks` (by simulating two mice, even if it is not always handled as expected by the app, it's better than disabling it completely).

I'm not sure if it is good enough, or if we can do better. Maybe what you suggested [here](https://github.com/Genymobile/scrcpy/issues/3568#issuecomment-1314396383) could be better, but I'm a bit afraid of side effects (and complexity). What do you think?

paweljasinski:
During last testing session discovered problems with middle mouse button. Let me first figure out what is going on.

rom1v:
OK.

FYI, I will probably release a new version (1.25) soon, especially for #3497.
I plan to merge this one too (it improves the behavior, even if it's not perfect), unless there are major problems with it.
It could be changed later if necessary.

PR #3646, opened 2022-12-30
use of while loop in strncpy function

divyeshdhole:


rom1v:
No reason for this change, and it's not equivalent (undefined behavior).

PR #3673, opened 2023-01-12
add devices msgbox prompt

ncnnnnn:


rom1v:

PR #3697, opened 2023-01-28
sc_mutex_held should be guarded by NDEBUG

Mr-JingShi:


rom1v:
What's the issue? What is the compiler error message?

Mr-JingShi:
Sorry, I was wrong, I tested it, it was not a compiler issue.
I just think sc_mutex_held should be guarded by NDEBUG.

like:
sc_mutex_unlock
https://github.com/Genymobile/scrcpy/blob/87da1372380ebddb60e4d89cff9a251c866e21c5/app/src/util/thread.c#L70

PR #3703, opened 2023-01-29
Forward audio

yume-chan:
Related to #14

Android version support:

* Android 10 or older: Not supported. `com.android.shell` package doesn't have `RECORD_AUDIO` permission declared.
* Android 11: Using the "Share heap dump" hack to bypass Android foreground check. The dialog is closed after AudioRecord started. Plus the `FakePackageNameContext` hack to "fix" Android permission check.
* Android 12 or newer: Using another variant of `FakePackageNameContext` hack to "fix" Android permission check. No "Share heap dump" hack required.

New option:

`--no-forward-audio` on client side, `forward_audio` on server side. Forwarding is enabled by default.

Socket order:

video -> [control] -> [audio]

Control and audio sockets may be absent if `--no-control` or `--no-forward-audio` is specified, but the order is maintained.

Audio compression:

No compression used, so the audio stream can be easily played at client side. The bandwidth is 48000Hz x 16bits x 2 channels = 1536000 bits =192000 bytes = 192KB = 187.5KiB per second, so I don't think it needs compression (at least on wired connection).

Latency:

The latency is around 30ms, tested in a rhythm game. And in one hour of continue using, the latency is seemingly increasing. Maybe it still requires packet header and time-stamping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Its features include:
  - [physical keyboard simulation (HID)](#physical-keyboard-simulation-hid)
  - [physical mouse simulation (HID)](#physical-mouse-simulation-hid)
  - [OTG mode](#otg)
+ - Audio forwarding
  - and more…
 
 ## Requirements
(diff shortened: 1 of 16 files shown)


yume-chan:
Replaced by #3757

PR #3704, opened 2023-01-29
Fix chrome when --forward-all-clicks is enabled

### LINKED ISSUE #3635: 1.25 forward-all-clicks doesn't work in Chrome web content

**Environment**
 - OS: Windows 11
 - scrcpy version: 1.25
 - installation method: Windows release
 - device model: XiaoMi Mi 11
 - Android version: 12

**Describe the bug**

When using `--forward-all-clicks` argument, mouse clicks doesn't work in Chrome web content area. It works fine in Chrome's own UI (address bar, menu, etc.) and other apps.

Comparing the events with a real OTG mouse, Scrcpy is missing events with `MotionEvent.ACTION_BUTTON_PRESS` and `MotionEvent.ACTION_BUTTON_RELEASE` actions. Normally they are delivered between `MotionEvent.ACTION_DOWN` and `MotionEvent.ACTION_UP`.

image

In Chromium source code, it ignores `MotionEvent.ACTION_DOWN` and `MotionEvent.ACTION_UP` events when `toolType` is set to `MotionEvent.TOOL_TYPE_MOUSE` (changed in Scrcpy 1.25):

https://source.chromium.org/chromium/chromium/src/+/main:ui/android/java/src/org/chromium/ui/base/EventForwarder.java;l=323-331;drc=3decef66bc4c08b142a19db9628e9efe68973e64

```
        // For mousedown and mouseup events, we use ACTION_BUTTON_PRESS
        // and ACTION_BUTTON_RELEASE respectively because they provide
        // info about the changed-button.
        if (eventAction == MotionEvent.ACTION_DOWN || eventAction == MotionEvent.ACTION_UP) {
            // While we use the action buttons for the changed state it is important to still
            // consume the down/up events to get the complete stream for a drag gesture, which
            // is provided using ACTION_MOVE touch events.
            return true;
        }
```

rom1v:
This is a mix of two branches made by @yume-chan to fix #3635:
 - https://github.com/Genymobile/scrcpy/issues/3635#issuecomment-1371833716
 - https://github.com/Genymobile/scrcpy/issues/3635#issuecomment-1375001124

---

@yume-chan Thank you, great work :+1:

On `fix-chrome-mouse` (the branch with events synthesized on the server), I like that the client don't have to handle the Android-specific sequence `ACTION_DOWN`, `ACTION_BUTTON_PRESS`, `ACTION_BUTTON_RELEASE`, `ACTION_UP` (IMO, this should be done on the server).

However, on `fix-chrome-mouse-v2` (the branch with events synthesized on the client), I like that the `action_button` forwarded by the client to the server (since it provided by the client, it's better than let the server "guess" using a diff of buttons state).

This is a mix of these two branches.

It works fine on my device. Please review and test before I merge.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Controller.java
@@ -1,5 +1,7 @@
 package com.genymobile.scrcpy;
 
+import com.genymobile.scrcpy.wrappers.InputManager;
+
 import android.os.Build;
 import android.os.SystemClock;
 import android.view.InputDevice;
@@ -99,7 +101,7 @@ public class Controller {
                 break;
             case ControlMessage.TYPE_INJECT_TOUCH_EVENT:
                 if (device.supportsInputEvents()) {
-                    injectTouch(msg.getAction(), msg.getPointerId(), msg.getPosition(), msg.getPressure(), msg.getButtons());
+                    injectTouch(msg.getAction(), msg.getPointerId(), msg.getPosition(), msg.getPressure(), msg.getActionButton(), msg.getButtons());
                 }
                 break;
             case ControlMessage.TYPE_INJECT_SCROLL_EVENT:
@@ -179,7 +181,7 @@ public class Controller {
         return successCount;
     }
 
-    private boolean injectTouch(int action, long pointerId, Position position, float pressure, int buttons) {
+    private boolean injectTouch(int action, long pointerId, Position position, float pressure, int actionButton, int buttons) {
         long now = SystemClock.uptimeMillis();
 
         Point point = device.getPhysicalPoint(position);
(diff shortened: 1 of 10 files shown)


rom1v:
[`ACTION_BUTTON_PRESS`](https://developer.android.com/reference/android/view/MotionEvent#ACTION_BUTTON_PRESS) and [`ACTION_BUTTON_RELEASE`](https://developer.android.com/reference/android/view/MotionEvent#ACTION_BUTTON_RELEASE) have been added in API 23.

So we should probably add:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/Controller.java b/server/src/main/java/com/genymobile/scrcpy/Controller.java
index 6c5fd573..ff78d998 100644
--- a/server/src/main/java/com/genymobile/scrcpy/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Controller.java
@@ -228,7 +228,7 @@ public class Controller {
             }
         }
 
-        /* If the input device is a mouse:
+        /* If the input device is a mouse (on API >= 23):
          *   - the first button pressed must first generate ACTION_DOWN;
          *   - all button pressed (including the first one) must generate ACTION_BUTTON_PRESS;
          *   - all button released (including the last one) must generate ACTION_BUTTON_RELEASE;
@@ -236,7 +236,7 @@ public class Controller {
          *
          * Otherwise, Chrome does not work properly: <https://github.com/Genymobile/scrcpy/issues/3635>
          */
-        if (source == InputDevice.SOURCE_MOUSE) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && source == InputDevice.SOURCE_MOUSE) {
             int previousButtons = pointer.getButtons();
             pointer.setButtons(buttons);
 
```

yume-chan:
> [ACTION_BUTTON_PRESS](https://developer.android.com/reference/android/view/MotionEvent#ACTION_BUTTON_PRESS) and [ACTION_BUTTON_RELEASE](https://developer.android.com/reference/android/view/MotionEvent#ACTION_BUTTON_RELEASE) have been added in API 23.

Considering this, they should definitely be generated at server side.

I also like the idea to keep the client simple and avoid sending redundant information (as they can be inferred).

server/src/main/java/com/genymobile/scrcpy/Controller.java
```diff
@@ -225,6 +228,65 @@ private boolean injectTouch(int action, long pointerId, Position position, float
...
+         *   - the last button released must in addition generate ACTION_UP.
+         *
+         * Otherwise, Chrome does not work properly: <https://github.com/Genymobile/scrcpy/issues/3635>
+         */
+        if (source == InputDevice.SOURCE_MOUSE) {
+            int previousButtons = pointer.getButtons();
+            pointer.setButtons(buttons);
+
+            if (action == MotionEvent.ACTION_DOWN) {
+                if (previousButtons == 0) {
```
yume-chan, line 244: In my v2 branch I tested `buttons == actionButton` to check whether it's the first pressed button, so no need to save previous `buttons` in pointer.

rom1v: Good point, fixed :+1:

yume-chan:
Verified all mouse buttons work in Chrome on my Mi 11 running Android 12.

But I found the MIUI Photos App still doesn't respond to mouse clicks. It responds to real mouse, but none of 1.24, 1.25 and this fix. I guess it's less an issue than Chrome, but will do more experiments.

yume-chan:
image

Turns out that injected events always have deviceId = -1, and this App ignores those events when action is down (0).

Checked Android source code, it might be possible to use InputFilter for injecting events (which preserves device ID), but that's used for accessibility service, so I don't think it's a good idea.

In conclusion, there is nothing Scrcpy can do for this App.

rom1v:
Thank you for your feedback :+1:

Merged into `dev`.

PR #1709, opened 2020-08-26
WIP: Implement CI/CD for scrcpy

dylanmtaylor:

PR #3721, opened 2023-02-05
Add GitHub actions for scrcpy server

qmfrederik:
This PR adds a GitHub action which:
- Builds the latest version of the scrcpy server, using a Circle CI container
- Publishes that as a build artifact

This can be useful as CI (making sure the server compiles) and for folks that want to build the scrcpy server using CI infrastructure.

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -9,8 +9,8 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build
       run: |
-        ANDROID_BUILD_TOOLS="33.0.1" BUILD_DIR="./out" ./server/build_without_gradle.sh
+        ./gradlew -p server assembleRelease
     - uses: actions/upload-artifact@v2
       with:
         name: scrcpy
-        path: ./out
+        path: ./server/build/outputs/apk/release/
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -1,0 +1,16 @@
+on:
+  push:
+
+jobs:
+  server:
+    runs-on: ubuntu-latest
+    container: cimg/android:2022.12.1
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: |
+        ANDROID_BUILD_TOOLS="33.0.1" BUILD_DIR="./out" ./server/build_without_gradle.sh
+    - uses: actions/upload-artifact@v2
+      with:
+        name: scrcpy
+        path: ./out
--- a/server/build_without_gradle.sh
+++ b/server/build_without_gradle.sh
@@ -50,6 +50,7 @@ echo "Compiling java sources..."
 cd ../java
 javac -bootclasspath "$ANDROID_JAR" -cp "$CLASSES_DIR" -d "$CLASSES_DIR" \
     -source 1.8 -target 1.8 \
+    -encoding UTF-8 \
     com/genymobile/scrcpy/*.java \
     com/genymobile/scrcpy/wrappers/*.java
 


qmfrederik:
The action won't run because it's defined in a PR, but you can see it in action here: https://github.com/qmfrederik/scrcpy/actions/runs/4097118741

.github/workflows/server.yml
```diff
@@ -0,0 +1,16 @@
+on:
+  push:
+
+jobs:
+  server:
+    runs-on: ubuntu-latest
+    container: cimg/android:2022.12.1
```
rom1v, line 7: Who created this image? Can we get an image with the necessary to build both the server and the client? (sorry, I'm a newbie in this domain)

rom1v (commented):
Thank you for your work.

There were a previous PR for github actions (#1709), but there was a problem with `gradle check` for an unknown reason. Can it work with with MR?

Would it be possible to build everything (server, client on linux, client on Windows…), i.e. just running `./release.sh`? Or should it be done in separate jobs?

server/build_without_gradle.sh
```diff
@@ -50,6 +50,7 @@ echo "Compiling java sources..."
 cd ../java
 javac -bootclasspath "$ANDROID_JAR" -cp "$CLASSES_DIR" -d "$CLASSES_DIR" \
     -source 1.8 -target 1.8 \
+    -encoding UTF-8 \
```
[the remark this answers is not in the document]
qmfrederik: `build_without_gradle.sh` would fail when being invoked from within a vanilla Ubuntu container. I think this is because `C_LANG` and friends were set to ASCII.  Since this repository uses UTF-8, it seemed to make sense to enforce that here as well, and this fixed the issue.
The PR as it currently stands no longer calls `build_without_gradle` directly so this is not needed for this MR.  I can undo the change if you want, but I don't think it's wrong ;-).

.github/workflows/server.yml
qmfrederik: These images are generated and maintained by Circle CI, a provider of CI tooling & infrastructure.  I had a look at the container images which can be used for building Android apps, and this seemed to be one of the most maintained ones.  The source repo is here: https://github.com/CircleCI-Public/cimg-android.

That's a long way to say I think they are fairly trustworthy.  I tried to start from ubuntu:jammy and then install ADK from the Ubuntu package repositories, but the ADK that ships with Ubuntu is too old.  The other option is to start from a vanilla ubuntu:jammy image (or the vanilla image for your favorite distro) and manually install the ADK.  Or not run in a container at all, but directly on the GitHub actions VM, which ship with the ADK as well: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#android.

In general, I think you're better of using containers because it shields you from any changes on the underlying build machines.

For the app, I would recommend running the build for the app in a separate job in a separate container, as the build dependencies are quite different: the server requires the ADK, the app itself requires headers for libav*, libusb, mingw and a bunch of other dependencies.

I originally had a job which would build the client as well, but then decided against it as I didn't have the time to fully complete that work (done is better than perfect, I guess).

yume-chan: My dev container dockerfile is also based on the cimg/android image, just only contains the essential components for building Scrcpy: https://github.com/yume-chan/scrcpy-devcontainer/blob/9a7b20391574d25617241d07b75457c9f22dd1a2/.devcontainer/Dockerfile#L30-L51

yume-chan:
I think there should also be a task to run unit tests.

Maybe cache `~/.gradle` to speed up subsequent runs. Found gradle has an official GitHub action for that: https://github.com/gradle/gradle-build-action#caching

rom1v:
Implemented by #5306 (without docker).

PR #3758, opened 2023-02-27
Fix hardcode deviceId while inject touch event.

baker915:
AOSP:
<https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/android10-release/cmds/input/src/com/android/commands/input/Input.java#359>

fix touch events not working on some unity game

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Controller.java
@@ -7,6 +7,8 @@ import android.view.KeyCharacterMap;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
 
+import com.genymobile.scrcpy.wrappers.InputManager;
+
 import java.io.IOException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -226,10 +228,12 @@ public class Controller {
         }
 
         MotionEvent event = MotionEvent
-                .obtain(lastTouchDown, now, action, pointerCount, pointerProperties, pointerCoords, 0, buttons, 1f, 1f, DEFAULT_DEVICE_ID, 0, source,
+                .obtain(lastTouchDown, now, action, pointerCount, pointerProperties, pointerCoords, 0, buttons, 1f, 1f, InputManager.getInputDeviceId(source), 0, source,
                         0);
         return device.injectEvent(event, Device.INJECT_MODE_ASYNC);
     }
+
+
 
     private boolean injectScroll(Position position, float hScroll, float vScroll, int buttons) {
         long now = SystemClock.uptimeMillis();
--- a/server/src/main/java/com/genymobile/scrcpy/wrappers/InputManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/wrappers/InputManager.java
@@ -2,6 +2,7 @@ package com.genymobile.scrcpy.wrappers;
 
 import com.genymobile.scrcpy.Ln;
 
+import android.view.InputDevice;
 import android.view.InputEvent;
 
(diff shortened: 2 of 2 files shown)


rom1v:
Thank you for the PR.

This is the alternative mentioned in 40febf4a91f0d9c9d9e1a6390930639b3bf5a70e, so it's probably ok if it fixes the problem for some games.

Could you mention the game(s) that you tested which is (are) fixed by this change?

baker915:
Thanks for you reply

This is the download link of game i've tested: https://ugame.9game.cn/game/downloadGame?pack.cooperateModelId=235099&pack.id=47093428

yume-chan:
> This is the download link of game i've tested: https://ugame.9game.cn/game/downloadGame?pack.cooperateModelId=235099&pack.id=47093428

I didn't see any issue in this game with Scrcpy 1.25.

Plus I don't think it will work. In my touch event tester app the device id is still -1

image

This is because Android set device id of all injected events to -1, unless the event is from InputFilter (accessibility service).

https://cs.android.com/android/platform/superproject/+/master:frameworks/native/services/inputflinger/dispatcher/InputDispatcher.cpp;l=4197-4206;drc=f3f3a95030a6811f722aced141e15e29e2105d11

```cpp
    // For all injected events, set device id = VIRTUAL_KEYBOARD_ID. The only exception is events
    // that have gone through the InputFilter. If the event passed through the InputFilter, assign
    // the provided device id. If the InputFilter is accessibility, and it modifies or synthesizes
    // the injected event, it is responsible for setting POLICY_FLAG_INJECTED_FROM_ACCESSIBILITY.
    // For those events, we will set FLAG_IS_ACCESSIBILITY_EVENT to allow apps to distinguish them
    // from events that originate from actual hardware.
    int32_t resolvedDeviceId = VIRTUAL_KEYBOARD_ID;
    if (policyFlags & POLICY_FLAG_FILTERED) {
        resolvedDeviceId = event->getDeviceId();
    }
```

The only method to get real device id is using AOA HID or uinput.

rom1v:
@baker915 Did you test that the change proposed in this MR actually works for the game you mentioned?

baker915:
> @baker915 Did you test that the change proposed in this MR actually works for the game you mentioned?

I did,and i tested other apps,everything works fine.

rom1v:
OK so without this MR, it does not work, with this MR, it works? If so, I think this could be merged.

On which Android version?

rom1v:
> If so, I think this could be merged.

(But I'm a bit afraid that it could break other use cases…)

yume-chan:
> Is it possible that you provide me touch event tester app apk you metioned?

Source code: https://github.com/yume-chan/android-motion-event-tester

APK: [com.companyname.AndroidApp1.zip](https://github.com/Genymobile/scrcpy/files/10852991/com.companyname.AndroidApp1.zip)

Touch or click in the empty area, it will display properties of received events.

baker915:
> OK so without this MR, it does not work, with this MR, it works? If so, I think this could be merged.
> 
> On which Android version?

on Android 10

baker915:
> > If so, I think this could be merged.
> 
> (But I'm a bit afraid that it could break other use cases…)

Agreed!I'm going to test more devices and read the source code @yume-chan metioned to figure out this MR is necessary.

baker915:
> > Is it possible that you provide me touch event tester app apk you metioned?
> 
> Source code: https://github.com/yume-chan/android-motion-event-tester
> 
> APK: [com.companyname.AndroidApp1.zip](https://github.com/Genymobile/scrcpy/files/10852991/com.companyname.AndroidApp1.zip)
> 
> Touch or click in the empty area, it will display properties of received events.

Thanks. I will test it in my environment.

PR #3816, opened 2023-03-14
Update Gentoo instructions

voyageur:
scrcpy is available directly in the distro, drop link to the overlay (which only contains older versions)


rom1v:
Thank you, merged: 6ad037ea043debb1fb240912a8a4b639a5e2d96e :+1:

PR #3817, opened 2023-03-14
Fix linux desktop files validation

### LINKED ISSUE #3633: desktop-file-validate fails on provided .desktop files

 - scrcpy version: 1.25

During the packaging of scrcpy for Fedora COPR, I noticed both `scrcpy.desktop` and `scrcpy-console.desktop` do not pass the [desktop-file-validate](https://www.freedesktop.org/wiki/Software/desktop-file-utils/) validation process:

> scrcpy.desktop

```log
$ desktop-file-validate scrcpy.desktop
scrcpy.desktop: error: value "/bin/sh -c '"$SHELL" -i -c scrcpy'" for key "Exec" in group "Desktop Entry" contains a reserved character ''' outside of a quote
scrcpy.desktop: error: value "/bin/sh -c '"$SHELL" -i -c scrcpy'" for key "Exec" in group "Desktop Entry" contains a non-escaped character '$' in a quote, but it should be escaped with two backslashes ("\\$")
scrcpy.desktop: error: value "/bin/sh -c '"$SHELL" -i -c scrcpy'" for key "Exec" in group "Desktop Entry" contains a reserved character ''' outside of a quote
scrcpy.desktop: hint: value item "RemoteAccess" in key "Categories" in group "Desktop Entry" can be extended with another category among the following categories: Network
```

> scrcpy-console.desktop

```log
$ desktop-file-validate scrcpy-console.desktop 
scrcpy-console.desktop: error: value "/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'" for key "Exec" in group "Desktop Entry" contains a reserved character ''' outside of a quote
scrcpy-console.desktop: error: value "/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'" for key "Exec" in group "Desktop Entry" contains a non-escaped character '$' in a quote, but it should be escaped with two backslashes ("\\$")
scrcpy-console.desktop: error: value "/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'" for key "Exec" in group "Desktop Entry" contains a reserved character '|' outside of a quote
scrcpy-console.desktop: error: value "/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'" for key "Exec" in group "Desktop Entry" contains a reserved character '|' outside of a quote
scrcpy-console.desktop: error: value "/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'" for key "Exec" in group "Desktop Entry" contains a reserved character ''' outside of a quote
scrcpy-console.desktop: hint: value item "RemoteAccess" in key "Categories" in group "Desktop Entry" can be extended with another category among the following categories: Network
```

voyageur:
Follow quoting rules from:
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables Also use /bin/sh for scrpy-console

This fixes #3633 <https://github.com/Genymobile/scrcpy/issues/3633>
Downstream Gentoo bug: https://bugs.gentoo.org/888133

It adds a few backslashes but makes desktop-file-validate happy

### DIFF
--- a/app/data/scrcpy-console.desktop
+++ b/app/data/scrcpy-console.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'
+Exec=/bin/bash --norc --noprofile -i -c "\"\\$SHELL\" -i -c scrcpy || read -p 'Press any key to quit...'"
 Icon=scrcpy
 Terminal=true
 Type=Application
--- a/app/data/scrcpy.desktop
+++ b/app/data/scrcpy.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/sh -c '"$SHELL" -i -c scrcpy'
+Exec=/bin/sh -c "\"\\$SHELL\" -i -c scrcpy"
 Icon=scrcpy
 Terminal=false
 Type=Application


rom1v:
Thank you for the PR.

I'm ok to change the quotes so that it pass validation by `desktop-file-validate`, but the command for `scrcpy-console.desktop` use the bash-specific version of the built-in `read` function (and since it uses bash it also pass `--norc` and --noprofile`), so we cannot just change `bash` to `sh`.

I suggest:

```diff
diff --git a/app/data/scrcpy-console.desktop b/app/data/scrcpy-console.desktop
index 47a63ec99..f99217827 100644
--- a/app/data/scrcpy-console.desktop
+++ b/app/data/scrcpy-console.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/bash --norc --noprofile -i -c '"$SHELL" -i -c scrcpy || read -p "Press any key to quit..."'
+Exec=/bin/bash --norc --noprofile -i -c "\"\\$SHELL\" -i -c scrcpy || read -p 'Press any key to quit...'"
 Icon=scrcpy
 Terminal=true
 Type=Application
diff --git a/app/data/scrcpy.desktop b/app/data/scrcpy.desktop
index 082b75e0f..1be86a2ba 100644
--- a/app/data/scrcpy.desktop
+++ b/app/data/scrcpy.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/sh -c '"$SHELL" -i -c scrcpy'
+Exec=/bin/sh -c "\"\\$SHELL\" -i -c scrcpy"
 Icon=scrcpy
 Terminal=false
 Type=Application
```

What do you think?

voyageur:
Sorry for the noise, I now remember I switched back my /bin/sh to bash (from dash)... So of course using /bin/sh worked

Let me update the PR restoring bash in scrcpy-console

rom1v:
Thank you :+1:

Rebased on `dev` and merged: d2b7315ba693e11bd9b9a0421a75b7df8a31c5cc

sixg0000d:
Hello, do you check the `scrcpy-console.desktop` if works? I tried it (from dev branch) but it shows me didn't find `/bin/bash -i`. I suggest this one, it both works and pass the validation:

```
Exec=/bin/sh -i -c "\\$SHELL -i -c scrcpy || echo 'Press any key to quit...';read"
```

ratijas:
Hi,

I noticed this MR because after recent update I saw new warnings in terminal:

```
kf.config.core: "KConfigIni: In file /usr/share/applications/scrcpy-console.desktop, line 8: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/scrcpy-console.desktop, line 8: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/scrcpy.desktop, line 8: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/scrcpy.desktop, line 8: " "Invalid escape sequence \"\\\"\"."
```

On a closer inspection, those lines might be valid after all, and we have a bug in KConfig parser instead.

Nevertheless, these lines look a bit over-engineered, but also inconsistent at the same time. Why hardcoding /bin/sh in one .desktop file, /bin/bash in another, and then asking them to execute whatever the `\"\\$SHELL\"` is pointing to.


Besides, the whole comment indicates that it is a workaround broken user setup in the first place:

```
# For some users, the PATH or ADB environment variables are set from the shell
# startup file, like .bashrc or .zshrc… Run an interactive shell to get
# environment correctly initialized.
```

If user needs their environment variables be accessible from apps, they need to put it into appropriate .profile config. Expecting each and every app to "fix" it for them in gazzilion different creative ways isn't going to fly.

rom1v:
> Why hardcoding /bin/sh in one .desktop file, /bin/bash in another

Because in one case (`scrcpy.desktop`) we just execute `scrcpy`.
But in the other case, we execute `scrcpy`, and only if it fails we want to keep the terminal open (calling `read -p 'Press Enter to quit...'`). But this last part depends on the shell used, so we must force bash.

> If user needs their environment variables be accessible from apps, they need to put it into appropriate .profile config. 

That's true, but in practice, scrcpy is a command-line app, so intended to be started from a terminal.

The point of `.desktop` in that case is to open a terminal with their default configuration (e.g. `.bashrc`), so that the behavior is the same as when they open a terminal then type `scrcpy`.

I agree that this is not pretty. In a sense, it seems that the `.desktop` files are a workaround to use scrcpy "as if" it was not a command line app (but it is).

ratijas:
Meanwhile I sent a ["readability patch"](https://invent.kde.org/frameworks/kconfig/-/merge_requests/203) to KConfig and filed a [bug report about missing specialized Exec= parsing algorithm](https://bugs.kde.org/show_bug.cgi?id=471533) in KService/KDesktopFile.

rom1v:
Just to be sure, the current version of scrcpy causes the bug in KConfig, but the previous one was working fine?

ratijas:
Sure. The old variant didn't contain any escape sequences after all.

rom1v:
@ratijas Please review #4130, it should not fix your problem but at least is simplifies the `-console` versions.

PR #3865, opened 2023-03-27
Fix debug build by adding compat.c to tests

metayan:
Linking of tests that needed something from compat.c failed.


rom1v:
LGTM.

On which platform did you experience a build issue?

metayan:
It also failed linking `test_vecdeque` and `test_vector` if I recall correctly, so added a more general and future-proof solution.

Found it on macOS Catalina.

Issue #1824, opened 2020-10-10
Media control Shortcuts


MonDV:
I have been using vysor to control my FireTV devices and since they upgraded to v3 it's been a complete cluster****.  Found your tool yesterday and it does 99% of everything I need and it is GREAT!  

There is one issue/request.  Can you add media control shortcuts?  Specifically to rewind, play/pause, forward media.  For the FireTV the keycodes are:

KEYCODE_MEDIA_PLAY_PAUSE
KEYCODE_MEDIA_REWIND
KEYCODE_MEDIA_FAST_FORWARD

In vysor, I modify the settings so that my current laptop buttons for these 'functions' is linked to the keycode (currently F9, F10 and F11 buttons).  

It would be great to add shortcuts that would pass those keycodes so that it is possible to control media play.

FYI, I am not a programmer and I don't know the complexity of this request.  Best, MonDV (Thanks for the great tool!!!)

rom1v:
Just to test, this works:

```diff
diff --git a/app/src/event_converter.c b/app/src/event_converter.c
index ab48898d..dc88c239 100644
--- a/app/src/event_converter.c
+++ b/app/src/event_converter.c
@@ -96,6 +96,9 @@ convert_keycode(SDL_Keycode from, enum android_keycode *to, uint16_t mod,
         MAP(SDLK_RCTRL,        AKEYCODE_CTRL_RIGHT);
         MAP(SDLK_LSHIFT,       AKEYCODE_SHIFT_LEFT);
         MAP(SDLK_RSHIFT,       AKEYCODE_SHIFT_RIGHT);
+        MAP(SDLK_F9,           AKEYCODE_MEDIA_PLAY_PAUSE);
+        MAP(SDLK_F10,          AKEYCODE_MEDIA_REWIND);
+        MAP(SDLK_F11,          AKEYCODE_MEDIA_FAST_FORWARD);
     }
 
     if (!(mod & (KMOD_NUM | KMOD_SHIFT))) {
```

But there is no customizable mapping on scrcpy (and media keys are not correctly received by SDL, for example I never receive `SDLK_AUDIOPLAY` if I click on my mediakey PLAY).

hafizidev:
> Just to test, this works:
> 
> ```diff
> diff --git a/app/src/event_converter.c b/app/src/event_converter.c
> index ab48898d..dc88c239 100644
> --- a/app/src/event_converter.c
> +++ b/app/src/event_converter.c
[...]

PR #3877, opened 2023-03-29
Add Music Controls Shortcut (like pause, next song, play, previous song)

FF55100:
I've been using Scrcpy to play music on my phone while studying on my computer. Music helps me focus but my problem is that there's no keyboard shortcut for music controls. I would be happy if you would add music control shortcuts to Scrcpy.


rom1v:
You should open an issue, not a pull request.

And in fact there is already an existing issue: #1824

Gottox:
related: #4080

Issue #3475, opened 2022-08-29
Is it possible to optionally display the copied and pasted text in the console window?


mdell-seradex:
Currently the console displays the fact that text was copied/synchronized between the PC and the Android device.
I was wondering if an option could be added that would display that text in the console?

Another option would be for it to be logged to a file.

I have looked at the Readme and FAQ, and at the command line parameters, but did not see anything like this.

Thank you.

ssailly:
I just started working on this issue, and noticed there were two messages showing despite copying only once:
```
INFO: Device clipboard copied
DEBUG: Computer clipboard unchanged
```
However, computer clipboard is effectively changed, and this might correspond to two messages sent to the client. I will report it in a new issue and try to fix it.

PR #3914, opened 2023-04-13
Add new option to show clipboard updates in verbose mode

ssailly:
This implements the feature suggested by #3475.

When copying from the device, then copy-pasting from the computer to the device using <kbd>Ctrl</kbd>+<kbd>V</kbd>, and running with `-Vdebug --show-clipboard` :
* Before:
```
INFO: Device clipboard copied
DEBUG: Computer clipboard set

[server] INFO: Device clipboard set
DEBUG: Computer clipboard unchanged
```

* After:
```
INFO: Device clipboard copied: This application mirrors Android devices
DEBUG: Computer clipboard set: This application mirrors Android devices

[server] INFO: Device clipboard set: It works on Linux
DEBUG: Computer clipboard unchanged
```

Running without `--show-clipboard` has the same behavior as "Before".

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -72,6 +72,7 @@ enum {
     OPT_REQUIRE_AUDIO,
     OPT_AUDIO_BUFFER,
     OPT_AUDIO_OUTPUT_BUFFER,
+    OPT_SHOW_CLIPBOARD,
 };
 
 struct sc_option {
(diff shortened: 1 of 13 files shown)


rom1v:

Small remarks inline after a quick review.

rom1v:
By default, the clipboard content must not be printed (it may contain sensitive information).

However, I'm wondering if we could just log it in verbose mode (without adding yet another very-specific niche option).

Issue #3796, opened 2023-03-12
Not work audio streaming on android 11


bropines:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Windows 11
 - scrcpy version: scrcpy 2.0
 - installation method: Windows release...
 - device model: Redmi Note 8T
 - Android version: 11

**Describe the bug**
The sound just doesn't work. When you try to make it forcibly this happens

```
PS Programs\scrcpy-win64-v2.0> .\scrcpy-console.bat -d --require-audio
scrcpy 2.0 <https://github.com/Genymobile/scrcpy>
C:\Users\bropi\Programs\scrcpy-win64-v2.0\scrcpy-server: 1 file pushed, 0 skipped. 62.6 MB/s (52867 bytes in 0.001s)
[server] INFO: Device: Xiaomi Redmi Note 8T (Android 11)
INFO: Renderer: direct3d
INFO: Initial texture: 1080x2336
[server] ERROR: Failed to start audio capture
[server] ERROR: On Android 11, it is only possible to capture in foreground, make sure that the device is unlocked when starting scrcpy.
WARN: Demuxer 'audio': stream explicitly disabled by the device
ERROR: Demuxer error
WARN: Killing the server... ```


Please do not post screenshots of your terminal, just post the content as text instead.

rom1v:
Dis you follow the instructions specific to Android 11? https://github.com/Genymobile/scrcpy/blob/master/doc/audio.md

bropines:
> Do you follow the instructions specific to Android 11? https://github.com/Genymobile/scrcpy/blob/master/doc/audio.md

Yep. The screen is on, the developer settings do not block access to the sound. I also tried it with the `--require-audio` parameter the result is zero. It just won't start.

rom1v:
For testing, could you please replace the server in the scrcpy 2.0 release folder:

- [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/3796/1/scrcpy-server) <sub>`SHA-256: 68f9c478177e543a180ee9460eb63b6969c0501ef22280fe00fe51d4310fe80f`</sub>

It disables the workaround for Android 11:

<details><summary>diff</summary>

```diff
[...]

PR #3930, opened 2023-04-18
Update AudioCapture.java

PrantikBarik:
can you please update this? i cant create a new app. I am facing the " Demuxer 'audio': stream explicitly disabled by the device " problem


rom1v:
https://github.com/Genymobile/scrcpy/issues/3796

PR #3944, opened 2023-04-23
I see no reason why the rotation variable is an unsigned instead of a…

Sapientia-PT:
… uint8_t which saves memory


consp1racy:
you seem to have forgotten to update screen.c and the function call sites

rom1v:
Obsolete since #4441, there is no rotation variable anymore.

PR #3959, opened 2023-04-30
Minimal VNC server implementation

### LINKED ISSUE #3473: VNC server

It would be nice if scrcpy could provide a VNC server so things like vncdotool could be used with it

DavidVentura:
Fixes #3473.
I wanted to have a direct VNC server sharing my scrcpy screen -- this is a work in progress towards that.

Code-quality wise it's pretty bad; a lot of copy-pasting the screen/display and just chucked the example libvnc eventloop thread, but it does function!

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -109,7 +109,6 @@ if not crossbuild_windows
         dependency('libavutil'),
         dependency('libswresample'),
         dependency('sdl2', version: '>= 2.0.5'),
-        #cc.find_library('swscale-5', dirs: ffmpeg_bin_dir),
         dependency('libswscale'),
         dependency('libvncserver'),
     ]
--- a/app/meson.build
+++ b/app/meson.build
@@ -31,6 +31,7 @@ src = [
     'src/screen.c',
     'src/server.c',
     'src/version.c',
+    'src/vnc_sink.c',
     'src/trait/frame_source.c',
     'src/trait/packet_source.c',
     'src/util/acksync.c',
@@ -108,6 +109,9 @@ if not crossbuild_windows
         dependency('libavutil'),
         dependency('libswresample'),
         dependency('sdl2', version: '>= 2.0.5'),
+        #cc.find_library('swscale-5', dirs: ffmpeg_bin_dir),
+        dependency('libswscale'),
+        dependency('libvncserver'),
     ]
 
     if v4l2_support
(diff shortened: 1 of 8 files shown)


rom1v:
Thank you for the prototype :+1: It's interesting. I keep it in some branch for now.

I'm not sure whether I want to add a VNC server directly in scrcpy (it is a "side" feature, and it typically does not provide the same performances).

I think it might require scrcpy to be a bit more "modular", so that we can compile independent "plugins" (I'm not very happy with the V4L2 sink and its bunch of ifdefs too). But on the other side, I don't want it to be as modular as VLC for example (this would require too much decoupling which add many constraints, that in practice prevent to implement some "simple" features because the abstractions need to be "too general"). An alternative would be to expose a libscrcpy and use this to create a VNC server binary.

These are just random thoughts, I keep the feature/need in mind. Thank you.

DavidVentura:
@rp1231 a VNC server is an implementation of the "remote-framebuffer" protocol, in this case, it allows you to display and interact with the android device from another computer, not necessarily the one that's running scrcpy (and not the one where the device is plugged in).

@rom1v it's not the _same_ performance, but it is pretty good, it feels "the same" when using a device for general purposes (did not try games / video).
It could be made more efficient in certain scenarios; as I didn't figure out a way to detect the "damage" in the video frames, every time something changes, the entire framebuffer is sent.

A libscrcpy would be pretty cool, and yeah, I did feel adding more ifdefs isn't really great. I'll think a bit for a way to refactor these.
I'm fine keeping the PR open - I have my own build so I'm covered

DavidVentura:
@rp1231 I have a device that I need to access sporadically. I've left it connected to my server and can use VNC to access / interact with it via my laptop.

DavidVentura:
@rom1v I had something like this in mind
https://github.com/DavidVentura/scrcpy/pull/1/files
which can be extended to cover also V4L2 -- mostly it moves all the `ifdef`'d state into a `plugin`, which bypasses _most_ of the `ifdef`s by having the plugin-specific data as an opaque `void*`

mudebug77:
Thank you for sharing, this is a great suggestion. With VNC, it can be better expanded.

AndrewO:
I'd love to see this feature. However, I see that [`libvncserver` is GPL2 licensed](https://github.com/LibVNC/libvncserver/blob/master/COPYING) so, correct me if I'm wrong, I don't think the Apache v2 licensed server code can just include it this way, right?

PR #4005, opened 2023-05-16
Installing on Windows with winget

TrangOul:


doc/windows.md
```diff
@@ -15,7 +15,14 @@ Download the [latest release]:
 
 and extract it.
 
-Alternatively, you could install it from packages manager, like [Chocolatey]:
+Alternatively, you could install it from packages manager, like [Winget]:
+
+```bash
+winget install scrcpy
+winget install adb    # if you don't have it yet
```
rom1v, line 22: Is `adb` available via Winget? I did not found it in https://github.com/microsoft/winget-pkgs

yume-chan: `winget install scrcpy` should be enough, it uses the official Windows release which already includes adb.exe.

adb does exist in winget packages, but can only be installed by its ID:

```sh
winget install Google.PlatformTools
```

rom1v: OK, so I will just remove the line about adb.

Do you know if it is necessary to install adb separately with scoop and chocolatey package managers, or if the scrcpy package is sufficient?

yume-chan: I don't know, I don't use other package managers.

rom1v:
Thank you, merged: 958f22490b3ab86677bd8b6126ddaa4de4082798

(I removed the line about `adb`)

TrangOul:
My bad about `adb`, thanks for correcting that!

Issue #4028, opened 2023-05-23
Regarding the necessity of killing the adb service in the otg mode on Windows platform


HWZen:
- [ yes ] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Regarding the necessity of killing the adb service in the otg mode on Windows platform**
We can not use HID and mirror mode on Windows platform beacuse Windows not possible to open a USB device by more than one process, and mirror mode will invoke adb server to connect device.  Adb will open our Android USB device.

But really ?

In fact, I managed to use the HID at the same time with the mirroring screen turned on successfully, on Windows platform, used scrcpy.

The key to the way is to use the remote debugging feature of adb, so that adb does not need to use USB to connect the device

First, we open a scrcpy

```script.exe --otg -K```

I will kill adb server and conncet to device, simulate HID input.

Then, we open another scrcpy, use TCP/IP  to connect the device

```script.exe --tcpip=192.168.12.34:5678```

If all goes well, you can now have HID input while mirroring is enabled

![20230523202813](https://github.com/Genymobile/scrcpy/assets/46065109/256163d1-d3b2-4316-bf36-4aedb431d3dd)

**My soluction: try connect to device by usb first?**
My mind is, on OTG mode, try to make a connection to the device first instead of killing adb first, If the connection fails, it is not too late to try to kill adb.

On mirror mode, if user use tcpip to connect to device, maybe we can allow simultaneous HID input from the user.

**I am proud to contribute code**
If you think what I'm saying makes sense, I'd be happy to contribute code to complete this functionality. <(￣︶￣)↗[GO!]

WuDi-ZhanShen:
I think there is a better way to solve this problem, see https://github.com/Genymobile/scrcpy/issues/4034

HWZen:
> I think there is a better way to solve this problem, see #4034

eh~This does seem like a more elegant solution. But are you sure this can be done via ADB without installing any app on the device? Sorry I don't know much about how to call Linux Api on Android devices

rom1v:
Hi,

Please keep talking in English ;)

> My mind is, on OTG mode, try to make a connection to the device first instead of killing adb first, If the connection fails, it is not too late to try to kill adb.

Maybe adding an option (`--no-kill-adb-before-hid`, specific to Windows), as suggested in #4032, would be simpler. What do you think?

[...]

Issue #4032, opened 2023-05-24
Add an option to keep the adb daemon alive in OTG mode


wh201906:
- [x] ~~I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.~~

**Is your feature request related to a problem? Please describe.**
According to the [doc](https://github.com/Genymobile/scrcpy/blob/2d3059e1ab687a18123636784f1e98ac653d3469/doc/hid-otg.md#physical-keyboard-simulation), users cannot use keyboard simulation with mirroring over the same USB connection on Windows. To monitor the phone and control it at the same time, I need to use adb over TCP and keyboard simulation in OTG mode.
However, If I start scrcpy for monitoring first and start another instance in otg mode for keyboard simulation, the latter will kill the adb daemon. So I have to start the scrcpy in otg mode first, then start scrcpy for monitoring.

**Describe the solution you'd like**
To start the two instances mentioned above in any sequence, I'd like to add an option to prevent killing adb daemon in otg mode([code there](https://github.com/Genymobile/scrcpy/blob/2d3059e1ab687a18123636784f1e98ac653d3469/app/src/usb/scrcpy_otg.c#L83-L90)).

**Describe alternatives you've considered**
+ For users like me, start the instances in specified order.
+ Kill the adb daemon only if the target device is connected via USB.
+ #4028 (I haven't test it yet.)

**Additional context**
I know the code for killing the adb daemon is for accessing the device via otg. However, a running adb daemon doesn't means that the target device is connected via USB. I think adding an extra option is easy to be implemented without changing the current behavior. I'm also ready to implement this if you'd like to.

rom1v:
Thank you for your suggestion.

I'm closing to avoid several threads, the discussion continues here: https://github.com/Genymobile/scrcpy/issues/4028#issuecomment-1562390025

PR #4035, opened 2023-05-25
Add `--no-kill-adb-before-hid` option

wh201906:
Hi. This is for https://github.com/Genymobile/scrcpy/issues/4028#issuecomment-1562390025
I'm not sure if the option name is good or not.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -71,6 +71,7 @@ enum {
     OPT_LIST_DISPLAYS,
     OPT_REQUIRE_AUDIO,
     OPT_AUDIO_BUFFER,
+    OPT_NO_KILL_ADB_BEFORE_HID,
 };
 
 struct sc_option {
@@ -657,6 +658,13 @@ static const struct sc_option options[] = {
         .argdesc = "value",
         .text = "Set the initial window height.\n"
                 "Default is 0 (automatic).",
+    },
+    {
+        .longopt_id = OPT_NO_KILL_ADB_BEFORE_HID,
+        .longopt = "no-kill-adb-before-hid",
+        .text = "By default, scrcpy kills the adb daemon on Windows"
+                "if --otg is specified.\n"
+                "This option keeps the adb daemon open."
     },
 };
 
@@ -1831,6 +1839,8 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                     return false;
                 }
                 break;
+            case OPT_NO_KILL_ADB_BEFORE_HID:
+                opts->kill_adb_before_hid = false;
             default:
                 // getopt prints the error message on stderr
                 return false;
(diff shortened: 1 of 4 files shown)


rom1v:
Thank you. I made some changes/fixes, and renamed hid to otg: branch [`no_kill_adb_before_otg`](https://github.com/Genymobile/scrcpy/commits/no_kill_adb_before_otg).

However, after thinking about it, I think this option is useless. In #4032, you say:

> To monitor the phone and control it at the same time, I need to use adb over TCP and keyboard simulation in OTG mode.
However, If I start scrcpy for monitoring first and start another instance in otg mode for keyboard simulation, the latter will kill the adb daemon. So I have to start the scrcpy in otg mode first, then start scrcpy for monitoring.
> […]
> To start the two instances mentioned above in any sequence, I'd like to add an option to prevent killing adb daemon in otg mode

If you start the adb daemon first, and don't kill it, your device will not be detected in otg mode, so `scrcpy --otg` will fail. That's the reason why it kills adb (to make it work). So this new option will not help.

wh201906:
You mean the scrcpy still uses adb to detect the Android device even with the `--otg`, right?

yume-chan:
Maybe Scrcpy can call `adb -s <serial> detach` or `adb -t <transport-id> detach` to let ADB server release the USB device, so it doesn't need to kill the ADB server.

rom1v:
> You mean the scrcpy still uses adb to detect the Android device even with the `--otg` option, right?

No, but if the USB device is open by adb, then scrcpy will not be able to open it separately for OTG. On other OS, it is perfectly ok to open a USB device already open by another program, but not on Windows.

@yume-chan maybe, to be tested. 😃

wh201906:
> 

How about the adb over TCP? I know it's impossible to use mirroring and otg on the same phone via USB, but I can use mirroring over TCP and otg over USB

rom1v:
> Do you mean the device becomes undetectable if it is connected by adb, even via TCP?

By default, if you plug the device and USB debugging is enabled, the adb daemon will detect it and open it, even if you don't use it.

wh201906:
> By default, IIRC, if you plug the device and USB debugging is enabled, the adb daemon will detect it and open it, even if you don't use it.

If so, this option is useless indeed. Thank you for checking this.
Should I close this pull request now?

yume-chan:
Tried the `adb detach` command. First it outputs:

```
> adb detach
adb.exe: failed to detach: attach/detach only implemented for libusb backend
NativeCommandExitException: Program "adb.exe" ended with non-zero exit code: 1.
```

Launch adb with libusb backend:

```
> $env:ADB_LIBUSB="1"
> adb server nodaemon
adb.exe F 06-02 21:49:59 14088 15940 usb_libusb.cpp:972] failed to register libusb hotplug callback
NativeCommandExitException: Program "adb.exe" ended with non-zero exit code: -1073740791.
```

it crashes immediately (due to unknow error, not the error it logged).

PR #4045, opened 2023-05-31
fix stream offset

shuax:
read is not bytes


rom1v:
Wow, good catch :+1:

I rebased on `dev` branch, edited the commit message and merged: b2d860382fd6ccefe85c45343da622ffbed8b3fa

Out of curiosity, did you find the problem by just reading the code, or because you noticed abnormally audible audio glitches on underflow?

shuax:
I tried to apply the volume using SDL_MixAudioFormat, and the sound was terrible, and I realized that `read` is not bytes, and then I found this problem.

PR #4057, opened 2023-06-04
Fixing Typo

suryawanshishantanu6:


rom1v:
I'm not native English, but I think "in" is correct, isn't it?

ChatGPT agrees:

> In English, we typically use the preposition "in" when referring to the availability of a package in a package manager. Therefore, the correct phrase would be: "This package is available in [package manager]."

PR #4130, opened 2023-06-29
Add --pause-on-error to simplify launchers

rom1v:
`scrcpy-console.bat` and `scrcpy-console.desktop` must keep the terminal open if scrcpy fails with an error.

It was implemented in cmd and shell respectively, which added complexity.

Add an option `--pause-on-error` in scrcpy to simplify.

One possible problem is that an error in command line parsing could occur before `--pause-on-error` was taken into account. For that purpose, if getopt did not detect it, this parameter is searched in the whole argv array "manually" (without getopt) so that it works in all cases.

Refs #3817 #3822


app/scrcpy.1
```diff
@@ -267,6 +267,12 @@ Set the TCP port (range) used by the client to listen.
 
 Default is 27183:27199.
 
+.TP
+.B \-\-pause\-on\-error
+Make scrcpy pause before exiting when an error occurs.
+
+This prevents the terminal to close so that error messages can be read.
```
ratijas, line 274: nit:

```suggestion
This prevents the terminal window from automatically closing, so that error messages can be read.
```

ratijas (commented):
yea, this sounds like a sensible idea

rom1v:
I refactor to support `--pause-on-exit` with 3 values: `true`, `false` and `if-error`.

`--pause-on-exit` is equivalent to `--pause-on-exit=true`.
If the option is not given, then it is equivalent to `--pause-on-exit=false`. 
In `scrcpy-console.bat` and `scrcpy-console.desktop`, `--pause-on-exit=if-error` is used.

rom1v:
Let's merge it (I thought it was already done).

PR #4154, opened 2023-07-10
Add Socket for standalone server

trickymagus:
Currently, the scrcpy-server uses LocalSocket for communication.
However, when attempting to run the scrcpy-server as a standalone server and connect to the LocalSocket through another android application, a permission error occurs.
We can add the capability to use Socket instead of LocalSocket, allowing other applications to communicate with the scrcpy-server.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/DesktopConnection.java
+++ b/server/src/main/java/com/genymobile/scrcpy/DesktopConnection.java
@@ -3,6 +3,7 @@ package com.genymobile.scrcpy;
 import android.net.LocalServerSocket;
 import android.net.LocalSocket;
 import android.net.LocalSocketAddress;
+import android.os.ParcelFileDescriptor;
 
 import java.io.Closeable;
 import java.io.FileDescriptor;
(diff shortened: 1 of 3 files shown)


rom1v:
Hi,

Although `scrcpy-server` exposes some options to simplify using it as a [standalone server](https://github.com/Genymobile/scrcpy/blob/master/doc/develop.md#standalone-server) (because "why not"), their implementation is trivial and non invasive.

Its primary purpose is still to serve the scrcpy client. I don't want to expose unrelated features just for external usages of scrcpy-server as a standalone server (it brings more complexity, more code to maintain, untested code…).

server/src/main/java/com/genymobile/scrcpy/DesktopConnection.java
```diff
@@ -60,40 +105,76 @@ private static String getSocketName(int scid) {
...
             throws IOException {
         String socketName = getSocketName(scid);
 
         LocalSocket videoSocket = null;
         LocalSocket audioSocket = null;
         LocalSocket controlSocket = null;
+
+        Socket videoExtSocket = null;
+        Socket audioExtSocket = null;
+        Socket controlExtSocket = null;
```
Tryanks, line 118: I believe this is not a good design, which adds unnecessary complexity and redundant branching judgments to the program. In order to implement Socket support, I'd be inclined to implement a Socket wrapper to provide a unified interface, and subsequently implement Socket and LocalSocket separately within it. however, your code contribution here is definitely a good enough PoC.

Issue #4011, opened 2023-05-18
Android 14 beta lock orientation doesn't work as expected when the device goes into landscape mode.


ssalenik:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: macOS
 - scrcpy version: 2.0
 - installation method: brew
 - device model: Pixel 7, 
 - Android version: Android 14 Beta (UpsideDownCake)

**Describe the bug**
Android 14 beta lock orientation doesn't work as expected when the device goes into landscape mode.
eg: if you launch: `scrcpy --lock-video-orientation=0` then go into landscape in an app, you get this:
![image](https://github.com/Genymobile/scrcpy/assets/1129490/302b3db9-0094-443a-bf88-7c2a6264d0da)

Looks like it no longer rotates the display output to fit the surface of the codec, but instead it keeps it in its orientation and scales it. Not clear if its a "bug" which will be fixed in the release, but seems possibly unlikely since its already in beta.

vikingden8:
me too, I think this is a problem on new Android 14

ssalenik:
OK, so now that the 14 source is released, I think the issue is due to this change in the `SurfaceControl.setDisplayProjection` fn:
Android 13: https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/android13-dev/core/java/android/view/SurfaceControl.java#2355

Android 14: https://android.googlesource.com/platform/frameworks/base.git/+/refs/tags/android-14.0.0_r2/core/java/android/view/SurfaceControl.java#2331

This fn now ignores the orientation param:
```
/**
--
  | * Because this API is now going through {@link DisplayManager}, orientation and displayRect
  | * will automatically be computed based on configuration changes. Because of this, the params
  | * orientation and displayRect are ignored
  | *
  | * @hide
  | */
  ```

rom1v:
@ssalenik Thank you :+1:

Reported [here](https://issuetracker.google.com/issues/303565669#comment2).

ssalenik:
Thanks. I think it may be possible to find a work around by re-implementing this `public static VirtualDisplay createVirtualDisplay` fn here:
https://android.googlesource.com/platform/frameworks/base.git/+/refs/tags/android-14.0.0_r2/core/java/android/hardware/display/DisplayManager.java#1567

[...]

PR #4213, opened 2023-08-03
Add option to mirror camera

### LINKED ISSUE #241: Add the option to Cast the Phone Camera instead of the hole phone screen
A lot of us are looking for apps like:

- IPWebcam 
- DroidCam
- ManyCam
- Kinoni
- WO Webcam

When you have an already working app that Casts all the phone in a better resolution and with much less lag! I can just open the camera app in my phone but you can imagine the problems  with it.

please add this functionality (Y)

yume-chan:
Closes #241

I can imagine this becoming another compatibility hell. 

I can only test on emulators and XiaoMi devices, so there isn't much more I can do, for example to the bug report in https://github.com/Genymobile/scrcpy/issues/241#issuecomment-1662119167.


server/src/main/java/com/genymobile/scrcpy/SurfaceEncoder.java
```diff
@@ -0,0 +1,288 @@
+package com.genymobile.scrcpy;
```
yume-chan, line 1: You can review this file by comparing it to (the original) `ScreenEncoder.java`. It didn't change much except extracting some logics to abstract methods.

server/src/main/java/com/genymobile/scrcpy/Workarounds.java
```diff
@@ -154,13 +167,29 @@ private static void fillAppContext() {
         }
     }
 
+    @SuppressLint("PrivateApi")
+    public static Context getSystemContext() throws ReflectiveOperationException {
+        if (systemContext == null) {
+            try {
+                // Hide warnings on XiaoMi devices
```
yume-chan, line 174: I'd like to include this to reduce noise on XiaoMi devices, and test it on other devices to make sure it doesn't cause any side effects.

server/src/main/java/com/genymobile/scrcpy/Workarounds.java
```diff
@@ -302,4 +331,44 @@ public static AudioRecord createAudioRecord(int source, int sampleRate, int chan
             throw new RuntimeException("Cannot create AudioRecord");
         }
     }
+
+    public static CameraManager getCameraManager() {
+        if (cameraManager == null) {
+            try {
+                fillBaseContext();
+                cameraManager = CameraManager.class.getDeclaredConstructor(Context.class)
```
yume-chan, line 339: Android 11 requires correct `Context.getOpPackageName()`. `getSystemContext().getService("camera")` doesn't work.

server/src/main/java/com/genymobile/scrcpy/CameraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+
+        cameraThread = new HandlerThread("camera");
+        cameraThread.start();
+        cameraHandler = new Handler(cameraThread.getLooper());
+    }
+
+    @SuppressLint("MissingPermission")
+    private CameraDevice openCamera(String id)
+            throws CameraAccessException, InterruptedException {
+        Semaphore semaphore = new Semaphore(0);
```
yume-chan, line 52: I don't know how to properly do this. Maybe with `ExecutorService` and `Future`s?

yume-chan:
I bought a Meizu mblu 10 (Android 11) for testing.

It turns out that, unlike audio capture, camera capture requires the app to keep being at foreground. The capture session stops once the foreground workaround was stopped. That's a pretty huge inconvenience for Android 11 users.

yume-chan:
I know two other methods to become foreground:

1. Create an overlay window: Shell app doesn't have this permission on Android 11
2. Foreground service: Shell app doesn't declare the required foreground service type on Android 11 so it can't access cameras.

rom1v:
> It turns out that, unlike audio capture, camera capture requires the app to keep being at foreground. The capture session stops once the foreground workaround was stopped. That's a pretty huge inconvenience for Android 11 users.

IMO, it is totally ok to require Android 12+ for the camera feature (the workaround for audio was "simple" and not too intrusive, so it was worth it).

yume-chan:
OK, now I bail out camera mirroring on Android 11 and older, and reverted changes to the foreground workaround.

To be precise, the camera capturing stops when either:

* `com.android.shell` app is forcefully stopped (the method we use to stop foreground workaround)
* The user manually dismisses the heap dump dialog (by tapping any one of three buttons, or tapping outside the dialog), **then** switches to another app (camera capturing continues to work while the user is staying in the current app)

So, it's barely usable, if the user doesn't use the device while mirroring. However, I have to admit it's too much a restriction.

chenxiaolong:
Thanks for working on this! It's been working very well for me (Android 13 + Google Pixel 7 Pro).

I've made an additional change in my branch to add support for specifying the camera resolution and frame rate: https://github.com/chenxiaolong/scrcpy/commit/edb8f5b0f81e6b0d93c67378ade07e23798fc419. In particular, this also adds support for the constrained high speed capture mode, allowing 120 and 240 fps camera streams. (This makes scrcpy the only program capable of streaming the camera at a high frame rate on Android :slightly_smiling_face:)

rom1v:
Just a small message to say that, despite the lack of feedback for 1 month, the feature from this PR has the highest priority for the next scrcpy version. 😉 

I'm just quite busy currently on other subjects at work, which require learning new stuff (which is very interesting, but leaves less time/effort available for scrcpy on my free time).

So it may take some time, but it will be merged 😄 Thank you again for your work 👍 

(Recently, I just started to review/rebase/adapt some of your refactor commits, branch [camera.2](https://github.com/Genymobile/scrcpy/commits/camera.2))

fstecker:
This is such a cool feature, thanks for making it! I just gave it a try on my Pixel 5a (Android 13) and it works nicely with `--camera=1` (the front facing camera) but produces no image when I try `--camera=0` (the back facing one). The output is the same in both cases, even with `--V verbose`. Do you have any idea what the problem could be, or which tests I could do to track it down?

rom1v:
Some update.

This evening, I took 1 or 2 hours to dive back into the subject. I implemented the `SurfaceCapture` interface for the camera (`CameraCapture`), on [camera.6](https://github.com/Genymobile/scrcpy/commits/camera.6) (I reused your code, I just made it match the interface I introduced).

The options are not plugged (except `--list-camera`), everything is hardcoded, it's just for making tests.

What's interesting is that it works regardless of the capture size, even if the size is not listed in the "active array size" list.

For example, here are the sizes supported by my back camera (as listed by `scrcpy --list-camera`):

```
    --video-source=camera --camera=0    (back)
        - 4000x3000
        - 4000x1824
        - 4000x1800
        - 3840x2160
        - 3264x2448
        - 3264x1472
        - 3264x1468
        - 3200x2400
        - 3000x3000
        - 2976x2976
        - 2688x1512
        - 2592x1944
        - 2448x2448
        - 2176x1080
        - 2120x1080
        - 2080x960
        - 1920x864
        - 1920x1080
        - 1440x1080
        - 1280x960
        - 1280x768
        - 1280x720
        - 1080x1080
        - 1024x738
        - 1024x768
        - 864x480
        - 800x600
        - 800x480
        - 720x480
        - 640x480
        - 352x288
        - 320x240
        - 176x144
```

But I can capture camera at 1920×1504 (random value) if I want to, and it works fine:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/CameraCapture.java b/server/src/main/java/com/genymobile/scrcpy/CameraCapture.java
index 709c122a2..6ee69188a 100644
--- a/server/src/main/java/com/genymobile/scrcpy/CameraCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CameraCapture.java
@@ -85,7 +85,7 @@ public class CameraCapture extends SurfaceCapture {
 
     @Override
     public Size getSize() {
-        return new Size(1920, 1080);
+        return new Size(1920, 1504);
     }
 
     @Override
```

Even non-multiple of 8 work! (but I will keep multiple of 8).

Therefore, I think I will just add an option `--camera-aspect-ratio=x:y` (default to 16:9 for example), and just let the second dimension be automatically computed, without checking the "size list". What do you think?

I haven't find a way to capture in the portrait orientation (even if I pass a size like 1080x1920, it captures a vertical video but from the device hold in landscape).

Other things not to forget:
 - detect end of capture (on camera disconnected, we get the callback but nothing is done yet)
 - disable input events
 - (automatically use microphone by default when audio is enabled?)
 - …

chenxiaolong:
Thanks for continuing to work on this! I'm not the original author of this PR (only worked on [the high-frame-rate bit](https://github.com/chenxiaolong/scrcpy/commit/edb8f5b0f81e6b0d93c67378ade07e23798fc419)), but hopefully you don't mind me commenting.

Unfortunately, specifying resolutions outside of the list is not universally supported. On my Pixel 7 Pro, it doesn't work at all.

With a standard `CameraCaptureSession` (<= 60 fps), it'll fail with:

```
Caused by: android.hardware.camera2.CameraAccessException: The camera device is currently in the error state; no further calls to it will succeed.
        at com.genymobile.scrcpy.CameraEncoder$2.onConfigureFailed(CameraEncoder.java:197)
        at android.hardware.camera2.impl.CallbackProxies$SessionStateCallbackProxy.lambda$onConfigureFailed$1(CallbackProxies.java:64)
        at android.hardware.camera2.impl.CallbackProxies$SessionStateCallbackProxy.$r8$lambda$4x6kfFGELk6RYRHf2tdedVsg2ag(Unknown Source:0)
        at android.hardware.camera2.impl.CallbackProxies$SessionStateCallbackProxy$$ExternalSyntheticLambda4.run(Unknown Source:4)
        at android.os.Handler.handleCallback(Handler.java:958)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:205)
        at android.os.Looper.loop(Looper.java:294)
        at android.os.HandlerThread.run(HandlerThread.java:67)
```

With a `CameraConstrainedHighSpeedCaptureSession` (which I realize the `camera.6` branch doesn't implement yet), it fails with at least a more useful message:

```
Caused by: java.lang.IllegalArgumentException: Surface size 1920x1072 is not part of the high speed supported size list [1280x720, 1920x1080]
        at android.hardware.camera2.utils.SurfaceUtils.checkConstrainedHighSpeedSurfaces(SurfaceUtils.java:266)
        at android.hardware.camera2.impl.CameraDeviceImpl.createCaptureSessionInternal(CameraDeviceImpl.java:756)
        at android.hardware.camera2.impl.CameraDeviceImpl.createCaptureSession(CameraDeviceImpl.java:679)
        at com.genymobile.scrcpy.CameraEncoder.createCaptureSession(CameraEncoder.java:201)
        at com.genymobile.scrcpy.CameraEncoder.setSurfaceInternal(CameraEncoder.java:316)
        ... 5 more
```

I think it probably makes the most sense to ensure that the `Surface` size is always something from the list.

> Therefore, I think I will just add an option `--camera-aspect-ratio=x:y` (default to 16:9 for example), and just let the second dimension be automatically computed, without checking the "size list". What do you think?

UX-wise, I personally prefer specifying a resolution directly. It's nice to be able to just copy/paste the resolution from `--list-camera`. scrcpy currently already uses pixel resolutions for everything, like `--crop`. I feel like an aspect ratio option would require the user to do more math.

(Of course, if the aspect ratio argument is optional and specifying both pixel dimensions is possible, then this doesn't apply at all.)

> I haven't find a way to capture in the portrait orientation (even if I pass a size like 1080x1920, it captures a vertical video but from the device hold in landscape).

Is this something we need given that there's already the `--rotation` option?

rom1v:
> Unfortunately, specifying resolutions outside of the list is not universally supported. On my Pixel 7 Pro, it doesn't work at all.

That was what I feared :disappointed:

> I think it probably makes the most sense to ensure that the `Surface` size is always something from the list.

OK, will do that.

> UX-wise, I personally prefer specifying a resolution directly.

But for simplicity (and for consistency with screen mirroring), I want `-m 1024` (i.e. passing only one dimension) to behave as expected, that is, it should select a size:
 - which is at most 1024 in both dimensions;
 - with a consistent aspect ratio (it should not select 16:9 with `-m 1024` and 4:3 on `-m 1200`, unless it has no other choice)

Typically, people don't want to compute and specify the exact size, they just want to limit the capture size for performance (like for screen capture).

As far as I remember, the heuristic from this PR was quite good, but it would have been even better to accept any dimensions (unfortunately, as you reported, that's not possible on all devices apparently).

Moreover, there is a "retry on error" mechanism which automatically downscales on encoding error, so in that case it should be able to select the other dimension automatically.

> (Of course, if the aspect ratio argument is optional and specifying both pixel dimensions is possible, then this doesn't apply at all.)

My idea was to reuse `-m` to pass a single size, and `--camera-aspect-ratio` (with a default value) could give any behavior. For example, for 1920x1080, `scrcpy -m1920 --camera-aspect-ratio=1920:1080` would have worked (a bit convoluted though).

If we add an option to give the exact camera size, should we ignore the `-m` option? For example:
 - `scrcpy --video-source=camera -m1024`: select a valid 1024×HEIGHT based on some heuristic
 - `scrcpy --video-source=camera --camera-size=1024x768`: only use the explicit value provided by the user
 - `scrcpy --video-source=camera -m1024 --camera-size=1024x768`: error, could not pass both `-m` and `--camera-size` (?)

An alternative could be to extend `-m` to optionnally pass two dimensions (`scrcpy --video-source=camera -m1024x768`), but `-m`/`--max-size` is really a maximum size, not the target size, so that does not work.

> > I haven't find a way to capture in the portrait orientation (even if I pass a size like 1080x1920, it captures a vertical video but from the device hold in landscape).
>
> Is this something we need given that there's already the `--rotation` option?

[Rotation](https://github.com/Genymobile/scrcpy/blob/master/doc/video.md#rotation) only impacts the window content, not recording.

We could reuse part of `--lock-video-orientation` and propagate the information to the recorded file during muxing on the client size (but this would also interact with `--rotation`). Anyway, as a first version, ignoring this limitation would be acceptable. Moreover, `--lock-video-orientation` is [broken in Android 14](https://github.com/Genymobile/scrcpy/issues/4011), so I prefer to see what happens with that problem before reusing it for recording.

chenxiaolong:
Thanks for explaining further. I understand what you're going for now.

> Moreover, there is a "retry on error" mechanism which automatically downscales on encoding error, so in that case it should be able to select the other dimension automatically.

That's really cool and I didn't know that existed.

> My idea was to reuse `-m` to pass a single size, and `--camera-aspect-ratio` (with a default value) could give any behavior. For example, for 1920x1080, `scrcpy -m1920 --camera-aspect-ratio=1920:1080` would have worked (a bit convoluted though).
> 
> If we add an option to give the exact camera size, should we ignore the `-m` option? For example:
> 
>   * `scrcpy --video-source=camera -m1024`: select a valid 1024×HEIGHT based on some heuristic
>   * `scrcpy --video-source=camera --camera-size=1024x768`: only use the explicit value provided by the user
>   * `scrcpy --video-source=camera -m1024 --camera-size=1024x768`: error, could not pass both `-m` and `--camera-size` (?)

I like the idea of `--camera-aspect-ratio` acting as the heuristic for `-m` and `--camera-size` being mutually exclusive with `-m`.

But if implementing that is a pain, I don't really mind dealing with `-m1920 --camera-aspect-ratio=1920:1080`.

> An alternative could be to extend `-m` to optionnally pass two dimensions (`scrcpy --video-source=camera -m1024x768`), but `-m`/`--max-size` is really a maximum size, not the target size, so that does not work.

Agreed on the maximum size vs. target size point.

> [Rotation](https://github.com/Genymobile/scrcpy/blob/master/doc/video.md#rotation) only impacts the window content, not recording.
> 
> We could reuse part of `--lock-video-orientation` and propagate the information to the recorded file during muxing on the client size (but this would also interact with `--rotation`). Anyway, as a first version, ignoring this limitation would be acceptable. Moreover, `--lock-video-orientation` is [broken in Android 14](https://github.com/Genymobile/scrcpy/issues/4011), so I prefer to see what happens with that problem before reusing it for recording.

Ahh, I see.

yume-chan:
I agree an aspect ratio option would be useful, the default value can be native aspect ratio of camera sensor so it has the most pixel count, and users can choose another one (like 16:9) so it fits the display without black borders.

rom1v:
Here is a new version: [`camera.26`](https://github.com/Genymobile/scrcpy/commits/camera.26)


### List cameras

To list the cameras available (with their declared valid sizes):

```
scrcpy --list-cameras
scrcpy --list-camera-sizes
```

### Mirror camera

```
scrcpy --video-source=camera
```

By default, it automatically switches audio source to microphone (as if `--audio-source=mic` were also passed).

```bash
scrcpy --video-source=display  # default is --audio-source=output
scrcpy --video-source=camera   # default is --audio-source=mic
scrcpy --video-source=display --audio-source=mic    # force display AND microphone
scrcpy --video-source=camera --audio-source=output  # force camera AND device audio output
```

### Camera selection

It is possible to pass an explicit camera id (as listed by `--list-cameras`):

```
scrcpy --video-source=camera --camera-id=0
```

Alternatively, the camera may be selected automatically:

```bash
scrcpy --video-source=camera                           # use the first camera
scrcpy --video-source=camera --camera-facing=front     # use the first front camera 
scrcpy --video-source=camera --camera-facing=back      # use the first back camera
scrcpy --video-source=camera --camera-facing=external  # use the first external camera
```

If `--camera-id` is specified, then `--camera-facing` is forbidden (the id already determines the camera):

```bash
scrcpy --video-source=camera --camera-id=0 --camera-facing=front  # error
```

### Camera size selection

It is possible to pass an explicit camera size:

```
scrcpy --video-source=camera --camera-size=1920x1080
```

The given size may be listed among the declared valid sizes (`--list-camera-sizes`), but may also be anything else (some devices support arbitrary sizes):

```
scrcpy --video-source=camera --camera-size=1840x444
```

Alternatively, a declared valid size (among the ones listed by `list-camera-sizes`) may be selected automatically.

Two constraints are supported:
 - `-m`/`--max-size` (already used for display mirroring), for example `-m1920`
 - `--camera-ar` to specify an aspect ratio (`<num>:<den>`, `<value>` or `sensor`)

Some examples:

```bash
scrcpy --video-source=camera                          # use the largest width and the largest associated height
scrcpy --video-source=camera -m1920                   # use the largest width not above 1920 and the largest associated height
scrcpy --video-source=camera --camera-ar=4:3          # use the largest size with an aspect ratio of 4:3 (+/- 10%)
scrcpy --video-source=camera --camera-ar=1.6          # use the largest size with an aspect ratio of 1.6 (+/- 10%)
scrcpy --video-source=camera --camera-ar=sensor       # use the largest size with the aspect ratio of the camera sensor (+/- 10%)
scrcpy --video-source=camera -m1920 --camera-ar=16:9  # use the largest width not above 1920 and the closest to 16:9 aspect ratio
```

If `--camera-size` is specified, then `-m`/`--max-size` and `--camera-ar` are forbidden (the size is determined):

```bash
scrcpy --video-source=camera --camera-size=1920x1080 -m3000  # error
```

chenxiaolong:
Thanks for continuing to work on this!

Everything seems to work great. All of the new CLI options seem to work as described, including the conflicting arguments. Audio from the microphone is working too.

The only issue I noticed is that `--list-cameras` and `--list-camera-sizes` take a very long time to exit after printing out the list (11+ minutes in my case with adb 34.0.5-10900879 + Pixel 8 Pro). It doesn't affect the `--list-displays` and `--list-encoders` options though. I'm still working on troubleshooting this--I have no idea why this might be happening.

(For folks who are interested in high frame rate capture, I've rebased my changes on top of `camera.26`: https://github.com/Genymobile/scrcpy/commit/dd6c275d6cacbfa7f3cb5af2a2079cea25e1abd5)

yume-chan:
@chenxiaolong I wonder if you could use the existing `--max-fps` option to toggle high frame rate capture, and if it could work to limit FPS in normal frame rate captures?

chenxiaolong:
Oh, good point, I overlooked that option. The only use case I can think of where having the capture frame rate separate from the encoding frame rate is when the user explicitly wants to record low-FPS video.

For example, some devices only support 24/30/60 (normal) and 120/240 (high speed), but the user would be able to record eg. 15 FPS video with `scrcpy <...> --camera-fps 30 --max-fps 15`. Capturing at 15 FPS directly would throw an exception if the camera does not explicitly support it.

> if it could work to limit FPS in normal frame rate captures?

Yep, limiting capture frame rate in normal mode also works, but it must be a valid item from the `CONTROL_AE_AVAILABLE_TARGET_FPS_RANGES` list (similar to how high-speed mode requires a value from the `getHighSpeedVideoFpsRangesFor()` list). I just updated https://github.com/chenxiaolong/scrcpy/commit/12e85875ad019293550f078b49f01080b12f0f7d to include the normal FPS list in `--list-camera-sizes`.

rom1v:
> The only issue I noticed is that `--list-cameras` and `--list-camera-sizes` take a very long time to exit after printing out the list (11+ minutes in my case with adb 34.0.5-10900879 + Pixel 8 Pro). It doesn't affect the `--list-displays` and `--list-encoders` options though.

Thank you, I reproduced on a Pixel. Fixed by 7a66c4d0a4b4709db465d424995c3fa815e1d0bc.

> I wonder if you could use the existing `--max-fps` option to toggle high frame rate capture, and if it could work to limit FPS in normal frame rate captures?

I think `--max-fps` and `--camera-fps` must be 2 differents options, since no "max fps" means "unlimited fps", and by default, we don't want to capture at 240fps even if it is available (thus, `--max-fps` is more appropriate for display capture, where the capture rate is variable).

> I just updated https://github.com/chenxiaolong/scrcpy/commit/12e85875ad019293550f078b49f01080b12f0f7d to include the normal FPS list in `--list-camera-sizes`.

Thank you :+1:

I think your commit implements two separate things:
 1. the `--camera-fps` option
 2. the high speed capture mode

I took the first part on my branch [`camera.28`](https://github.com/Genymobile/scrcpy/commits/camera.28) (and added bash/zsh completion and manpage): 8db995aaeed0092f752c988c7d7be6e779cf598c

Btw, I can capture at different frame rates that those reported. For example, one camera reports `fps=[15, 23, 28, 30]` in `--list-cameras`, but I can use `--camera-fps=60` and it really captures at 60fps (as shown with `--print-fps`, and the difference is visible). Even `--camera-fps=120` captures at 120 fps, without even handling high speed capture. `--camera-fps=5` also works. Maybe we should retrieve the available frame rates differently to get accurate values?

_I will review/test high speed capture later._

chenxiaolong:
> Btw, I can capture at different frame rates that those reported. For example, one camera reports `fps=[15, 23, 28, 30]` in `--list-cameras`, but I can use `--camera-fps=60` and it really captures at 60fps (as shown with `--print-fps`, and the difference is visible). Even `--camera-fps=120` captures at 120 fps, without even handling high speed capture. `--camera-fps=5` also works. Maybe we should retrieve the available frame rates differently to get accurate values?

Huh, I haven't seen any device that behaves like that. On my Pixel 8 Pro and Samsung Tab S8 Ultra, only the items in the list are supported (I tried everything in [1, 120] exhaustively). Seems like yet another weird Android inconsistency like you saw with the camera resolutions where some devices seem to allow arbitrary values.

Unfortunately, I don't think there's any way to query the supported FPS values outside of `CONTROL_AE_AVAILABLE_TARGET_FPS_RANGES`. Most OEM camera apps hardcode the supported FPS values per device. I wouldn't be surprised if they overlooked making the camera2 API report the same values.

For `--camera-fps=120`, would you mind trying to record that to a file and see if each frame is actually unique (vs. 60 fps with each frame duplicated)? Android added the high speed mode because it batches frames together and has a happy path when working with `MediaCodec` surfaces to make it possible to run reliably on mobile SoCs. I'd be super interested if there are devices where 120 fps works in normal mode--that'd mean the SoC is fast enough to process each frame individually.

rom1v:
> Seems like yet another weird Android inconsistency like you saw with the camera resolutions where some devices seem to allow arbitrary values.

On a Pixel 6a, the reported fps were `[15, 30, 60]`, but `--camera-fps=5` worked for example. I will try other values when I have access to the device.

On a One Plus 7 Pro, the reported fps were `[15, 23, 28, 30]`, but basically all values work, even for example `--camera-fps=113`, it produced exactly 113 fps. If I request `--camera-fps=240`, it produces 30 fps though.

> For `--camera-fps=120`, would you mind trying to record that to a file and see if each frame is actually unique (vs. 60 fps with each frame duplicated)?

[120fps.mp4](https://tmp.rom1v.com/scrcpy/4213/120fps.mp4)

I extracted the frames with:

```bash
mkdir frames
ffmpeg -i 120fps.mp4 frames/frame-%03d.png
```

At the beginning, some frames are duplicated, but then it seems it's really 120fps.

chenxiaolong:
Thanks! I wish the newer Pixels allowed arbitrary FPS in normal mode like that.

rom1v:
So it's not easy to determine if we must use high speed or not based only on the fps. In particular, we cannot reject fps values just because they're not declared:
https://github.com/chenxiaolong/scrcpy/commit/12e85875ad019293550f078b49f01080b12f0f7d#diff-8bd110fa92b613c45b0bf94599022900e2edea78bfbccb243f8f1d04c5f88855R149

Maybe we should just check whether it is valid for some high speed size?

I don't really understand the fps ranges. Why is each entry a range, not a single value? Is it possible to capture using the camera using a variable frame rate?

For example, on a One Plus 7 Pro, the high speed sizes have two ranges each:
 - 1280×720
   - [60, 480]
   - [480, 480]
 - 1920×1080
   - [30, 240]
   - [240, 240]

Why are there two ranges with the same maximum value? And the range `[30, 240]` includes frame rates which are possible without high speed mode… The meaning is not very clear to me.

chenxiaolong:
> Maybe we should just check whether it is valid for some high speed size?

Agreed. We should use high-speed mode if the parameters match a high-speed resolution and frame rate. Otherwise, just try it in the normal mode and let it fail if it doesn't work.

> The meaning is not very clear to me.

Yep... It's definitely [not well documented](https://developer.android.com/reference/android/hardware/camera2/CameraConstrainedHighSpeedCaptureSession#createHighSpeedRequestList(android.hardware.camera2.CaptureRequest)). My understanding is that the upper limit is always used for the recording `MediaCodec` `Surface` and the lower limit is relevant for other features, like an on-screen preview in a camera app, color filters, etc.

rom1v:
> We should use high-speed mode if the parameters match a high-speed resolution and frame rate.

But for example on my One Plus 7 Pro which declared 2 ranges `[60, 480]` and `[480, 480]` for 1280×720, if we consider only the range upper value, then high speed is only supported for 480fps (which is not true). If we also consider the range lower value, the `[30, 240]` overlaps the non-high speed ranges.

I think a new option `--camera-high-speed` would not be that bad:
 - it would avoid to guess using heuristics,
 - if we guess from the fps, changing the camera fps could implicitly result in a the selected size change, which is confusing since it's not explicit which value causes high speed to be enabled.

zx2c4:
Practically speaking, what `--camera-high-speed` means is that users first try without that switch and then try again with that switch when it fails. Can't we just programmatically do that? Try one, see if it works, fallback to the next mode.

rom1v:
> Practically speaking, what `--camera-high-speed` means is that users first try without that switch and then try again with that switch when it fails.

Not necessarily, the high speed sizes and frame rates are listed in `--list-camera-sizes`, so users can decide in advance which mode to use. The high speed frame rates has more constraints (typically it requires a specific size) so it might make sense to make this flag explicit.

> Try one, see if it works, fallback to the next mode.

Sometimes, failure is not trivial to detect. For example, an unsupported configuration might cause a deadlock (the configuration works, but never sends frame to the encoder). Sometimes, it may fail later: what to do if the first frame works, but then it fails, is it a failure? How the user could switch to the other mode manually?

zx2c4:
I suppose the fancy way would be detecting whether you get a second of two of mostly unique frames. I can understand wanting to avoid this kind of complexity, though, so point taken.

rom1v:
On [`camera.29`](https://github.com/Genymobile/scrcpy/commits/camera.29), I implemented `--camera-high-speed`. When specified, `--camera-fps` is mandatory.

In this version, `--list-camera-sizes` lists the high speed sizes in a separate section, for 2 reasons:
 - in the unlikely case where a high speed size does not exist in the normal sizes, it would not be printed at all
 - now that `--camera-high-speed` is an explicit flag, it makes sense to list the supported sizes separately

```
scrcpy --video-source=camera --camera-high-speed --camera-fps=120
```

The reported sizes are not always accurate. For example, on a Pixel 4a:

```console
$ scrcpy --list-camera-sizes
...
      High speed capture:
        - 1280x720 (fps=[120, 240])
        - 1920x1080 (fps=[120, 240])
```

```console
$ scrcpy --video-source=camera --camera-high-speed --camera-size=1920x1080 --camera-fps=240
...

ERROR: Encoding error: java.lang.IllegalArgumentException: Surface size 1920x1080 is not part of the high speed supported size list [1280x720]
```

:upside_down_face:

chenxiaolong:
Works great, thanks!

> In this version, `--list-camera-sizes` lists the high speed sizes in a separate section, for 2 reasons:
> 
>   * in the unlikely case where a high speed size does not exist in the normal sizes, it would not be printed at all
>   * now that `--camera-high-speed` is an explicit flag, it makes sense to list the supported sizes separately

Agreed.

> The reported sizes are not always accurate.

I think I'm starting to see why people _really_ don't like Android's Camera APIs :slightly_smiling_face:

rom1v:
Here is a final branch (hopefully), ready to be merged: [`camera.36`](https://github.com/Genymobile/scrcpy/commits/camera.36).

I added some documentation: https://github.com/Genymobile/scrcpy/commit/c369f83d91589fec3c0036fa0c72661c065ddc75 |  [rendered](https://github.com/Genymobile/scrcpy/blob/camera.36/doc/camera.md)

rom1v:
Here is a final branch (hopefully), ready to be merged: [`camera.36`](https://github.com/Genymobile/scrcpy/commits/camera.36).

I added some documentation: bd2da81dd8b03482b1997233a308d2d095fc6e3b | [rendered](https://github.com/Genymobile/scrcpy/blob/camera.36/doc/camera.md)

rom1v:
@yume-chan about `Hide warnings on XiaoMi devices` (f75f9b9b46e38ef3d6e0ff139b35c540e5e91daf), it seems it does not work for me (tested on commit d0d0ece9786792dd19a25808aca64282a3bc6394):

```
$ ./run d --list-cameras
scrcpy 2.1.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     --> (tcpip)  xxx.xxx.xxx.xxx:5555               device  M2101K7AG
DEBUG: Device serial: xxx.xxx.xxx.xxx:5555
DEBUG: Using SCRCPY_SERVER_PATH: d/server/scrcpy-server
d/server/scrcpy-server: 1 file pushed, 0 skipped. 100.2 MB/s (87430 bytes in 0.001s)
java.io.FileNotFoundException: /data/system/theme_config/theme_compatibility.xml: open failed: ENOENT (No such file or directory)
	at libcore.io.IoBridge.open(IoBridge.java:492)
	at java.io.FileInputStream.<init>(FileInputStream.java:160)
	at java.io.FileInputStream.<init>(FileInputStream.java:115)
	at java.io.FileReader.<init>(FileReader.java:58)
	at miui.content.res.ThemeCompatibilityLoader.getVersion(ThemeCompatibilityLoader.java:108)
	at miui.content.res.ThemeCompatibilityLoader.getConfigDocumentTree(ThemeCompatibilityLoader.java:126)
	at miui.content.res.ThemeCompatibilityLoader.loadConfig(ThemeCompatibilityLoader.java:59)
	at miui.content.res.ThemeCompatibility.<clinit>(ThemeCompatibility.java:31)
	at miui.content.res.ThemeCompatibility.isThemeEnabled(ThemeCompatibility.java:111)
	at android.content.res.MiuiResourcesImpl.<clinit>(MiuiResourcesImpl.java:41)
	at android.content.res.Resources.<init>(Resources.java:353)
	at android.content.res.MiuiResources.<init>(MiuiResources.java:49)
	at android.content.res.Resources.getSystem(Resources.java:233)
	at android.app.LoadedApk.<init>(LoadedApk.java:239)
	at android.app.ContextImpl.createSystemContext(ContextImpl.java:2625)
	at android.app.ActivityThread.getSystemContext(ActivityThread.java:2533)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.genymobile.scrcpy.Workarounds.fillBaseContext(Workarounds.java:173)
	at com.genymobile.scrcpy.Workarounds.getCameraManager(Workarounds.java:320)
	at com.genymobile.scrcpy.LogUtils.buildCameraListMessage(LogUtils.java:73)
	at com.genymobile.scrcpy.Server.main(Server.java:203)
	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:407)
Caused by: android.system.ErrnoException: open failed: ENOENT (No such file or directory)
	at libcore.io.Linux.open(Native Method)
	at libcore.io.ForwardingOs.open(ForwardingOs.java:166)
	at libcore.io.BlockGuardOs.open(BlockGuardOs.java:254)
	at libcore.io.IoBridge.open(IoBridge.java:478)
	... 22 more
[server] INFO: List of cameras:
    --video-source=camera --camera=0    (Back, 4000x3000)
    --video-source=camera --camera=1    (Front, 4208x3120)
```

One dirty way to get rid of these messages is:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/Server.java b/server/src/main/java/com/genymobile/scrcpy/Server.java
index a8e8e36a2..89b800225 100644
--- a/server/src/main/java/com/genymobile/scrcpy/Server.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Server.java
@@ -3,7 +3,9 @@ package com.genymobile.scrcpy;
 import android.os.BatteryManager;
 import android.os.Build;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -222,7 +224,14 @@ public final class Server {
             }
             if (options.getListCameras() || options.getListCameraSizes()) {
                 Workarounds.apply(false, true);
-                Ln.i(LogUtils.buildCameraListMessage(options.getListCameraSizes()));
+                PrintStream out = System.out;
+                PrintStream err = System.err;
+                System.setOut(new PrintStream(new ByteArrayOutputStream()));
+                System.setErr(new PrintStream(new ByteArrayOutputStream()));
+                String s = LogUtils.buildCameraListMessage(options.getListCameraSizes());
+                System.setOut(out);
+                System.setErr(err);
+                Ln.i(s);
             }
             // Just print the requested data, do not mirror
             return;
```

Not sure if we can leverage this to improve `Ln.i()` (and others) without side effects (I'm not sure).

rom1v:
@yume-chan I propose this to get rid of all internal errors printed to the console on Xiaomi devices: 3c7a6d363ede7806530c99621b4b1cf4bc7da3ed. It seems to work well.

What do you think?

yume-chan:
I can confirm my workaround works on my Mi 11 running stock MIUI 14 and Android 13. But again, my workaround only hides one of the two system outputs, so your method of replacing output streams looks like a better solution.

I also checked all your changes on the `camera.36` branch, everything looks good. In 120FPS high-speed capture mode, I can get around 50ms of latency, it's much better than 30 FPS normal-speed capture. About arbitrary resolutions and frame rates, my MI 11 only supports the resolutions and frame rates listed in `--list-camera-sizes`.

rom1v:
Thank you for your feedback. :+1:

I simplified the commit to get rid of internal errors, and applied it to all devices: 24a24c9c5b2bfa2b9234933fe2ad30c18ebfa68a ([`camera.38`](https://github.com/Genymobile/scrcpy/commits/camera.38).

rom1v:
Merged and published scrcpy v2.2 :rocket: 

Thank you very much @yume-chan and @chenxiaolong :heart:

Mufanc:
I tried to create a transparent and not touchable floaty window in scrcpy-server to make camera mirroring work for Android 11 devices. I tested it in AVD and it works fine, but I don't have any more Android 11 devices to test it on, and I can't find any reason in the AOSP why such a design would work. Here are some of the changes I made:
https://github.com/Genymobile/scrcpy/compare/master...Mufanc:scrcpy:feat/camera-a11?expand=1

yume-chan:
@Mufanc It works on Meizu Mblu 10 (Flyme 9, Android 11), but on OnePlus 6 (HydrogenOS, Android 11) it throws an error:

```text
> ./scrcpy -s 9719052f --video-source=camera
scrcpy v2.2 <https://github.com/Genymobile/scrcpy>
INFO: Camera video source: control disabled
INFO: Camera video source: microphone audio source selected
INFO: ADB device found:
INFO:           (usb)              73294bba            device  M2011K2C
INFO:     -->   (usb)              9719052f            device  ONEPLUS_A6000
INFO:           (usb)  M01046406Y2031500412            device  mblu10
C:\Users\simon\Downloads\scrcpy-win64-v2.2\scrcpy-server: 1 file pushed, 0 skipped. 98.7 MB/s (72892 bytes in 0.001s)
[server] INFO: Device: [OnePlus] OnePlus ONEPLUS A6000 (Android 11)
[server] INFO: Using camera '0'
java.lang.SecurityException: Given calling package android does not match caller's uid 2000
        at android.os.Parcel.createExceptionOrNull(Parcel.java:2373)
        at android.os.Parcel.createException(Parcel.java:2357)
        at android.os.Parcel.readException(Parcel.java:2340)
        at android.os.Parcel.readException(Parcel.java:2282)
        at android.app.IActivityManager$Stub$Proxy.getContentProvider(IActivityManager.java:5855)
        at android.app.ActivityThread.acquireProvider(ActivityThread.java:7169)
        at android.app.ContextImpl$ApplicationContentResolver.acquireProvider(ContextImpl.java:2951)
        at android.content.ContentResolver.acquireProvider(ContentResolver.java:2509)
        at android.provider.Settings$ContentProviderHolder.getProvider(Settings.java:2608)
        at android.provider.Settings$NameValueCache.getStringForUser(Settings.java:2762)
        at android.provider.Settings$Secure.getStringForUser(Settings.java:6052)
        at android.provider.Settings$Secure.getIntForUser(Settings.java:6249)
        at android.provider.Settings$Secure.getInt(Settings.java:6243)
        at android.view.ViewRootImplInjector.handleGestureConfigCheck(ViewRootImplInjector.java:411)
        at android.view.ViewRootImplInjector.<init>(ViewRootImplInjector.java:330)
        at android.view.ViewRootImpl.<init>(ViewRootImpl.java:855)
        at android.view.ViewRootImpl.<init>(ViewRootImpl.java:791)
        at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:411)
        at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:110)
        at com.genymobile.scrcpy.CameraCapture.prepareOverlayWindow(CameraCapture.java:270)
        at com.genymobile.scrcpy.CameraCapture.init(CameraCapture.java:94)
        at com.genymobile.scrcpy.SurfaceEncoder.streamScreen(SurfaceEncoder.java:55)
        at com.genymobile.scrcpy.SurfaceEncoder.lambda$start$0$com-genymobile-scrcpy-SurfaceEncoder(SurfaceEncoder.java:253)
        at com.genymobile.scrcpy.SurfaceEncoder$$ExternalSyntheticLambda0.run(Unknown Source:4)
        at java.lang.Thread.run(Thread.java:923)
        at com.android.server.am.ActivityManagerService.getContentProvider(ActivityManagerService.java:8671)
        at android.app.IActivityManager$Stub.onTransact(IActivityManager.java:2443)
        at com.android.server.am.ActivityManagerService.onTransact(ActivityManagerService.java:3321)
        at android.os.Binder.execTransactInternal(Binder.java:1165)
        at android.os.Binder.execTransact(Binder.java:1134)
```

IIRC there is no way for creating content provider in shell process yet.

PR #4357, opened 2023-10-15
Update scrcpy-noconsole.vbs

NilsLeo:
Doesnt Open a new Scrcpy window if one is already open. This prevents echoing when playing Audio on pc. In addition, if the window is already open, it will be brought to foregroud (unless it is minimized). This wont quite work yet, as the window title is dependent on the device. Maybe one could get this name dynamically or rename the window to scrcpy on all devices. just a suggestion


phonon112358:
+1

PR #4427, opened 2023-11-15
Enhance CI Workflow: Automate `.deb` Package Build and Release for Multiple Distributions

hacksysteam:
## Overview
This pull request introduces enhancements to the CI/CD pipeline by automating the build process of `.deb` packages for `scrcpy` across multiple Linux distributions. The updated GitHub Actions workflow builds the packages and publishes them as assets to a GitHub release, streamlining the release process and ensuring consistent delivery of the latest builds to users.

## Changes
- Implemented a matrix strategy in the GitHub Actions workflow to build `scrcpy` for Debian Bullseye, Bookworm, and Ubuntu Focal.
- Configured caching for APT packages and Gradle dependencies to speed up the build process.
- Added steps to automatically create a GitHub release with the commit hash as a tag and upload the built `.deb` packages to this release.

## Testing
The updated workflow has been tested in a fork of the repository, where it successfully built `.deb` packages and published them to a release. The packages were then downloaded and tested in their respective distributions to confirm their integrity and functionality.

This still might need some more changes, feel free to comment.

Thanks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,185 @@
+name: Build and Publish Release
+
+on:
+  push:
+    branches:
+      - master
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - distro: "debian:bookworm"
+            os_name: "bookworm"
+          - distro: "debian:bullseye"
+            os_name: "bullseye"
+          - distro: "ubuntu:focal"
+            os_name: "focal"
+
+    container:
+      image: docker://${{ matrix.distro }}
+
+    env:
+      SCRCPY_USER: scrcpy
+      SCRCPY_BUILD_DIR: ./build-auto
+      ANDROID_SDK_ROOT: /opt/android-sdk
+      ANDROID_HOME: /opt/android-sdk
+      DISTRIBUTION: ${{ matrix.os_name }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Cache APT packages, Android SDK and Gradle dependencies
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            /var/cache/apt
+            /var/lib/apt/lists
+            /opt/android-sdk
+            /home/${{ env.SCRCPY_USER }}/.gradle
+          key: ${{ runner.os }}-${{ matrix.os_name }}-apt-gradle-${{ hashFiles('**/build.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.os_name }}-apt-gradle-
+            ${{ runner.os }}-${{ matrix.os_name }}-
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you. That looks cool :+1:

I really need to take the time to read the doc about GitHub actions (my CI knowledge is very limited). By the way, I would like to also build the Windows releases (just executing `./release.sh` from the CI), and maybe to build macOS binaries too (later). There were other PR for that in the past, but IIRC there were always unsolved problems for me.

I have many things to do on my TODO list before that, but that would be great to publish .deb with scrvpy releases :+1:

.github/workflows/build.yml
```diff
@@ -0,0 +1,185 @@
...
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            "https://api.github.com/repos/${{ github.repository }}/releases/tags/latest")
+          release_id=$(echo $response | jq .id)
+          echo "release_id=$release_id" >> $GITHUB_ENV
+        continue-on-error: true
+
+      - name: Create Release
+        if: env.release_id == ''
+        uses: ncipollo/release-action@v1
```
rom1v, line 171: This adds a dependency from a "random" user in the process, doesn't it? Ditto below.

hacksysteam: Github actions allow you to use community actions. The one you highlighted uses `ncipollo/release-action@v1` to create a release.

You can check the release here: https://github.com/hacksysteam/scrcpy/releases
The workflow file was designed to just create 1 release per push, we can make changes and adapt this to the project requirements. I just tested the bookworm deb file and it works, both server and app are included in the same deb file.

hacksysteam: I'll get it fixed.

1fexd:
Looking forward to this PR being merged, I'm running Mint 21.3 (based on Ubuntu 22) and had a lot of issues with building due the ffmpeg/libavcodec version being very outdated - just forked this PR, added focal, started the action and now it works - thanks!

hacksysteam:
@rom1v Kudos on merging https://github.com/Genymobile/scrcpy/pull/5306 but I do not see the deb files being built? It's much easier to use the deb file as we can include the dependencies information and it will be auto-downloaded and installed.

rom1v:
> but I do not see the deb files being built

We could add deb files for the latest Ubuntu/Debian versions in addition.

The GitHub Action script from this PR would need to be adapted to:
 - retrieve the server binary built for the other action
 - be minimal (no unnecessary steps)
 - do not depend on external actions

But if the static binary works well, it's maybe not necessary (a single version, and .deb would not offer much benefits since they would not auto-update).

What would be great is for the package to be provided upstream, but it's not possible anymore due to broken/obsolete android SDK packages in Debian: https://tracker.debian.org/pkg/scrcpy

> It's much easier to use the deb file as we can include the dependencies information and it will be auto-downloaded and installed.

(The provided binary for Linux is static, so it does not depend on installing FFmpeg, SDL or libusb anymore.)

Issue #3927, opened 2023-04-17
Android 14 turn screen off issue


Da2MinuteNoodle:
On the Android 14 beta, launching scrcpy from the lockscreen reboots the device, ~copy-paste is broken, and~ --turn-screen-off no longer works.

Launch logs:
```
PS C:\DATA\Utilities\scrcpy> .\scrcpy.exe --no-audio --turn-screen-off
scrcpy 2.0 <https://github.com/Genymobile/scrcpy>
C:\DATA\Utilities\scrcpy\scrcpy-server: 1 file pushed, 0 skipped. 79.3 MB/s (52867 bytes in 0.001s)
[server] INFO: Device: Google Pixel 6 (Android 13)
[server] ERROR: Could not invoke method
java.lang.NoSuchMethodException: android.content.IClipboard$Stub$Proxy.addPrimaryClipChangedListener [interface android.content.IOnPrimaryClipChangedListener, class java.lang.String, class java.lang.String, int]
        at java.lang.Class.getMethod(Class.java:2890)
        at java.lang.Class.getMethod(Class.java:2470)
        at com.genymobile.scrcpy.wrappers.ClipboardManager.getAddPrimaryClipChangedListener(ClipboardManager.java:129)
        at com.genymobile.scrcpy.wrappers.ClipboardManager.addPrimaryClipChangedListener(ClipboardManager.java:139)
        at com.genymobile.scrcpy.Device.<init>(Device.java:100)
        at com.genymobile.scrcpy.Server.scrcpy(Server.java:64)
        at com.genymobile.scrcpy.Server.main(Server.java:381)
        at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method)
        at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:359)
[server] ERROR: Could not invoke method
java.lang.NoSuchMethodException: android.view.SurfaceControl.getPhysicalDisplayIds []
        at java.lang.Class.getMethod(Class.java:2890)
        at java.lang.Class.getMethod(Class.java:2470)
        at com.genymobile.scrcpy.wrappers.SurfaceControl.getGetPhysicalDisplayIdsMethod(SurfaceControl.java:137)
        at com.genymobile.scrcpy.wrappers.SurfaceControl.getPhysicalDisplayIds(SurfaceControl.java:144)
        at com.genymobile.scrcpy.Device.setScreenPowerMode(Device.java:282)
[...]

PR #4441, opened 2023-11-20
Improve orientation support

rom1v:
## Problems

There are currently two options to rotate the video:
 - `--lock-video-orientation`: it rotates the capture, but does not work for camera capture, and does not work at all for Android 14+ (#4011)
 - `--rotation`: it only rotates the display, so the rotation is not applied to a file recorded with `--record`
 - (less important) there was no way to support flipping the video (#1380, #3819)

As a consequence, in particular, it was not possible to record a vertical video from a camera capture.

### New orientation option

Introduce a new option `--orientation` (and separate options `--display-orientation` and `--record-orientation` if needed) which supports the 8 possible orientations (4 rotations × 2 flips):

```
scrcpy --orientation=0
scrcpy --orientation=90       # 90° clockwise
scrcpy --orientation=180      # 180°
scrcpy --orientation=270      # 270° clockwise
scrcpy --orientation=flip0    # hflip
scrcpy --orientation=flip90   # hflip + 90° clockwise
scrcpy --orientation=flip180  # vflip (hflip + 180°)
scrcpy --orientation=flip270  # hflip + 270°
```

The rotation is expressed in degrees clockwise so that `--orientation=90` works for vertical video from a camera (it's more natural than `--orientation=270` or `--orientation=-90`).

It is possible to change the display orientation dynamically using [shortcuts](https://github.com/Genymobile/scrcpy/blob/orientation/doc/shortcuts.md):
 - rotate display left: <kbd>MOD</kbd>+<kbd>←</kbd>
 - rotate display right: <kbd>MOD</kbd>+<kbd>→</kbd>
 - flip display horizontally: <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>←</kbd> or <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>→</kbd>
 - flip display vertically: <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>↑</kbd> or <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>↓</kbd>

The rotation is applied to a recorded file by writing a display transformation to the MP4 or MKV target file.
Flipping is not supported for recording, so only the 4 first values are allowed when recording.

For MKV files, unfortunately, some players currently do not handle it properly:
 - [I proposed a fix for VLC](https://code.videolan.org/videolan/vlc/-/merge_requests/4558)
 - [I reported an issue for MPV](https://github.com/mpv-player/mpv/issues/12928)

[...]

PR #4446, opened 2023-11-21
Fix turn screen off on Android 14

rom1v:
This PR is on top of #4441, so just consider the last commits.

On Android 14, execute a separate process with a different classpath and LD_PRELOAD to execute the methods required to turn the device screen off.
    
Fixes https://github.com/Genymobile/scrcpy/issues/3927
Refs https://github.com/Genymobile/scrcpy/issues/3927#issuecomment-1790031953

Tested on Pixel 8 with Android 14.

Binary for Windows:
- [`scrcpy-turnscreenoff-android14.v1.zip`](https://tmp.rom1v.com/scrcpy/4446/1/scrcpy-turnscreenoff-android14.v1.zip) <sub>`SHA-256: d74020bd6e094168d84f1fa3e031341fb5850e078d9299e8ea583a5537c69c4a`</sub>
  (old: 1 process per request)

- [`scrcpy-turnscreenoff-android14.v2.zip`](https://tmp.rom1v.com/scrcpy/4446/2/scrcpy-turnscreenoff-android14.v2.zip) <sub>`SHA-256: dbef9c49d165eb96b7a098d9e9bbbf7f0ffe1668ad7e7825380dafd20e426972`</sub>  
  (new: 1 process for all, better reactivity)


Sy-CoDe:
screen off "mod o" works on my pixel 6 Android 14

f870103:
Confirm worked on Poco F5 (Redmi Note 12 Turbo)
Android 14 Evolution X

f870103:
> but command line flags "-Swf" don't work
I don't have problem with `-Swf` flags.

Sy-CoDe:
> > but command line flags "-Swf" don't work
> 
> I don't have problem with `-Swf` flags.

correction, flags do work for me as well, mistake on my end

Corentin-Ch:
working on Google Pixel 6 Pro Android 14

corentin-c:
Fixed on Pixel 6 Pro Android 14

rom1v:
It takes few hundred milliseconds to execute a new java process.

If I add `services.jar` to the classpath and `libandroid_servers.so` to `LD_PRELOAD` for the main scrcpy-server process, and execute the code directly, it is slightly faster to turn the screen off an on when using shortcuts (it is noticeable). But it fails to run scrcpy on older devices:

```
CANNOT LINK EXECUTABLE: library "/system/lib64/libandroid_servers.so" not found
page record for 0xb6f4f00c was not found (block_size=64)
```

Maybe I could start the process once (the first time the device screen is turned off), and keep it alive for further requests.

Or just accept the few hundred milliseconds latency when turning the screen on and off.

rom1v:
> Maybe I could start the process once (the first time the device screen is turned off), and keep it alive for further requests.

I just did this: e844c6810b38c8455d4afb3a53e3bf1ce3b385e6

Please test the new binary :)

rom1v:
> Maybe I could start the process once (the first time the device screen is turned off), and keep it alive for further requests.

I just did this: c5c2734db9c843708041307b7890b79b173cd8ed

Please test the new binary :)

yume-chan:
This might work (I only tested in emulator) without new processes, LD_PRELOAD and multiple CLASSPATH

```java
var classLoaderFactoryClass = Class.forName("com.android.internal.os.ClassLoaderFactory");
var createClassLoaderMethod = classLoaderFactoryClass.getDeclaredMethod("createClassLoader", String.class, String.class, String.class, ClassLoader.class, int.class, boolean.class, String.class);
var classLoader = (PathClassLoader) createClassLoaderMethod.invoke(null, "/system/framework/services.jar", "", "", ClassLoader.getSystemClassLoader(), 10000, true, null);

var displayControlClass = Class.forName("com.android.server.display.DisplayControl", true, classLoader);

var loadLibraryMethod = Runtime.class.getDeclaredMethod("loadLibrary0", Class.class, String.class);
loadLibraryMethod.setAccessible(true);
loadLibraryMethod.invoke(Runtime.getRuntime(), displayControlClass, "android_servers");
```

rom1v:
@yume-chan Wow, that works! Thanks :tada:

I will rework my PR without a separate process then :smile:  (even if this adds a lot of private calls)

Could you please detail why this particular ClassLoader with these arguments? (I will investigate later)

---

Just a note to match the parameter names:

```java
    public static ClassLoader createClassLoader(String dexPath,
            String librarySearchPath, String libraryPermittedPath, ClassLoader parent,
            int targetSdkVersion, boolean isNamespaceShared, String classLoaderName) {
```

https://github.com/aosp-mirror/platform_frameworks_base/blob/0e28d56b1493b67ed0e1524065c07408f7608f99/core/java/com/android/internal/os/ClassLoaderFactory.java#L110-L112

yume-chan:
First I found Android has its own dynamic library linker (libdl) in bionic, and it has a feature called [linker namespace](https://fadeevab.com/android-linker-namespace-security-flaws/), that can limit which libraries a caller can load.

Android Runtime utilizes this feature by [associating class loaders with linker namespaces](https://cs.android.com/android/platform/superproject/main/+/main:art/libnativeloader/library_namespaces.h;l=38-41;drc=e741dfd18dcd15f002bc1db9bd6634322e4eeef8). app_process will create a class loader for `services.jar` [with a shared linker namespace](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/com/android/internal/os/ZygoteInit.java;l=496-510;drc=5ec0f632e5e2587091ab8d118cf29c6887081c0d), but create a class loader for user code (for example `scrcpy-server.jar`) with an isolated and limited linker namespace. That's why calling `System.loadLibrary("android_servers")` in our code will fail. (by using LD_PRELOAD, the library is loaded immediately using the "default" linker namespace, so it succeeds)

A shared namespace means it [can load all libraries from its parent namespace](https://cs.android.com/android/platform/superproject/main/+/main:bionic/linker/linker.cpp;l=2480-2501;drc=5ec0f632e5e2587091ab8d118cf29c6887081c0d). For app_process, the parent namespace should be "default", which can load libraries from any locations.

In Android Runtime, the call stack is `System.loadLibrary` -> [`JavaVMExt::LoadNativeLibrary`](https://cs.android.com/android/platform/superproject/main/+/main:art/runtime/jni/java_vm_ext.cc;l=931;drc=5ec0f632e5e2587091ab8d118cf29c6887081c0d) -> [`android::OpenNativeLibrary`](https://cs.android.com/android/platform/superproject/main/+/main:art/libnativeloader/native_loader.cpp;l=209;drc=b94401e6e15903b64de709fd69863b6880cedd7c). The last method finds the associated linker namespace for `class_loader` parameter (or create an isolated one if it doesn't have one, this is the case for the class loader that loads user code), then calls `android_dlopen_ext` in bionic to load the library using that linker namespace.

So we need a class loader that 1) has a shared linker namespace so it can load `libandroid_servers.so` and its dependencies in /apex folder and 2) has Java classes from `services.jar` so JNI initialization in `libandroid_servers.so` can find the classes it wants.

The `dexPath` argument includes "/system/framework/services.jar" to satisfy 2), and `isNamespaceShared` argument is `true` to satisfy 1). Other arguments don't seem to matter, with `parent: null` and `targetSdkVersion: 0` it still works. 

We also need to actually load `libandroid_servers.so` using that class loader. [`com.android.server.SystemServer.main`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/services/java/com/android/server/SystemServer.java;l=664;drc=d34869871fe5df392d97a80ffb34d52b9bda7384) contains [`System.loadLibrary("android_servers");`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/services/java/com/android/server/SystemServer.java;l=871;drc=d34869871fe5df392d97a80ffb34d52b9bda7384), but calling it using reflection fails with initializing system services.`System.loadLibrary` (internally [`Runtime.loadLibrary`](https://cs.android.com/android/platform/superproject/main/+/main:libcore/ojluni/src/main/java/java/lang/Runtime.java;l=982-985;drc=6b0ec6aa3395d4799ec10063f7e96ec07f1ebfca)) uses caller class's class loader, so I have to call `Runtime.loadLibrary0` using reflection, as it accepts a custom class (and class loader).

rom1v:
Superseded by #4456

yume-chan:
FYI, using direct communication with SurfaceFlinger

```java
var serviceManagerClass = Class.forName("android.os.ServiceManager");
var getServiceMethod = serviceManagerClass.getDeclaredMethod("getService", String.class);
var surfaceComposer = (IBinder) getServiceMethod.invoke(null, "SurfaceFlingerAIDL");

long[] displayIds;
{
    var data = Parcel.obtain();
    var reply = Parcel.obtain();
    try {
        data.writeInterfaceToken("android.gui.ISurfaceComposer");
        surfaceComposer.transact(IBinder.FIRST_CALL_TRANSACTION + 5, data, reply, 0);
        reply.readException();
        var size = reply.readInt();
        displayIds = new long[size];
        for (var i = 0; i < size; i += 1) {
            displayIds[i] = reply.readLong();
        }
    } finally {
        data.recycle();
        reply.recycle();
    }
}

var surfaceControlClass = Class.forName("android.view.SurfaceControl");
var setDisplayPowerModeMethod = surfaceControlClass.getDeclaredMethod("setDisplayPowerMode", IBinder.class, int.class);

for (long displayId : displayIds) {
    IBinder token;
    {
        var data = Parcel.obtain();
        var reply = Parcel.obtain();
        try {
            data.writeInterfaceToken("android.gui.ISurfaceComposer");
            data.writeLong(displayId);
            surfaceComposer.transact(IBinder.FIRST_CALL_TRANSACTION + 6, data, reply, 0);
            reply.readException();
            token = reply.readStrongBinder();
        } finally {
            data.recycle();
            reply.recycle();
        }
    }

    setDisplayPowerModeMethod.invoke(null, token, 0);
}
```

rom1v:
Thank you for the sample. It's great to have several ways to access it :+1:

But I'm afraid that it would be too "fragile" (and will fail on some devices without useful error messages if the transaction id doesn't match).

Moreover, this [commit](https://android.googlesource.com/platform/frameworks/native/+/9e84f339bf%5E%21/) says:

> Additonal service, named as "SurfaceFlingerAIDL", is added to surfaceflinger during the process of migrating ISurfaceComposer interface to AIDL. New changes are put into namespace, android::gui. Once migration is complete, this service will be deleted.

So IIUC, this solution may not be permanent.

PR #4448, opened 2023-11-21
Fix Linux desktop files

### LINKED ISSUE #4367: MR #3817 Breaks launching scrcpy-console using the .desktop file
- [x ] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [ x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Arch Linux
 - scrcpy version: Master
 - installation method: From source via AUR
 -  model: N/A
 - Android version: N/A

**Describe the bug**
https://github.com/Genymobile/scrcpy/pull/3817 breaks `scrcpy-console`, because there are one too many backslashes in the `Exec` line. 

`/bin/bash --norc --noprofile -i -c "\"\\$SHELL\" -i -c scrcpy || read -p 'Press Enter to quit...'"` 

errors out, because it reads it as `\/bin/zsh` (or `\/bin/bash`, `\/bin/fish`, etc) and gives you a `\/bin/zsh: no such file or directory` error. 

Removing one of the backslashes before `$SHELL` fixes the problem completely and you can launch scrcpy-console from the .desktop file as intended.  

```
bash: \/bin/zsh: No such file or directory
Press Enter to quit...
```
Here's what happens when I run `/bin/bash --norc --noprofile -i -c "\"\$SHELL\" -i -c scrcpy || read -p 'Press Enter to quit...'"`:

```
scrcpy 2.1.1 <https://github.com/Genymobile/scrcpy>
ERROR: Could not find any ADB device
Press Enter to quit...
```
Obviously I don't have adb running so that error can be ignored, when I have adb running and a device connected, it functions. 

This may be a new issue due to Plasma updating or , but during my update I got this warning and many similar other warnings about other desktop files:

```
kf.config.core: "KConfigIni: In file /usr/share/applications/scrcpy-console.desktop, line 8: " "Invalid escape sequence \"\\\"\"."

kf.config.core: "KConfigIni: In file /usr/share/applications/scrcpy-console.desktop, line 8: " "Invalid escape sequence \"\\\"\"."
```

kidonng:
Fix #4367

I have checked the new desktop files with `desktop-file-validate`.

### DIFF
--- a/app/data/scrcpy-console.desktop
+++ b/app/data/scrcpy-console.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/sh -c "\"\\$SHELL\" -i -c scrcpy --pause-on-exit=if-error"
+Exec=/bin/sh -c "\"\$SHELL\" -i -c scrcpy --pause-on-exit=if-error"
 Icon=scrcpy
 Terminal=true
 Type=Application
--- a/app/data/scrcpy-console.desktop
+++ b/app/data/scrcpy-console.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/sh -c "\"\$SHELL\" -i -c scrcpy --pause-on-exit=if-error"
+Exec=/bin/sh -c "\\$SHELL -i -c scrcpy --pause-on-exit=if-error"
 Icon=scrcpy
 Terminal=true
 Type=Application
--- a/app/data/scrcpy.desktop
+++ b/app/data/scrcpy.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/sh -c "\"\\$SHELL\" -i -c scrcpy"
+Exec=/bin/sh -c "\"\$SHELL\" -i -c scrcpy"
 Icon=scrcpy
 Terminal=false
 Type=Application
--- a/app/data/scrcpy.desktop
+++ b/app/data/scrcpy.desktop
@@ -5,7 +5,7 @@ Comment=Display and control your Android device
 # For some users, the PATH or ADB environment variables are set from the shell
 # startup file, like .bashrc or .zshrc… Run an interactive shell to get
 # environment correctly initialized.
-Exec=/bin/sh -c "\"\$SHELL\" -i -c scrcpy"
+Exec=/bin/sh -c "\\$SHELL -i -c scrcpy"
 Icon=scrcpy
 Terminal=false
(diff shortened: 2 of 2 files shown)


kidonng:
I'm not sure why `$SHELL` is additionally quoted in the original PR (https://github.com/Genymobile/scrcpy/pull/3351/commits/953531fc39bb7e3ae3869c3fa57cea6fd5da2f54), but I think it's too cautious (`$SHELL` should not contain characters that needs to be quoted anyway) and just complicates the config. I'm removing the extra quotes so it passes `desktop-file-validate` AND does work in DE.

rom1v:
cc @voyageur #3817

kidonng:
I was referring to #3351, when the desktop files were initially added. (still, thanks for the effort!)

The PR breaking the desktop files (#3817) is technically not wrong, by `desktop-file-validate` standard. It's just that this check utility, the desktop environment, and an actual shell do not share the same behavior regarding quoting and escaping.

rom1v:
> I'm removing the extra quotes so it passes desktop-file-validate AND does work in DE.

If you don't remove them, what is the problem?

rom1v:
> If you don't remove them, what is the problem?

It seems to work fine in both cases. I prefer the version which quotes `$SHELL`, so I propose to merge only the first part (without removing the quotes): https://github.com/Genymobile/scrcpy/commit/d720c34a28f8456b8d4213e5630a9625314fcf6f

Is it ok for you?

kidonng:
> If you don't remove them, what is the problem?

I don't see there's a `Exec` that

1. Keep the extra quotes around `$SHELL`
2. Pass `desktop-file-validate` (the motivation for the original change)
3. **Actually work in desktop environments**

And as I mentioned above, the first point has not practical value (if your `$SHELL` needs to be quoted, you have much bigger trouble awaiting), and just being a hassle for the latter two points.

> It seems to work fine in both cases.

How did you test it? Because it certainly didn't work in KDE (on my Arch Linux machine).

Testing it with a shell does not help, as I said above, since desktop environments can have different behavior. It has to work in launchers provided by desktop environments, that's what the desktop file is for!

rom1v:
> How did you test it? Because it certainly didn't work in KDE (on my Arch Linux machine).

I tested on my machine, with XFCE. I `ninja install` scrcpy, and clicked on `scrcpy` and `scrcpy-console` in the XFCE menu, and it worked.

So IIUC https://github.com/Genymobile/scrcpy/commit/d720c34a28f8456b8d4213e5630a9625314fcf6f does not work on KDE, right?

voyageur:
Yes, in a previous issue we had confirmed that KDE did not care much about the spec (https://github.com/Genymobile/scrcpy/pull/3822#issuecomment-1471478819 and below comments), this can be seen if tested with gtk-launch or gio launch.
Maybe time to add the mentioned "--wait-on-error" option?

kidonng:
I'll admit it's my fault for not pointing out I'm testing in KDE earlier (I had intented to mention it, but discarded it as irrelevant as KDE is a big player worth supporting anyway).

@voyageur Thanks for confirming.

rom1v:
Ok, so I removed the quotes around `$SHELL`.

Then I noticed that `scrcpy-console` just did not pause on error as expected, because `scrcpy --pause-on-exit=if-error` must be passed as a single argument (so I added quotes).

I prepared branch [`pr4448`](https://github.com/Genymobile/scrcpy/commits/pr4448) ready to be merged. Is it ok for you?

rom1v:
Merged into `dev`.

PR #4460, opened 2023-11-24
Improve crossbuild

rom1v:
Install all the prebuilt dependencies for Windows to a specific folder, and use meson command line options to specify their location.

This removes crossbuild-specific code from the meson scripts and will simplify dependency upgrades.


rom1v:
67f356f881898e21edabd4ecd2fb509ad4c92362

PR #4473, opened 2023-11-28
[WIP] UHID mouse/keyboard input

### LINKED ISSUE #3110: OTG / HID mode using Bluetooth

OTG / Keyboard HID / Mouse HID modes require a physical USB connection to the computer, even though Android supports connecting to Bluetooth HIDs.

Implement an optional Bluetooth HID connection method in addition to the existing libusb method

Emulating a USB connection - would either require additional hardware (e.g. a USB dongle) or emulation software on the Android device as well.
Sharing devices over USB/IP - no client software available for android devices.
Bluetooth HID emulation (e.g. https://github.com/xenogenesi/hidclient) - requires manual binding & unbinding through the command line and only works on Linux (depends on libbluetooth / BlueZ)

### LINKED ISSUE #3913: HID/OTG on localhost to enable IME input
- [x ] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

I would like to use --hid-keyboard but with a virtual device (ReDroid; Android in docker container) running on the same machine as scrcpy. So there is no real USB connection and thus it currently falls back:
`scrcpy -s localhost:5555 --window-width 400 --audio-codec=aac --audio-encoder='OMX.google.aac.encoder' --hid-keyboard`
```
/usr/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 483.0 MB/s (52867 bytes in 0.000s)
[server] INFO: Device: redroid redroid12_x86_64 (Android 12)
ERROR: Could not find USB device localhost:5555
ERROR:         0000:2d:00.3       (1d6b:0003)  Linux 6.2.8-x64v1-xanmod1-1 xhci-hcd xHCI Host Controller
ERROR:         31C6BCA0           (046d:0825)  (null) (null)
ERROR:         0000:2d:00.3       (1d6b:0002)  Linux 6.2.8-x64v1-xanmod1-1 xhci-hcd xHCI Host Controller
ERROR:         0000:02:00.0       (1d6b:0003)  Linux 6.2.8-x64v1-xanmod1-1 xhci-hcd xHCI Host Controller
ERROR:         A02020051203       (1462:7c94)  MSI MYSTIC LIGHT 
ERROR:         0000:02:00.0       (1d6b:0002)  Linux 6.2.8-x64v1-xanmod1-1 xhci-hcd xHCI Host Controller
ERROR: Fallback to default keyboard injection method (-K/--hid-keyboard ignored)
INFO: Renderer: opengl
INFO: OpenGL version: 4.6 (Compatibility Profile) Mesa 22.3.4
INFO: Initial texture: 1080x2160
[server] WARN: Could not inject char u+3063
```

A parameter that can be added to make --hid-input functionality work for that case.

fcitx5 on Android but it's by far not ready yet

**Additional context**
I use the input method framework fcitx5 on my Arch Linux desktop. I switch layouts pretty often which only takes a simple key shortcut.

### LINKED ISSUE #4034: A better method to simulate HID-Keyboard and HID-Mouse for Android device
https://github.com/torvalds/linux/blob/master/samples/uhid/uhid-example.c

We can utilize the Linux uhid API to create a HID input device, and having ADB permission is sufficient to achieve this.

This method doesn't require an actual USB wire to be connected to the Android device, and allows the HID keyboard and mouse functionality to coexist with other features of scrcpy.

However, this method requires scrcpy to load an ELF file, as the uhid API must be called in JNI.

yume-chan:
Fixes #4034 
Fixes #3110
Fixes #3913

Currently only the keyboard part is implemented to get the overall architecture reviewed.

As described in https://github.com/Genymobile/scrcpy/issues/4034#issuecomment-1826714479, the client is responsible for creating and parsing UHID messages, and the server only do the reads and writes to UHID  file descriptors. This allows reusing existing AoA mouse/keyboard code to do the HID device emulation, and use native C code to handle UHID messages (which is also defined in C headers).

The new option is named `--keyboard-input-mode`, after its internal name. Available values are `disable`, `inject`, `aoa` and `uhid`. `aoa` deprecates the `-K`/`--hid-keyboard` option. When `--otg` is not specified, all values are accepted, except `aoa` doesn't work on Windows. When `--otg` is specified, only `disable` and `aoa` are accepted. When `--keyboard-input-mode` is not speficied, it defaults to `inject` in normal mode and `aoa` in OTG mode.

Three new control messages are added: `UHID_OPEN`, `UHID_WRITE` and `UHID_CLOSE`. Each UHID device is a separate file descriptor, so all messages have an `int id` field, generated by client, maps to a file descriptor on server.

Currently `UHID_OPEN` and `UHID_WRITE` both have a `byte[] data` field (containing a `uhid_event` structure), so opening file and registering device can be done in one message. But this design is a little bit strange. Maybe we don't need `UHID_OPEN` at all.

One new device messages are added: `UHID_DATA`. It contains a `uhid_event` structure read from the file descriptor. In theory the client should wait for a `UHID_DATA` messages with `START` and `OPEN` UHID message types before starting to send inputs, but in practice these two arrives pretty quickly, so when user starts typing, UHID already finished initializing.

`UHID_DATA` can also contain `OUTPUT` UHID messages, these are HID output reports defined in the device's report descriptor, for example LED status for keyboards and rumbles for gamepads. The code to handle these messages haven't been added to the client.


rom1v:
Thank you very much :+1:

Sorry for the delay, I don't have much time currently. After a quick glance, it looks good :+1:, but I need to review in more details to better understand how everything works.

One problem I encounter however is that it does not use the correct keyboard mapping.

If I use AOA, I can select the physical keyboard mapping once ([doc](https://github.com/Genymobile/scrcpy/blob/master/doc/hid-otg.md#physical-keyboard-simulation) | https://github.com/Genymobile/scrcpy/pull/2632#issuecomment-923756915), and it works fine. But with UHID it is not used, so when I press `é` (French keyboard) it injects `2` (US keyboard). Do you know a solution for this problem?

yume-chan:
Keyboard layout setting is also available for UHID virtual keyboards.

image

I think Android system can distinguish different keyboards and save different layout settings for each device.

The display name is set here: https://github.com/Genymobile/scrcpy/pull/4473/files#diff-ee5bd2180bba9b212ca2eb2799d4272e64029b26751a4e305beac776124aee93R25

rom1v:
Awesome :ok_hand:

yume-chan:
I added keyboard output report parsing. After the virtual keyboard is registered, or when keyboard modifier status changes, the device will send an output report for keyboard LED status. The client can combine this information with computer keyboard modifier key status to make sure they are in sync.

For example, if the user toggles caps lock outside Scrcpy, on the next key event, the client can detect the local modifier status is different from what is on device, and send an extra caps lock key press event to correct that.

rom1v:
As a preliminary work, I started to extract the replacement of `-K`/`--hid-keyboard` and `-M`/`--hid-mouse` by `--keyboard-input-mode=…` and `--mouse-input-mode=…` to a separate commit (before the introduction of uhid).

But I think that `--keyboard-input-mode=` and `--mouse-input-mode=` are too long as command line parameter names.

Maybe `--keyboard-im=` and `--mouse-im=`? Or even `--keyboard=` and `--mouse=` (but probably not explicit enough)? What do you think?

yume-chan:
I think only `--keyboard` and `--mouse` is fine. Maybe they would block us from adding some unrelated `--keyboard-XX` options in the future, but I can't think of any concrete examples.

rom1v:
One concrete example could be an option to install and switch to a specific keyboard on the device (AdbKeyboard, NullKeyboard…). Even if such an option is not planned, it could be named `--keyboard=…`.

rom1v:
I like `--keyboard` and `--mouse`, so as preliminary commits I added these options with 3 possible values: `disabled`, `sdk` and `aoa`.

Branch [`uhid.3`](https://github.com/Genymobile/scrcpy/commits/uhid.3/) (3 last commits).

app/src/hid/uhid.h
```diff
@@ -0,0 +1,207 @@
...
+	uint8_t phys[64];
+	uint8_t uniq[64];
+	uint16_t rd_size;
+	uint16_t bus;
+	uint32_t vendor;
+	uint32_t product;
+	uint32_t version;
+	uint32_t country;
+	uint8_t rd_data[HID_MAX_DESCRIPTOR_SIZE];
+} __attribute__((__packed__));
```
rom1v, line 63: This is a structure compiled for the computer architecture, it should not be used as is to pass it to the device architecture (in theory, endianness could be different for example).

IMO, the client should only send the descriptor, and the device construct the expected uhid structure on the server side.

rom1v (commented):
Currently, the AOA/HID keyboard handles two different things:
 1. create the HID events
 2. send them over AOA

To reuse the HID part but using another protocol, your PR abstracts (2) (how to send them):
 - introduce a new trait `sc_hid_interface`, implemented by `sc_aoa` and `sc_uhid`
 - an instance is passed to construct a `sc_hid_keyboard`

This works, but it brings some complexity:
 - two levels of interfaces implementations (something implementing a `sc_key_processor`, with one implementation receiving something implementing a `sc_hid_interface`)
 - the need for an `async_message` flag (because not all implementations behave exactly the same)
 - all implementation have to `create()`/`destroy()` on the client-side (for UHID, I would prefer to create the UHID struct on the server side, the client should not have to handle all these details, like `vendor_id`/`product_id`, and the UHID handle should probably be closed on exit from the server side, without any message from the client)

I think "abstracting" the other way around ((1) extracting how to create the HID events) leads to a simpler solution (with less coupling).

My idea is to extract a component to create HID events (regardless of how they will be sent), so that separate keyboard implementations could reuse that same HID tools (without interfaces/traits) when they need to.

That way, we could get 3 separate keyboard/mouse "independant" implementations:
 - `sc_keyboard_sdk` (previously `sc_keyboard_inject")
 - `sc_keyboard_aoa` (previously `sc_hid_keyboard`)
 - `sc_keyboard_uhid`

Concretly:
 - `app/src/hid` would contain HID tools
 - `app/src/usb` would contain the AOA keyboard/mouse implementations
 - `app/src/uhid` would contain the UHID keyboard/mouse implementations

I implemented preliminary commits which perform all these refactors with the current features: [`uhid.10`](https://github.com/Genymobile/scrcpy/commits/uhid.10/)

From this branch, you could more easily add a UHID implementation (independently of the AOA keyboard).

What do you think?

yume-chan:
> I think "abstracting" the other way around ((1) extracting how to create the HID events) leads to a simpler solution (with less coupling).

I agree with that. In fact, my app is using this architecture.

app/src/hid/uhid.h
yume-chan: > IMO, the client should only send the descriptor, and the device construct the expected uhid structure on the server side.

I'd like letting the client to specify device names (like `Scrcpy UHID mouse`/`Scrcpy UHID keyboard`), because the server doesn't know the device's type.

Client also needs to be able to specify `vendor` and `product`, because my gamepad rumble implementation only works with a specific pair of those (by emulating a Bluetooth Xbox One controller to load a special driver).

Because little-endian is so dominant on modern PC and Android, I don't know whether I need to deal with endianness even if those structures are created in Java. The header file is copied from Linux kernel so using C to handle them is the simplest way.

rom1v: > I'd like letting the client to specify device names (like Scrcpy UHID mouse/Scrcpy UHID keyboard), because the server doesn't know the device's type.

Sure. I'd pass all the necessary fields in a control message structure, with serialization/deserialization ((like the other control messages).

rom1v: > I don't know whether I need to deal with endianness even if those structures are created in Java.

I guess so, you must pass a `byte[]` with the expected representation to your `UHidManager.write()` method.
You should probably use a `ByteBuffer` to build it (ref `ByteBuffer.order(…)`, `ByteOrder.nativeOrder()`).

To explain a bit more, I think it is incorrect that the client knows the UHID implementation details (including endianness, some fields that the client does not care about…). Moreover, it is possible to use scrcpy-server standalone (and control it from python for example). Even if we do not absolutely need to consider this usage, it still seems weird to expect to pass a Linux C-struct for UHID messages.

castillofrancodamian:
How do I change the keyboard language? I already tried changing the language of the physical keyboard on Android but it still doesn't work.

app/src/hid/uhid.h
rom1v: I implemented that on [uhid.15](https://github.com/Genymobile/scrcpy/commits/uhid.15/) (cf `UhidManager`):
 - the client sends UHID_CREATE and UHID_INPUT control messages containing HID data
 - the server (device) creates the binary UHID structures to the file descriptor

It works well, but I haven't taken the commits for CAPSLOCK synchronization (HID output) yet.

server/src/main/java/com/genymobile/scrcpy/UHidManager.java
```diff
@@ -0,0 +1,98 @@
...
+            Ln.e("Failed to open uhid", e);
+            return;
+        }
+
+        mHandles.put(id, fd);
+        queue.addOnFileDescriptorEventListener(fd, MessageQueue.OnFileDescriptorEventListener.EVENT_INPUT | MessageQueue.OnFileDescriptorEventListener.EVENT_OUTPUT, (fd2, event) -> {
+            if (event == MessageQueue.OnFileDescriptorEventListener.EVENT_INPUT) {
+                try {
+                    buffer.rewind();
+                    Os.read(fd2, buffer);
```
rom1v, line 57: Do you know why it reads 4380 bytes (with only the first byte being non-zero)?

That does not match `struct uhid_output_req` either ([documentation](https://raw.githubusercontent.com/torvalds/linux/master/Documentation/hid/uhid.rst)):
https://github.com/torvalds/linux/blob/603c04e27c3e9891ce7afa5cd6b496bfacff4206/include/uapi/linux/uhid.h#L81-L85

app/src/hid/hid_keyboard.c
```diff
@@ -381,20 +401,39 @@ sc_key_processor_process_key(struct sc_key_processor *kp,
...
 }
 
+static void
+sc_hid_keyboard_process_output(struct sc_hid_device* device, int16_t report_id, const uint8_t *data, size_t size) {
+    (void) report_id;
+    (void) size;
+
+    struct sc_hid_keyboard *kb = DOWNCAST_HID_DEVICE(device);
+    kb->device_mod = 0;
+    if (data[0] & 0x01) {
```
rom1v, line 418: Where did you find these values?

server/src/main/java/com/genymobile/scrcpy/UHidManager.java
yume-chan: 4380 is the size of `uhid_event`. It always reads/writes full `uhid_event`s.

app/src/hid/hid_keyboard.c
yume-chan: From HID usage table.

> # 11 LED Page (0x08)
>
> | Usage ID | Usage Name  | Usage Types | Section |
> | -------- | ----------- | ----------- | ------- |
> | 00       | Undefined   |             |         |
> | 01       | Num Lock    | OOC         | 11.1    |
> | 02       | Caps Lock   | OOC         | 11.1    |
> | 03       | Scroll Lock | OOC         | 11.1    |
> | 04       | Compose     | OOC         | 11.1    |
> | 05       | Kana        | OOC         | 11.1    |
> | 06       | Power       | OOC         | 11.6    |
>
> [...]

rom1v:
Here is my current branch: [uhid.20](https://github.com/Genymobile/scrcpy/commits/uhid.20/).

UHID and mod synchronization work. Please review and test. :)

I also added a new shortcut MOD+k to open physical keyboard settings easily.

Next steps:
 - documentation
 - HID mouse

rom1v:
I fixed several issues, and implemented mouse HID: [uhid.24](https://github.com/Genymobile/scrcpy/commits/uhid.24/)

rom1v:
Note: on a Nexus 5 with Android 6, I get a permission denied when opening `/dev/uhid` (branch [`uhid.25`](https://github.com/Genymobile/scrcpy/commits/uhid.25/)):

```
java.io.IOException: android.system.ErrnoException: open failed: EACCES (Permission denied)
```

```console
$ adb shell ls /dev/uhid
/dev/uhid: Permission denied
```

rom1v:
Oops, indeed some files were not committed.

I forced pushed `uhid.27` with the missing files.

rom1v:
> diff1

Thank you, fixed locally.

> diff2

This is due to not exceed 80 chars (that's maybe a stupid rule, but for consistency I stick to it).

> I'm probably wrong but I'm wondering, maybe this could allow you to also handle touch events even though the mouse is in relative coordinates?

I'm pretty sure it should be possible to simulate a HID touch screen (with absolute coordinates). To be investigated…

rom1v:
https://github.com/Genymobile/scrcpy/pull/4473#issuecomment-1963004171

So apparently it fails on Android 6. It works on Android 11.

Could you please test on Android 7, 8, 9 and 10 to determine from which Android version it works?

yume-chan:
I can only confirm it works on Android 10.

`uhid.32` branch LGTM, I only corrected some typos: https://github.com/yume-chan/scrcpy/commit/503d0f7f9128015f0f34f77d3fc5d32865fd02d1

rom1v:
Thank you for your review :+1: I fixed on `uhid.34`.

I'll do another review pass before merging.

rom1v:
`uhid.38` merged into `dev` :rocket:

chldbwnstm:
Hi @yume-chan ,

Thanks for the update. I was just wondering if I could send a software message to scrcpy server like
click(300,300) and it should click the coordinates using the scrcpy server's mouse input structure.
Can this be achieved with the new UHID mouse feature?

PR #4502, opened 2023-12-04
Add workaround for Android U

SwallowJoe:
On android U, InputManager.getInstance() calls ActivityThread.currentApplication(), which requires a non-null Application.


SwallowJoe:
Just checked again, version 2.3.1 has been adapted to Android 14.

```java
    public static Class<?> getInputManagerClass() {
        try {
            // Parts of the InputManager class have been moved to a new InputManagerGlobal class in Android 14 preview
            return Class.forName("android.hardware.input.InputManagerGlobal");
        } catch (ClassNotFoundException e) {
            return android.hardware.input.InputManager.class;
        }
    }
```
so close.

PS. I use v2.0 for test so make this mistake.

PR #4529, opened 2023-12-13
Simulate tilt multitouch event by pressing Shift

tillrathmann:
I missed simulating tilt multitouch events (two fingers moving up or down) for map apps.
Here is an implementation for that.

### DIFF
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -76,6 +76,8 @@ sc_input_manager_init(struct sc_input_manager *im,
     im->sdl_shortcut_mods.count = shortcut_mods->count;
 
     im->vfinger_down = false;
+    im->vfinger_invert_x = false;
+    im->vfinger_invert_y = false;
 
     im->last_keycode = SDLK_UNKNOWN;
     im->last_mod = 0;
@@ -347,9 +349,14 @@ simulate_virtual_finger(struct sc_input_manager *im,
 }
 
 static struct sc_point
-inverse_point(struct sc_point point, struct sc_size size) {
-    point.x = size.width - point.x;
-    point.y = size.height - point.y;
+inverse_point(struct sc_point point, struct sc_size size,
+              bool invert_x, bool invert_y) {
+    if (invert_x) {
+        point.x = size.width - point.x;
+    }
+    if (invert_y) {
+        point.y = size.height - point.y;
+    }
     return point;
 }
 
@@ -605,7 +612,9 @@ sc_input_manager_process_mouse_motion(struct sc_input_manager *im,
         struct sc_point mouse =
            sc_screen_convert_window_to_frame_coords(im->screen, event->x,
                                                     event->y);
-        struct sc_point vfinger = inverse_point(mouse, im->screen->frame_size);
+        struct sc_point vfinger = inverse_point(mouse, im->screen->frame_size,
+                                                im->vfinger_invert_x,
+                                                im->vfinger_invert_y);
         simulate_virtual_finger(im, AMOTION_EVENT_ACTION_MOVE, vfinger);
     }
 }
@@ -726,7 +735,7 @@ sc_input_manager_process_mouse_button(struct sc_input_manager *im,
         return;
     }
 
-    // Pinch-to-zoom simulation.
+    // Pinch-to-zoom, rotate and tilt simulation.
     //
     // If Ctrl is hold when the left-click button is pressed, then
     // pinch-to-zoom mode is enabled: on every mouse event until the left-click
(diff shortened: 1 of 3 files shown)


rom1v:
Thank you. I like the idea.

However, <kbd>Alt</kbd> + mouse move is captured by my window manager (to move the window), so I can't use it.

Since there are 2 axis, we can use for example:
 - <kbd>Ctrl</kbd> to mirror the mouse horizontally (to slide vertically) (or the reverse)
 - <kbd>Shift</kbd> to mirror the mouse vertically (to slide horizontally) (or the reverse)
 - <kbd>Ctrl</kbd>+<kbd>Shift</kbd> to mirror in both directions (what already exists in the current version)

One drawback is that this cause a shortcut change for the existing action, but it seems more logical to me.

What do you think?

tillrathmann:
Ah, good point.
Two more alternatives:
- We could keep the current behavior for <kbd>Ctrl</kbd> and use <kbd>Shift</kbd> for mirroring horizontally (to slide vertically / tilt gesture). Mirroring only vertically could be dropped, as sliding horizontally with two fingers is no common gesture I think.
- The same, but additionally provide <kbd>Alt</kbd> for this uncommon vertical only mirroring. If this is rarely used it is probably no big deal that it is not reachable if prevented by the window manager. But if someone really needs to use such a gesture it can be circumvented by temporarily switching to a different window manager or disabling the <kbd>Alt</kbd> + mouse move capturing.

So, what to do? Your proposal or one of these alternatives?

rom1v:
> We could keep the current behavior for <kbd>Ctrl</kbd> and use <kbd>Shift</kbd> for mirroring horizontally (to slide vertically / tilt gesture). Mirroring only vertically could be dropped, as sliding horizontally with two fingers is no common gesture I think.

I like this one :+1: Indeed I think that horizontal sliding is not common.

If <kbd>Ctrl</kbd>+<kbd>Shift</kbd> are pressed, either:
 - do nothing (disable virtual finger)
 - only consider the first pressed one (<kbd>Ctrl</kbd> then <kbd>Shift</kbd> is the same <kbd>Ctrl</kbd>, <kbd>Shift</kbd> then <kbd>Ctrl</kbd> is the same <kbd>Shift</kbd>)

(choose the one with the simplest implementation)

(<kbd>Ctrl</kbd>+<kbd>Shift</kbd> could be reserved for the less common horizontal sliding for later if necessary)

tillrathmann:
Alright, let's do that!

rom1v:
:ok_hand: 

I just reformulated a bit because technically it's an axial symmetry:

```diff
diff --git a/app/src/input_manager.c b/app/src/input_manager.c
index 13f5a4908..76cfbd928 100644
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -745,9 +745,10 @@ sc_input_manager_process_mouse_button(struct sc_input_manager *im,
     // In other words, the center of the rotation/scaling is the center of the
     // screen.
     //
-    // To simulate a tilt gesture Shift can be used instead of Ctrl. The "virtual
-    // finger" has the x coordinate inverted through the center of the screen
-    // while the y coordinate is unchanged.
+    // To simulate a tilt gesture (a vertical slide with two fingers), Shift
+    // can be used instead of Ctrl. The "virtual finger" has a position
+    // inverted with respect to the vertical axis of symmetry in the middle of
+    // the screen.
     const SDL_Keymod keymod = SDL_GetModState();
     const bool ctrl_pressed = keymod & KMOD_CTRL;
     const bool shift_pressed = keymod & KMOD_SHIFT;
```

(Please do not hesitate to suggest a reformulation if my English is incorrect.)

I also updated some more documentation:

```diff
diff --git a/app/scrcpy.1 b/app/scrcpy.1
index 8ca4a7730..beaa99ab4 100644
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -642,7 +642,11 @@ Enable/disable FPS counter (print frames/second in logs)
 
 .TP
 .B Ctrl+click-and-move
-Pinch-to-zoom from the center of the screen
+Pinch-to-zoom and rotate from the center of the screen
+
+.TP
+.B Shift+click-and-move
+Tilt (slide vertically with two fingers)
 
 .TP
 .B Drag & drop APK file
diff --git a/app/src/cli.c b/app/src/cli.c
index fd4525f55..c580c9597 100644
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -947,7 +947,11 @@ static const struct sc_shortcut shortcuts[] = {
     },
     {
         .shortcuts = { "Ctrl+click-and-move" },
-        .text = "Pinch-to-zoom from the center of the screen",
+        .text = "Pinch-to-zoom and rotate from the center of the screen",
+    },
+    {
+        .shortcuts = { "Shift+click-and-move" },
+        .text = "Tilt (slide vertically with two fingers)",
     },
     {
         .shortcuts = { "Drag & drop APK file" },
diff --git a/doc/shortcuts.md b/doc/shortcuts.md
index c0fc28421..21bccbd95 100644
--- a/doc/shortcuts.md
+++ b/doc/shortcuts.md
@@ -49,7 +49,8 @@ _<kbd>[Super]</kbd> is typically the <kbd>Windows</kbd> or <kbd>Cmd</kbd> key._
  | Synchronize clipboards and paste⁵           | <kbd>MOD</kbd>+<kbd>v</kbd>
  | Inject computer clipboard text              | <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>v</kbd>
  | Enable/disable FPS counter (on stdout)      | <kbd>MOD</kbd>+<kbd>i</kbd>
- | Pinch-to-zoom                               | <kbd>Ctrl</kbd>+_click-and-move_
+ | Pinch-to-zoom/rotate                        | <kbd>Ctrl</kbd>+_click-and-move_
+ | Tilt (slide vertically with 2 fingers)      | <kbd>Shift</kbd>+_click-and-move_
  | Drag & drop APK file                        | Install APK from computer
  | Drag & drop non-APK file                    | [Push file to device](control.md#push-file-to-device)
 
```

The resulting commit is here: d2ed4510a76e0d273f9088bce6e85334b922c9cb

tillrathmann:
Looks fine! :thumbsup:

rom1v:
Merged into `dev` :rocket:

Issue #4179, opened 2023-07-20
scrcpy --otg TWO devices AT ONCE?


achimghita:
Is it possible to control 2 (ore more devices) via OTG in the same time? Assuming I have 2 identical phones connected through a usb hub to the computer, one with broken screen, can I mirror the mouse actions on the 2 devices? Plan is to activate usb debugging.

rom1v:
Not at the same time.

In the future, I might consider implementing a feature to record a HIID/OTG session to a file, and replay it later, for this exact use case.

achimghita:
That would be awesome! Meanwhile, do you think a hacked cable would work for this purpose? 1 usb in (mouse) and 2 usb out (one for each phone)

sukumar1210:
@rom1v Hi, as Hacktober Fest is approaching, i wanted to know that if this repo is going to be there in the fest, if so i would like to contribute to this issue.

PR #4556, opened 2023-12-31
Add --hid-capture & --hid-replay options

Rob--W:
This PR introduces the `--hid-record` and `--hid-replay` flags to support the use case of capturing events and replaying it (#4468). I have designed it such that it can also replay from a stream, currently through a fifo pipe to support the use case of targeting multiple devices at once (#4179). FIFO pipes are UNIX only, but the implementation is easily extensible to support streaming over a local socket (e.g. TCP, which would also work on Windows).

Documentation and unit tests are included.

Tested as follows:
- The file format is plain text, line-based `[timestamp] [accessory id] [hex-encoded space-separated event data]`. This simple format is chosen to make it easier to edit the event data manually if desired. Blank lines and lines with `#` are ignored, everything else is rejected.
- The parsing and serialization of the file format is fully covered by unit tests.
- I have manually tested the functionality with real devices connected to a Linux laptop via USB.
- Test devices were: Google Pixel 8, Samsung Galaxy A52s, Samsung Galaxy A53.
- Test scenarios: record to one file, replay from one file, record (overwrite) existing file, replay from non-existing file, replay and record in one command, replay from a FIFO pipe, where input is recorded to. In all cases I also tested the effect of interrupting the replay early, with the expected behavior (program closes without issues).
- I made sure that there are no memory leaks or memory safety issue by compiling with ASAN:
  `meson setup x --buildtype=debug --strip -Db_lto=true -Dprebuilt_server=/usr/share/scrcpy/scrcpy-server -Db_sanitize=address,undefined && ninja -Cx`

A note on the design:
- The implementation utilizes up to three extra threads to minimize the impact on the main and event thread (aoa).
- To record events, one thread is used (to perform I/O - write to file).
- To replay data, two threads are used: one to perform I/O (read from file/stream) and another to replay (waiting as needed to ensure that enough time has elapsed to replay the event). I chose two separate threads to make sure that the I/O and replay-sleep do not block each other from progressing as quick as possible. For ordinary local files, this does not really matter, but for data streamed from a fifo pipe (or in the future: sockets), blocking the pipe or socket could result in undesired interruptions or blocked pipes or closed sockets.

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -88,6 +88,7 @@ usb_support = get_option('usb')
 if usb_support
     src += [
         'src/usb/aoa_hid.c',
+        'src/usb/hid_event.c',
         'src/usb/hid_event_serializer.c',
         'src/usb/hid_keyboard.c',
         'src/usb/hid_mouse.c',
@@ -259,6 +260,16 @@ if get_option('buildtype') == 'debug'
             'tests/test_vector.c',
         ]],
     ]
+    if usb_support
+        tests += [
+          ['test_hid_event_serializer', [
+              'tests/test_hid_event_serializer.c',
+              'src/usb/hid_event.c',
+              'src/usb/hid_event_serializer.c',
+              'src/util/memory.c',
+          ]],
+        ]
+    endif
 
     foreach t : tests
         sources = t[1] + ['src/compat.c']
--- a/app/meson.build
+++ b/app/meson.build
@@ -88,8 +88,10 @@ usb_support = get_option('usb')
 if usb_support
     src += [
         'src/usb/aoa_hid.c',
+        'src/usb/hid_event_serializer.c',
         'src/usb/hid_keyboard.c',
         'src/usb/hid_mouse.c',
+        'src/usb/hid_replay.c',
         'src/usb/scrcpy_otg.c',
         'src/usb/screen_otg.c',
         'src/usb/usb.c',
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -172,6 +172,23 @@ By default, right-click triggers BACK (or POWER on) and middle-click triggers HO
 Print this help.
 
 .TP
+.BI "\-\-hid-record " output.log
+Record HID input to a file.
+
+Only HID input is recorded, which requires the \fB\-\-hid\-keyboard\fR, \fB\-\-hid\-mouse\fR and/or \fB\-\-otg\fR options.
+When \fB\-\-hid\-replay\fR is used simulatenously, any replayed input is also recorded to the target specified by \fB\-\-hid\-replay\fR.
+
+To mirror input to multiple devices, specify a named pipe as the filename (see mkfifo for UNIX) and pass the same filename to \fB\-\-hid\-replay\fR of a second (parallel) scrcpy instance.
+
+.TP
+.BI "\-\-hid-replay " input.log
+Replay HID input from a file created by \fB\-\-hid-record\fR.
+
+Events are only replayed when HID is used to simulate input, which requires the \fB\-\-hid\-keyboard\fR, \fB\-\-hid\-mouse\fR and/or \fB\-\-otg\fR options.
+
+See \fB\-\-hid\-record\fR for more information.
+
+.TP
 .B \-\-kill\-adb\-on\-close
 Kill adb when scrcpy terminates.
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -58,6 +58,8 @@ enum {
     OPT_RAW_KEY_EVENTS,
     OPT_NO_DOWNSIZE_ON_ERROR,
     OPT_OTG,
+    OPT_HID_RECORD,
+    OPT_HID_REPLAY,
     OPT_NO_CLEANUP,
     OPT_PRINT_FPS,
     OPT_NO_POWER_ON,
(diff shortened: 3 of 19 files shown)


Rob--W:
@rom1v Note: I have tested it extensively on Linux, but I haven't tried cross-compiling to Windows yet. I'll check whether it can compile successfully on Windows after your code review.

Rob--W:
@rom1v The branch currently has merge conflicts due to changes in the past month. Before I rebase, I'd like to ask whether you have unsubmitted feedback or any other blockers to merging the PR.

rom1v:
First of all, thank you for your PR. It's a lot of work, with a lot attention to details (as per your PR description) :heart:

Sorry for the delay (the UHID feature took priority :wink:).


## Streaming

Initially, I did not consider the "live" recording/replay of events, so loading the whole file (which should be small) then replaying it would be ok. But I like the idea to support this use case :+1:


## Replay

During the replay, keyboard and mouse should be disabled (a mouse move during a replay breaks everything).
I think that replay does not even require mirroring, and there is even no need for a window (like OTG).
This could be a separate code path which just configures AOA/HID and replays synchronously (with some feedback in the console).

For me, the most important use case for this feature is to enable USB debugging on a device with a broken screen (but other people might have other use cases).


## Threading

To minimize timing errors, the wait should be as close as possible as the input injection. So IMO, the thread waiting during `ms_to_sleep` (btw we should use µS instead of ms) should not post an event to another thread (`sc_aoa_push_hid_event()`) but should call `sc_aoa_send_hid_event()` directly immediately. In other words, the thread waiting for the deadline should be the thread which writes AOA over USB directly.

The thread which parses the file might be a separate thread posting the parsed events in a queue along with their timings, as you said:

> I chose two separate threads to make sure that the I/O and replay-sleep do not block each other from progressing as quick as possible.

TBH, I think it might as well be the same thread, since the "injection" needs to wait for I/O to get the data anyway, so unless parsing takes a long time (in which case pipelining could improve the progress), I'm not sure it could make a difference. But why not.


## Format

I think a binary format would be simpler. Parsing text in C is a PITA. I already parse some text because I have no choice in [`adb_parser.c`](https://github.com/Genymobile/scrcpy/blob/master/app/src/adb/adb_parser.c), but if we can avoid that (more complexity, more code to maintain, more bugs…), that would be great. In another language (like Rust or Python), parsing text would be ok, but in C I really really prefer binary.

What do you think?

And I'm not sure there is an important need to edit HID event data manually in text.

What I have in mind is something like:

```
mouse:    |xx xx xx xx xx xx xx|02|xx xx xx xx|
keyboard: |xx xx xx xx xx xx xx|01|xx xx xx xx xx xx xx xx|
           \------------------/ ^^ \---------------------/
                 timestamp    device    HID event
```

_(the size of the HID event might be deduced from the device)_

I think we should add a protocol version at the beginning of the file, to be able to make changes later but still be able to replay old recorded files.

Also, a minor detail: the timestamps should start at 0 in the recorded file.

## AOA/UHID

When you implemented the PR, HID was only supported via AOA. Now there is UHID.
But I think the feature should still target AOA specifically, because timings are captured and replayed on the client side (UHID would require to capture and replay on the server side, so the code could not be common).


---

Your contribution is great! I'm sorry that the codebase changed a lot since your PR was written (causing conflicts).

Also, I'm aware that changing from a text format to a binary format is quite a big change, but the binary version should be simpler (I'd really prefer not to maintain more code which parses text in C when this is not absolutely necessary).

Thank you again :wink:

Podingo1224:
how this work please ? I absolutely need it

Rob--W:
Because this PR has not been merged you would have to check out the branch locally and compile the code yourself.

Or you could wait until I found time to update the patch to implement the changes requested by the maintainer of this project.

Der-Lappen:
I have an issue using the functionality and cannot get this to work. I recorded some mouse activity with  

> ./run x --otg --hid-record=onyx.log 

and then tried to replay. 
Unfortunately, i get this error: 

> ERROR: Invalid HID replay data in onyx.log

This is the log file created with the tool. Any ideas what might cause the issue?
[onyx.log](https://github.com/user-attachments/files/16643919/onyx.log)

achimghita:
I hope this gets implemented SOON. Also is there any possibility for --hid-record & --otg-replay?

Rob--W:
> I hope this gets implemented SOON.

I have to update this patch to apply the review feedback.

> Also is there any possibility for --hid-record & --otg-replay?

What do you mean by that? The flag name or some kind of functionality?

achimghita:
I mean possibility to record touch events (on one device) and output/emulate them as otg mouse events (on another device without USB debugging).

glober6:
How do I make it work? 
I get an error message :
~/scrcpy/build/app/scrcpy --otg --hid-replay=recording.log
scrcpy 2.3.1 <https://github.com/Genymobile/scrcpy>
INFO: USB device found:
INFO:     --> 9FT9K21C08901679   (12d1:107f)  HUAWEI DRA-LX9
ERROR: Could not open icon image: /usr/local/share/icons/hicolor/256x256/apps/scrcpy.png
WARN: Could not load icon
INFO: Replaying HID input from: recording.log
ERROR: Invalid HID replay data in recording.log
INFO: Finished HID replay from recording.log
[recording.log](https://github.com/user-attachments/files/21170014/recording.log)

glober6:
> I have an issue using the functionality and cannot get this to work. I recorded some mouse activity with
> 
> > ./run x --otg --hid-record=onyx.log
> 
> and then tried to replay. Unfortunately, i get this error:
> 
> > ERROR: Invalid HID replay data in onyx.log
> 
> This is the log file created with the tool. Any ideas what might cause the issue? [onyx.log](https://github.com/user-attachments/files/16643919/onyx.log)

Hi. How did you solve this problem???

Der-Lappen:
> > I have an issue using the functionality and cannot get this to work. I recorded some mouse activity with
> > > ./run x --otg --hid-record=onyx.log
> > 
> > 
> > and then tried to replay. Unfortunately, i get this error:
> > > ERROR: Invalid HID replay data in onyx.log
> > 
> > 
> > This is the log file created with the tool. Any ideas what might cause the issue? [onyx.log](https://github.com/user-attachments/files/16643919/onyx.log)
> 
> Hi. How did you solve this problem???

I didn't. If you manage to, let me know

PR #4579, opened 2024-01-10
Add a missing separator between patterns

vladmovchan:
Without this change the bash-completion script produces the following error:
```
bash: /usr/local/etc/bash_completion.d/scrcpy: line 118: syntax error near unexpected token `newline'
bash: /usr/local/etc/bash_completion.d/scrcpy: line 118: `        --orientation'
```


rom1v:
Thank you, but it is already fixed on `dev` branxh by #4532 😉

vladmovchan:
Oh, I didn't see that. Thank you!

PR #4645, opened 2024-02-03
Force fullscreen when using the KMSDRM video driver

LHLaurini:
When running from a virtual console on a Raspberry Pi Zero 2W, SDL will use the KMSDRM video driver. Without any extra options, it automatically switches to ~800x600~ 640x480, while the `--fullscreen` flag keeps the current resolution.

This patch automatically enables fullscreen (and prevents it from being disabled) if the KMSDRM driver is used.

PS: I couldn't test on other systems at the moment, so I'd also like some help in testing whether this behavior is present on other systems and whether this patch fixes it.

### DIFF
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -661,6 +661,7 @@ void
 sc_screen_switch_fullscreen(struct sc_screen *screen) {
     if (screen->is_kmsdrm && screen->fullscreen) {
         LOGD("Ignored mode switch");
+        return;
     }
 
     uint32_t new_mode = screen->fullscreen ? 0 : SDL_WINDOW_FULLSCREEN_DESKTOP;
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -362,12 +362,14 @@ sc_screen_init(struct sc_screen *screen,
     screen->maximized = false;
     screen->minimized = false;
     screen->mouse_capture_key_pressed = 0;
+    screen->is_kmsdrm = strcmp(SDL_GetCurrentVideoDriver(), "KMSDRM") == 0;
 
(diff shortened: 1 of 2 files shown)


rom1v:
> I couldn't test on other systems at the moment

On my laptop, if I switch to a console (Ctrl+Alt+F1), and I run scrcpy, it does not change the resolution (but I can't control anything, keyboard and mouse input are not captured).

> it automatically switches to 800x600

What if you start fullscreen then disable fullscreen, does it switch to 800x600?

LHLaurini:
**Note:** it seems I misremembered, it actually switches to 640x480, which is the smallest resolution supported by my monitor.

I just tried testing stock scrcpy again, it seems to be pretty hit or miss. On my first test, it changed the resolution, but worked fine otherwise. After that, it would always get stuck at `INFO: Initial texture: ...`. Adding `-V debug` made it work again, for some reason (maybe some race condition in SDL2?). The patched version seems to always work, however.

> On my laptop, if I switch to a console (Ctrl+Alt+F1), and I run scrcpy, it does not change the resolution

This could be specific to some systems then... Maybe just Raspberry Pis? Would be great if somebody could help us test that hypothesis.

> but I can't control anything, keyboard and mouse input are not captured

For me, whether the resolution changes or not, input is still captured. When running in a virtual console [SDL2 uses evdev](https://discourse.libsdl.org/t/testing-experimental-kms-drm-backend-no-keyboard-events/23247/2) for input, so the user must be in the `input` group (or have equivalent permissions) for that to work.

> What if you start fullscreen then disable fullscreen, does it switch to 800x600?

Yes, it does.

PR #4712, opened 2024-03-01
Upgrade SDL 2.30.0 (causes issue on Windows)

rom1v:
I tried upgrading from SDL 2.28.5 to SDL 2.30.0, but then scrcpy does not work anymore on Windows: `SDL_WaitEvent()` fails with an empty `SDL_GetError()`.

It still works with SDL 2.30.0 on Linux.

I reported the problem here: https://github.com/libsdl-org/SDL/issues/9177

But I cannot really debug/investigate since the problem is specific to Windows. Any help would be welcome to find the issue in SDL :wink:


rom1v:
Thank you for your help.

I finally configured a builddir to cross-build SDL for Windows, and I bisected (sending the `SDL2.dll` to the Windows test machine via ssh), and I found the `bisect/bad` that I reported.

PR #4748, opened 2024-03-09
Add a shortcut to pause/unpause display

### LINKED ISSUE #1632: Pause and resume screen mirroring using a shortcut
- [X] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

Oftentimes, I find myself wanting to take a temporary screenshot of the tablet display to work with something I've written on my tablet, such as notes or annotations. However, it seems apparent to me that such a feature currently does not exist.

It would be great if there was a *shortcut that "freezes" the current display of the Android device, i.e. pauses mirroring. When entering the shortcut again, mirroring would be resumed*.

As mentioned above, taking temporary screenshots seemed to be the most feasible, yet cumbersome, alternative. Soon enough, there's a bunch of clutter on disk due to the several screenshots that one could take for temporary use only to delete them afterwards, which is, quite frankly, a banal task.

Once _could_ simply record whatever notes one may use, but this would not take into account any content that comes to mind after recording. Furthermore, the issue of removing the recording when done still stands analogously to deleting temporary screenshots.

**Additional context**
I'm not sure if additional context is warranted, as I believe the above explanation was pretty sufficient. However, further discussion on the matter is more than welcome. If this feature request has already been suggested before (despite my best efforts in looking for a similar one) then a link to it would be greatly appreciated.

rom1v:
Pause/unpause display on <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>p</kbd>.

It only impacts rendering, the device is still captured, the video stream transmitted to the device, and recorded if recording is enabled.

Fixes #1632

---

Here is a binary (replace `scrcpy.exe` in your scrcpy 2.4 release folder):

- [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/4748/2/scrcpy.exe) <sub>`SHA-256: 64a44eb6f871bb752f1436c3bc78456747a1ec8ddb69c7af05dc5f7dc405c23`</sub>

<details><summary>v1 (old)</summary>

- [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/4748/1/scrcpy.exe) <sub>`SHA-256: c7661c5971145e0037f727568ecc61a8d2f6596f0a7bd9b99d45e381b0992d5`</sub>
- 
</details>

### DIFF
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -578,6 +578,14 @@ Flip display horizontally
 Flip display vertically
 
 .TP
+.B MOD+z
+Pause or re-pause display
+
+.TP
+.B MOD+Shift+z
+Unpause display
+
+.TP
 .B MOD+g
 Resize window to 1:1 (pixel\-perfect)
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -901,6 +901,14 @@ static const struct sc_shortcut shortcuts[] = {
         .text = "Flip display vertically",
     },
     {
+        .shortcuts = { "MOD+z" },
+        .text = "Pause or re-pause display",
+    },
+    {
+        .shortcuts = { "MOD+Shift+z" },
+        .text = "Unpause display",
+    },
+    {
         .shortcuts = { "MOD+g" },
         .text = "Resize window to 1:1 (pixel-perfect)",
     },
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -402,6 +402,7 @@ sc_input_manager_process_key(struct sc_input_manager *im,
                              const SDL_KeyboardEvent *event) {
     // controller is NULL if --no-control is requested
     bool control = im->controller;
+    bool paused = im->screen->paused;
 
     SDL_Keycode keycode = event->keysym.sym;
     uint16_t mod = event->keysym.mod;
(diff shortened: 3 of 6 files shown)


rom1v:
@ttschnz (https://github.com/Genymobile/scrcpy/issues/1632#issuecomment-1986899236) please test the updated PR, keyboard and mouse should be disabled while in pause.

rom1v:
Maybe <kbd>Alt</kbd>+<kbd>z</kbd> wiuld be better (in reference to Ctrl+z which stops a process on Linux)? What do you think?

ttschnz:
Here are two features I discovered to still work during the freeze:
1) volume can be changed via shortcut
2) files can be transferred (also apps)

I would argue for these two to stay, because:
1) We freeze the screen to pause updating what's happening on the screen, I see it as a feature to still be able to control volume.
2) For this action to be triggered, something bust be done on purpose, and as a end-user I wouldn't expect a screen-freeze to also disable this.

During my (short) tests, I came to one or two occurences where the frame would not update as soon as I unfreezed, allthough it would then continue as soon as there was another update (through direct input on the device or through an interaction with the stream). However, I could not reproduce it on purouse, I will reach out as soon as I know more about this - I will certainly use the feature extensively during the next week.

As of the shortcut, I've come to like <kbd>mod</kbd>+<kbd>z</kbd>, one could persume "z", as in free**z**e.

rom1v:
> volume can be changed via shortcut

Oops, I forgot this one. For consistency with other shortcuts actions, I think they should be disabled on pause.

> Files can be transferred (also apps)

I think this one can be kept, it is quite independent.

> During my (short) tests, I came to one or two occurences where the frame would not update as soon as I unfreezed

Oops, I did not updated correctly. It should be fixed in the last version (I did not created a new binary, I hope you can test from the PR directly).

ttschnz:
Another flash of insight I just had: during the freeze you could add a shortcut that updates the screen, but then refreezes, as if I pressed <kbd>mod</kbd>+<kbd>z</kbd> twice. Since the inputs don't redirect to the device during the paused display, a simple <kbd>r</kbd> would be enough (allthough inconsistent with the default behaviour).
This could however be challenging to implement, and I also don't want to push this feature too far, KISS principle applies...

rom1v:
Then I propose the following:
 - <kbd>MOD</kbd>+<kbd>z</kbd>: pause (or refresh if already paused)
 - <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>z</kbd>: unpause

Another benefit is that we don't need to know the current state (which is not always obvious) to pause or unpause. I updated the PR.

ttschnz:
I've been testing commit b9cc6f7 during last and this week - I haven't found anything to complain about, so I'd say it's good like that.

For the shortcuts, I'd leave it the way it is, I find it convenient.

rom1v:
Let's merge this :rocket:

Helaer:
@rom1v  Hello, I found that although the screen is paused, there is still a large amount of CPU occupied on the computer side. Can you stop rendering processing on the computer side when the screen is paused?

rom1v:
> Can you stop rendering processing on the computer side when the screen is paused?

This feature just prevents the frame to be refreshed in the display. It does not impact other components.

The video stream is still transmitted, demuxed and decoded. The intermediate frames (at least a few ones) need to be decoded anyway, because the current frame (on unpause) may depend on them. In addition, other components may need the decoded frames (v4l2).

The feature is very local (non intrusive) to the display component, it does not impact other components. It's just a mean to "hide" a transition during a presentation, it does not release any resources.

Issue #4760, opened 2024-03-14
Unable to create window on sway desktop environment and crashes


KairuiLiu:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: ArchLinux Sway (Wayland)
 - scrcpy version: 2.4
 - installation method: AUR
 - device model: ASUS Zenbook UX3404VA
 - Android version: 10

**Describe the bug**

On Sway (Wayland) desktop environment, when I try to run scrcpy, a window flashes on the screen and then immediately disappears. I attempted to limit the window resolution with `--max-size=1080`, and also tried to set windows with `title=scrcpy` and `title=GM1910` to floating mode, but these attempts were unsuccessful.

```bash
❯ scrcpy -Vdebug

scrcpy 2.4 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  72800be1                        device  GM1910
DEBUG: Device serial: 72800be1
DEBUG: Using server: /usr/share/scrcpy/scrcpy-server
/usr/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 290.2 MB/s (69007 bytes in 0.000s)
[server] INFO: Device: [OnePlus] OnePlus GM1910 (Android 10)
[server] WARN: Audio disabled: it is not supported before Android 11
[server] DEBUG: Audio encoder stopped
DEBUG: Server connected
DEBUG: Starting controller thread
DEBUG: Starting receiver thread
[server] DEBUG: Using video encoder: 'OMX.qcom.video.encoder.avc'
INFO: Renderer: opengl
INFO: OpenGL version: 4.6 (Compatibility Profile) Mesa 24.0.2-arch1.2
INFO: Trilinear filtering enabled
DEBUG: Using icon: /usr/share/icons/hicolor/256x256/apps/scrcpy.png
DEBUG: Demuxer 'video': starting thread
DEBUG: Demuxer 'audio': starting thread
WARN: Demuxer 'audio': stream explicitly disabled by the device
INFO: Texture: 1440x3120
[server] DEBUG: Display: using SurfaceControl API
wp_viewport@35: error 2: source rectangle out of buffer bounds
DEBUG: User requested to quit
DEBUG: quit...
DEBUG: Demuxer 'video': end of frames
[...]

PR #4761, opened 2024-03-15
fix: window crash on tiling wayland WM

KairuiLiu:
Add a system-resize parameter to hand over window resize to the window manager.

#4760


rom1v:
Thank you.

Is it possible to detect automatically that the window is "tiled" in a window manager (rather than adding a new option)?

Issue #4727, opened 2024-03-05
UHID does not work the same as OTG


shiumano:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Windows, Android (Termux)
 - scrcpy version: 2.4
 - installation method: Windows Release, Termux pkg
 - device model: XQ-DQ44
 - Android version: 14

**Describe the bug**

If you try to enter in UHID, you will not be able to use it as a replacement for OTG.

In UHID mode, all mouse buttons except left mouse click are treated as special keys (right click is BACK, middle click is HOME).

Also, if there is no video/audio playback, the program will be forced to exit.
```
~ $ scrcpy --mouse=uhid --keyboard=uhid --no-video --no-audio
scrcpy 2.4 <https://github.com/Genymobile/scrcpy>
ERROR: No video, no audio, no OTG: nothing to do
```

rom1v:
> In UHID mode, all mouse buttons except left mouse click are treated as special keys (right click is BACK, middle click is HOME).

Add `--forward-all-clicks`.

> Also, if there is no video/audio playback, the program will be forced to exit.

Yes, the reason why OTG exists is to control the device without using USB debugging (adb) and a server on the device. UHID mode requires all that, so in a sense it behaves like the "normal mode".

But it is true that a mode without mirroring could be added, since in UHID mode the mouse is relative (this is not possible with the default mouse mode, because the mouse is absolute, this would be unusable). It would require quite a lot of changes to support a control-only mode with UHID though.

rom1v:
Please test #4868.

shiumano:
Excellent performance!
Tried it with Termux, very low CPU usage, practical enough.

rom1v:
Thank you for your test :+1: :tada:

> However, the other issue, the mouse click specification, I still think that all clicks should be forwarded automatically in UHID mode.

In many cases, it turns out that pressing BACK with the mouse is more useful than a real right click on an Android device.

In the very first public version of scrcpy (2018), all clicks were forwarded. A few days later, in v1.1, I added this commit: 675704c71c0d105af2242d9bbeee83127097fa0d (suggested by a user and independently by a colleague). In practice, I always use BACK with the mouse. So in my opinion, this is a good default.

[...]

PR #4877, opened 2024-04-23
Forward all clicks by default for UHID and AOA

rom1v:
In relative mouse modes (when the cursor appears on the device), it makes more sense to forward all clicks (disable HOME/BACK shortcuts on secondary clicks).

Transform the boolean option `--forward-all-clicks` into a 3-state option:
 - `--forward-all-clicks=true` or `--forward-all-clicks`: enable
 - `--forward-all-clicks=false`: disable
 - `--forward-all-clicks=auto` (default): enable only for UHID and AOA

That way, by default, all clicks are forwarded with `scrcpy -M`.

Refs <https://github.com/Genymobile/scrcpy/issues/4727#issuecomment-2069869750>


rom1v:
Or maybe it would be more consistent to keep `--forward-all-clicks` as a boolean option, and add a new `--no-forward-all-clicks`:

```bash
# SDK mouse: forward all clicks is disabled
scrcpy
# UHID mouse: forward all clicks is enabled
scrcpy -M

# SDK mouse: force enabled
scrcpy --forward-all-clicks
# UHID mouse: force disabled
scrcpy -M --no-forward-all-clicks
```

What do you think?

rom1v:
As an alternative, I think `--forward-all-clicks` could be deprecated;

It could be replaced by a new parameter `--mouse-bind=xxxx`. The argument would be exactly 4 characters, one for each secondary click:

```
--mouse-bind=xxxx
             ^^^^
             ||||
             ||| `- 5th click
             || `-- 4th click
             | `--- right click
              `---- left click
```

The characters could be:
 - `+`: forward the click
 - `-`: disable the click
 - `b`: trigger shortcut BACK
 - `h`: trigger shortcut HOME
 - `s`: trigger shortcut APP_SWITCH
 - `n`: trigger shortcut "expand notification panel"

For example:
 - `--mouse-bind=bhsn` (the default mode with SDK mouse)
 - `--mouse-bind=++++` forwards all clicks (would be the default mode in HID/AOA)
 - `--mouse-bind=++bh` forwards middle and right click, and use 4th button for BACK and 5th button for HOME

This would provide a solution for https://github.com/Genymobile/scrcpy/pull/2258#issuecomment-2182394460.

What do you think?

rom1v:
Superseded by #5022.

rom1v:
@123SONIC321 Implemented by #5022 :rocket:  Please test :wink:

PR #4930, opened 2024-05-16
Fix: the server stopped when requesting the maximum integer value(2147483647) for 'videoBitRate'

kyoung-mo-yang:
If the value of --video-bit-rate exceeds the supported range, the server is stopped.

```
$> scrcpy -s R59N102E8HT --video-bit-rate=2147483647 
scrcpy 2.4 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  R59N102E8HT                     device  SM_A305N
x/server/scrcpy-server: 1 file pushed, 0 skipped. 28.9 MB/s (69343 bytes in 0.002s)
[server] INFO: Device: [samsung] samsung SM-A305N (Android 11)
INFO: Renderer: metal
INFO: Texture: 1080x2336
WARN: Device disconnected
Aborted 
```

I have added exception handling code for this issue.


server/src/main/java/com/genymobile/scrcpy/SurfaceEncoder.java
```diff
@@ -48,11 +49,34 @@ public SurfaceEncoder(SurfaceCapture capture, Streamer streamer, int videoBitRat
...
+    /**
+     * Fixing the issue that the server crashes when requesting the maximum integer value(2147483647) for 'videoBitRate'
+     */
+    private void adjustMediaFormatValuesToSupported(MediaFormat format, MediaCodec mediaCodec) {
+        // Adjust VideoBitRate
+        try {
+            String mimeType = format.getString(MediaFormat.KEY_MIME);
+            if (mimeType != null) {
+                MediaCodecInfo mediaCodecInfo = mediaCodec.getCodecInfo();
+                Range<Integer> bitrateRange = mediaCodecInfo.getCapabilitiesForType(mimeType).getVideoCapabilities().getBitrateRange();
```
rom1v, line 61: In theory, it would be better to check all video capabilities like you are doing there.

But in the past, I noticed that in practice (on some devices), the reported values were not always accurate (especially for sizes), where "valid" values don't work or "invalid" values works. That's why I don't use them to filter parameters.

server/src/main/java/com/genymobile/scrcpy/SurfaceEncoder.java
```diff
@@ -48,11 +49,34 @@ public SurfaceEncoder(SurfaceCapture capture, Streamer streamer, int videoBitRat
         this.downsizeOnError = downsizeOnError;
     }
 
+    /**
+     * Fixing the issue that the server crashes when requesting the maximum integer value(2147483647) for 'videoBitRate'
```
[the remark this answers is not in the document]
kyoung-mo-yang: The program did not terminate abnormally due to a segmentation fault; instead, it exited with an 'aborted' log message.

server/src/main/java/com/genymobile/scrcpy/SurfaceEncoder.java
kyoung-mo-yang: As you mentioned, if there are cases where the attribute values provided by the equipment are invalid, a comprehensive test of the entire equipment (within a specific range) will be required.

PR #4947, opened 2024-05-26
server: add root support

vvb2060:
The service can now run with root. It will auto change the main thread's euid to 2000 for binder calls. If a binder call requires root user, notify the main thread to switch to euid 0.

Tested on Android 14, can display secure content.

[scrcpy-server.zip](https://github.com/Genymobile/scrcpy/files/15448218/scrcpy-server.zip)

usage:
```sh
adb root
scrcpy
```

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/ScreenCapture.java
@@ -4,10 +4,19 @@ import com.genymobile.scrcpy.wrappers.ServiceManager;
 import com.genymobile.scrcpy.wrappers.SurfaceControl;
 
 import android.graphics.Rect;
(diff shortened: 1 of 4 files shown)


adam12639:
scrcpy --adb root
scrcpy

vvb2060:
@rom1v It conflicts now, any comments? If you think this PR is bad, please let me know.

vvb2060:
this PR only modifies the service part, I don't care how to notify the exe to start the service with root. (For me adb root is enough.

BobbyWibowo:
@rom1v Any updates on this? I grabbed 2.4 client and just replaced the server with the attached file in this PR, and I can confirm it simply works on my Android 14 phone. But yeah, latest client is already 3.1 by now.

PR #4952, opened 2024-05-28
使用代理服务器下载安装包 先安装依赖所需软件

zhangyiming748:
+  first install the required software.
These are necessary software on debian bookworm
+ Use a proxy server to download thenstallation package 
In some countries and regions we must use a proxy to access github

### DIFF
--- a/install_release.sh
+++ b/install_release.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+sudo apt install -y meson libavdevice-dev libusb-1.0-0-dev
 set -e
 
 BUILDDIR=build-auto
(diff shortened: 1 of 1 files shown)


rom1v:
> first install the required software.
> These are necessary software on debian bookworm

This is documented here: https://github.com/Genymobile/scrcpy/blob/master/doc/linux.md#latest-version
(there are other packages)

Their installation does not belong to the `install_release.sh` script, which is not Debian-specific.

> Use a proxy server to download thenstallation package
> In some countries and regions we must use a proxy to access github

This is your specific proxy, not something to submit upstream.

PR #4974, opened 2024-06-06
Update AudioCapture.java

venomRK0:
with audio tracker


venomRK0:
i thought is best for gamers

rom1v:
You play the capture… but what you are playing is re-captured (so it is not played on the device).

It might "work" if you set another "usage" (like notifications, see this poc from Simon Chan: 2dfc869af77fb3b6711e14014f9ba44361a08d40) but it's quite hacky.

PR #5008, opened 2024-06-21
Modify `virtualDisplay` API priority and use `DisplayManager` first.

DerekWuYD:
Hi @rom1v 
This PR is about modify the ScreenCapture `virtualDisplay` API priority.

The Android kernel team removed the `createDisplay` method on December 21, 2023. [CL: Delete Unsupported APIs](https://cs.android.com/android/_/android/platform/frameworks/base/+/e35fd6654f3ff6572b52fe125110af4d716a4974)\
Official ROM updates for Pixel devices (Pixel 5a and newer) released after March 2024 incorporate this kernel change.\
Considering the wide range of affected devices – Pixel 5a, 6, 7, 8, and the upcoming Pixel 9.\
Prioritizing the `virtualDisplay` method over `createDisplay` is recommended, rather than triggering an exception with `createDisplay` before falling back to `virtualDisplay`.

Refer to the [Pixel Official ROM list](https://developers.google.com/android/images?hl=en#barbet), `(AP1A.240305.019.A1, Mar 2024)` already merged to Pixel 5a and newer devices.

Thank you.
Derek.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/ScreenCapture.java
@@ -45,18 +45,18 @@ public class ScreenCapture extends SurfaceCapture implements Device.RotationList
         }
 
         try {
-            display = createDisplay();
-            setDisplaySurface(display, surface, videoRotation, contentRect, unlockedVideoRect, layerStack);
-            Ln.d("Display: using SurfaceControl API");
-        } catch (Exception surfaceControlException) {
             Rect videoRect = screenInfo.getVideoSize().toRect();
-            try {
-                virtualDisplay = ServiceManager.getDisplayManager()
+            virtualDisplay = ServiceManager.getDisplayManager()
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you for the PR.

I split into 2 commits and adapted commit messages: branch [`pr5008`](https://github.com/Genymobile/scrcpy/commits/pr5008/). Is it ok for you?

DerekWuYD:
It looks good to me.\
Thank you rom1v.

rom1v:
Thank you, merged into `dev`.

lukeaschenbrenner:
Hi, sorry to necropost but I was looking through how scrcpy creates virtual displays that are considered "secure" by creating as shell in Android < 12 back when it was allowed. Does this commit negate the ability to view these secure apps with scrcpy on Android 10 and 11, since these two versions also support virtualDisplay API?

rom1v:
> Does this commit negate the ability to view these secure apps with scrcpy on Android 10 and 11, since these two versions also support virtualDisplay API?

A priori, Android 10 and 11 do not support the virtualDisplay API. IIRC it was introduced in Android 14.

PR #5014, opened 2024-06-23
Add --root option to bypass AID_SHELL secure display restriction starting in A12

Giantvince1:
Hello everyone! I am creating this pull request in hopes of merging this feature before the 2.5.x release window, as many would benefit from being able to use ```scrcpy``` with root access, without being able to run ```adb root``` since Android's new restrictions on the ```adbd``` daemon now being an APEX module. The initial idea and most of the groundwork was done by @RiggiG and I mainly just fixed up a few things that would have prevented it from merging with the current state of the ```dev``` branch, as well as a few things that caused other problems.

During testing, I have found that on an Android 14 AVD rooted with Magisk (instead of ```adb root```), I was able to see bidirectional clipboard access working, along with display, audio, and even settings changes working without problems. The only thing that is somewhat janky about the settings changes is that since ```AID_SYSTEM``` does NOT have write access to the settings database, actual root access is required by ```AID_SYSTEM``` so the server-side executable can handle cleanup on exit properly. This was just rebased using ```git cherry-pick``` to be able to get the necessary changes pulled in, without having to see nearly 100 commits that duplicated the progress on the origin ```dev``` branch in the meantime.

### DIFF
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -73,6 +73,8 @@ arguments=(
     '--record-orientation=[Set the record orientation]:orientation values:(0 90 180 270)'
     '--render-driver=[Request SDL to use the given render driver]:driver name:(direct3d opengl opengles2 opengles metal software)'
     '--require-audio=[Make scrcpy fail if audio is enabled but does not work]'
+    '--root[Launch the server as root]'
+    '--rotation=[Set the initial display rotation]:rotation values:(0 1 2 3)'
     {-s,--serial=}'[The device serial number \(mandatory for multiple devices only\)]:serial:($("${ADB-adb}" devices | awk '\''$2 == "device" {print $1}'\''))'
     {-S,--turn-screen-off}'[Turn the device screen off immediately]'
     '--shortcut-mod=[\[key1,key2+key3,...\] Specify the modifiers to use for scrcpy shortcuts]:shortcut mod:(lctrl rctrl lalt ralt lsuper rsuper)'
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -693,15 +693,15 @@ static const struct sc_option options[] = {
                 "is enabled but does not work."
     },
     {
+        .longopt_id = OPT_ROOT,
+        .longopt = "root",
+        .text = "Launch the server as root (disabled by default).",
+    },
+    {
         // deprecated
         .longopt_id = OPT_ROTATION,
         .longopt = "rotation",
         .argdesc = "value",
-    },
-    {
-        .longopt_id = OPT_ROOT,
-        .longopt = "root",
-        .text = "Launch the server as root (disabled by default).",
     },
     {
         .shortopt = 's',
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -2498,6 +2498,7 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                 break;
             case OPT_CAMERA_HIGH_SPEED:
                 opts->camera_high_speed = true;
+                break;
             case OPT_ROOT:
                 opts->root = true;
                 break;
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -701,8 +701,7 @@ static const struct sc_option options[] = {
     {
         .longopt_id = OPT_ROOT,
         .longopt = "root",
-        .text = "Try to launch the server as root.\n"
-                "Disabled by default.",
+        .text = "Launch the server as root (disabled by default).",
     },
     {
         .shortopt = 's',
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -98,6 +98,7 @@ enum {
     OPT_HID_KEYBOARD_DEPRECATED,
     OPT_HID_MOUSE_DEPRECATED,
     OPT_NO_WINDOW,
+    OPT_ROOT,
 };
 
 struct sc_option {
@@ -696,6 +697,12 @@ static const struct sc_option options[] = {
         .longopt_id = OPT_ROTATION,
         .longopt = "rotation",
         .argdesc = "value",
+    },
+    {
+        .longopt_id = OPT_ROOT,
+        .longopt = "root",
+        .text = "Try to launch the server as root.\n"
+                "Disabled by default.",
     },
     {
         .shortopt = 's',
@@ -2492,6 +2499,8 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                 break;
             case OPT_CAMERA_HIGH_SPEED:
                 opts->camera_high_speed = true;
+            case OPT_ROOT:
+                opts->root = true;
                 break;
             case OPT_NO_WINDOW:
                 opts->window = false;
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -2499,11 +2499,11 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             case OPT_CAMERA_HIGH_SPEED:
                 opts->camera_high_speed = true;
                 break;
-            case OPT_ROOT:
-                opts->root = true;
-                break;
             case OPT_NO_WINDOW:
                 opts->window = false;
+                break;
+            case OPT_ROOT:
+                opts->root = true;
                 break;
             default:
                 // getopt prints the error message on stderr
--- a/cross_win32.txt
+++ b/cross_win32.txt
@@ -6,7 +6,7 @@ c = 'i686-w64-mingw32-gcc'
 cpp = 'i686-w64-mingw32-g++'
 ar = 'i686-w64-mingw32-ar'
 strip = 'i686-w64-mingw32-strip'
-pkg-config = 'i686-w64-mingw32-pkg-config'
+pkgconfig = 'i686-w64-mingw32-pkg-config'
 windres = 'i686-w64-mingw32-windres'
 
 [host_machine]
--- a/cross_win32.txt
+++ b/cross_win32.txt
@@ -6,7 +6,7 @@ c = 'i686-w64-mingw32-gcc'
 cpp = 'i686-w64-mingw32-g++'
 ar = 'i686-w64-mingw32-ar'
 strip = 'i686-w64-mingw32-strip'
-pkgconfig = 'i686-w64-mingw32-pkg-config'
+pkg-config = 'i686-w64-mingw32-pkg-config'
 windres = 'i686-w64-mingw32-windres'
 
 [host_machine]
--- a/cross_win64.txt
+++ b/cross_win64.txt
@@ -6,7 +6,7 @@ c = 'x86_64-w64-mingw32-gcc'
 cpp = 'x86_64-w64-mingw32-g++'
 ar = 'x86_64-w64-mingw32-ar'
 strip = 'x86_64-w64-mingw32-strip'
-pkg-config = 'x86_64-w64-mingw32-pkg-config'
+pkgconfig = 'x86_64-w64-mingw32-pkg-config'
 windres = 'x86_64-w64-mingw32-windres'
 
 [host_machine]
--- a/cross_win64.txt
+++ b/cross_win64.txt
@@ -6,7 +6,7 @@ c = 'x86_64-w64-mingw32-gcc'
 cpp = 'x86_64-w64-mingw32-g++'
 ar = 'x86_64-w64-mingw32-ar'
 strip = 'x86_64-w64-mingw32-strip'
-pkgconfig = 'x86_64-w64-mingw32-pkg-config'
+pkg-config = 'x86_64-w64-mingw32-pkg-config'
 windres = 'x86_64-w64-mingw32-windres'
 
 [host_machine]
--- a/server/src/main/java/com/genymobile/scrcpy/FakeContext.java
+++ b/server/src/main/java/com/genymobile/scrcpy/FakeContext.java
@@ -6,6 +6,7 @@ import android.content.Context;
 import android.content.ContextWrapper;
 import android.os.Build;
 import android.os.Process;
+import android.system.Os;
 
 public final class FakeContext extends ContextWrapper {
 
@@ -24,11 +25,24 @@ public final class FakeContext extends ContextWrapper {
 
     @Override
     public String getPackageName() {
+        if (Os.getuid() == 1000) {
+            return "android";
+        }
         return PACKAGE_NAME;
     }
 
     @Override
     public String getOpPackageName() {
+        if (Os.getuid() == 1000) {
+            return "android";
+        }
+        return PACKAGE_NAME;
+    }
+
+    public static String getPackageNameStatic() {
+        if (Os.getuid() == 1000) {
+            return "android";
+        }
         return PACKAGE_NAME;
     }
 
@@ -37,6 +51,9 @@ public final class FakeContext extends ContextWrapper {
     public AttributionSource getAttributionSource() {
         AttributionSource.Builder builder = new AttributionSource.Builder(Process.SHELL_UID);
         builder.setPackageName(PACKAGE_NAME);
+        if (Os.getuid() == 1000) {
+            builder.setPackageName("android");
+        }
         return builder.build();
     }
 
--- a/server/src/main/java/com/genymobile/scrcpy/FakeContext.java
+++ b/server/src/main/java/com/genymobile/scrcpy/FakeContext.java
@@ -10,7 +10,7 @@ import android.system.Os;
 
 public final class FakeContext extends ContextWrapper {
 
-    public static final String PACKAGE_NAME = "com.android.shell";
+    public static final String PACKAGE_SHELL = "com.android.shell";
     public static final int ROOT_UID = 0; // Like android.os.Process.ROOT_UID, but before API 29
 
     private static final FakeContext INSTANCE = new FakeContext();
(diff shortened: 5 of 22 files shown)


app/data/zsh-completion/_scrcpy
```diff
@@ -73,6 +73,8 @@ arguments=(
     '--record-orientation=[Set the record orientation]:orientation values:(0 90 180 270)'
     '--render-driver=[Request SDL to use the given render driver]:driver name:(direct3d opengl opengles2 opengles metal software)'
     '--require-audio=[Make scrcpy fail if audio is enabled but does not work]'
+    '--root[Launch the server as root]'
+    '--rotation=[Set the initial display rotation]:rotation values:(0 1 2 3)'
```
diogotcorreia, line 77: Why are you adding rotation here?

app/src/cli.c
```diff
@@ -697,6 +698,11 @@ static const struct sc_option options[] = {
         .longopt = "rotation",
         .argdesc = "value",
     },
+    {
+        .longopt_id = OPT_ROOT,
+        .longopt = "root",
```
diogotcorreia, line 703: Nit: move this above "rotation" to preserve alphabetical order

app/src/cli.c
```diff
@@ -2493,6 +2499,9 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             case OPT_CAMERA_HIGH_SPEED:
                 opts->camera_high_speed = true;
                 break;
+            case OPT_ROOT:
```
diogotcorreia, line 2502: Nit: move this under `OPT_NO_WINDOW` to preserve order of the enum at the beginning of the file

cross_win32.txt
```diff
@@ -6,7 +6,7 @@ c = 'i686-w64-mingw32-gcc'
 cpp = 'i686-w64-mingw32-g++'
 ar = 'i686-w64-mingw32-ar'
 strip = 'i686-w64-mingw32-strip'
-pkg-config = 'i686-w64-mingw32-pkg-config'
+pkgconfig = 'i686-w64-mingw32-pkg-config'
```
diogotcorreia, line 9: Seems unrelated to this PR

cross_win64.txt
```diff
@@ -6,7 +6,7 @@ c = 'x86_64-w64-mingw32-gcc'
 cpp = 'x86_64-w64-mingw32-g++'
 ar = 'x86_64-w64-mingw32-ar'
 strip = 'x86_64-w64-mingw32-strip'
-pkg-config = 'x86_64-w64-mingw32-pkg-config'
+pkgconfig = 'x86_64-w64-mingw32-pkg-config'
```
diogotcorreia, line 9: Seems unrelated to this PR

server/src/main/java/com/genymobile/scrcpy/FakeContext.java
```diff
@@ -32,6 +34,7 @@ public String getOpPackageName() {
         return PACKAGE_NAME;
     }
 
+
```
diogotcorreia, line 37: Nit: remove unnecessary diff

server/src/main/java/com/genymobile/scrcpy/ScreenCapture.java
```diff
@@ -100,8 +101,11 @@ public void onRotationChanged(int rotation) {
     private static IBinder createDisplay() throws Exception {
         // Since Android 12 (preview), secure displays could not be created with shell permissions anymore.
         // On Android 12 preview, SDK_INT is still R (not S), but CODENAME is "S".
-        boolean secure = Build.VERSION.SDK_INT < Build.VERSION_CODES.R || (Build.VERSION.SDK_INT == Build.VERSION_CODES.R && !"S".equals(
-                Build.VERSION.CODENAME));
+        boolean secure = Build.VERSION.SDK_INT < Build.VERSION_CODES.R || (Build.VERSION.SDK_INT == Build.VERSION_CODES.R && !"S"
+                .equals(Build.VERSION.CODENAME));
+        if (Os.getuid() == 1000 || Os.getuid() == 1003) {
```
diogotcorreia, line 106: Perhaps add this to the condition above instead of mutating the variable (and edit the comment accordingly)?
Just a suggestion, not sure if it gets too confusing.

server/src/main/java/com/genymobile/scrcpy/Server.java
```diff
@@ -2,7 +2,7 @@
 
 import android.os.BatteryManager;
 import android.os.Build;
-
+import android.system.Os;
```
diogotcorreia, line 5: No need for this change, revert

server/src/main/java/com/genymobile/scrcpy/Settings.java
```diff
@@ -2,7 +2,7 @@
 
 import com.genymobile.scrcpy.wrappers.ContentProvider;
 import com.genymobile.scrcpy.wrappers.ServiceManager;
-
+import android.system.Os;
```
diogotcorreia, line 5: Perhaps preserve the empty line that was here (and I think this import should go after line 6, to preserve alphabetical order; your IDE should do this perhaps)

server/src/main/java/com/genymobile/scrcpy/Settings.java
```diff
@@ -79,4 +101,4 @@ public static String getAndPutValue(String table, String key, String value) thro
         }
         return oldValue;
     }
-}
+}
```
diogotcorreia, line 104: Add back EOF newline

server/src/main/java/com/genymobile/scrcpy/SurfaceEncoder.java
```diff
@@ -7,6 +7,7 @@
 import android.os.Looper;
 import android.os.SystemClock;
 import android.view.Surface;
+import android.system.Os;
```
diogotcorreia, line 10: No need for this change, revert

diogotcorreia (commented):
I've left a few comments, mostly nitpicks and styling problems. Haven't actually tested the code.

PS: I haven't contributed to scrcpy before, but I've been following this issue for a while, so take my review with a grain of salt.

cross_win32.txt
Giantvince1: This was actually required for 64-bit Ubuntu-based machines to cross-build Win32 32-bit binaries; the build system was checking for 32-bit versions of packages only without that revert.

cross_win64.txt
Giantvince1: Same as above; the build system would only check for 32-bit versions of packages without the revert.

app/data/zsh-completion/_scrcpy
Giantvince1: I'm assuming upstream ```dev``` had left it alone, but it's just a cosmetic change for those running ```scrcpy``` under ```zsh```. The file in question is just an auto-complete/help file associated with the program. I double-checked the file and there was no double mention of ```--rotate=```.

app/src/cli.c
Giantvince1: Will be flipped around per your suggestion in a fixup commit.

server/src/main/java/com/genymobile/scrcpy/ScreenCapture.java
Giantvince1: I simply patched in the original check without mangling it, but now I did refactor it to include the new EUID check as part of the initial selection procedure.

server/src/main/java/com/genymobile/scrcpy/Server.java
Giantvince1: Fixed, thank you. For some reason I thought it was needed.

server/src/main/java/com/genymobile/scrcpy/Settings.java
Giantvince1: Added the newline and reordered the import statements. Thank you for the help.

server/src/main/java/com/genymobile/scrcpy/Settings.java
Giantvince1: Fixed.

server/src/main/java/com/genymobile/scrcpy/SurfaceEncoder.java
Giantvince1: Thank you, fixed.

server/src/main/java/com/genymobile/scrcpy/FakeContext.java
Giantvince1: Fixed; somehow an unnecessary newline got in.

app/src/cli.c
Giantvince1: Fixed.

app/data/zsh-completion/_scrcpy
Giantvince1: That, and I also figured it'd be a good idea to include it until such time that the ```--rotate``` flag actually gets un-implemented in upstream ```dev```.

cross_win32.txt
Giantvince1: Fixed. Used ```meson``` from ```pip``` instead of ```apt``` as the ```pip``` version is FAR more up-to-date than what's currently packaged as a .deb file.

cross_win64.txt
Giantvince1: Same as the other one; updated meson by switching to pip version. Undid revert.

Giantvince1:
Is there anyone able to test the ```--root``` option on real, rooted hardware running production builds of Android, and make sure things like bidirectional clipboard actually work, especially on Android 13 and lower? I know it works on an Android Studio AVD patched to use Magisk as its root solution, but AVDs are always generated as either ```eng``` or ```userdebug``` builds, meaning they have less restriction than an actual ```user-release``` build of Android.

diogotcorreia:
I've tested this on Android 14 rooted (real hardware), but it doesn't seem to work (apps with the secure flag are still black).
However, my ROM (DerpFest) already has a flag to ignore the secure flag (I've disabled it to test), so it might be interfering. I know that the previous PR trying to implement this also did not work.

Giantvince1:
That's strange. I had used the previous PR made by RiggiG to workaround ```FLAG_SECURE``` on a Pixel 7 Pro running LineageOS 21, which is also A14, rooted via Magisk, and it worked perfectly; the easiest way to test this is to try to open your phone's settings screen while making sure the ```Allow overlays on Settings app``` toggle is off in Developer Options. If it's off, Settings will require a display with ```FLAG_SECURE```, but if the toggle is ON, the Settings app ignores the flag. I can confirm from previous experience that the Settings app did in fact black out when ```scrcpy``` wasn't run in ```AID_SYSTEM``` context with the toggle off on my device, and that it showed without issue with the ability to gain ```AID_SYSTEM``` permissions (which can create secure displays).

Giantvince1:
I'm gonna try to see if I can find the capabilities of different system-level users (UIDs below 10000), and compare them from stock AOSP and/or LineageOS to DerpFest's own code. Maybe they stripped the perms for creating a ```FLAG_SECURE``` display from UID 1000? Maybe UID 1000 isn't ```AID_SYSTEM``` on DerpFest, and instead got renumbered? I can't know yet, but I'm trying to sift through the Android source trees.

RiggiG:
> I can't know yet, but I'm trying to sift through the Android source trees.
It's only `AID_GRAPHICS` and `AID_SYSTEM`, and `AID_GRAPHICS` is not sufficient for any other scrcpy functionality.

[The relevant code is in SurfaceFlinger](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/native/services/surfaceflinger/SurfaceFlinger.cpp;l=583?q=%22Only%20privileged%20processes%20can%20create%20a%20secure%22), and has not changed in 4 years.

Giantvince1:
@RiggiG thank you for providing that info, that's actually really helpful. However, I'm not sure what you mean by "stock A14 Pixel 5", do you mean no root access, or do you mean it's running the stock ROM with Magisk installed? Either way, I'll try out a different app to test ```FLAG_SECURE``` capability with.

However, with the Settings part, it would actually black out on A14 without ```--root``` passed to enable ```FLAG_SECURE``` on the display on LineageOS, but with your patchset, it no longer blacked out, at least based on my experience with it. Either way, I plan to test it in more detail as it should be able to instantiate a secure display without restrictions with ```AID_SYSTEM``` permissions, and it should produce visible results. I'll be doing testing tonight, and hopefully I can have some conclusive results tomorrow.

RiggiG:
### dumpsys display:
<details>
  <summary>Pixel 5 A14 scrcpy with --root: </summary>

```
DisplayDeviceInfo{"Built-in Screen": uniqueId="local:4630946523002478721", 1080 x 2340, modeId 2, renderFrameRate 90.0, defaultModeId 2, supportedModes [{id=1, width=1080, height=2340, fps=60.000004, alternativeRefreshRates=[90.0], supportedHdrTypes=[2, 3]}, {id=2, width=1080, height=2340, fps=90.0, alternativeRefreshRates=[60.000004], supportedHdrTypes=[2, 3]}], colorMode 0, supportedColorModes [0, 7, 9], hdrCapabilities HdrCapabilities{mSupportedHdrTypes=[2, 3], mMaxLuminance=500.0, mMaxAverageLuminance=500.0, mMinLuminance=0.0}, allmSupported false, gameContentTypeSupported false, density 440, 435.428 x 433.839 dpi, appVsyncOff 2333333, presDeadline 11500000, cutout DisplayCutout{insets=Rect(0, 136 - 0, 0) waterfall=Insets{left=0, top=0, right=0, bottom=0} boundingRect={Bounds=[Rect(0, 0 - 0, 0), Rect(0, 0 - 145, 136), Rect(0, 0 - 0, 0), Rect(0, 0 - 0, 0)]} cutoutPathParserInfo={CutoutPathParserInfo{displayWidth=1080 displayHeight=2340 physicalDisplayWidth=1080 physicalDisplayHeight=2340 density={2.75} cutoutSpec={M 41,77 a 40,40 0 1 0 80,0 40,40 0 1 0 -80,0 Z @left} rotation={0} scale={1.0} physicalPixelDisplaySizeRatio={1.0}}}}, touch INTERNAL, rotation 0, type INTERNAL, address {port=129, model=0x40446d53ad64a0}, deviceProductInfo DeviceProductInfo{name=, manufacturerPnpId=QCM, productId=1, modelYear=null, manufactureDate=ManufactureDate{week=27, year=2006}, connectionToSinkType=0}, state ON, committedState ON, frameRateOverride , brightnessMinimum 0.0, brightnessMaximum 1.0, brightnessDefault 0.2901961, hdrSdrRatio NaN, roundedCorners RoundedCorners{[RoundedCorner{position=TopLeft, radius=108, center=Point(108, 108)}, RoundedCorner{position=TopRight, radius=108, center=Point(972, 108)}, RoundedCorner{position=BottomRight, radius=108, center=Point(972, 2232)}, RoundedCorner{position=BottomLeft, radius=108, center=Point(108, 2232)}]}, FLAG_ALLOWED_TO_BE_DEFAULT_DISPLAY, FLAG_ROTATES_WITH_CONTENT, FLAG_SECURE, FLAG_SUPPORTS_PROTECTED_BUFFERS, FLAG_TRUSTED, installOrientation 0, displayShape DisplayShape{ spec=784481599 displayWidth=1080 displayHeight=2340 physicalPixelDisplaySizeRatio=1.0 rotation=0 offsetX=0 offsetY=0 scale=1.0}}
```
```
DisplayDeviceInfo{"scrcpy": uniqueId="virtual:com.google.SSRestartDetector,1000,scrcpy,0", 1080 x 2340, modeId 56, renderFrameRate 60.0, defaultModeId 56, supportedModes [{id=56, width=1080, height=2340, fps=60.0, alternativeRefreshRates=[], supportedHdrTypes=[]}], colorMode 0, supportedColorModes [0], hdrCapabilities null, allmSupported false, gameContentTypeSupported false, density 1, 1.0 x 1.0 dpi, appVsyncOff 0, presDeadline 16666666, touch NONE, rotation 0, type VIRTUAL, deviceProductInfo null, state ON, committedState UNKNOWN, owner com.google.SSRestartDetector (uid 1000), frameRateOverride , brightnessMinimum 0.0, brightnessMaximum 0.0, brightnessDefault 0.0, hdrSdrRatio NaN, FLAG_PRIVATE, installOrientation 0, displayShape DisplayShape{ spec=-1084927657 displayWidth=1080 displayHeight=2340 physicalPixelDisplaySizeRatio=1.0 rotation=0 offsetX=0 offsetY=0 scale=1.0}}
```
</details>

Pixel 5 A13 AVD (Magisk): scrcpy does not appear



### dumpsys SurfaceFlinger:

<details>
  <summary>Pixel 5 A14 scrcpy with --root: </summary>

```
Display 4630946523002478721 (physical, "")
   Composition Display State:
   isEnabled=true isSecure=true usesDeviceComposition=true
```

```
Display 11529215050129715134 (virtual, "scrcpy")
   Composition Display State:
   isEnabled=true isSecure=false usesDeviceComposition=false
...
Layer Display 40 name="scrcpy"#82010 pid:2094 uid:1000 handleAlive
```
</details>

<details>
  <summary>Pixel 5 A13 AVD (Magisk):</summary>

```
Display 4619827259835644672 (physical, "EMU_display_0")
   Composition Display State:
   isEnabled=true isSecure=true usesDeviceComposition=true
```

```
Display 11529215050238090333 (virtual, "scrcpy")
   Composition Display State:
   isEnabled=true isSecure=true usesDeviceComposition=false
```
</details>

Giantvince1:
That explains it; the display is not being marked with ```FLAG_SECURE```. However ```AID_SYSTEM``` DOES own the screen, which means we just have to figure out HOW to specify the flag we need. Most likely we would also need to specify the display to be public instead of containing ```FLAG_PRIVATE```, so that Android will just mirror to it on its own, but that's just speculation on my part.

Giantvince1:
Unless someone can take the time to figure out how to specify the desired display flag(s) needed for this to work, I'm gonna close this PR before it gets to be far too outdated and nothing more than a cumbersome patch set for a feature few would use. I will still keep my repo open to the public in case anyone wants to create a PR against it, but as it stands, it's already a fair bit behind upstream, and a rebase isn't in my immediate plans, especially considering only half the ```FLAG_SECURE``` puzzle has been solved (getting system rights rather than shell).

PR #5022, opened 2024-06-24
Add option to configure mouse bindings

rom1v:
The argument must be exactly 4 characters, one for each secondary click:

    --mouse-bind=xxxx
                 ^^^^
                 ||||
                 ||| `- 5th click
                 || `-- 4th click
                 | `--- middle click
                  `---- right click
 
Each character must be one of the following:

 - `+`: forward the click to the device
 - `-`: ignore the click
 - `b`: trigger shortcut BACK (or turn screen on if off)
 - `h`: trigger shortcut HOME
 - `s`: trigger shortcut APP_SWITCH
 - `n`: trigger shortcut "expand notification panel"

For example:

```bash
scrcpy --mouse-bind=bhsn  # the default mode
scrcpy --mouse-bind=++++  # forward all clicks
scrcpy --mouse-bind=++bh  # forward right and middle clicks,
                          # use 4th and 5th for BACK and HOME
```
    
This deprecates `--forward-all-clicks` (use `--mouse-bind=++++` instead).

Refs https://github.com/Genymobile/scrcpy/pull/2258#issuecomment-2182394460

---

Also forward all clicks by default for UHID/AOA
    
By default, only the left click is forwarded to the device, and secondary clicks trigger shortcuts (the behavior can now be configured by `--mouse-bind=xxxx`).
    
But when the mouse mode is relative (AOA and UHID modes), forward all clicks by default. This makes more sense since the cursor is handled on the device side, the user expects all mouse buttons to be forwarded.
    
Refs https://github.com/Genymobile/scrcpy/issues/4727#issuecomment-2069869750

Supersedes #4877

---

Here is a binary for Windows:

- [`scrcpy-win64-pr5022.zip`](https://tmp.rom1v.com/scrcpy/5022/1/scrcpy-win64-pr5022.zip) <sub>`SHA-256: 7329375bb91a588e7d5f588ea9e189093cf0992fd379b61cc5b92e06f0972d1`</sub>


rom1v:
I'm not sure about the second part (forwarding all clicks by default only for AOA and UHID), that might add confusion.

Maybe keeping the shortcuts `bhsn` for all secondary clicks by default whatever the mouse mode is better, what do you think?

rom1v:
> I've tested it, and I'm glad to report that everything worked as expected. I didn't encounter any issues.

Thank you for the feedback :+1:

> Once I've set the keybinds, do I need to set them again the next time I use the program, or does it remember them?

Scrcpy does not remember anything, all configuration is passed via command line arguments.

But you can just add the arguments you want in a script: https://github.com/Genymobile/scrcpy/blob/master/doc/windows.md#run

rom1v:
:rocket:

Issue #5140, opened 2024-07-26
scrcpy crashes with NullPointerException displaying screen on TCL C805 Smart TV


algrimes:
## Environment

 - **OS:** macOS
 - **Scrcpy version:** 2.5
 - **Installation method:** brew
 - **Device model:** TCL C805 Smart TV
 - **Android version:** 12

## Describe the bug

Reproduces on all valid codec/encoder combinations
Framework jar attached

```
scrcpy --video-codec=h264 --video-encoder='OMX.google.h264.encoder' -b2M --no-audio -m1024  
scrcpy 2.5 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     --> (tcpip)  192.168.x.xxx:5555              device  Smart_TV_Pro
/usr/local/Cellar/scrcpy/2.5/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 47.6 MB/s (69624 bytes in 0.001s)
[server] INFO: Device: [TCL] TCL Smart TV Pro (Android 12)
[server] ERROR: Exception on thread Thread[video,5,main]
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
	at android.media.MediaCodec.saveSetData(MediaCodec.java:2314)
	at android.media.MediaCodec.configure(MediaCodec.java:2345)
	at android.media.MediaCodec.configure(MediaCodec.java:2268)
	at com.genymobile.scrcpy.SurfaceEncoder.streamCapture(SurfaceEncoder.java:70)
	at com.genymobile.scrcpy.SurfaceEncoder.lambda$start$0$com-genymobile-scrcpy-SurfaceEncoder(SurfaceEncoder.java:257)
	at com.genymobile.scrcpy.SurfaceEncoder$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
	at java.lang.Thread.run(Thread.java:920)
INFO: Renderer: metal
ERROR: Controller error
```
[framework.jar.zip](https://github.com/user-attachments/files/16394074/framework.jar.zip)

Scrcpy works on TCL C745 running android 11 with different processor, so possible to be confined to Android 12 devices with the new Mediatek pentonic 700 chip.

algrimes:
Can confirm scrcpy was working successfully on earlier Android 11-based TCL TVs (e.g C745), so might be specific to their android 12 implementation.

[...]

PR #5148, opened 2024-07-28
Add workaround for TCL Android 12 Smart TVs.

algrimes:
Fixes genymobile/scrcpy#5140

A fix for a bug which caused scrcpy to crash generating a video stream on certain TCL Android TVs.

Manufacturer's custom code expects to have access information about the activity. 
Fix follows similar pattern to workarounds for other manufacturers, providing them the fake context.

Scoped only to android 12.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/Workarounds.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Workarounds.java
@@ -72,6 +72,11 @@ public final class Workarounds {
             //  - <https://github.com/Genymobile/scrcpy/issues/3805#issuecomment-1596148031>
             mustFillAppInfo = true;
             mustFillAppContext = true;
+        } else if (Build.BRAND.equalsIgnoreCase("tcl") && Build.VERSION.SDK_INT == Build.VERSION_CODES.S) {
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you for your PR.

I'm considering enabling all workarounds by default (I'm not sure yet), that would also fix your issue: #5154.

rom1v:
in any case, I will merge your commit first, in case the commit enabling all workarounds needs to be reverted in the future.

I edited your commit a bit to use the same if-branch as the other devices, and removing the Android version check (too specific I guess): https://github.com/Genymobile/scrcpy/commits/pr5148/

algrimes:
Hi @rom1v ! Definitely cleaner to move it into the same branch but on removing the Android 12 scoping, its difficult to say as I don't have access to any other TCL devices (tvs, phones etc) for testing, but up to you - happy to merge like this if you are.

Defaulting to prefilling the fake context sounds sensible as it may be a smaller list of devices that in fact aren't compatible and perhaps a config flag could skip that step in a future design to avoid further code changes being needed.

rom1v:
:+1: merged f691ebb1b4fde44789292d1e8ed41adaee8e6854 into `dev`.

Issue #4451, opened 2023-11-23
Request: multiple cameras at once


rektide:
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
This is a feature request!

**Describe the solution you'd like**
Ideally I could run multiple copies of `scrcpy` with different `--camera-id`'s specified and get multiple different cameras streaming at once into different windows.

**Describe alternatives you've considered**
One alternative that likely would be acceptable for me is to stream multiple cameras as multiple different windows. This is not-optimal for me, as I can't turn cameras on and off on the fly; both cameras need to be configured ahead of time and stay on. But it would be better than what I have now.

One thought I had but wasn't able to follow up on is maybe it's possible to create multiple different multiple serial links to my phone and that maybe that might help. I don't know `adb` very well nor it's technical limitations, but I poked around for a bit & I don't think I got anywhere in terms of setting up multiple connections. One thing I tried here was seeing if I could use -P port to maybe launch multiple adb servers at once and have them each pair, but it was hard to tell what was really happening.

Nightmare scary mode (to me)_might be trying to make one `scrcpy` somehow blit multiple cameras or screens together in some way. That seems like it would be hard to configure well.

**Additional context**
I am using adb-over-wifi serial to talk to a Samsung S22, with Android 13. Right now I'm on a windows box. This is what 

I'm trying to run scrcpy twice, as so:

```sh-session
$ scrcpy -s adb-R5CRC3MLEFK-Po0SPe._adb-tls-connect._tcp. --no-audio --video-bit-rate=2M --video-source=camera --camera-size=960x720 --camera-id=0 --window-title=phonecam-back
$ scrcpy -s adb-R5CRC3MLEFK-Po0SPe._adb-tls-connect._tcp. --no-audio --video-bit-rate=2M --video-source=camera --camera-size=960x720 --camera-id=1 --window-title=phonecam-front
```

When I start the second scrcpy, the first instance dies as so, closing the window, and then the new scrcpy window starts up with the desired video camera (as normal, without trouble):

```
[server] WARN: Camera disconnected
[server] WARN: Camera capture failed: frame 121
[server] WARN: Camera capture failed: frame 122
[server] WARN: Camera capture failed: frame 123
[server] WARN: Camera capture failed: frame 124
[...]

PR #5153, opened 2024-07-30
Add ability to setup server by config file

### LINKED ISSUE #2841: Stack corruption detected when using version 1.21

**Environment**
 - OS: MacOS
 - scrcpy version: 1.21 (last)
 - installation method: Downloaded the last prebuild server (https://github.com/Genymobile/scrcpy/blob/master/BUILD.md#option-2-use-prebuilt-server)
 - device model: Samsung Galaxy A40
 - Android version: 11

**Describe the bug**

I was using for some time scrcpy 1.17 in order to retrieve frames from the device, and decided today to update to the last version. In the previous version, I was able in Python3, following a code a bit like this one https://github.com/DawningW/swy-bot/blob/main/scrcpy.py to: 

- Push the server on my phone
- Launch the server
- Forward server port : 
- Then get from the video socket the dummy byte, the device name, and the screen resolution 

After that, I was able to receive all my frames from the video socket, and it was working perfectly fine.

In 1.21, I saw that the method to launch the server changed. In 1.17, I was using : 

```
subprocess.Popen(
                ['adb', '-s', device_id, 'shell',
                 'CLASSPATH=/data/local/tmp/scrcpy-server',
                 'app_process', '/', 'com.genymobile.scrcpy.Server 1.17 verbose {} {} {} -1 true - false false 0 false false - -'.format(max_width, bitrate, max_fps)],
```

But in 1.21, there are now named parameters, and also new one, based on what I saw in the Server.java src code. I modified my code to something like : 

```
['adb', '-s', device_id, 'shell',
                 'CLASSPATH=/data/local/tmp/scrcpy-server',
                 'app_process', '/',
                 'com.genymobile.scrcpy.Server {} log_level=verbose max_size=0 bit_rate=8000000 max_fps=0 lock_video_orientation=-1 tunnel_forward=true send_frame_meta=false control=false display_id=0 show_touches=false stay_awake=false power_off_on_close=false clipboard_autosync=false'], 
                 cwd=SERVER_ROOT)
```

After launching the server, I'm still able to get access to the dummy byte, the device name, and the screen resolution from the socket, but then I don't receive any frame anymore. I have this error happening : 

`stack corruption detected (-fstack-protector)
Aborted 
`

My question is, is it still possible to retrieve frames this way ? And if it is the case, is there a parameter that is not working like that anymore, and that would explain why I have a problem retrieving the frames from the video socket now ?

Thanks a lot for your time.

kuzalekon:
Greetings!

My colleagues and I are working on a "Mobile Farm" project.
While working, I came across a problem with Samsung devices described in #2841. If a certain number of characters was exceeded during an attempt to start the scrcpy-server, the process crashed with an error. And to solve this problem, I decided to make it possible to transfer the list of required options through a server configuration file.

Along the way, I reworked the parsing of options, using `FileReader`/`StringReader` + `Properties` classes.

Closes #2841

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -19,6 +19,7 @@
 
 enum {
     OPT_BIT_RATE = 1000,
+    OPT_SERVER_CONFIG,
     OPT_WINDOW_TITLE,
     OPT_PUSH_TARGET,
     OPT_ALWAYS_ON_TOP,
@@ -284,6 +285,17 @@ static const struct sc_option options[] = {
                 "The values are expressed in the device natural orientation "
                 "(typically, portrait for a phone, landscape for a tablet). "
                 "Any --max-size value is computed on the cropped size.",
+    },
+    {
+        .longopt_id = OPT_SERVER_CONFIG,
+        .longopt = "server-config",
+        .argdesc = "/path/to/config",
+        .text = "Server configuration file path.\n"
+                "The file must be in Properties format.\n"
+                "Exmaple:\n"
+                "# scrcpy server configuration\n"
+                "power_off_on_close=true\n"
+                "cleanup=true\n",
     },
     {
         .shortopt = 'd',
@@ -2144,6 +2156,9 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                 break;
             case OPT_CROP:
                 opts->crop = optarg;
+                break;
+            case OPT_SERVER_CONFIG:
+                opts->server_config = optarg;
                 break;
             case OPT_DISPLAY:
                 LOGW("--display is deprecated, use --display-id instead.");
--- a/app/src/options.c
+++ b/app/src/options.c
@@ -2,6 +2,7 @@
 
 const struct scrcpy_options scrcpy_options_default = {
     .serial = NULL,
+    .server_config = NULL,
     .crop = NULL,
     .record_filename = NULL,
     .window_title = NULL,
--- a/app/src/options.h
+++ b/app/src/options.h
@@ -204,6 +204,7 @@ struct sc_port_range {
 
 struct scrcpy_options {
     const char *serial;
+    const char *server_config;
     const char *crop;
     const char *record_filename;
     const char *window_title;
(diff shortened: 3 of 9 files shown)


rom1v:
Thank you for the PR.

One problem I see is that the server parameters are "internal", so the user should not create a file containing the "private" server key/value pairs to pass to `--server-config`.

In theory, the client is in charge to construct and pass the necessary parameters to the server, so it should not be exposed via a command line argument.

I have no good solution for now.

One possibility would be to make the client always write a file, push it to the server, and the server would parse it (as you do). But I don't like pushing a separate file, it would add more problems (possible race conditions, typically when multiple scrcpy instances are started "simultaneously" for the same device, etc.). And in practice it would add a small starting delay, while it is almost always unnecessary.

Another possibility would be to pass only some parameters via command line arguments to the server (the one necessary to correctly establish the communication with the client), and pass the remaining ones on a socket (in addition to the video, audio or control socket). But it adds some complexity, and complicates using a standalone scrcpy-server (unless we allow both).

Yet another possibility could be to "compress" the arguments keys and values (currently, they are particularly verbose, for example `true` for a `boolean`, etc.).

kuzalekon:
Thank You for the answer @rom1v 

I agree that these are internal server details. but no one forces the user to write such a configuration file. They are may not use this opportunity at all. But quite a lot of projects use only the scrcpy-server and encounter problems from #2841.

I have these suggestions:
1. Remove the ability to specify a configuration file using the client command line. But leave this option in the scrcpy-server.
2. Implement the functionality of the optional “conversion” of command line arguments into a file and its subsequent transfer before starting the server. To activate this method of starting the server, you will also need to add a command line argument (I think what it could be called).

yume-chan:
If Scrcpy supports multiple video sockets (for multiple cameras https://github.com/Genymobile/scrcpy/issues/4451#issuecomment-1823912220) and audio sockets (for multiple audio sources https://github.com/Genymobile/scrcpy/issues/5412#issuecomment-2644517093), options will naturally be passed to each socket, and workaround the command line length limit on Samsung devices.

PR #5171, opened 2024-08-05
Fix typo

yangfl:
Requested by https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1077968#27


rom1v:
Thank you :+1:

Merged into `dev`: dd47cefa47f6cbd2e453c14df8e52de592ea6661

Does it mean that it a new package will be available in Debian? How is the server built?

yangfl:
I'm waiting for the final review before upload. Still unfortunately users have to download the separate server binary, but hopefully it would ease any kind of compilation from end users.

Issue #4098, opened 2023-06-20
install_release.sh wrong SERVER_URL and SHA256


pupitetris:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Debian bookworm (latest)
 - scrcpy version: 2.0
 - installation method: manual build
 - device model: not relevant
 - Android version: not relevant

**Describe the bug**
After a successful installation, scrcpy won´t run because install_release.sh pulled an old version of the scrcpy-server binary.

On errors, please provide the output of the console (and `adb logcat` if relevant).

```
$ scrcpy --no-audio
scrcpy 2.0 <https://github.com/Genymobile/scrcpy>
/usr/local/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 20.2 MB/s (42151 bytes in 0.002s)
[server] ERROR: Exception on thread Thread[main,5,main]
java.lang.IllegalArgumentException: The server version (1.25) does not match the client (2.0)
	at com.genymobile.scrcpy.Server.createOptions(Server.java:166)
	at com.genymobile.scrcpy.Server.main(Server.java:330)
	at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method)
	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:355)
ERROR: Server connection failed
```

The solution was to get the url for the v2.0 binary, calculate the SHA256 hash and replace the values on the script:

```
PREBUILT_SERVER_URL=https://github.com/Genymobile/scrcpy/releases/download/v1.25/scrcpy-server-v1.25
PREBUILT_SERVER_SHA256=ce0306c7bbd06ae72f6d06f7ec0ee33774995a65de71e0a83813ecb67aec9bdb
```

to:

```
PREBUILT_SERVER_URL=https://github.com/Genymobile/scrcpy/releases/download/v2.0/scrcpy-server-v2.0
PREBUILT_SERVER_SHA256=9e241615f578cd690bb43311000debdecf6a9c50a7082b001952f18f6f21ddc2
```

And run the script again. Now scrcpy runs fine.

I suppose those values on the script should be generated automatically using a template or something when generating the release.

Cheers!

rom1v:
[...]

PR #5172, opened 2024-08-06
Add optional versioned server file checker

yangfl:
If users have downloaded prebuilt server file, they may forget to update the server file simply because they are all named "scrcpy-server". It is already known to cause issues when server and client versions mismatch. Let it check for "scrcpy-server-v<VERSION>" first (just as how it is named in Github release page), then fall back to "scrcpy-server".

This mechanism is fully optional: users may still use the old unversioned name "scrcpy-server"; or they can directly download server file from release page without renaming it, and they will be notified to update their server file because client cannot find either files (with the correct version).


rom1v:
I am not keen on checking several filenames for the server.

I'm not sure to understand why you need this: this code is for the case `#ifdef PORTABLE` (i.e. in practice only for Windows, where there is no problem anyway), which should not be the case for your Debian package, isn't it?

For downloading a separate server binary (https://github.com/Genymobile/scrcpy/pull/5171#issuecomment-2269204118), could you include a script in your package, similar to `install_release.sh`, but which would only download the server for the expected version and save it to `/usr/share/scrcpy-server`?

For example, something like this (to be installed to `/usr/bin/scrcpy-update-server`, to be exected after a package upgrade):

```bash
#!/usr/bin/env bash
set -e

VERSION=2.6.1
PREBUILT_SERVER_URL="https://github.com/Genymobile/scrcpy/releases/download/v$VERSION/scrcpy-server-v$VERSION"
PREBUILT_SERVER_SHA256=ca7ab50b2e25a0e5af7599c30383e365983fa5b808e65ce2e1c1bba5bfe8dc3b

if echo "$PREBUILT_SERVER_SHA256  /usr/bin/scrcpy-server" | sha256sum --check
then
    echo "[scrcpy] Server $VERSION already up to date"
    exit 0
fi

echo "[scrcpy] Downloading prebuilt server..."
wget "$PREBUILT_SERVER_URL" -O scrcpy-server  # adapt to use some temporary directory
echo "[scrcpy] Verifying prebuilt server..."
echo "$PREBUILT_SERVER_SHA256  scrcpy-server" | sha256sum --check

sudo cp scrcpy-server /usr/share/
echo "[scrcpy] Server $VERSION installed"
```

yangfl:
My proposed Debian update does use PORTABLE build.

However, I'd like to make generalized feature proposals for all potential users, and in this case, non-Windows users may accidentally use mismatched server if they copy self-built client everywhere. This version checker helps to prevent this event. For Windows build I'd keep current behavior intact.

rom1v:
> My proposed Debian update does use PORTABLE build.

Why? The purpose of portable is basically to put `scrcpy-server` in the same directory as `scrcpy` (whichever it is). But your package will still install scrcpy it to `/usr/bin/`, right?

acbits:
> If users have downloaded prebuilt server file, they may forget to update the server file simply because they are all named "scrcpy-server". It is already known to cause issues when server and client versions mismatch. Let it check for "scrcpy-server-v" first (just as how it is named in Github release page), then fall back to "scrcpy-server".
> 
> This mechanism is fully optional: users may still use the old unversioned name "scrcpy-server"; or they can directly download server file from release page without renaming it, and they will be notified to update their server file because client cannot find either files (with the correct version).

I just hit this bug. I tried to build 2.6.1, but the version is still pointing to 2.6, so scrcpy refused to start.

Looks like it needs to be retagged?

`git diff v2.6.1 -- install_relase.sh` shows the required update.

I built it by running `git checkout v2.6.1; ./install_relase.sh`. I hope that is the correct way?

rom1v:
The correct way currently is to always execute `./install_release.sh` from the `master` branch.

The other valid way is to checkout the tag `v2.6.1-install-release`.

See https://github.com/Genymobile/scrcpy/issues/4098#issuecomment-1600332180.

acbits:
> The correct way currently is to always execute `./install_release.sh` from the `master` branch.
> 
> The other valid way is to checkout the tag `v2.6.1-install-release`.
> 
> Also see [#4098 (comment)](https://github.com/Genymobile/scrcpy/issues/4098#issuecomment-1600332180).

Thanks. I was not familiar with tag naming conventions.

PR #5175, opened 2024-08-06
[feat] add support to create virtual display

rsanjuan87:
with option ` --create-new-display=WxH ` you may cast a new virtual screen

based on the fork https://github.com/anirudhb/scrcpy/ 

imagen


app/src/screen.c
```diff
@@ -283,9 +285,38 @@ event_watcher(void *data, SDL_Event *event) {
...
-        // that specific case. Anyway, it's just a workaround.
-        sc_screen_render(screen, true);
+        // idk
+        if (screen->is_virtual) { //((sz.width % 16) == 0 && (sz.height % 16) == 0) {
+            // FIXME: probably slow
+            struct sc_size sz = get_window_size(screen);
+            struct sc_control_msg msg;
+            msg.type = SC_CONTROL_MSG_TYPE_RESIZE_DISPLAY;
+            msg.resize_display.width = sz.width;
+            msg.resize_display.height = sz.height;
```
rom1v, line 295: When I resize my client window, the captured display remains at the initial resolution (which is fine, but I think this code was expected to resize on the device, right?). In the end, it should probably be an option.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -166,7 +168,7 @@ private boolean encode(MediaCodec codec, Streamer streamer) throws IOException {
                 alive = false;
                 break;
             }
-            int outputBufferId = codec.dequeueOutputBuffer(bufferInfo, -1);
+            int outputBufferId = codec.dequeueOutputBuffer(bufferInfo, 100000);
```
rom1v, line 171: Why?

rom1v (commented):
Thank you for the prototype :+1:

I tested, it works on my device :sunglasses: 

I keep this feature on my TODO list (beforehand, I'd like to add gamepad support).

rom1v:
Superseded by #5370.

Issue #5137, opened 2024-07-26
Scrcpy Virtual Display Mirroring Issue and Resolution for Android Devices


mengyanshou:
## Environment

 - **OS: macOS**
 - **Scrcpy version: 2.5**
 - **Installation method: brew**
 - **Device model: Xiaomi14 MEIZU 21**
 - **Android version: 14(both)**

## Describe the bug

I encountered a problem similar to one in another issue, related issue #4598

Some devices have virtual displays, such as Samsung's Dex, Smartisan's TNT. When using scrcpy to mirror these devices, the screen touch events become ineffective.

I have two devices, Xiaomi 14 and Meizu 21, both are Android 14, and the same issue exists.

If I cast to a display with an ID of 2 or 3,

when the target display is a VirtualDisplay (I haven't tested with a physical display yet), the screen control events do not work.

But if I direct all the events' display ID to the display ID created by the scrcpy-server's VirtualDisplay,

everything works fine.

The scrcpy-server created the virtual display here:

```java
try {
    Rect videoRect = screenInfo.getVideoSize().toRect();
    virtualDisplay = ServiceManager.getDisplayManager()
            .createVirtualDisplay("scrcpy", videoRect.width(), videoRect.height(), device.getDisplayId(), surface);
    Ln.d("Display: using DisplayManager API");
} catch (Exception displayManagerException) {
    try {
        display = createDisplay();
        setDisplaySurface(display, surface, videoRotation, contentRect, unlockedVideoRect, layerStack);
        Ln.d("Display: using SurfaceControl API");
    } catch (Exception surfaceControlException) {
        Ln.e("Could not create display using DisplayManager", displayManagerException);
        Ln.e("Could not create display using SurfaceControl", surfaceControlException);
        throw new AssertionError("Could not create display");
    }
}
```
Here are my operations:

**1. Create a virtual display using Android API**

Assuming the obtained display ID is 2.

**2. Run Task on the specified display**

```shell
am display move-stack taskId $displayId
```
or

```shell
adb shell am start-activity -W -S --activityType 0 --activity-brought-to-front --display 2 -n com.nightmare.sula/.MainActivity
This will run the Android task on this virtual display.
```

**3. Use scrcpy --display-id 2 to mirror**

The screen can be rendered normally, but the events do not take effect.

I made the following modifications to the scrcpy server code:

Device.java

```java
public final class Device {
    private final int displayId;
    private int inputDisplayId;
    // code

[...]

PR #5214, opened 2024-08-18
add dispatch_to_vd option to server

mengyanshou:
Related issue #5137
Fix input event do not work when screen virtual display


mengyanshou:
As far as I know, the corresponding scrcpy client also needs to add the corresponding parameters.

rom1v:
Please test #5370: does it fixes this problem (without using the `--new-display` feature)?

mengyanshou:
> Please test #5370: does it fixes this problem (without using the `--new-display` feature)?

thanks~see https://github.com/Genymobile/scrcpy/pull/5370#issuecomment-2411152853

rom1v:
Superseded by #5370.

Issue #4080, opened 2023-06-10
Integrate with mediakeys dbus extension (MPRIS)


Gottox:
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

Thanks for scrcpy! I'm always surprised how seamless this tool works!

**Is your feature request related to a problem? Please describe.**
With the introduction of audio forward in scrcpy 2.0, I want to start using my smartphone as my main media player. While sound and mouse/keyboard control works nicely I wished that it could integrate with my Linux Desktop even better.

**Describe the solution you'd like**

It would be awesome if scrcpy could interact with the [MPRIS](https://specifications.freedesktop.org/mpris-spec/latest/) interface to support media buttons and maybe even report the current play state/artist/title from android (I'm not sure if that's possible though)

**Describe alternatives you've considered**

Currently I'm using the spotify webinterface as kind of "remote control" to make the media buttons work on linux. While this is enough for my use case, it is clumsy and only works with spotify.

AzlanCoding:
Hello, it's been a year.<br>
I too use my smartphone as my main media player and really wanted this feature. So I wrote a python script to help forward mediakeys in the dbus extension to the media key controls on the phone. Check out [here](https://github.com/AzlanCoding/scrcpyMediaController) and let me know what you think about it.<br>
![Screenshot of scrcpyMediaController in use](https://raw.githubusercontent.com/AzlanCoding/scrcpyMediaController/main/Screenshots/Screenshot_30-May_20-25-30_20743.png)<br>
I wrote this script while I was bored studying for Organic Chemistry in preparation for my GCE O'level examination. Slowly, I kept fixing some bugs until I was satisfied with it. I might compile a proper release binary or maybe even learn C and make a pull request to scrcpy after my exams.

PR #5255, opened 2024-09-01
initial MPRIS support

Gottox:
This PR adds support for the MPRIS Dbus protocol as described in #4080. My first approach was to directly use the low level API, but unfortunately the usage of the library is discouraged:

> This manual documents the low-level D-Bus C API. **If you use this low-level API directly, you're signing up for some pain.**

So I felt back on the glib API and tried to get an glib event loop running in a worker thread.

For now the only event that has been wired up is the [Raise](https://specifications.freedesktop.org/mpris-spec/latest/Media_Player.html#Method:Raise) method.

The goal for the first iteration is to only get the media keys working. Later iterations could also update the current track metadata if they can be extracted from the device. I haven't looked at the server implementation yet.

This implementation is unfinished and meant as an opportunity to see if @rom1v considers the approach acceptable.

addresses #1824, #4080


rom1v:
Thank you for your PR, and for posting an early draft :+1:

AFAIU, dbus/mpris is useful to control a player with a feedback, i.e. the client knows the state of the media player and get asynchronous updates.

Here, IIUC this is not the case (we are not able to receive the Android player state to synchronize with the client). In that case, maybe shortcuts are sufficient (#1824)?

Also, how does it interact with a local media player? For example, in my case, I don't want the media keys to control the Android device but my audio player on the computer.

> So I felt back on the glib API and tried to get an glib event loop running in a worker thread.

I would like to avoid adding a dependency to glib. The low-level API can still be used, like in VLC:

https://code.videolan.org/videolan/vlc/-/blob/master/modules/control/dbus/dbus.c

But without state feedback from the device, I'm not sure the feature is useful. What do you think?

Gottox:
My usecase: Since scrcpy supports audio forwarding (❤️❤️❤️), I really enjoy putting my phone into the dock and my desktop and seamlessly continue listening to music/podcasts on my desktop. While scrcpy is pretty good at forwaring, the integration of my android players into my desktop environment would be a real cherry on top.

> Also, how does it interact with a local media player? For example, in my case, I don't want the media keys to control the Android device but my audio player on the computer.

That depends on the desktop environment. Gnome for example can handle multiple sources:

image

Also, in the final version, I would hide it behind a cmdline option: `--mpris`

> I would like to avoid adding a dependency to glib. The low-level API can still be used, like in VLC:

Thanks! It looks quite massive, but I'll have a look.

> But without state feedback from the device, I'm not sure the feature is useful. What do you think?

I believe that it would already have a benefit if a user can control the music on the android device without having scrcpy in focus. At least that would be a nice start for me.

I'm currently looking into the [MediaSessionManager](https://developer.android.com/reference/android/media/session/MediaController) [(src)](https://android.googlesource.com/platform/frameworks/base/+/master/media/java/android/media/MediaServiceManager.java) API of android.

Up to now I'm unable to instantiate it, but if I understand correctly, it would allow control over all MediaSessions of a device and also can request metainformation about these sessions. I'm not that familiar with the Android API.

Do you think that's feasible?

Gottox:
Update: I finally got access to a functional [`MediaSessionManager`](https://developer.android.com/reference/android/media/session/MediaSessionManager) instance. That would make it possible to create proxy controls for most Android Media Applications in mpris. That's exciting!

I haven't looked into the libdbus itself. So for now it'll remain in libgdbus. Now to figure out how to wire everything up...

Gottox:
I updated the PR. Currently I'm facing the issue, that the MediaSessionManager and the MediaController do not emit any events. Any pointers what the issue could be here?

app/src/device_msg.c
```diff
@@ -56,21 +63,43 @@ sc_device_msg_deserialize(const uint8_t *buf, size_t len,
...
+        case DEVICE_MSG_TYPE_MEDIA_REMOVE: {
+            if (len < 3) {
+                // at least id
+                return 0; // not available
+            }
+            uint16_t id = sc_read16be(&buf[1]);
+            msg->media_remove.id = id;
+            return 3;
+            }
+        }
```
MultisampledNight, line 102: Is it intentional that these closing brackets don't match their indentation level? The first one closes the case of `DEVICE_MSG_TYPE_MEDIA_REMOVE` (henceforth _m_) while the second one closes the case of `DEVICE_MSG_TYPE_UHID_OUTPUT` (henceforth _o_).

(I'm not even sure if the `case` of _m_ works since it's *inside* the block opened by _o_, but I don't know C well enough to say for sure. Maybe `case`s inside a switch can be in nested blocks and blocks aren't hard scopes.)

app/src/mpris.c
```diff
@@ -0,0 +1,377 @@
...
+
+static void
+method_call_player(G_GNUC_UNUSED GDBusConnection *connection,
+				   G_GNUC_UNUSED const char *sender,
+				   G_GNUC_UNUSED const char *_object_path,
+				   G_GNUC_UNUSED const char *interface_name,
+				   const char *method_name, G_GNUC_UNUSED GVariant *parameters,
+				   GDBusMethodInvocation *invocation, G_GNUC_UNUSED gpointer user_data) {
+	if (g_strcmp0(method_name, "Pause") == 0) {
+		g_dbus_method_invocation_return_value(invocation, NULL);
```
MultisampledNight, line 166: Is there a reason the `Pause` method is not logged like the others?

```suggestion
	if (g_strcmp0(method_name, "Pause") == 0) {
        LOGD("mpris: Pause");
		g_dbus_method_invocation_return_value(invocation, NULL);
```

MultisampledNight (commented):
First of all: Thanks!! This is amazing!! :tada:
Then: noted a few small things I noticed while skimming through. I don't know C that well though, so take them with some caution!

KodzghlyCZ:
Hello

Any chance to make MPRIS work with current versions? I'm trying to build the version of the PR, but I'm not so sure I'll be able to fully implement it myself. Thank you for the amazing work so far <3

KodzghlyCZ:
I have tried to run the MPRIS implementation from the PR branch, but I've hit a brickwall of _MediaSessionManager_ probably not being available on newer Androids and require some extra permissions. 

I get

`WARN: MediaManager not available: java.lang.NullPointerException: Attempt to invoke virtual method 'android.media.MediaServiceManager$ServiceRegisterer android.media.MediaServiceManager.getMediaSession

ServiceRegisterer()' on a null object reference

`

I'm using LineageOS 21 with microG (Android 14), which doesn't help much, since the error might be also related to some services not being implemented / enabled in microG. 
Right now I have no other way to try out other versions of Android, but I'm pretty suspicious this issue occurs because of some permissions missmatch. 

If anyone finds a way around this, I'd be really grateful, since all my attempts to make this work failed terribly.

yume-chan:
> ```
> WARN: MediaManager not available: java.lang.NullPointerException: Attempt to invoke virtual method 'android.media.MediaServiceManager$ServiceRegisterer android.media.MediaServiceManager.getMediaSessionServiceRegisterer()' on a null object reference
> ```

Due to changes in [864d0b7220410d5fd89efc39a3be83b8cb5e46c3](https://cs.android.com/android/_/android/platform/frameworks/base/+/864d0b7220410d5fd89efc39a3be83b8cb5e46c3), from Android 12 you need to run [`MediaFrameworkPlatformInitializer.setMediaServiceManager(new MediaServiceManager());`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/app/ActivityThread.java;l=8968;drc=61197364367c9e404c7da6900658f1b16c42d0da) before [`FakeContext.get().getSystemService(MediaSessionManager.class);`](https://github.com/Genymobile/scrcpy/pull/5255/files#diff-59a9c1034516ab3ff0785c669227ceba46015cf29fb548879839d17f9b6a8eaaR36)

Or just call [`android.app.ActivityThread.initializeMainlineModules`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/app/ActivityThread.java?q=symbol%3A%5Cbandroid.app.ActivityThread.initializeMainlineModules%5Cb%20case%3Ayes) directly [as `telecom` command does](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/cmds/telecom/src/com/android/commands/telecom/Telecom.java;l=45;drc=61197364367c9e404c7da6900658f1b16c42d0da)

kerollosy:
Hey @Gottox, are you still actively working on this? And do you have any plans to add SMTC support for Windows alongside MPRIS?

I've been working on something similar from the Windows side, I have a proof of concept (https://github.com/kerollosy/sndsync) that already handles the Android metadata forwarding (title, artist, album art, playback state, timeline) and has SMTC working end-to-end on Windows.

On the SMTC implementation side: SMTC is a WinRT API and the only practical way to consume it is C++/WinRT (https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/). It's header-only and already bundled with the Windows SDK so zero new dependencies, but it does mean one .cpp file. The raw COM/C ABI alternative exists but Microsoft themselves describe it as painful and it produces strictly larger and slower binaries. My plan is to keep it fully isolated behind a plain C interface so none of scrcpy's existing C code changes. (https://github.com/kerollosy/SMTC-C)

yume-chan's fix above (calling `ActivityThread.initializeMainlineModules`) unblocks `MediaSessionManager`  on Android 11+


Happy to collaborate if you're open to it, and also curious what @rom1v thinks about the .cpp file approach cause I know y'all wanna keep it 100% C.

PR #2130, opened 2021-02-18
Support game controllers

LHLaurini:
closes #99 

## What it does

- Handles events from connected game controllers;
- Transfers game controller events to Android device;
- Emulates an Xbox 360 game controller input device.

## Important stuff

- In order to emulate input devices, we need a few system calls (`open`, `ioctl`, `write` and `close`). Since we can't use most of these directly from Java (AFAIK), we need either native code or a library that allows us to call these functions. To avoid depending on the NDK, I'm using the Java Native Access library;
- JNA requires a native library (libjnidispatch.so). For some reason, `app_process` doesn't want to load it, so right now it's extracted to `/data/local/tmp` on initialization and cleaned up later.

## Things you can help with

- Testing:
    - Comment whether it works on your device or not, so I can keep a list;
    - Axes behavior;
    - Multiple controllers.
- Find a way to use the `ioctl` syscall without native code;
- Load native libraries without extracting them;
- Figure out multiple controllers with event injection.

## To-do

- [x] Add command line option for disabling game controllers
- [x] Write tests
- [x] Isolate uinput related stuff into another class (so other types of devices can be implemented)
- [x] Fix input not working when uinput fails
- [x] Support legacy uinput
- [x] Look into event injection (for when uinput is unavailable)
- [ ] Look into getting games to recognize axis movement when using event injection
- [ ] Implement controller through event injection

## Out of scope

I probably won't be adding some special features to this pull request, such as:
- Vibration
- Battery reporting
- Lights
- Touchpad

## Feedback

If you have any feedback (suggestions, complaints, new info, ...), please don't hesitate to write it down below.

## Tested devices

Manufacturer | Model | Android version | Last tested on | Tested by | Supported | Supported with root
:-:|:-:|:-:|:-:|:-:|:-:|:-:
Motorola | Moto G<sup>5</sup> Plus | 8.1 | 2021-04-30 | @LHLaurini | :x: | :grey_question:
Motorola | moto g(6) play | 9 | 2021-04-30 | @LHLaurini | :x: | :grey_question:
Samsung | Galaxy A7 2018 | 10 | 2021-04-06 | @Fablo020 | :x: | :grey_question:
Samsung | Galaxy M21s | 11 | 2021-04-30 | @LHLaurini | :heavy_check_mark: | :grey_question:
Huawei | Honor Play COR-L29 | 9 | 2021-05-01 | @Cannahawk | :x: | :grey_question:
[...]

Issue #3088, opened 2022-03-06
High polling rate makes mouse(HID) sluggish


nothingshiru:
- [ ] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [ ] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Windows 10
 - scrcpy version: 1.23
 - installation method: Windows release
 - device model: Huawei Matepad 10.4
 - Android version: 10

**Describe the bug**
The polling rate of my mouse can be set to 125hz, 500hz, 1000hz in driver. When it is set to 500hz or 1000hz, mouse(HID) is sluggish.

rom1v:
Any relevant logs in the console output?

nothingshiru:
Sorry. I'm not a programmer that might not understand you well. Where can I find the console output?

rom1v:
Run scrcpy by double-clicking on `scrcpy-console.bat`. The console is the black window 😉

rom1v:
Double-click on `open_a_terminal_here.bat`, type `scrcpy --otg`. Some logs will be printed in this terminal while scrcpy is running.

nothingshiru:
`scrcpy -KM` doesn't work. `scrcpy --otg` works. 

And here is the log.
```
D:\Program Files\scrcpy-win64-v1.23>scrcpy -KM -s TFMNU20429107644
scrcpy 1.23 <https://github.com/Genymobile/scrcpy>
ERROR: On Windows, it is not possible to open a USB device already open by another process (like adb).
ERROR: Therefore, -K/--hid-keyboard and -M/--hid-mouse may only work in OTG mode (--otg).

D:\Program Files\scrcpy-win64-v1.23>scrcpy --otg -s TFMNU20429107644
scrcpy 1.23 <https://github.com/Genymobile/scrcpy>
INFO: Killing adb daemon (if any)...
INFO: USB device: TFMNU20429107644 (12d1:107e) HUAWEI BAH3-W09
WARN: On this platform, libusb does not have hotplug capability; device disconnection will not be detected properly
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
WARN: Could request HID event
[...]

PR #5270, opened 2024-09-06
Add gamepad support

### LINKED ISSUE #99: Support game controllers
I know its note quite the goal of this project but would it be possible to use an Xbox 360 controller connected to the pc on my android phone trough scrcpy

rom1v:
This PR adds support for game controllers (gamepads).

There was a PR (#2130) implementing gamepad support using uinput on the device. However, this required to add a native dependency or to add native code (cf https://github.com/Genymobile/scrcpy/pull/2130#issuecomment-2276249226).

This PR uses HID instead, and provides support for AOA and UHID (like for keyboard and mouse).
It does not add any dependency, and it also works in OTG mode (via AOA), so even without USB debugging.

To enable it:

```bash
# UHID
scrcpy --gamepad=uhid
scrcpy -G  # short option, equivalent

# AOA
scrcpy --gamepad=aoa

# OTG
scrcpy --otg --gamepad=aoa
```

It is not enabled by default because UHID is not supported on all devices (especially with old Android versions, I don't remember which version exactly, due to a permission error).

I tested with PS4 and PS5 controllers connected by USB on my laptop, with gamepad tester apps, it seems to work.

Rumble is not implemented, because I don't know how to make it work. It probably requires to add the correct bytes to the report descriptor, and to parse the received HID outputs (the mechanism to receive them is already there). @yume-chan (or others) your help is welcome :wink:

@LHLaurini I marked you as co-author of the commit named `Handle SDL gamepad events` (I don't give the SHA1 because it will change on every rebase), because your PR helped me to write this commit :+1:

## Tests

Here is a binary for Windows:

- [`scrcpy-win64-gamepad_draft4.zip`](https://tmp.rom1v.com/scrcpy/5270/4/scrcpy-win64-gamepad_draft4.zip) <sub>`SHA-256: 14a62baa9879bb4eb299a422aed337ab28ddf47667edd61fa641ce732738de2`</sub>

Please test with your controllers.
Also with 2 (or more) gamepads simultaneously (which games support multiple controllers?).

## TODO

 - [ ] ~Rumble~
 - [x] Review/rework/fix patchset
 - [x] User documentation

Fixes #99


ahosker:
This would be very cool

LHLaurini:
I don't have any games with controller support installed, so I tested using https://greggman.github.io/html5-gamepad-test/ and a Gamepad Tester app.

**OS:** Arch Linux (up-to-date)
**Phone:** Samsung Galaxy M21s
**Controller:** Sony Dualsense (PS5 controller)

**What works**:
- All digital buttons
- Analog sticks

**What kinda works**:
- Analog trigger values flicker between the correct value and half (e.g. if I press it halfway, it alternates between 0.5 and 0.25).

**What doesn't work**:
- Pressing (not just touching) the touchpad shows the following message: `[server] WARN: Unknown UHID id: 5760`. If repeated, `scrcpy` eventually crashes.
- Pressing the mute button shows the following message: `[server] WARN: Unknown UHID id: 6864`. If repeated, `scrcpy` eventually crashes.

rom1v:
> **What kinda works:**
>
>   - Analog trigger values flicker between the correct value and half (e.g. if I press it halfway, it alternates between 0.5 and 0.25 -- kinda hard to tell exactly, it flickers fast).

Oh, I wrote in the wrong endianness:

```diff
diff --git a/app/src/hid/hid_gamepad.c b/app/src/hid/hid_gamepad.c
index ca4dc67be..f78cfec73 100644
--- a/app/src/hid/hid_gamepad.c
+++ b/app/src/hid/hid_gamepad.c
@@ -322,12 +322,12 @@ sc_hid_gamepad_event_from_slot(uint16_t hid_id,
     uint8_t *data = hid_input->data;
     // Buttons are individual bits in the report descriptor, so the resulting
     // u16 must be written in little-endian
-    sc_write16be(data, slot->axis_left_x);
-    sc_write16be(data + 2, slot->axis_left_y);
-    sc_write16be(data + 4, slot->axis_right_x);
-    sc_write16be(data + 6, slot->axis_right_y);
-    sc_write16be(data + 8, slot->axis_left_trigger);
-    sc_write16be(data + 10, slot->axis_right_trigger);
+    sc_write16le(data, slot->axis_left_x);
+    sc_write16le(data + 2, slot->axis_left_y);
+    sc_write16le(data + 4, slot->axis_right_x);
+    sc_write16le(data + 6, slot->axis_right_y);
+    sc_write16le(data + 8, slot->axis_left_trigger);
+    sc_write16le(data + 10, slot->axis_right_trigger);
     sc_write16le(data + 12, slot->buttons & SC_HID_BUTTONS_MASK);
     data[14] = sc_hid_gamepad_get_dpad_value(slot->buttons);
 }
```

I wondered why it did not affect analog sticks, the reason is because the source is on 8 bits, so when converted to a 16-bit value by SDL it gives 2 identical 8-bit values, so big/little endian makes no difference.

> What doesn't work:
>
>   - Pressing (not just touching) the touchpad shows the following message: [server] WARN: Unknown UHID id: 5760. If repeated, scrcpy eventually crashes.
>   - Pressing the mute button shows the following message: [server] WARN: Unknown UHID id: 6864. If repeated, scrcpy eventually crashes.

Hmm, I cannot reproduce. It looks like control message corruption.

Can you reproduce over AOA?

SuperKali:
Hi @rom1v, Thank you very much for your work you are doing, I tried the binary you released but when I try to start the session it gives me permission error:

```bash
C:\Users\super\Downloads\scrcpy-win64-gamepad_draft1\scrcpy-win64-gamepad_draft1>scrcpy.exe -G
scrcpy 2.6.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     --> (tcpip)  youyeetoo-r1-v3:5555            device  redroid13_arm64
C:\Users\super\Downloads\scrcpy-win64-gamepad_draft1\scrcp... file pushed, 0 skipped. 85.9 MB/s (71328 bytes in 0.001s)
[server] INFO: Device: [redroid] redroid redroid13_arm64 (Android 13)
INFO: Renderer: direct3d
INFO: Texture: 1080x1920
[server] ERROR: Controller error
java.io.IOException: android.system.ErrnoException: open failed: EACCES (Permission denied)
        at com.genymobile.scrcpy.control.UhidManager.open(UhidManager.java:68)
        at com.genymobile.scrcpy.control.Controller.handleEvent(Controller.java:213)
        at com.genymobile.scrcpy.control.Controller.control(Controller.java:102)
        at com.genymobile.scrcpy.control.Controller.lambda$start$0$com-genymobile-scrcpy-control-Controller(Controller.java:110)
        at com.genymobile.scrcpy.control.Controller$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
Caused by: android.system.ErrnoException: open failed: EACCES (Permission denied)
        at libcore.io.Linux.open(Native Method)
        at libcore.io.ForwardingOs.open(ForwardingOs.java:563)
        at libcore.io.BlockGuardOs.open(BlockGuardOs.java:274)
        at android.system.Os.open(Os.java:494)
        at com.genymobile.scrcpy.control.UhidManager.open(UhidManager.java:51)
        ... 5 more
WARN: Device disconnected

C:\Users\super\Downloads\scrcpy-win64-gamepad_draft1\scrcpy-win64-gamepad_draft1>
```

How can fix it?

rom1v:
Oh, you don't have the permissions to open /dev/uhid on your device. Typically, this happens on old Android versions, but you are on Android 13. I have no solution for UHID. However, maybe AOA works over USB (`scrcpy --gamepad=aoa`).

LHLaurini:
Rebuilt with `-Db_sanitize=address`, I can't get it to crash anymore, even without the flag...

> Please run scrcpy -G --no-audio -Vverbose, and post the content before you get Unknown UHID id please.

When pressing either the touchpad or the mute button, I get:

> VERBOSE: input: UHID input [0] 
> [server] WARN: Unknown UHID id: 0
> VERBOSE: input: UHID input [0] 
> [server] WARN: Unknown UHID id: 0

The first two lines show up on press, the other two on release.

So the crash seems to be due to trying to read from an empty buffer (just a guess, I didn't look at the code)? We should just ignore these.

> Can you reproduce over AOA?

No, AOA seems fine (other than the issue with the triggers).

app/src/uhid/gamepad_uhid.c
```diff
@@ -0,0 +1,133 @@
...
+    sc_gamepad_uhid_send_input(gamepad, &hid_input, "gamepad axis");
+}
+
+static void
+sc_gamepad_processor_process_gamepad_button(struct sc_gamepad_processor *gp,
+                                const struct sc_gamepad_button_event *event) {
+    struct sc_gamepad_uhid *gamepad = DOWNCAST(gp);
+
+    struct sc_hid_input hid_input;
+    sc_hid_gamepad_generate_input_from_button(&gamepad->hid, &hid_input, event);
```
LHLaurini, line 97: Here, you don't check whether `sc_hid_gamepad_generate_input_from_button` succeeded or not. When either the mute button or the touchpad are pressed, the button is `-1`, which causes the function to fail.

So here's the bug, I guess.

app/src/usb/gamepad_aoa.c
```diff
@@ -0,0 +1,91 @@
...
+static void
+sc_gamepad_processor_process_gamepad_button(struct sc_gamepad_processor *gp,
+                                const struct sc_gamepad_button_event *event) {
+    struct sc_gamepad_aoa *gamepad = DOWNCAST(gp);
+
+    struct sc_hid_input hid_input;
+    if (!sc_hid_gamepad_generate_input_from_button(&gamepad->hid, &hid_input,
+                                                   event)) {
+        return;
+    }
```
LHLaurini, line 65: The same issue doesn't happen for AOA, as the check is present here.

app/src/uhid/gamepad_uhid.c
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+static void
+sc_gamepad_processor_process_gamepad_axis(struct sc_gamepad_processor *gp,
+                                const struct sc_gamepad_axis_event *event) {
+    struct sc_gamepad_uhid *gamepad = DOWNCAST(gp);
+
+    struct sc_hid_input hid_input;
+    sc_hid_gamepad_generate_input_from_axis(&gamepad->hid, &hid_input, event);
```
LHLaurini, line 86: Missing check here too.

app/src/uhid/gamepad_uhid.c
rom1v: Exactly, I just found it thanks to your comment, I was just about to post the patch :+1:

Thank you!

app/src/hid/hid_gamepad.c
```diff
@@ -0,0 +1,456 @@
...
+    0x95, 0x02,
+    // Input (Data, Variable, Absolute): 2 bytes (L2, R2)
+    0x81, 0x02,
+
+    // Usage Page (Buttons)
+    0x05, 0x09,
+    // Usage Minimum (1)
+    0x19, 0x01,
+    // Usage Maximum (15)
+    0x29, 0x0E,
```
yume-chan, line 93: Right stick button doesn't work. This should be `0x29, 0x0F`

rom1v: Thank you :+1:

rom1v:
I reworked the PR to fix all known issues. I also updated the binaries for Windows in the first post (draft 2).

Thank you for your feedback :heart:

app/src/hid/hid_gamepad.c
rom1v: In fact, what was important was:
```diff
-    // Report Count (15)
-    0x95, 0x0F,
+    // Report Count (16)
+    0x95, 0x10,
```

Here, the value should even be `0x29, 0x10` (not `0x0F`). But it seems harmless. If fix it locally.

LHLaurini:
> I reworked the PR to fix all known issues.

Everything's working correctly for me with `-G`/`--gamepad=uhid`. Only thing I was unable to test was using a second controller.

---

With `--gamepad=aoa`, changes to axes seem noticeably delayed. Also, if I turn the analog sticks in circles, I get the following message (repeated):

> WARN: Could not push AOA HID input (gamepad axis)

`-Vverbose` doesn't reveal anything interesting.

yume-chan:
This is how my gamepad rumble currently works:

```diff
diff --git a/app/src/scrcpy.c b/app/src/scrcpy.c
index c45f88ad..8509b32b 100644
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -756,6 +756,7 @@ aoa_complete:
         if (options->gamepad_input_mode == SC_GAMEPAD_INPUT_MODE_UHID) {
             sc_gamepad_uhid_init(&s->gamepad_uhid, &s->controller);
             gp = &s->gamepad_uhid.gamepad_processor;
+            uhid_gamepad = &s->gamepad_uhid;
         }
 
         struct sc_uhid_devices *uhid_devices = NULL;
diff --git a/app/src/uhid/gamepad_uhid.c b/app/src/uhid/gamepad_uhid.c
index 5f19a9d0..5a85816d 100644
--- a/app/src/uhid/gamepad_uhid.c
+++ b/app/src/uhid/gamepad_uhid.c
@@ -119,7 +119,15 @@ sc_gamepad_uhid_process_hid_output(struct sc_gamepad_uhid *gamepad,
         LOGI("==== HID output [%" PRIu16 "]", hid_id);
     }
 
-    // TODO
+    struct sc_hid_gamepad *gp = &gamepad->hid;
+    struct sc_hid_gamepad_slot* slot = &gp->slots[hid_id - SC_HID_ID_GAMEPAD_FIRST];
+    SDL_GameController *gc = SDL_GameControllerFromInstanceID(slot->gamepad_id);
+    // |           | HID Report | SDL Parameter |
+    // |-----------|------------|---------------|
+    // | Intensity | 0 - 100    | 0 - 65535     |
+    // | Duration  | 0 - 255    | 0 - 1000      |
+    SDL_GameControllerRumble(gc, data[4] * 0xFFFF / 100, data[5] * 0xFFFF / 100, data[6] * 1000 / 0xFF);
 }
 
 void
diff --git a/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java b/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
index 408dbf5d..2db043bc 100644
--- a/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
@@ -25,6 +25,7 @@ public final class UhidManager {
     private static final int UHID_INPUT2 = 12;
 
     // Linux: include/uapi/linux/input.h
+    private static final short BUS_BLUETOOTH = 0x05;
     private static final short BUS_VIRTUAL = 0x06;
 
     private static final int SIZE_OF_UHID_EVENT = 4380; // sizeof(struct uhid_event)
@@ -174,9 +175,9 @@ public final class UhidManager {
         buf.put("scrcpy".getBytes(StandardCharsets.US_ASCII));
         buf.put(empty, 0, 256 - "scrcpy".length());
         buf.putShort((short) reportDesc.length);
-        buf.putShort(BUS_VIRTUAL);
-        buf.putInt(0); // vendor id
-        buf.putInt(0); // product id
+        buf.putShort(BUS_BLUETOOTH);
+        buf.putInt(0x045e); // product id
+        buf.putInt(0x02fd); // vendor id
         buf.putInt(0); // version
         buf.putInt(0); // country;
         buf.put(reportDesc);

```

It uses this Linux driver: https://github.com/torvalds/linux/blob/d1f2d51b711a3b7f1ae1b46701c769c1d580fa7f/drivers/hid/hid-microsoft.c#L450-L461, so the device bus must be Bluetooth and the vender ID/product ID must match one of the list items.

I changed the server to always send those values, but this will break HID keyboards and mice, so it should be specified by the client.

This driver hard coded the output event format: https://github.com/torvalds/linux/blob/d1f2d51b711a3b7f1ae1b46701c769c1d580fa7f/drivers/hid/hid-microsoft.c#L44-L57. So the report descriptor (the `Usage Page (Physical Input Device)` part in https://github.com/Genymobile/scrcpy/pull/2130#issuecomment-2279467944) doesn't matter.

It doesn't support trigger rumbles (byte 2 and 3 in the output report) (notice the `MAGNITUDE_STRONG` enum value starts at 2, so `MAGNITUDE_NUM` is 4, but 0 and 1 are unused).

I believe there are other methods to support rumbles, for example with proper HID `Set Effect Report` usage. I can't find the reference documentation now, but this usage is super complicated. The host needs to first set several rumble presents (the device/Scrcpy virtual device needs to store them), then the host specifies which present and some options to play.

Also I believe wired Xbox/PlayStation controllers uses other specific drivers (and packet formats) to read inputs and play rumbles, but those can't be simulated by UHID.

yume-chan:
I also tested two Xbox controllers on Windows and it works:

image

rom1v:
@LHLaurini 

> With `--gamepad=aoa`, changes to axes seem noticeably delayed. Also, if I turn the analog sticks in circles, I get the following message (repeated):
>
>    WARN: Could not push AOA HID input (gamepad axis)

OK, I cannot reproduce with my device/gamepad, but we would need an input resampler to handle this case: #3088.

rom1v:
@yume-chan Thank you very much!

However, I fail to generate any rumble request (HID output) from Android.

I tested for example with "Game Controller Tester" (`uk.co.powgames.gamecondiag`), if I enable "Rumble left shoulder", it says: "no rumble support":

no_rumble

I tested from branch `gamepad.rumblewip` (which is the current gamepad branch + your diff, I squashed your first change with the previous commit).

Withoutruless:
Im on A14, UNICA Rom (OneUI 6.1)
Windows 11, using a PDP Xbox One Controller.

Here's some feedback:


- UHID works perfectly, everything works correctly except for vibration obv (tested using 'scrcpy -G --no-playback --mouse=disabled --keyboard=disabled')

 - On AOA, the Triggers dont do anything and the Right Stick acts as both of the Triggers instead.

Both "Triggers" start points are 0.5.

Left depresses LTRIGGER, Right presses LTRIGGER.
Up depresses RTRIGGER and Down presses RTRIGGER.

ive had this problem before on OTG but reversed (so the Triggers would act as the right stick instead) and i fixed that by modifying the "Generic.kl" in /system/usr/keylayout (ive tried vendor specific .kl many times, it just wouldn't work) so im guessing that's why its doing this.

this is the code in my modified Generic.kl if you need it:
```
Triggers
axis 0x02 LTRIGGER
axis 0x05 RTRIGGER

Left and right stick
axis 0x00 X flat 4096
axis 0x01 Y flat 4096
axis 0x03 Z flat 4096
axis 0x04 RZ flat 4096 
```


- On AOA you cant *only* use the gamepad, for example: "scrcpy --otg --gamepad=aoa --mouse=disabled --keyboard=disabled" gives you "Could not disable both keyboard and mouse in OTG mode."

- And finally, on AOA moving the mouse pins the Right Stick top left and the Left Stick as well, but the left stick moves with the mouse and teleports based on the direction you move your mouse:
Left = Left Stick tp to the right and start moving left.
Right = Left Stick tp to the left and start moving right.

Other than that, all buttons work fine on AOA and im pretty excited for this, thanks for dropping test binaries.

and i hope my gibberish is useful to you lol

rom1v:
> Here's some feedback:
>
>    UHID works perfectly, everything works correctly except for vibration obv (tested using 'scrcpy -G --no-playback --mouse=disabled --keyboard=disabled')
>

> On AOA, the Triggers dont do anything and the Right Stick acts as both of the Triggers instead.
> Both "Triggers" start points are 0.5.

Did you test draft1 or draft2?

These two problems you report for AOA should be fixed in draft2, but the behavior should be the same for UHID (broken in draft1, fixed in draft2).

> Left depresses LTRIGGER, Right presses LTRIGGER.
> Up depresses RTRIGGER and Down presses RTRIGGER.

Cannot reproduce. Probably related to the previous errors in draft1.

> On AOA you cant only use the gamepad, for example: "scrcpy --otg --gamepad=aoa --mouse=disabled --keyboard=disabled" gives you "Could not disable both keyboard and mouse in OTG mode."

Thank you, fixed :+1:

> And finally, on AOA moving the mouse pins the Right Stick top left and the Left Stick as well, but the left stick moves with the mouse and teleports based on the direction you move your mouse:
Left = Left Stick tp to the right and start moving left.
Right = Left Stick tp to the left and start moving right.

Oh thank you, stupid mistake, I assigned the same HID_ID to the mouse and to the first controller (off-by-one error in the hardcoded values :smile:).

yume-chan:
> I tested for example with "Game Controller Tester" (uk.co.powgames.gamecondiag), if I enable "Rumble left shoulder", it says: "no rumble support":

Is your device running Android 12+? Controller rumble support was added in Android 12.

This app works for me on an Android 13 device, but it always vibrates both motors no matter which one I choose (it also does this when my Xbox controller is connected to the device directly over Bluetooth).

From the permission prompt it shows on connection, it must use Android USB API to directly talk to controllers, so it can support all motors on all Android versions. (I think it's not possible to emulate USB devices through software)

I test my implementation using Chrome and this code I wrote: https://codepen.io/yume-chan/pen/qBzLGNd. Move the sticks to control vibration. On PC it supports all four motors, but on Android it only supports two. (again, same as when the controller is connected over Bluetooth or USB). My code can vibrate left and right motors separately.

I checked again, at least for Xbox controllers, neither `hid-microsoft` nor `xpad` (https://github.com/paroj/xpad/blob/3a215825f981643d164aa4e10fd2ab584fd9987c/xpad.c#L1740-L1756) driver supports trigger rumble, so I think that's why Chrome doesn't support 4 motors no matter how the controller is connected (over USB, over Bluetooth, or virtual device using `hid-microsoft` driver).

rom1v:
> Is your device running Android 12+? Controller rumble support was added in Android 12.

Yes, Android 14.

I just tested, in fact it works like you say on a Pixel 8 with Android 14, but not on a Samsung Galaxy Tab with Android 14.

Btw, if I set the vid:pid to 02fd:045e, the device is shown as a `Xbox One S Controller`. If I don't, I get "scrcpy" with the actual gamepad name as reported by SDL:

gamepad

Given that:
 - rumble only works on some devices
 - not all motors work
 - individual motors are not taken into account (only both vibrate)
 - the feature is probably not used very much in Android games
 - it impacts how the device is recognized

I think I will not support rumble for now. Maybe in the future if things improve.

Thank you very much for your research and documentation in this thread :+1:

rom1v:
Do you know any (free) Android game which supports 2 or more gamepads (a two-player fighting game for example)?

Withoutruless:
> Do you know any (free) Android game which supports 2 or more gamepads (a two-player fighting game for example)?

afaik "Vita Fighters" supports 2 controllers

rom1v:
@Withoutruless Thank you, it works :+1:

Withoutruless:
> > On AOA, the Triggers dont do anything and the Right Stick acts as both of the Triggers instead.
> > Both "Triggers" start points are 0.5.
> 
> Did you test draft1 or draft2?

Ok so i copied the .kl for the PDP Xbox One Controller (should be the same for X360 and the normal XOne too) and made a custom .kl called "Vendor_ffff_Product_ffff.kl"

and changed the text from:

```
# Triggers.
axis 0x02 LTRIGGER --
axis 0x05 RTRIGGER --

# Left and right stick.
# The reported value for flat is 128 out of a range from -32767 to 32768, which is absurd.
# This confuses applications that rely on the flat value because the joystick actually
# settles in a flat range of +/- 4096 or so.
axis 0x00 X flat 4096
axis 0x01 Y flat 4096
axis 0x03 Z flat 4096 --
axis 0x04 RZ flat 4096 --

# This works on the physical OTG (no Scrcpy)
```
To:
```
# Triggers.
axis 0x0a LTRIGGER --
axis 0x09 RTRIGGER --

# Left and right stick.
# The reported value for flat is 128 out of a range from -32767 to 32768, which is absurd.
# This confuses applications that rely on the flat value because the joystick actually
# settles in a flat range of +/- 4096 or so.
axis 0x00 X flat 4096
axis 0x01 Y flat 4096
axis 0x02 Z flat 4096 --
axis 0x05 RZ flat 4096 --

# This works on Scrcpy AOA.
```

so now i can use both AOA and UHID just fine.

everything else i mentioned has been fixed so thank you for that haha.

rom1v:
> So i copied the .kl for the PDP Xbox One Controller (should be the same for X360 and the normal XOne too) and made a custom .kl called "Vendor_ffff_Product_ffff.kl" (which is what scrcpy uses on AOA)

Where do you copy that file? Is there something to change in this PR so that it works the same "out-of-the-box"?

Withoutruless:
> Where do you copy that file?

i copy that file to "/system/usr/keylayout" and it changes the way android handles input for the device with that specific vendor & product id.

> Is there something to change in this PR so that it works the same "out-of-the-box"?

if its possible to change the vendor & product id so it matches something like an Xbox 360 Wired Controller 
(Vendor = 045e, Product = 028e)

then that'd make android handle the received input differently and you wouldnt have to accomodate for the generic.kl's bugged RT & RStick 

which im assuming you do by sending:

LTRIGGER as axis 0x0a instead of 0x02
RTRIGGER as axis 0x09 instead of 0x05

Z as axis 0x02 instead of 0x03
and
RZ as axis 0x05 instead of 0x04

so you could try changing those values and the emulated controller vendor & product id and see if it works.
but im not an expert so... sorry if im wrong.

also would it be possible to add an optional background capture for UHID & AOA so it captures input regardless if the window is in focus or not?

rom1v:
> if its possible to change the vendor & product id so it matches something like an Xbox 360 Wired Controller
(Vendor = 045e, Product = 028e)

That's what I did here: https://github.com/Genymobile/scrcpy/pull/5270#issuecomment-2337792505
But then, of course, the controller is shown as a Xbox 360 controller, even if the physical one is a PS4 controller.

I might use the real vid:pid of the physical device, but that might also cause inconsistencies between controllers I guess (I pass a custom report descriptor).

> also would it be possible to add an optional background capture for UHID & AOA so it captures input regardless if the window is in focus or not?

The problem is the same as with mouse and keyboard: I can't receive input events from SDL without a focused window (AFAIK).

LHLaurini:
> The problem is the same as with mouse and keyboard: I can't receive input events from SDL without a focused window (AFAIK).

There's [SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS](https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS).

Withoutruless:
> > also would it be possible to add an optional background capture for UHID & AOA so it captures input regardless if the window is in focus or not?
> 
> The problem is the same as with mouse and keyboard: I can't receive input events from SDL without a focused window (AFAIK).

on LHLaurini's build i was able to kinda "hack" it by clicking on another window so the scrcpy one wouldnt gain focus and that uses SDL too no?

> But then, of course, the controller is shown as a Xbox 360 controller, even if the physical one is a PS4 controller.
> 
> I might use the real vid:pid of the physical device, but that might also cause inconsistencies between controllers I guess (I pass a custom report descriptor).

yeah, using the real vid:pid wouldnt work as is, cause you're passing the correct axis values for Generic.kl not the vendor specific one:

> > LTRIGGER as axis 0x0a instead of 0x02
> > RTRIGGER as axis 0x09 instead of 0x05
> 
> > Z as axis 0x02 instead of 0x03
> > and
> > RZ as axis 0x05 instead of 0x04

so on an xbox 360 controller, for which there's a keylayout for ( "/system/usr/keylayout/Vendor_045e_Product_028e.kl"), 
the correct axis values sent from pc would be:

```
axis 0x02 LTRIGGER
axis 0x05 RTRIGGER

axis 0x03 Z
axis 0x04 RZ
```

and if those were incorrect, then you'd be met with the same issue i had.

also im calling it "value" here but im definitely saying it wrong, so what is this called "0x02".

rom1v:
I did a full review, fixed issues, added support for gamepad events in background (https://github.com/Genymobile/scrcpy/pull/5270#issuecomment-2339360460)…

Please test this new version (branch `gamepad_draft4`), I also published a new binary in the first post.

One remaining problem that I cannot explain: in AOA mode, 2 physical gamepads are detected as a single one, and buttons/axis from one controller do the same as from the other controller.

```bash
scrcpy --otg -G
scrcpy --gamepad=aoa
```

It works well in UHID:

```bash
scrcpy -G
scrcpy -KMG   # UHID keyboard, mouse and gamepad
```

Any idea?

rom1v:
Removing "draft". Ready to be merged.

rom1v:
Ref issue #5362.

Issue #4489, opened 2023-12-03
MacOS Binary request !!!


Gary-Cod:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?

Apple M1 Mac
MacOS Sonoma 14.1.2

Installing scrcpy on MacOS is very bad experience unlike windows. Installing using Homebrew or Macports involves installing a lot of unnecessary packages & wasting a lot of storage. 

**Describe the solution you'd like**
Please provide binaries for MacOS (Apple Silicon) just like the Windows ...

PR #5306, opened 2024-09-18
Add GitHub Actions workflow for release builds

### LINKED ISSUE #4490: CI for release builds
Since the beginning, I wanted a single script to build and release everything, so from a Linux machine I can execute `./release.sh` and produce all the binaries to publish.

But it is too restrictive: although the Windows release can be cross-compiled from Linux, it's not the same for macOS.

I'm considering (for a long time) to write one script per target (`release_windows.sh`, `release_macos.sh`, `release_deb.sh`…), that will be run on CI (e.g. GitHub actions) on different platforms (`release_macos.sh` will be run on a macOS VM for example). This CI will produce the release binaries for all platforms.

I started to refactor a bit the cross-compilation script in 67f356f881898e21edabd4ecd2fb509ad4c92362, and ~I plan to integrate [scrcpy-deps](https://github.com/rom1v/scrcpy-deps/) directly in the scrcpy repo~ (done by #4713), with scripts to build minimal binaries of the FFmpeg libs for Windows and macOS (for Linux, scrcpy will depend on the FFmpeg from the distribution).

In practice, there are 2 parts to build:
 - `scrcpy-server`, which requires the Android SDK, and which is common to every platform
 - the scrcpy client, one per platform/architecture

I don't want every CI job to build its own `scrcpy-server` (it is unnecessary, it would produce different binaries for each target, and it would make them all require the Android SDK). So I would like a separate job just for the server. But in the end, `scrcpy-server` must be included in the final release for each target (in `scrcpy-win64.zip`, in `scrpy-win32.zip`, in `scrcpy-ubuntu23.10.deb`, etc…) which are produced by the CI.

What is the correct way to achieve this kind of job dependency (each one must depend on the output of the job which produces `scrcpy-server`)? (I'm am very inexperienced in CI)

Thank you for your help and suggestions :slightly_smiling_face: 

Refs https://github.com/Genymobile/scrcpy/issues/2256#issuecomment-1218375121 and next comments (cc @dur-randir @LeeBinder @Coool)
Refs #1709 (cc @dylanmtaylor)
Refs #3721 (cc @qmfrederik)
Refs #4427 (cc @hacksysteam)
Refs #4489 (cc @Gary-Cod)

rom1v:
Better late than never (#4490), I finally worked on a GitHub Actions workflow to build scrcpy releases.

I had to refactor the `release.mk` makefile to parallelize server and client build jobs. It is only run manually to build a release (not on push or pull request, I don't need it for now).

ga

In the end, it produces a zip artifact containing:
 - `scrcpy-server-XXX`
 - `scrcpy-win32-XXX.zip`
 - `scrcpy-win64-XXX.zip`
 - `SHA256SUMS.txt`

where `XXX` is the input name you give when you manually start the workflow (if empty, it uses the ref name, for example the branch or tag name).

To test it:
 - fork the repo to your own account
 - checkout the branch from this PR, and push it to your `master` branch
 - enable GitHub Actions in your repo settings
 - go to the Actions tab
 - select "Build"
 - select "Run workflow"
 - keep `master` branch, choose a version name (for example `v42`)
 - click on Run workflow

workflow

---

For now, it generates the same targets as `./release.sh` did locally (scrcpy-server, release win32, release win64).

The goal now is to add more jobs, especially:
 - macOS releases
 - .deb packages for Ubuntu and Debian recent versions (#4427)

Your help is welcome! :heart:

Now that scrcpy-server is generated separately once and for all, it will be easier to write the jobs for building the scrcpy client for different targets (Android SDK is not needed).

As an important constraint, the jobs MUST only reference "official" github actions (or dockers images), not actions/images from random users. Keep it as simple as possible.

---

Fixes #4490
Refs https://github.com/Genymobile/scrcpy/issues/2256#issuecomment-1218375121 and next comments (cc @dur-randir @LeeBinder @Coool)
Refs #1709 (cc @dylanmtaylor)
Refs #3721 (cc @qmfrederik)
Refs #4427 (cc @hacksysteam)
Refs #4489 (cc @Gary-Cod)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,147 @@
+name: Build
+
+on:
+  workflow_dispatch:
+    inputs:
+      name:
+        description: 'Version name (default is ref name)'
+
+jobs:
+  build-scrcpy-server:
+    runs-on: ubuntu-latest
+    env:
+      GRADLE: gradle  # use native gradle instead of ./gradlew in release.mk
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '17'
+
+      - name: Test scrcpy-server
+        run: make -f release.mk test-server
+
+      - name: Build scrcpy-server
+        run: make -f release.mk build-server
+
+      - name: Upload scrcpy-server artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: scrcpy-server
+          path: build-server/server/scrcpy-server
+
+  test-client:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y meson ninja-build nasm ffmpeg libsdl2-2.0-0 \
+             libsdl2-dev libavcodec-dev libavdevice-dev libavformat-dev \
+             libavutil-dev libswresample-dev libusb-1.0-0 libusb-1.0-0-dev
+
+      - name: Build
+        run: |
+          meson setup d -Db_sanitize=address,undefined
+
+      - name: Test
+        run: |
+          meson test -Cd
+
+  build-win32:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y meson ninja-build nasm ffmpeg libsdl2-2.0-0 \
+             libsdl2-dev libavcodec-dev libavdevice-dev libavformat-dev \
+             libavutil-dev libswresample-dev libusb-1.0-0 libusb-1.0-0-dev \
+             mingw-w64 mingw-w64-tools libz-mingw-w64-dev
+
+      - name: Workaround for old meson version run by Github Actions
+        run: sed -i 's/^pkg-config/pkgconfig/' cross_win32.txt
+
+      - name: Build scrcpy win32
+        run: make -f release.mk build-win32
+
+      - name: Upload build-win32 artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: build-win32-intermediate
+          path: build-win32/dist/
+
+  build-win64:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y meson ninja-build nasm ffmpeg libsdl2-2.0-0 \
+             libsdl2-dev libavcodec-dev libavdevice-dev libavformat-dev \
+             libavutil-dev libswresample-dev libusb-1.0-0 libusb-1.0-0-dev \
+             mingw-w64 mingw-w64-tools libz-mingw-w64-dev
+
+      - name: Workaround for old meson version run by Github Actions
+        run: sed -i 's/^pkg-config/pkgconfig/' cross_win64.txt
+
+      - name: Build scrcpy win64
+        run: make -f release.mk build-win64
+
+      - name: Upload build-win64 artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: build-win64-intermediate
+          path: build-win64/dist/
+
+  package:
+    needs:
+      - build-scrcpy-server
+      - build-win32
+      - build-win64
+    runs-on: ubuntu-latest
+    env:
+      # $VERSION is used by release.mk
+      VERSION: ${{ github.event.inputs.name || github.ref_name }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Download scrcpy-server
+        uses: actions/download-artifact@v4
+        with:
+          name: scrcpy-server
+          path: build-server/server/
+
+      - name: Download build-win32
+        uses: actions/download-artifact@v4
+        with:
+          name: build-win32-intermediate
+          path: build-win32/dist/
+
+      - name: Download build-win64
+        uses: actions/download-artifact@v4
+        with:
+          name: build-win64-intermediate
+          path: build-win64/dist/
+
+      - name: Package
+        run: make -f release.mk package
+
+      - name: Upload release artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: scrcpy-release-${{ env.VERSION }}
+          path: release-${{ env.VERSION }}
--- a/release.mk
+++ b/release.mk
@@ -9,7 +9,7 @@
 # the server to the device.
 
 .PHONY: default clean \
-	test \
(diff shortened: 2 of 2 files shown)


LeeBinder:
Awesome, great work, Romain. I'm recovering health-wise right now which might take a while so will be mostly offline for the near future :/

dur-randir:
I've added macos build job + modified makefile in my fork at https://github.com/dur-randir/scrcpy/commit/6c4c2c33b34d028de4227f6e5c93b6ad57b610fe. The next question is, this generates unsigned binaries - e.g. you either have toprovide a signing key (see https://docs.github.com/en/actions/use-cases-and-examples/deploying/installing-an-apple-certificate-on-macos-runners-for-xcode-development) or a user have to perform additional steps after download (like brew does automatically when installing from casks).

Genxster1998:
> I've added macos build job + modified makefile in my fork at [dur-randir@6c4c2c3](https://github.com/dur-randir/scrcpy/commit/6c4c2c33b34d028de4227f6e5c93b6ad57b610fe). The next question is, this generates unsigned binaries - e.g. you either have to provide a signing key (see https://docs.github.com/en/actions/use-cases-and-examples/deploying/installing-an-apple-certificate-on-macos-runners-for-xcode-development) or a user have to perform additional steps after download (like brew does automatically when installing from casks).

amd64 aka x86_64 build actions config ? I can't see them there.

rom1v:
@dur-randir Thank you very much!

A few questions/remarks:
 - Could `app/deps/ffmpeg-macos.sh` be "merged" into `app/defs/ffmpeg.sh` to avoid duplication (a common file for all platforms)?
 - It seems only FFmpeg is built for macOS, not the other dependencies (SDL and libusb)? (and an `adb` binary for macOS)
 - Is `macpack` absolutely necessary? Can we build a release without this tool?

I'll read about signing keys later.

dur-randir:
>Could app/deps/ffmpeg-macos.sh be "merged" into app/defs/ffmpeg.sh to avoid duplication (a common file for all platforms)?

Sure, I've done it separately because of didn't know in advance what'd be needed to be changed. There's dievergence just in two places - additional args list for configure `disable-debug..enable-videotoolbox` and `export PATH`, which can be put under `if $HOST` in deps/common.

>It seems only FFmpeg is built for macOS, not the other dependencies (SDL and libusb)? (and an adb binary for macOS)

They're taken from `brew install` in binary form.

>Is macpack absolutely necessary? Can we build a release without this tool?

Yes and no. I've tried (again!) to make a fully static build (link ffmpeg + libusb + libsdl) but I don't know meson at all and it broke somewhere down the road while double-linking _ssh_$whatever while I was trying to shove them all together and then I ceased trying. Without static linking with those dependencies, one must relocate all libraries to be searched at relative paths by DYLD (they're baked by absolute paths by default on macos). IIRC this can be done with native instruments, but they're non-recursive, so I've picked one tool that works recursively out-of-the box. Another one tool is https://github.com/auriamg/macdylibbundler which can be installed from brew. Or create this recursive hand-written loop for path change. Or ideally make a fully static build, but I've given up. If you know meson, maybe you can create a recipe for at least linux, so I can tweak it for macos?

rom1v:
> Sure, I've done it separately because of I didn't know in advance what'd be needed to be changed.

:+1:

> They're taken from `brew install` in binary form.

OK :+1: In the end, I think we want to build them without depending on brew.

> IIRC this can be done with native instruments, but they're non-recursive, so I've picked one tool that works recursively out-of-the box.

Indeed, the native tool seems to be `install_name_tool` (which is used by `macpack`).

For windows, the scrcpy builds have 6 dll (so these libs are linked dynamically):

 - `avcodec-61.dll`
 - `avformat-61.dll`
 - `avutil-59.dll`
 - `swresample-5.dll`
 - `libusb-1.0.dll`
 - `SDL2.dll`

I think the macOS build could do the same, there would be no need for finding dependencies recursively.

Genxster1998:
Here, i added for MacOS x86-64 thanks to @dur-randir [db01900](https://github.com/Genxster1998/scrcpy/commit/db019004aeda96f6083b6c25645f36a760af4de7)[
artifacts](https://github.com/Genxster1998/scrcpy/actions/runs/11261791172)

LeeBinder:
Awesome, you ROCK, @Genxster1998, thank you 👍 

These two script files come in handy 😉:

[assign icon.png to scrcpy binary.zip](https://github.com/user-attachments/files/17319344/assign.icon.png.to.scrcpy.binary.zip)

[scrcpy-launcher.zip](https://github.com/user-attachments/files/17319349/scrcpy-launcher.zip)

Unpack them into the same dir as scrcpy, scrcpy-server etc.
run 'assign icon.png to scrcpy binary' first
run 'scrcpy-launcher' to launch scrcpy

LeeBinder:
.. and for those who like it as a native macOS app 😀:

[scrcpy 2.7 (Intel) app.zip](https://github.com/user-attachments/files/17319453/scrcpy.2.7.Intel.app.zip)

Genxster1998:
@LeeBinder , you won't need a launcher to furnish path to scrcpy-server as it has been set default to executable_path already rather than /opt/... so no need envt. just adb,scrcpy,libs,server in same dir after patching binary for icon. i think *.dylib(s), adb and scrcpy-server can be moved to Frameworks in proper macOS app.
> run 'scrcpy-launcher' to launch scrcpy

LeeBinder:
@Genxster1998 ideally, but it's not working (yet). Because when I simply run scrcpy even if scrcpy-server IS present in the same directory, I get:

```
ERROR: Could not get local file path, using scrcpy-server from current directory
stat: No such file or directory
ERROR: 'scrcpy-server' does not exist or is not a regular file
```
Try it yourself, please.

Looks as if you need to add something in the compile/ build script so that scrcpy indeed DOES look in the same directory. Something is still missing.

rom1v:
> Looks as if you need to add something in the compile/ build script so that scrcpy indeed DOES look in the same directory.

The option is `build_portable`: https://github.com/Genymobile/scrcpy/blob/665ccb32f5306ebd866dc0d99f4d08ed2aeb91c3/meson_options.txt#L4

(add `-Dbuild_portable=true` to the meson command)

However, it the code to find the executable path is not implemented for macos:

https://github.com/Genymobile/scrcpy/blob/665ccb32f5306ebd866dc0d99f4d08ed2aeb91c3/app/src/sys/unix/file.c#L64-L67

Genxster1998:
@LeeBinder Thats true indeed , Launching app from launchpad or Finder is not working. I tried adding `LSEnvironment` dict keys in info.plist which is specifically made for same purpose. yet it still fails to acknowledge server path from there. btw it works with `open -a scrcpy.app` in terminal. @rom1v Yeah, it is compiling with portable boolean true for meson, i guess a static path could be assigned for MacOS GUI app in ../Resources or ../Libraries Directory.

LeeBinder:
I see. Agreed, should eventually be in ../Resources (neither in ../Libraries nor in ../Frameworks because it's neither of those but a Android executable binary)

rom1v:
The scripts to build dependencies are currently specific to cross-compile Windows from Linux. I'll have to adapt them to also support native build (will be useful for macOS). Meanwhile, I can merge this PR.

rom1v:
See #5515.

PR #5334, opened 2024-10-02
Added shortopt-flag '-z' for --shortcut-mod

zdn-gh:
I think it would be nice to have a shorter version of `--shortcut-mod`.
Since the `s` and `m` flags are already in use I chose `z`, I think `x` would fit as well.


rom1v:
Hi,

I would like to avoid short options for "advanced features", and reserve them for more common features.

(The first versions of scrcpy, I always assigned shortopts, but that's not the case anymore.)

Issue #1887, opened 2020-11-06
Single App Package Screen Mirroring instead of whole screen 


Genxster1998:
- [* ] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
No, but feature enhancement. Instead of using single screen i would like to mirror different app windows in free form mode independently of main canvas.

**Describe the solution you'd like**
an option like passing package=com.android.contacts to mirror specific app only screen instead of whole ui.

**Describe alternatives you've considered**

**Additional context**

guanzhangrtk:
If I understand the feature request correctly, you are asking for app-specific screens to be mirrored. If this indeed can be implemented I wonder if that means we can have different scrcpy windows per app so that you can sort of run and control multiple apps at once?

Genxster1998:
@guanzhangrtk yepp! Totally right , but I don't know it's possible to spawn single app window screen without system am policies implementation ? 
Like  clicking an app in special launcher app leads to open in new scrcpy window .

guanzhangrtk:
Yeah I don't know if it is possible but I know that with MuMu App Player (android emulator) you can open multiple apps and then drag them off the main window and they would continue running in the background and you switch between them as if they're both running simultaneously. This is the only android emulator I'm aware of that has this feature (at least on the Mac).

hltdev8642:
It isnt really a *solution*, but using a split terminal with windows terminal  and termux over usb/adb is pretty effective.


![tmxsetup](https://user-images.githubusercontent.com/39349712/114227252-bba55d00-9942-11eb-9b12-d78fef474145.png)




 Also, made a `.run` alias that generates a dialog list of apps and you can launch them. (can post that if anyone wants it). 

I get it though, it would be great to be able to launch individual apps kind of like the way an x11 GUI app runs in wsl or something...

guanzhangrtk:
@hltdev8642 What's the benefit of doing this instead of just running `adb shell` directly from Windows?

Genxster1998:
[...]

PR #5370, opened 2024-10-12
Add virtual display feature

### LINKED ISSUE #4598: Android 14: Samsung phones - mouse actions (left click/scroll) stop working on virtual display
Hi developers.
I am facing an issue when using scrcpy (v2.3.1) after updating my Samsung S22 ultra phone (SM-S908E) from Android 13 to 14.
An other display is automatically created when scrcpy connects with the phone.

Display id numbers before starting scrcpy
```
  mDisplayId=: 0 -> primary display (phone screen)
  mDisplayId=: 2 -> for Samsung DeX display
```

Display id numbers after starting scrcpy
```
  mDisplayId=: 0 -> primary display (phone screen)
  mDisplayId=: 2 -> for Samsung DeX display
  mDisplayId=: 80 -> a new display id is created
```
A new display (id=80) is created. But there is no issue with this if I want to mirror the phone screen. Everything works perfectly.

Similarly, when I want to mirror virtual display with option (--display-id 90) as some ways mentioned in https://github.com/Genymobile/scrcpy/issues/1887#issuecomment-1775028321 and https://github.com/Genymobile/scrcpy/issues/1887#issuecomment-1847235527, an new display is also created with id=91
```
  mDisplayId=: 0 -> primary display (phone screen)
  mDisplayId=: 2 -> for Samsung DeX display
  mDisplayId=:90 -> virtual display id
  mDisplayId=:91 -> a new display id is created
```

scrcpy can show the virtual display (id=90), input text from keyboard works but click action does not work. 
I try to perform input events by using adb and found that virtual display can receive inputs via display with id=91 instead of id=90

```
adb shell input -d 90 tap 100 800 -> not work
adb shell input -d 91 tap 100 800 -> work
```
-> I think the created display (id=91) is used for receiving the tap/click events. In order to prove that I start another scrcpy with --display-id=91 option. Another scrcpy can show the virtual display content and all click events work (they are injected to display with id=91).
I think this may be the intention of Samsung for Samsung DeX mode or something else.

Can you help give adivces on this case? Thanks,

rom1v:
This PR adds the possibility to start a new virtual display instead of mirroring the screen.

## Implementation

There are 3 steps:
 1. refactors
 2. rework event handling for virtual displays
 3. add virtual display features

### Refactors

The (7) firsts commits refactor to move code from `Device` to `Controller` and `ScreenCapture`, to prepare the next commits.

### Events handling

The commit `Inject display-related events to virtual display` aims to fix #4598 and #5137 (based on the work by @mengyanshou in #5137 and #5214), also necessary for making virtual display events work.

Basically, for event injection, there are two display ids:
 - the source display id (`--display-id=…`, default to 0 for the main display)
 - the id of the virtual display used for mirroring

(In case the ScreenCapture uses the "SurfaceControl API", then both ids are equals, but this is an implementation detail.)

In order to make events work correctly in all cases:
 - the virtual display id must be used for events relative to the display (mouse and touch events with coordinates);
 - the source display id must be used for other events (like key events).

### Virtual display

Then, based on the discussions in #1887 and the work by @yume-chan and @anirudhb:
 - https://github.com/yume-chan/scrcpy/commits/feat/virtual-display-3/ (https://github.com/Genymobile/scrcpy/issues/1887#issuecomment-1405184608)
 - https://github.com/anirudhb/scrcpy/commits/virtual-display/ (https://github.com/Genymobile/scrcpy/issues/1887#issuecomment-2240359603)

The commit `Add virtual display feature` implements a first version of virtual display support:

```bash
scrcpy --new-display=1920x1080
scrcpy --new-display=1920x1080/420  # force 420 dpi
scrcpy --new-display         # use the default screen size and density
scrcpy --new-display=/240    # use the default screen size and 240 dpi
scrcpy --new-display -m1920  # use the default screen size and density, scaled to fit in 1920
```

The size is fixed (not resizable).

On a Pixel 8, the virtual display contains a menu, so it is possible to open an app easily.

On other devices, it may be empty. In that case, you can start your app manually. Find the new display id is scrcpy logs (let's say it's 42), and start an app manually:

```bash
adb shell am start -a android.settings.SETTINGS --display 42
```

To start an app manually:

```bash
adb shell cmd package resolve-activity --brief org.mozilla.firefox | tail -n 1  # get the activity name
adb shell am start -S -n org.mozilla.firefox/.App --display 5
```

### Start app

The feature really needs a new command to start an app, for example:

```bash
# not implemented, just an idea
scrcpy --new-display --start-app=firefox
```

_I might need to rework how arguments are passed to the server (to start an app "My gallery" for example), because currently it does not support spaces or other special characters (see bec3321fff4c6dc3b3dbc61fdc6fd98913988a78)._

I don't know if the expected behavior is to just have a virtual display and open any app (like on Pixel 8, where there is a "launcher"), or expose something like a "single app" feature.

It probably depends on the device (I don't know if it's possible to limit a single app on Pixel 8 for example, or to navigate to other apps on device without a launcher). What do you think? (also ref discussion https://github.com/Genymobile/scrcpy/issues/1887#issuecomment-1407012535 @4nric)

### Detect failure

If I run `scrcpy --new-display` on a Nexus 5 with Android 6, it does not fail, it just waits indefinitely for a frame. How to detect that "it works"? Should we only use checks based on the Android version?

---

### Start app

Here is a draft v2, with new options to list and start apps (edit: changed by https://github.com/Genymobile/scrcpy/pull/5370#issuecomment-2424988000):

```bash
scrcpy --list-apps
scrcpy --new-display --start-app=org.mozilla.firefox  # start app by package name
scrcpy --new-display --start-app=?firefox             # start app by name with '?' prefix (takes more time)
scrcpy --start-app=org.mozilla.firefox   # also works for real display mirroring
scrcpy --start-app=+org.mozilla.firefox  # prefix by '+' to force-stop before starting the app
```

There is a performance issue on some devices to retrieve the app names (but not package names), because [`loadLabel()`](https://developer.android.com/reference/android/content/pm/PackageItemInfo#loadLabel(android.content.pm.PackageManager)) must be called for each app). As a consequence, `--list-apps` may take several seconds, and selecting an application by name may also take time. However, starting via package name is quick. Any solution welcome.

## Binary

Here is a Windows binary:

- [`scrcpy-win64-vd3.zip`](https://tmp.rom1v.com/scrcpy/5370/3/scrcpy-win64-vd3.zip) <sub>`SHA-256: 013a4b2a82d12d2a006f128c243130f939331cdd71e0c28886438da1a83c38b`</sub>

<details><summary>old binaries</summary>

- [`scrcpy-win64-vd.zip`](https://tmp.rom1v.com/scrcpy/5370/1/scrcpy-win64-vd.zip) <sub>`SHA-256: bda97902fe727dd236879c0c3850f62e0bdb85aac50f37aa5dc44f8fc91c65f`</sub>
- [`scrcpy-win64-vd2.zip`](https://tmp.rom1v.com/scrcpy/5370/2/scrcpy-win64-vd2.zip) <sub>`SHA-256: ab509d5943bc9e86739c1e6e1abab12f8bc75eb5f414b5c4b25cc7e1b1a3e11`</sub>

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ It focuses on:
 Its features include:
  - [audio forwarding](doc/audio.md) (Android 11+)
  - [recording](doc/recording.md)
+ - [virtual display](doc/virtual_display.md)
  - mirroring with [Android device screen off](doc/device.md#turn-screen-off)
  - [copy-paste](doc/control.md#copy-paste) in both directions
  - [configurable quality](doc/video.md)
@@ -91,6 +92,12 @@ Here are just some common examples.
     scrcpy --video-codec=h265 -m1920 --max-fps=60 --no-audio -K  # short version
     ```
 
+ - Start VLC in a new virtual display (separate from the device display):
+
+    ```bash
+    scrcpy --new-display=1920x1080 --start-app=org.videolan.vlc
+    ```
+
  - Record the device camera in H.265 at 1920x1080 (and microphone) to an MP4
    file:
 
@@ -134,6 +141,7 @@ documented in the following pages:
  - [Device](doc/device.md)
  - [Window](doc/window.md)
  - [Recording](doc/recording.md)
+ - [Virtual display](doc/virtual_displays.md)
  - [Tunnels](doc/tunnels.md)
  - [OTG](doc/otg.md)
  - [Camera](doc/camera.md)
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -79,6 +79,7 @@ _scrcpy() {
         -s --serial=
         -S --turn-screen-off
         --shortcut-mod=
+        --start-app=
         -t --show-touches
         --tcpip
         --tcpip=
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -33,6 +33,7 @@ _scrcpy() {
         --keyboard=
         --kill-adb-on-close
         --legacy-paste
+        --list-apps
         --list-camera-sizes
         --list-cameras
         --list-displays
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -46,6 +46,8 @@ _scrcpy() {
         --mouse-bind=
         -n --no-control
         -N --no-playback
+        --new-display
+        --new-display=
         --no-audio
         --no-audio-playback
         --no-cleanup
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -52,6 +52,7 @@ arguments=(
     '--mouse-bind=[Configure bindings of secondary clicks]'
     {-n,--no-control}'[Disable device control \(mirror the device in read only\)]'
     {-N,--no-playback}'[Disable video and audio playback]'
+    '--new-display=[Create a new display]'
     '--no-audio[Disable audio forwarding]'
     '--no-audio-playback[Disable audio playback]'
     '--no-cleanup[Disable device cleanup actions on exit]'
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -82,6 +82,7 @@ arguments=(
     {-s,--serial=}'[The device serial number \(mandatory for multiple devices only\)]:serial:($("${ADB-adb}" devices | awk '\''$2 == "device" {print $1}'\''))'
     {-S,--turn-screen-off}'[Turn the device screen off immediately]'
     '--shortcut-mod=[\[key1,key2+key3,...\] Specify the modifiers to use for scrcpy shortcuts]:shortcut mod:(lctrl rctrl lalt ralt lsuper rsuper)'
+    '--start-app=[Start an Android app]'
     {-t,--show-touches}'[Show physical touches]'
     '--tcpip[\(optional \[ip\:port\]\) Configure and connect the device over TCP/IP]'
     '--time-limit=[Set the maximum mirroring time, in seconds]'
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -40,6 +40,7 @@ arguments=(
     '--keyboard=[Set the keyboard input mode]:mode:(disabled sdk uhid aoa)'
     '--kill-adb-on-close[Kill adb when scrcpy terminates]'
     '--legacy-paste[Inject computer clipboard text as a sequence of key events on Ctrl+v]'
+    '--list-apps[List Android apps installed on the device]'
     '--list-camera-sizes[List the valid camera capture sizes]'
     '--list-cameras[List cameras available on the device]'
     '--list-displays[List displays available on the device]'
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -228,6 +228,10 @@ Inject computer clipboard text as a sequence of key events on Ctrl+v (like MOD+S
 This is a workaround for some devices not behaving as expected when setting the device clipboard programmatically.
 
 .TP
+.B \-\-list\-apps
+List Android apps installed on the device.
+
+.TP
 .B \-\-list\-camera\-sizes
 List the valid camera capture sizes.
 
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -498,6 +498,10 @@ Default is "lalt,lsuper" (left-Alt or left-Super).
 .BI "\-\-start\-app " name
 Start an Android app, by its exact package name.
 
+Add a '+' prefix to force-stop before starting the app:
+
+    scrcpy --new-display --start-app=+org.mozilla.firefox
+
 .TP
 .B \-t, \-\-show\-touches
 Enable "show touches" on start, restore the initial value on exit.
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -495,6 +495,10 @@ For example, to use either LCtrl or LSuper for scrcpy shortcuts, pass "lctrl,lsu
 Default is "lalt,lsuper" (left-Alt or left-Super).
 
 .TP
+.BI "\-\-start\-app " name
+Start an Android app, by its exact package name.
+
+.TP
 .B \-t, \-\-show\-touches
 Enable "show touches" on start, restore the initial value on exit.
 
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -498,9 +498,17 @@ Default is "lalt,lsuper" (left-Alt or left-Super).
 .BI "\-\-start\-app " name
 Start an Android app, by its exact package name.
 
+Add a '?' prefix to select an app whose name starts with the given name, case-insensitive (retrieving app names on the device may take some time):
+
+    scrcpy --start-app=?firefox
+
 Add a '+' prefix to force-stop before starting the app:
 
     scrcpy --new-display --start-app=+org.mozilla.firefox
+
+Both prefixes can be used, in that order:
+
+    scrcpy --start-app=+?firefox
 
 .TP
 .B \-t, \-\-show\-touches
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -315,6 +315,18 @@ Disable device control (mirror the device in read\-only).
 Disable video and audio playback on the computer (equivalent to \fB\-\-no\-video\-playback \-\-no\-audio\-playback\fR).
 
 .TP
+\fB\-\-new\-display\fR[=[\fIwidth\fRx\fIheight\fR][/\fIdpi\fR]]
+Create a new display with the specified resolution and density. If not provided, they default to the main display dimensions and DPI, and \fB\-\-max\-size\fR is considered.
+
+Examples:
+
+    \-\-new\-display=1920x1080
+    \-\-new\-display=1920x1080/420
+    \-\-new\-display         # main display size and density
+    \-\-new\-display -m1920  # scaled to fit a max size of 1920
+    \-\-new\-display=/240    # main display size and 240 dpi
+
+.TP
 .B \-\-no\-audio
 Disable audio forwarding.
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -811,7 +811,9 @@ static const struct sc_option options[] = {
         .longopt_id = OPT_START_APP,
         .longopt = "start-app",
         .argdesc = "name",
-        .text = "Start an Android app, by its exact package name.",
+        .text = "Start an Android app, by its exact package name.\n"
+                "Add a '+' prefix to force-stop before starting the app:\n"
+                "    scrcpy --new-display --start-app=+org.mozilla.firefox",
     },
     {
         .shortopt = 't',
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -812,8 +812,14 @@ static const struct sc_option options[] = {
         .longopt = "start-app",
         .argdesc = "name",
         .text = "Start an Android app, by its exact package name.\n"
+                "Add a '?' prefix to select an app whose name starts with the "
+                "given name, case-insensitive (retrieving app names on the "
+                "device may take some time):\n"
+                "    scrcpy --start-app=?firefox\n"
                 "Add a '+' prefix to force-stop before starting the app:\n"
-                "    scrcpy --new-display --start-app=+org.mozilla.firefox",
+                "    scrcpy --new-display --start-app=+org.mozilla.firefox\n"
+                "Both prefixes can be used, in that order:\n"
+                "    scrcpy --start-app=+?firefox",
     },
     {
         .shortopt = 't',
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -102,6 +102,7 @@ enum {
     OPT_NO_MOUSE_HOVER,
     OPT_AUDIO_DUP,
     OPT_GAMEPAD,
+    OPT_NEW_DISPLAY,
 };
 
 struct sc_option {
@@ -556,6 +557,21 @@ static const struct sc_option options[] = {
         .longopt = "no-playback",
         .text = "Disable video and audio playback on the computer (equivalent "
                 "to --no-video-playback --no-audio-playback).",
+    },
+    {
+        .longopt_id = OPT_NEW_DISPLAY,
+        .longopt = "new-display",
+        .argdesc = "[<width>x<height>][/<dpi>]",
+        .optional_arg = true,
+        .text = "Create a new display with the specified resolution and "
+                "density. If not provided, they default to the main display "
+                "dimensions and DPI, and --max-size is considered.\n"
+                "Examples:\n"
+                "    --new-display=1920x1080\n"
+                "    --new-display=1920x1080/420  # force 420 dpi\n"
+                "    --new-display         # main display size and density\n"
+                "    --new-display -m1920  # scaled to fit a max size of 1920\n"
+                "    --new-display=/240    # main display size and 240 dpi",
     },
     {
         .longopt_id = OPT_NO_AUDIO,
@@ -2668,6 +2684,9 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                     return false;
                 }
                 break;
+            case OPT_NEW_DISPLAY:
+                opts->new_display = optarg ? optarg : "";
+                break;
             default:
                 // getopt prints the error message on stderr
                 return false;
@@ -2848,6 +2867,25 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
         }
     }
 
+    if (opts->new_display) {
+        if (opts->video_source != SC_VIDEO_SOURCE_DISPLAY) {
+            LOGE("--new-display is only available with --video-source=display");
+            return false;
+        }
+
+        if (!opts->video) {
+            LOGE("--new-display is incompatible with --no-video");
+            return false;
+        }
+
+        if (opts->max_size && opts->new_display[0] != '\0'
+                           && opts->new_display[0] != '/') {
+            // An explicit size is defined (not "" nor "/<dpi>")
+            LOGE("Cannot specify both --new-display size and -m/--max-size");
+            return false;
+        }
+    }
+
     if (otg) {
         if (!opts->control) {
             LOGE("--no-control is not allowed in OTG mode");
@@ -2951,6 +2989,11 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             || opts->camera_high_speed
             || opts->camera_size) {
         LOGE("Camera options are only available with --video-source=camera");
+        return false;
+    }
+
+    if (opts->display_id != 0 && opts->new_display) {
+        LOGE("Cannot specify both --display-id and --new-display");
         return false;
     }
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -103,6 +103,7 @@ enum {
     OPT_AUDIO_DUP,
     OPT_GAMEPAD,
     OPT_NEW_DISPLAY,
+    OPT_LIST_APPS,
 };
 
 struct sc_option {
@@ -442,6 +443,11 @@ static const struct sc_option options[] = {
                 "on Ctrl+v (like MOD+Shift+v).\n"
                 "This is a workaround for some devices not behaving as "
                 "expected when setting the device clipboard programmatically.",
+    },
+    {
+        .longopt_id = OPT_LIST_APPS,
+        .longopt = "list-apps",
+        .text = "List Android apps installed on the device.",
     },
     {
         .longopt_id = OPT_LIST_CAMERAS,
@@ -2610,6 +2616,9 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                 break;
             case OPT_LIST_CAMERA_SIZES:
                 opts->list |= SC_OPTION_LIST_CAMERA_SIZES;
+                break;
+            case OPT_LIST_APPS:
+                opts->list |= SC_OPTION_LIST_APPS;
                 break;
             case OPT_REQUIRE_AUDIO:
                 opts->require_audio = true;
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -104,6 +104,7 @@ enum {
     OPT_GAMEPAD,
     OPT_NEW_DISPLAY,
     OPT_LIST_APPS,
+    OPT_START_APP,
 };
 
 struct sc_option {
@@ -805,6 +806,12 @@ static const struct sc_option options[] = {
                 "For example, to use either LCtrl or LSuper for scrcpy "
                 "shortcuts, pass \"lctrl,lsuper\".\n"
                 "Default is \"lalt,lsuper\" (left-Alt or left-Super).",
+    },
+    {
+        .longopt_id = OPT_START_APP,
+        .longopt = "start-app",
+        .argdesc = "name",
+        .text = "Start an Android app, by its exact package name.",
     },
     {
         .shortopt = 't',
@@ -2696,6 +2703,9 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             case OPT_NEW_DISPLAY:
                 opts->new_display = optarg ? optarg : "";
                 break;
+            case OPT_START_APP:
+                opts->start_app = optarg;
+                break;
             default:
                 // getopt prints the error message on stderr
                 return false;
@@ -3136,6 +3146,10 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
         }
         if (opts->power_off_on_close) {
             LOGE("Cannot request power off on close if control is disabled");
+            return false;
+        }
+        if (opts->start_app) {
+            LOGE("Cannot start an Android app if control is disabled");
             return false;
         }
     }
--- a/app/src/control_msg.c
+++ b/app/src/control_msg.c
@@ -183,6 +183,10 @@ sc_control_msg_serialize(const struct sc_control_msg *msg, uint8_t *buf) {
         case SC_CONTROL_MSG_TYPE_UHID_DESTROY:
             sc_write16be(&buf[1], msg->uhid_destroy.id);
             return 3;
+        case SC_CONTROL_MSG_TYPE_START_APP: {
+            size_t len = write_string_tiny(&buf[1], msg->start_app.name, 255);
+            return 1 + len;
+        }
         case SC_CONTROL_MSG_TYPE_EXPAND_NOTIFICATION_PANEL:
         case SC_CONTROL_MSG_TYPE_EXPAND_SETTINGS_PANEL:
         case SC_CONTROL_MSG_TYPE_COLLAPSE_PANELS:
@@ -308,6 +312,9 @@ sc_control_msg_log(const struct sc_control_msg *msg) {
         case SC_CONTROL_MSG_TYPE_OPEN_HARD_KEYBOARD_SETTINGS:
             LOG_CMSG("open hard keyboard settings");
             break;
+        case SC_CONTROL_MSG_TYPE_START_APP:
+            LOG_CMSG("start app \"%s\"", msg->start_app.name);
+            break;
         default:
             LOG_CMSG("unknown type: %u", (unsigned) msg->type);
             break;
@@ -332,6 +339,9 @@ sc_control_msg_destroy(struct sc_control_msg *msg) {
             break;
         case SC_CONTROL_MSG_TYPE_SET_CLIPBOARD:
             free(msg->set_clipboard.text);
+            break;
+        case SC_CONTROL_MSG_TYPE_START_APP:
+            free(msg->start_app.name);
             break;
         default:
             // do nothing
--- a/app/src/control_msg.h
+++ b/app/src/control_msg.h
@@ -41,6 +41,7 @@ enum sc_control_msg_type {
     SC_CONTROL_MSG_TYPE_UHID_INPUT,
     SC_CONTROL_MSG_TYPE_UHID_DESTROY,
     SC_CONTROL_MSG_TYPE_OPEN_HARD_KEYBOARD_SETTINGS,
+    SC_CONTROL_MSG_TYPE_START_APP,
 };
 
 enum sc_screen_power_mode {
@@ -110,6 +111,9 @@ struct sc_control_msg {
         struct {
             uint16_t id;
         } uhid_destroy;
+        struct {
+            char *name;
+        } start_app;
     };
 };
 
--- a/app/src/options.c
+++ b/app/src/options.c
@@ -104,6 +104,7 @@ const struct scrcpy_options scrcpy_options_default = {
     .mouse_hover = true,
     .audio_dup = false,
     .new_display = NULL,
+    .start_app = NULL,
 };
 
 enum sc_orientation
--- a/app/src/options.c
+++ b/app/src/options.c
@@ -103,6 +103,7 @@ const struct scrcpy_options scrcpy_options_default = {
     .window = true,
     .mouse_hover = true,
     .audio_dup = false,
+    .new_display = NULL,
 };
 
 enum sc_orientation
--- a/app/src/options.h
+++ b/app/src/options.h
@@ -304,6 +304,7 @@ struct scrcpy_options {
 #define SC_OPTION_LIST_DISPLAYS 0x2
 #define SC_OPTION_LIST_CAMERAS 0x4
 #define SC_OPTION_LIST_CAMERA_SIZES 0x8
+#define SC_OPTION_LIST_APPS 0x10
     uint8_t list;
     bool window;
     bool mouse_hover;
--- a/app/src/options.h
+++ b/app/src/options.h
@@ -310,6 +310,7 @@ struct scrcpy_options {
     bool mouse_hover;
     bool audio_dup;
     const char *new_display; // [<width>x<height>][/<dpi>] parsed by the server
+    const char *start_app;
 };
 
 extern const struct scrcpy_options scrcpy_options_default;
--- a/app/src/options.h
+++ b/app/src/options.h
@@ -308,6 +308,7 @@ struct scrcpy_options {
     bool window;
     bool mouse_hover;
     bool audio_dup;
+    const char *new_display; // [<width>x<height>][/<dpi>] parsed by the server
 };
 
 extern const struct scrcpy_options scrcpy_options_default;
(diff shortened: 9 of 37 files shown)


huynhtanloc2612:
Hi @rom1v, binary file cannot be downloaded

rom1v:
@huynhtanloc2612 Thanks, fixed.

huynhtanloc2612:
@rom1v 
There is a minor bug which is that the primary screen does not turn on again after closing scrcpy tool (something like when using with the option --no-cleanup)
My command: `scrcpy -SK --new-display=1920x1080/200`

vaddisrinivas:
tested this on mac(m1pro) - works!

Steps I followed - 
- clone and switch to branch
- install dependencies
- build 
- install 
- tested for -[this issue](https://github.com/Genymobile/scrcpy/issues/4598)

mengyanshou:

Test platform: Mac(M2)/Meizu 21

Test way:
```bash
git fetch origin pull/5370/head:pr-5370
git checkout pr-5370
meson setup x --buildtype=release --strip -Db_lto=true
ninja -Cx  # DO NOT RUN AS ROOT
./run x --new-display=1920x1080
```

I mainly tested the following scenarios:

- Without using the --new-display option, using scrcpy --display xxx to render the virtual display screen, the click events work normally.
- Using scrcpy --new-display=1920x1080, the created virtual display can be clicked normally.
Due to this flag

On the Meizu device I tested, the Launcher will start on this virtual display. This is not an issue, but rather an unexpected benefit.

Here is a screenshot.

截屏2024-10-14 20 49 52


By the way, I have ported scrcpy to Android.

Once I complete the remaining work, I invite you to experience this software. Thank you very much.

eiyooooo:
A small issue: IME will not appear on the virtual display unless the following setting is applied:

```
adb shell settings put global force_desktop_mode_on_external_displays 1
```

This is likely because the `ImePolicy` is not set to `DISPLAY_IME_POLICY_LOCAL`, as indicated by the [`getImePolicy()`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/services/core/java/com/android/server/wm/DisplayContent.java;l=4256) method.

---

Additionally, on some devices, the launcher displays incorrectly on non-default screens.

4nric:
> Additionally, on some devices, the launcher displays incorrectly on non-default screens.

Launcher I tested that works: Lawnchair ([playstore](https://play.google.com/store/apps/details?id=app.lawnchair.play), [github](https://github.com/LawnchairLauncher/lawnchair)). Another instance of the launcher can be opened on secondary displays without additional flags. Good enough for quickly launching apps for now

```
scrcpy --new-display
```
```
adb shell am start -n app.lawnchair.play/app.lawnchair.LawnchairLauncher --display 12
```
https://github.com/user-attachments/assets/85058b48-8086-43ab-99e9-cfd41efbfcd4

4nric:
@rom1v thank you for working on this feature!

I'm getting permission issue on android 12. @yume-chan's [proof-of-concept](https://github.com/Genymobile/scrcpy/issues/1887#issuecomment-1405184608) works on this device

```
[server] INFO: Device: [samsung] samsung SM-G975F (Android 12)
[server] ERROR: Could not create display
java.lang.SecurityException: Requires ADD_TRUSTED_DISPLAY permission to create a trusted virtual display.
        at android.os.Parcel.createExceptionOrNull(Parcel.java:2438)
        at android.os.Parcel.createException(Parcel.java:2422)
        at android.os.Parcel.readException(Parcel.java:2405)
        at android.os.Parcel.readException(Parcel.java:2347)
        at android.hardware.display.IDisplayManager$Stub$Proxy.createVirtualDisplay(IDisplayManager.java:2020)
        at android.hardware.display.DisplayManagerGlobal.createVirtualDisplay(DisplayManagerGlobal.java:694)
        at android.hardware.display.DisplayManagerGlobal.createVirtualDisplay(DisplayManagerGlobal.java:664)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1390)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1357)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1302)
        at com.genymobile.scrcpy.wrappers.DisplayManager.createNewVirtualDisplay(DisplayManager.java:138)
        at com.genymobile.scrcpy.video.NewDisplayCapture.start(NewDisplayCapture.java:95)
        at com.genymobile.scrcpy.video.SurfaceEncoder.streamCapture(SurfaceEncoder.java:89)
        at com.genymobile.scrcpy.video.SurfaceEncoder.lambda$start$0$com-genymobile-scrcpy-video-SurfaceEncoder(SurfaceEncoder.java:279)
        at com.genymobile.scrcpy.video.SurfaceEncoder$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
        at java.lang.Thread.run(Thread.java:920)
        at com.android.server.display.DisplayManagerService$BinderService.createVirtualDisplay(DisplayManagerService.java:4070)
        at android.hardware.display.IDisplayManager$Stub.onTransact(IDisplayManager.java:887)
        at android.os.Binder.execTransactInternal(Binder.java:1215)
        at android.os.Binder.execTransact(Binder.java:1179)

java.lang.AssertionError: Could not create display
        at com.genymobile.scrcpy.video.NewDisplayCapture.start(NewDisplayCapture.java:100)
        at com.genymobile.scrcpy.video.SurfaceEncoder.streamCapture(SurfaceEncoder.java:89)
        at com.genymobile.scrcpy.video.SurfaceEncoder.lambda$start$0$com-genymobile-scrcpy-video-SurfaceEncoder(SurfaceEncoder.java:279)
        at com.genymobile.scrcpy.video.SurfaceEncoder$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
        at java.lang.Thread.run(Thread.java:920)
INFO: Renderer: direct3d
WARN: Device disconnected
ERROR: Demuxer 'audio': stream disabled due to connection error
```

rom1v:
We probably need to remove this flag for Android < 13 (according to c7b04a6b19a86b11b8c857a5fb775b0e16b50ebf):

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
index 0622998ed..9440199ce 100644
--- a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
@@ -85,7 +85,6 @@ public class NewDisplayCapture extends SurfaceCapture {
                     | VIRTUAL_DISPLAY_FLAG_ROTATES_WITH_CONTENT
                     | VIRTUAL_DISPLAY_FLAG_DESTROY_CONTENT_ON_REMOVAL
                     | VIRTUAL_DISPLAY_FLAG_SHOULD_SHOW_SYSTEM_DECORATIONS
-                    | VIRTUAL_DISPLAY_FLAG_TRUSTED
                     | VIRTUAL_DISPLAY_FLAG_OWN_DISPLAY_GROUP
                     | VIRTUAL_DISPLAY_FLAG_ALWAYS_UNLOCKED
                     | VIRTUAL_DISPLAY_FLAG_TOUCH_FEEDBACK_DISABLED
```

eiyooooo:
> We probably need to remove this flag for Android < 13 (according to [c7b04a6](https://github.com/Genymobile/scrcpy/commit/c7b04a6b19a86b11b8c857a5fb775b0e16b50ebf)):
> 
> ```diff
> diff --git a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
> index 0622998ed..9440199ce 100644
> --- a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
> +++ b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
> @@ -85,7 +85,6 @@ public class NewDisplayCapture extends SurfaceCapture {
>                      | VIRTUAL_DISPLAY_FLAG_ROTATES_WITH_CONTENT
>                      | VIRTUAL_DISPLAY_FLAG_DESTROY_CONTENT_ON_REMOVAL
>                      | VIRTUAL_DISPLAY_FLAG_SHOULD_SHOW_SYSTEM_DECORATIONS
> -                    | VIRTUAL_DISPLAY_FLAG_TRUSTED
>                      | VIRTUAL_DISPLAY_FLAG_OWN_DISPLAY_GROUP
>                      | VIRTUAL_DISPLAY_FLAG_ALWAYS_UNLOCKED
>                      | VIRTUAL_DISPLAY_FLAG_TOUCH_FEEDBACK_DISABLED
> ```

According to my research, `VIRTUAL_DISPLAY_FLAG_TRUSTED`, `VIRTUAL_DISPLAY_FLAG_OWN_DISPLAY_GROUP`, `VIRTUAL_DISPLAY_FLAG_ALWAYS_UNLOCKED`, and `VIRTUAL_DISPLAY_FLAG_TOUCH_FEEDBACK_DISABLED` are included in Android 12L (Build.VERSION_CODES.S_V2).

`VIRTUAL_DISPLAY_FLAG_OWN_FOCUS` and `VIRTUAL_DISPLAY_FLAG_DEVICE_DISPLAY_GROUP` are included in Android 13 (Build.VERSION_CODES.TIRAMISU).

rom1v:
@eiyooooo The flag exists, but apparently in Android 12, shell does not have permission to create a virtual display with this flag enabled.

eiyooooo:
> @eiyooooo The flag exists, but apparently in Android 12, shell does not have permission to create a virtual display with this flag enabled.

Yes, at least Android 12L is required to use the `VIRTUAL_DISPLAY_FLAG_TRUSTED` flag.

rom1v:
> Yes, at least Android 12L is required to use the `VIRTUAL_DISPLAY_FLAG_TRUSTED` flag.

If you don't pass it, what happens?

eiyooooo:
> > Yes, at least Android 12L is required to use the `VIRTUAL_DISPLAY_FLAG_TRUSTED` flag.
> 
> If you don't pass it, what happens?

Based on my tests, the flags `VIRTUAL_DISPLAY_FLAG_SHOULD_SHOW_SYSTEM_DECORATIONS`, `VIRTUAL_DISPLAY_FLAG_OWN_DISPLAY_GROUP`, `VIRTUAL_DISPLAY_FLAG_OWN_FOCUS`, and `VIRTUAL_DISPLAY_FLAG_DEVICE_DISPLAY_GROUP` require a trusted virtual display to function properly.

eiyooooo:
```
scrcpy.exe --new-display
scrcpy 2.7 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  R5CN210NXME                     device  SM_G986U1
C:\q\software\scrcpy-win64-v2.4\scrcpy-win64-v2.7-35-g765e...file pushed, 0 skipped. 106.6 MB/s (72764 bytes in 0.001s)
[server] INFO: Device: [samsung] samsung SM-G986U1 (Android 13)
INFO: Renderer: direct3d
INFO: Texture: 1920x1080
[server] INFO: New display id: 2
```

```
adb shell am start --display 2 -n com.samsung.android.app.notes/com.samsung.android.app.notes.memolist.MemoListActivity
Starting: Intent { cmp=com.samsung.android.app.notes/.memolist.MemoListActivity }

Exception occurred while executing 'start':
java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Resources android.content.Context.getResources()' on a null object reference
        at com.android.server.wm.MultiTaskingTaskLaunchParamsModifier.calculate(MultiTaskingTaskLaunchParamsModifier.java:228)
        at com.android.server.wm.MultiTaskingTaskLaunchParamsModifier.onCalculate(MultiTaskingTaskLaunchParamsModifier.java:154)
        at com.android.server.wm.LaunchParamsController.calculate(LaunchParamsController.java:142)
        at com.android.server.wm.ActivityStarter.computeLaunchParams(ActivityStarter.java:3110)
        at com.android.server.wm.ActivityStarter.startActivityInner(ActivityStarter.java:2658)
        at com.android.server.wm.ActivityStarter.startActivityUnchecked(ActivityStarter.java:2470)
        at com.android.server.wm.ActivityStarter.executeRequest(ActivityStarter.java:1808)
        at com.android.server.wm.ActivityStarter.execute(ActivityStarter.java:1026)
        at com.android.server.wm.ActivityTaskManagerService.startActivityAsUser(ActivityTaskManagerService.java:1943)
        at com.android.server.wm.ActivityTaskManagerService.startActivityAsUser(ActivityTaskManagerService.java:1816)
        at com.android.server.am.ActivityManagerService.startActivityAsUserWithFeature(ActivityManagerService.java:3747)        at com.android.server.am.ActivityManagerShellCommand.runStartActivity(ActivityManagerShellCommand.java:656)
        at com.android.server.am.ActivityManagerShellCommand.onCommand(ActivityManagerShellCommand.java:222)
        at com.android.modules.utils.BasicShellCommandHandler.exec(BasicShellCommandHandler.java:97)
        at android.os.ShellCommand.exec(ShellCommand.java:38)
        at com.android.server.am.ActivityManagerService.onShellCommand(ActivityManagerService.java:11276)
        at android.os.Binder.shellCommand(Binder.java:1085)
        at android.os.Binder.onTransact(Binder.java:903)
        at android.app.IActivityManager$Stub.onTransact(IActivityManager.java:5643)
        at com.android.server.am.ActivityManagerService.onTransact(ActivityManagerService.java:3135)
        at android.os.Binder.execTransactInternal(Binder.java:1321)
        at android.os.Binder.execTransact(Binder.java:1280)
```

On the S20+ (Android 13), after multiple attempts to create virtual displays using `scrcpy.exe --new-display`, the display ID always remains 2, and apps fail to start in the virtual display created by scrcpy when the `VIRTUAL_DISPLAY_FLAG_OWN_FOCUS` flag is set. 😞
However, when this flag is not set, everything works fine.

rom1v:
I added features to list and start apps. I updated the main post, and also published a binary for this new version.

Please test and give feedback :wink:

rom1v:
@eiyooooo I added a commit to fix flags depending on Android versions: 5ee42eab85a3aac28d2cbcced001f011f5cd3be2

However, I have no device with Android 12 or 12L to test, so don't hesitate to suggest a patch with the correct flags.

eiyooooo:
> I added a commit to fix flags depending on Android versions: [5ee42ea](https://github.com/Genymobile/scrcpy/commit/5ee42eab85a3aac28d2cbcced001f011f5cd3be2)
> 
> However, I have no device with Android 12 or 12L to test, so don't hesitate to suggest a patch with the correct flags.

After digging deeper, I found that Android 12L still encounters the following issue:

```
java.lang.SecurityException: Requires ADD_TRUSTED_DISPLAY permission to create a trusted virtual display.
```

References:
- [Commit that give Shell to `ADD_TRUSTED_DISPLAY`](https://android.googlesource.com/platform/frameworks/base/+/990e3429636382175ca8e7bf04df054f48fbd130)
- [Try search this commit in Android 12L Release Code](https://android.googlesource.com/platform/frameworks/base/+log/refs/heads/android12L-release/?s=990e3429636382175ca8e7bf04df054f48fbd130)
- [Try search this commit in Android 13 Release Code](https://android.googlesource.com/platform/frameworks/base/+log/refs/heads/android13-release/?s=990e3429636382175ca8e7bf04df054f48fbd130)

Therefore, I think this part of the code:

https://github.com/Genymobile/scrcpy/blob/5ee42eab85a3aac28d2cbcced001f011f5cd3be2/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java#L89

should be updated to:

```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```

rom1v:
@eiyooooo OK.

> [Commit that granted Shell the ADD_TRUSTED_DISPLAY permission](https://android.googlesource.com/platform/frameworks/base/+/990e3429636382175ca8e7bf04df054f48fbd130)

Indeed, it has been merged in Android 13:

```bash
# in aosp/framework_base
git tag --contains 990e3429636382175ca8e7bf04df054f48fbd130
```

I first added readable Android version constants (already merged into `dev`): 3acffaae57238ee47e05f97f8e762a04550fdad8 (the Android way to select a version was a mess, I had to check the mapping every time).

Then I set the TRUSTED and other flags only since Android 13: e05be0ba17a1a48275eef19d92f1c729b35a0c6d

eiyooooo:
> I first added readable Android version constants (already merged into `dev`): [3acffaa](https://github.com/Genymobile/scrcpy/commit/3acffaae57238ee47e05f97f8e762a04550fdad8) (the Android way to select a version was a mess, I had to check the mapping every time).

@rom1v By the way, Android 15 has been released, so you'll need to add a new Android version constant. 😄

```java
public static final int API_35_ANDROID_15 = Build.VERSION_CODES.VANILLA_ICE_CREAM;
```

rom1v:
@eiyooooo Done: a46150f753c47d0fb180040448629d229ae74581

rom1v:
Given that loading the app names may take time (several seconds), by default, I reworked so that `--start-app` only accepts package names by default:

```
scrpcy --start-app=org.mozilla.firefox
```

For convenience, however, it is still possible to explicitly select an app by its name (but with some delay) by adding a `?` prefix:

```
scrcpy --start-app=?firefox
```

Like in the previous version, a `+` prefix force-stops the app before starting:

```bash
scrcpy --new-display --start-app=+org.mozilla.firefox   # by package name
scrcpy --new-display --start-app=+?firefox              # by name
```

yume-chan:
Seems like we need to also call `getLeanbackLaunchIntentForPackage` (https://developer.android.com/reference/android/content/pm/PackageManager#getLeanbackLaunchIntentForPackage(java.lang.String)) for Android TV apps (https://stackoverflow.com/questions/30446052/getlaunchintentforpackage-is-null-for-some-apps).

You can also call them in `listApps` to only return launchable apps.

rom1v:
> Seems like we need to also call getLeanbackLaunchIntentForPackage (https://developer.android.com/reference/android/content/pm/PackageManager#getLeanbackLaunchIntentForPackage(java.lang.String)) for Android TV apps (https://stackoverflow.com/questions/30446052/getlaunchintentforpackage-is-null-for-some-apps).

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/device/Device.java b/server/src/main/java/com/genymobile/scrcpy/device/Device.java
index f98988e8c..73f2d312c 100644
--- a/server/src/main/java/com/genymobile/scrcpy/device/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Device.java
@@ -261,7 +261,11 @@ public final class Device {
     }
 
     public static void startApp(String packageName, int displayId, boolean forceStop) {
-        Intent launchIntent = FakeContext.get().getPackageManager().getLaunchIntentForPackage(packageName);
+        PackageManager pm = FakeContext.get().getPackageManager();
+
+        // Use Leanback launch intent for TV
+        boolean hasLeanback = pm.hasSystemFeature(PackageManager.FEATURE_LEANBACK);
+        Intent launchIntent = hasLeanback ? pm.getLeanbackLaunchIntentForPackage(packageName) : pm.getLaunchIntentForPackage(packageName);
         if (launchIntent == null) {
             Ln.w("Cannot create launch intent for app " + packageName);
             return;
```

?

Or should we only use it if the "normal" launch intent failed? (I have to Android TV to test)

> You can also call them in `listApps` to only return launchable apps.

Oh, that's a great filter to only get relevant apps! I will do that.

Maybe there is something "lighter" to know if an app is launchable without actually creating the intent? (I have not found such a method).

yume-chan:
> Or should we only use it if the "normal" launch intent failed? (I have no Android TV to test)

From that StackOverflow question, it up to the app to have whether "normal" launch intent or lean launch intent, so normal apps on Android TVs should also use "normal" launch intent, and we should try both on all devices.

>Maybe there is something "lighter" to know if an app is launchable without actually creating the intent? 

Settings app (https://cs.android.com/android/platform/superproject/main/+/main:packages/apps/Settings/src/com/android/settings/applications/appinfo/AppButtonsPreferenceController.java;l=371-374;drc=2825589cb6ad9cdb4fa50367916cdad5614f0e40) (and launcher apps, I believe) also use `getLaunchIntentForPackage` to determine if an app is launchable, so it should be the only way.

rom1v:
:+1:

Then I will do something like:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/device/Device.java b/server/src/main/java/com/genymobile/scrcpy/device/Device.java
index f98988e8c..25005081b 100644
--- a/server/src/main/java/com/genymobile/scrcpy/device/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Device.java
@@ -261,7 +261,7 @@ public final class Device {
     }
 
     public static void startApp(String packageName, int displayId, boolean forceStop) {
-        Intent launchIntent = FakeContext.get().getPackageManager().getLaunchIntentForPackage(packageName);
+        Intent launchIntent = getLaunchIntent(packageName);
         if (launchIntent == null) {
             Ln.w("Cannot create launch intent for app " + packageName);
             return;
@@ -283,4 +283,14 @@ public final class Device {
         am.startActivity(launchIntent, options);
     }
 
+    public static Intent getLaunchIntent(String packageName) {
+        PackageManager pm = FakeContext.get().getPackageManager();
+
+        Intent launchIntent = pm.getLaunchIntentForPackage(packageName);
+        if (launchIntent != null) {
+            return launchIntent;
+        }
+
+        return pm.getLeanbackLaunchIntentForPackage(packageName);
+    }
 }
```

and

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/device/Device.java b/server/src/main/java/com/genymobile/scrcpy/device/Device.java
index 25005081b..a6758e616 100644
--- a/server/src/main/java/com/genymobile/scrcpy/device/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Device.java
@@ -219,7 +219,9 @@ public final class Device {
         List<DeviceApp> apps = new ArrayList<>();
         PackageManager pm = FakeContext.get().getPackageManager();
         for (ApplicationInfo appInfo : pm.getInstalledApplications(PackageManager.GET_META_DATA)) {
-            apps.add(toApp(pm, appInfo));
+            if (getLaunchIntent(appInfo.packageName) != null) {
+                apps.add(toApp(pm, appInfo));
+            }
         }
 
         return apps;
```

rom1v:
Done :heavy_check_mark: 

Btw, I don't know if the audio output is related to a specific display. For example, if I run Firefox in a new virtual display, and VLC on my physical device, I still get the VLC audio through scrcpy (I would like the audio of Firefox in scrcpy, and the audio of VLC on the physical device).

yume-chan:
> I don't know if the audio output is related to a specific display

I don't know that either.

Audio Policy API can capture audio from specific apps using [`RULE_MATCH_UID`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/media/java/android/media/audiopolicy/AudioMixingRule.java;l=81-86;drc=0da795facfad832108966687df52e27e6fb60ff5), replacing here:

https://github.com/Genymobile/scrcpy/blob/665ccb32f5306ebd866dc0d99f4d08ed2aeb91c3/server/src/main/java/com/genymobile/scrcpy/audio/AudioPlaybackCapture.java#L47-L50

https://github.com/legendsayantan/ShizuTools/blob/f0783bd76e87b28e8cb59e38ae8493f31f2a27e2/app/src/main/java/com/legendsayantan/adbtools/lib/PlayBackThread.kt#L62-L71 is an app that implements per-app volume control using Media Projection API (which uses Audio Policy API internally) (I know this app before, but I haven't tried it).

It would for the "single app mode", but if user launches another app on the virtual display later, it's complex to track and create new rules for them.

rom1v:
I reworked some commits. I pushed a new version and here is a new binary:

- [`scrcpy-win64-vd3.zip`](https://tmp.rom1v.com/scrcpy/5370/3/scrcpy-win64-vd3.zip) <sub>`SHA-256: 013a4b2a82d12d2a006f128c243130f939331cdd71e0c28886438da1a83c38b`</sub>

I removed the `Draft:` prefix, I consider that it is ready to be merged.

One question: I don't know if a new display works on Android 10, so I'm not sure 10fab000758d00bcd70767764b9a0fd13a0b256c is correct.

Reviews and tests welcome :wink:

This PR refactors a lot of things related to displays and events, so if you could use this version even without using the virtual display feature, that would be helpful to detect issues :sunglasses:

Withoutruless:
i dont know if this is intended but if you do "scrcpy.exe --new-display=1920x1080/240  --max-size=1280"
the stream resolution is 1080p instead of 720p as its ignoring the max size option.

streaming in high res increases the delay by a lot while choosing a lower virtual display res decreases the quality in games.
so the perfect solution would be to allow high VD res while streaming in a lower res.

tl;dr:
"--new-display" should respect the "--max-size" option.

rom1v:
> i dont know if this is intended but if you do "scrcpy.exe --new-display=1920x1080/240 --max-size=1280"
the stream resolution is 1080p instead of 720p as its ignoring the max size option.

The mirroring size is necessarily the same as the virtual disolay size (or we would need to create another VirtualDisplay instance which mirrors the first one, but with a different size).

> streaming in high res increases the delay by a lot while choosing a lower virtual display res decreases the quality in games.

I don't see the benefit of creating a higher res virtual display if you mirror it at a lower res. The game will render at the target resolution, which is better. You have to adjust the dpi proportionally to get the "same" result. This PR does this automatically if you don't specify a dpi value.

More details: `--max-size` is used only if no explicit size is specified, for example, if you pass `scrcpy --new-display -m1280`.
Also, if the encoder fails, it retries with a lower definition only if no explicit size is specified. Otherwise, the user would request a size, but it would open a disolay with a different size.

This is the same behavior as camera sizes.

However, it should fail with an error if both an explicit size and an max size is specified.

Withoutruless:
> I think the mirroring size is necessarily the same as the virtual disolay size (I have not tested though, and I can't test right now) (or we would need to create another VirtualDisplay instance which mirrors the first one, but with a different size).

it shouldnt be like that, --max-size should be a way to change the stream's resolution not the vd res as you can already change the vd res with --new-display.

> I don't see the benefit of creating a higher res virtual display if you mirror it at a lower res. The game will render at the target resolution, which is better. You have to adjust the dpi proportionally to get the "same" result. This PR does this automatically if you don't specify a dpi value.

there is a very big difference in games, UI looks way sharper and high-res and so does the game itself.
take nvidia dsr on PCs for example, you could render a game at 4k and output it on a 1080p screen, resulting in a way sharper image.

also on vd3 i noticed that you added the "--mouse=uhid not supported with --new-display" error, why is that?
i thought it worked fine.

rom1v:
For screen mirroring, there are two sizes: the physical size (not configurable, except with `adb shell wm size`), and the encoding size, which depends on `--max-size`.

For camera, in theory there are also two sizes: the camera capture size (with supported sizes listed by `--list-camera-sizes`), and the encoding size, which could in theory be different. Since it makes no sense to capture and encode at different sizes, `--max-size` determines the camera capture size if no explicit size is provided (it selects the best matching supported video size, and encodes at the same size).

In my opinion, a separate virtual display should behave like a camera: always be encoded at the native resolution, and `--max-size` determines the display size if no explicit size is given.

You challenge this position, and say that we should be able to create a virtual display with a given size and encode its content at a different size.

> there is a very big difference in games, UI looks way sharper and high-res and so does the game itself.

It is true that images may be rendered using (partially) higher resolution internally (e.g. multisampling in OpenGL). But IMO it still does not make sense to do this at the virtual display level. The game should know and use the target output size. It's a waste to render the whole content at a higher resolution then downscale.

I'm interested in what other people think about this.

> also on vd3 i noticed that you added the "--mouse=uhid not supported with --new-display" error, why is that?
> i thought it worked fine.

Because the mouse pointer does not appear on the virtual display (so it just does not work, at least for me).

Withoutruless:
> You challenge this position, and say that we should be able to create a virtual display with a given size and encode its content at a different size.

yes, because having more options is good.

> It is true that images may be rendered using (partially) higher resolution internally (e.g. multisampling in OpenGL). But IMO it still does not make sense to do this at the virtual display level. The game should know and use the target output size. It's a waste to render the whole content at a higher resolution then downscale.

thats the problem, some (or most in my case) games are dumb and have issues with the ui being pixelated at lower res and/or the game looking very bad, so i'd gladly waste more gpu power to help with that.

the main issue is the encoding & decoding time increase that adds so much input delay that it becomes unusable.

> Btw, if I start a new display and the encoder fails because the size is too big, then it retries with a lower resolution: in that case, I want a smaller display encoded at its native size, not a bigger display downscaled to a size supported by the encoder.

on that i agree, but that happens automatically, if you manually set the encoding res lower than the vd res then it would be on you.


> > also on vd3 i noticed that you added the "--mouse=uhid not supported with --new-display" error, why is that?
> > i thought it worked fine.

> Because the mouse pointer does not appear on the virtual display (so it just does not work, at least for me).

works fine for me, might be an OS issue.
im having problems with the keyboard though, only sdk works, aoa and uhid do nothing at all (this isnt only on the vd builds, but on scrcpy as a whole), it worked fine in the past, so im a bit confused.

rom1v:
> I think the mirroring size is necessarily the same as the virtual disolay size (I have not tested though, and I can't test right now) (or we would need to create another VirtualDisplay instance which mirrors the first one, but with a different size).

I just checked, and I can confirm: the surface passed to the new VirtualDisplay must have the same size (it cannot downscale, it just crops), so definitely the encoding size will be the same as the virtual display size with the `--new-display` feature.

rom1v:
> OK thank you, I will allow UHID mouse for virtual display then.

But UHID or AOA have no notion of displayId, I can't see how the mouse may work on a separate virtual display.

rom1v:
> But UHID or AOA have no notion of displayId, I can't see how the mouse may work on a separate virtual display.

I just retested, in fact it "works", but on the main display (which is not the expected behavior), so it does not work :confused:

rom1v:
> However, it should fail with an error if both an explicit size and an max size is specified.

It now fails if a new display size and a `-m`/`--max-size` option are both given.

@Withoutruless Can you apply this patch on top of this PR and retest with UHID and AOA mouse:

```diff
diff --git a/app/src/cli.c b/app/src/cli.c
index c4e69597c..09b2fabe2 100644
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -2910,16 +2910,6 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             LOGE("Cannot specify both --new-display size and -m/--max-size");
             return false;
         }
-
-        if (opts->mouse_input_mode == SC_MOUSE_INPUT_MODE_UHID) {
-            LOGE("--mouse=uhid not supported with --new-display");
-            return false;
-        }
-
-        if (opts->mouse_input_mode == SC_MOUSE_INPUT_MODE_AOA) {
-            LOGE("--mouse=aoa not supported with --new-display");
-            return false;
-        }
     }
 
     if (otg) {
```

Where is the mouse pointer if you run `scrcpy -M --new-display`?

Withoutruless:
> > However, it should fail with an error if both an explicit size and an max size is specified.
> 
> It now fails if a new display size and a `-m`/`--max-size` option are both given.
> 
> @Withoutruless Can you apply this patch on top of this PR and retest with UHID and AOA mouse:
> 
> ```diff
> diff --git a/app/src/cli.c b/app/src/cli.c
> index c4e69597c..09b2fabe2 100644
> --- a/app/src/cli.c
> +++ b/app/src/cli.c
> @@ -2910,16 +2910,6 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
>              LOGE("Cannot specify both --new-display size and -m/--max-size");
>              return false;
>          }
> -
> -        if (opts->mouse_input_mode == SC_MOUSE_INPUT_MODE_UHID) {
> -            LOGE("--mouse=uhid not supported with --new-display");
> -            return false;
> -        }
> -
> -        if (opts->mouse_input_mode == SC_MOUSE_INPUT_MODE_AOA) {
> -            LOGE("--mouse=aoa not supported with --new-display");
> -            return false;
> -        }
>      }
>  
>      if (otg) {
> ```
> 
> Where is the mouse pointer if you run `scrcpy -M --new-display`?

it works fine, the mouse is in the vd as intended.

sorry it took me a bit to respond though, i was testing the keyboard issue i mentioned earlier, turns out i was naming every uhid device "Microsoft X-Box 360 Pad" (#5362) and giving them xbox360 vid:pids...

rom1v:
> it works fine, the mouse is in the vd as intended.

How does it work if you run several instances on different displays in parallel?

```
scrcpy -M
scrcpy -M --new-display
scrcpy -M --new-display
```

Withoutruless:
> > it works fine, the mouse is in the vd as intended.
> 
> How does it work if you run several instances on different displays in parallel?
> 
> ```
> scrcpy -M
> scrcpy -M --new-display
> scrcpy -M --new-display
> ```

the mouse only works on the newest vd, clicking any other vd/the main screen still moves the mouse in the newest one.

closing the newest vd causes the mouse to teleport to the main display and only work there and closing a vd other than the newest one does nothing to the mouse.

> Which device is it?

Galaxy A52s 5G on A14 UN1CA Rom.

rom1v:
@Withoutruless I remove the restriction for UHID/AOA mouse.

Let's merge this. :rocket: 

I think this feature will deserve a new major version (3.0). Maybe next week or the week after, I have others changes to make beforehand.

yume-chan:
> I don't know if the audio output is related to a specific display

FYI, Actually there is per-display audio routing for companion service: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/media/java/android/media/PlayerBase.java;l=582;drc=0a293aabaef57434b7cde9f431bacf6dc3ee7fe2

If we create the VirtualDisplay using companion service: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/services/companion/java/com/android/server/companion/virtual/VirtualDeviceManagerService.java;l=494;drc=0a293aabaef57434b7cde9f431bacf6dc3ee7fe2

Companion service is a hidden API for projecting displays, inputs, audio and even cameras to companion devices (like smart watches). Normally it needs a companion device to work, but shell apps can use the testing API to create a fake one: https://cs.android.com/android/platform/superproject/main/+/main:cts/tests/tests/companion/common/src/android/companion/cts/common/AppHelper.kt;l=29;drc=831c3ced50d5574cd5645afc0277bc6c59e83278

I was particularly interested in the virtual camera injection feature, but it's hide behind a flag that's disabled by default.

rom1v:
I changed the behavior of `--max-size` for virtual displays: #5506.

Xtr126:
I wonder why the virtual display automatic resize feature was dropped [Controller.java#217](https://github.com/anirudhb/scrcpy/commit/5d4791ab4d481865d010571f24b212e2ea776b31#diff-1a24c2089218479cebb925527de82f65d07d38a6f431158d6999d51117ce6598R217) would be cool to have apps streaming with resizeable windows

mauriciofelippe:
I'm getting a issue, when the phone goes to lock screen i lose control(keyboard and left&right buttons, if a home button is pressed it is work, so just input actions related to lock screen works.) of virtual display.

rom1v:
@mauriciofelippe On Samsung phones, clicks stop working on virtual display when the device goes to sleep apparently: https://github.com/Genymobile/scrcpy/issues/5557#issuecomment-2506666017

PR #2231, opened 2021-04-02
update `brew` command

quyleanh:
`brew cask` was deprecated according to [this comment](https://github.com/Homebrew/discussions/discussions/340#discussioncomment-232364). Should be `brew install`.

PR #2004, opened 2021-01-02
Use --cask for latest versions of brew

clesiemo3:
brew cask <command> was deprecated in favor of brew <command> --cask in Homebrew 2.6.0. Now that 2.7.0 has been released, they have been disabled.

https://github.com/Homebrew/discussions/discussions/340

I am happy to update the other language readmes once appropriate format is settled on.

PR #5398, opened 2024-10-23
Update macos.md

Gutem:
update brew command

### DIFF
--- a/doc/macos.md
+++ b/doc/macos.md
@@ -13,7 +13,7 @@ brew install scrcpy
 You need `adb`, accessible from your `PATH`. If you don't have it yet:
 
 ```bash
-brew install android-platform-tools
+brew install --cask android-platform-tools
 ```
 
 Alternatively, Scrcpy is also available in [MacPorts], which sets up `adb` for you:


rom1v:
OK, so finally `--cask` is necessary? (cf this thread: https://github.com/Genymobile/scrcpy/pull/2231#issuecomment-812552226)

rom1v:
cc @quyleanh (#2231) @clesiemo3 (#2004)

Gutem:
> OK, so finally `--cask` is necessary? (cf this thread: [#2231 (comment)](https://github.com/Genymobile/scrcpy/pull/2231#issuecomment-812548167))

At least, on OSX 15.0.1 Sequoia, `--cask` was the only way to install `android-platform-tools`

quyleanh:
Hmm. I'm using `macOS 13.7 (22H123)` now and the `brew` command without `--cask` is still working. With `--cask` command works too. So I think this commit is fine.

rom1v:
Merged into `master`: f1f27116269ecbe422fd853ec192e3a80e168f76

PR #4469, opened 2023-11-27
Restart capture when screen resolution changes

yume-chan:
Fixes #1918
Fixes #161
Might affect #4152

Found this API by accident when researching https://github.com/Genymobile/scrcpy/pull/4446#issuecomment-1827164429.

Some devices have a resolution option in their Settings app, and on all devices the resolution can be changed by `adb shell wm size 1080x1920`. This event will react to both cases.

Tested on Xiaomi Mi 11 (Android 13, both system settings and `wm size`), Samsung Galaxy S9 (Android 10, both system settings and `wm size`), Onyx  BOOX Nova Pro (Android 6.0.1, only `wm size`).

Issue #4152, opened 2023-07-09
Support for automatically resizing between multiple displays


LuminarySage:
**Environment**
 - OS: Windows 11
 - scrcpy version:  v2.1
 - installation method: Windows release
 - device model: Google Pixel Fold
 - Android version: 13

**Describe the bug**
Right now if I switch between displays on something like a foldable phone, the screen capture won't scale to the correct aspect ratio at all

Transitioning from the outer display to the inner display
![scrcpy_47Mmbwv6nN](https://github.com/Genymobile/scrcpy/assets/24306400/2e3bad31-0327-40f4-aa30-f5b3bec407af)

Transitioning from the inner display to the outer display
![ezgif com-optimize](https://github.com/Genymobile/scrcpy/assets/24306400/54789884-8cc4-4980-b41a-0648186cbafd)

rom1v:
@AdoPi (as the author of #3979): could you reproduce?

AdoPi:
Sorry I can't reproduce. It works with the latest version available on master, on a Samsung Z Fold 4.
Note that I'm testing on Linux, Currently I don't have Windows.

AdoPi:
@Jnss98b Could you please add this little patch to help debugging this issue?

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/Device.java b/server/src/main/java/com/genymobile/scrcpy/Device.java
index f817a3ce..e9da6742 100644
--- a/server/src/main/java/com/genymobile/scrcpy/Device.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Device.java
@@ -103,6 +103,7 @@ public final class Device {
             ServiceManager.getWindowManager().registerDisplayFoldListener(new IDisplayFoldListener.Stub() {
                 @Override
                 public void onDisplayFoldChanged(int displayId, boolean folded) {
+                    Ln.i("==== onDisplayFoldChanged(" + displayId + ", " + folded + ")");
                     synchronized (Device.this) {
                         DisplayInfo displayInfo = ServiceManager.getDisplayManager().getDisplayInfo(displayId);
                         if (displayInfo == null) {

```

Thanks!

Jnss98b:
Will test it tomorrow when I get back. Currently out of town at the moment

Jnss98b:
@AdoPi Could you do me a favor and build a 64-bit windows release with the patch please? I'm having trouble building from source with msys2

rom1v:
@Jnss98b:
[...]

Issue #1918, opened 2020-11-19
Doesn't sync correctly after use adb shell wm size.


Sodium-Aluminate:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Ubuntu 18.04
 - scrcpy version: 1.14
 - installation method: apt
 - device model: Redmi note 8 pro (begonia) <- 1080x2336 default.
 - Android version: 10

**Describe the bug**
1. start scrcpy
2. run `adb shell wm size 1080x2336` or  `adb shell wm size 1080x1920`
3. scrcpy stdout nothing and the copied screen was not fully displayed.

terminal 1(scrcpy):
```
sodiumaluminate@AlsPad:~$ scrcpy
INFO: scrcpy 1.14 <https://github.com/Genymobile/scrcpy>
/usr/local/share/scrcpy/scrcpy-server:...shed. 5.7 MB/s (33142 bytes in 0.006s)
[server] INFO: Device: Xiaomi Redmi Note 8 Pro (Android 10)
INFO: OpenGL shaders: ENABLED
INFO: Created renderer: opengl
INFO: Renderer: opengl
INFO: OpenGL version: 4.6.0 NVIDIA 450.80.02
INFO: Trilinear filtering enabled
INFO: Initial texture: 1080x2336
^CWARN: Device disconnected
sodiumaluminate@AlsPad:~$ 
```

terminal 2(adb wm):
```
sodiumaluminate@AlsPad:~$ adb shell 
begonia:/ $ wm 
Window manager (window) commands:
  help
      Print this help text.
  size [reset|WxH|WdpxHdp] [-d DISPLAY_ID]
    Return or override display size.
    width and height in pixels unless suffixed with 'dp'.
  density [reset|DENSITY] [-d DISPLAY_ID]
    Return or override display density.
  folded-area [reset|LEFT,TOP,RIGHT,BOTTOM]
    Return or override folded area.
  overscan [reset|LEFT,TOP,RIGHT,BOTTOM] [-d DISPLAY ID]
    Set overscan area for display.
  scaling [off|auto] [-d DISPLAY_ID]
    Set display scaling mode.
  dismiss-keyguard
    Dismiss the keyguard, prompting user for auth if necessary.
  set-user-rotation [free|lock] [-d DISPLAY_ID] [rotation]
[...]

Issue #161, opened 2018-05-31
Screen problems after adjusting the resolution.


LimSH0731:
After decrease the resolution and resize window, the window will still leave a black space on the right and bottom. And show background screen beyond set screen size on home screen.

![55565](https://user-images.githubusercontent.com/37290854/40765502-23b5a2da-649c-11e8-9c28-b74bfcdf0b09.PNG)

rom1v:
Does it also happen with v1.1?

LimSH0731:
This is screenshot with 1.1 v.
FHD -> WQHD

![fhd-wqhd](https://user-images.githubusercontent.com/37290854/40766628-a4f33080-649f-11e8-974f-cfc1bb852a88.PNG)

WQHD -> FHD

![wqhd-fhd](https://user-images.githubusercontent.com/37290854/40766633-ab5e3708-649f-11e8-9f8c-fd6541aba617.PNG)

rom1v:
> After decrease the resolution and resize window

How do you decrease the resolution?

LimSH0731:
Samsung smartphones have that resolution change feature for saving battery or watching movie more clear. but, other smartphones can take this feature to 'ROOT'. So I think it is not problem only on Samsung.

rom1v:
> Samsung smartphones have that resolution change feature for saving battery or watching movie more clear.

OK, so it would require to listen to resolution changes in addition to [orientation changes](https://github.com/Genymobile/scrcpy/blob/v1.2/server/src/main/java/com/genymobile/scrcpy/Device.java#L26).

PR #5415, opened 2024-10-28
Listen to display changed events

rom1v:
Whenever the display size changes, scrcpy must reset the capture/encoding to stream a video at the new size.

The display size may change for several reasons:
 - device rotation
 - device folding/unfolding
 - resolution change
 - …

Currently, scrcpy only listened for device rotation and device folding/unfolding events.

This was sometimes insufficient:
 - #161
 - #1918
 - #4152
 - https://github.com/Genymobile/scrcpy/issues/5362#issuecomment-2416219316

Instead of listening to specific events that could cause a display size change, @yume-chan submitted #4469 to listen to display changed events directly. It was almost ready to be merged, but we noticed that it didn't work on Android 14: https://github.com/Genymobile/scrcpy/pull/4469#issuecomment-1836572786 (the events were never received)

In recent Android 14 upgrades, it seems the behavior is fixed (the display changed events are correctly received).

But we must still support the devices with a broken version of Android 14, so this PR implements the following behavior:
 - on Android != 14: only use display changed events
 - on Android 14: register rotation/fold listeners (like before) and listen to display changed events. On the first display changed event received, we know that the device has an Android 14 version where the `DisplayListener` works, so unregister rotation/fold listeners (to avoid resetting the capture/encoding twice)

The display changed events are triggered for many reasons, not only the size. This PR stores the current size and reset only if the new size is different.

@yume-chan Please review :wink:

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/device/Size.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Size.java
@@ -52,6 +52,6 @@ public final class Size {
 
     @Override
     public String toString() {
-        return "Size{" + "width=" + width + ", height=" + height + '}';
+        return "Size{" + width + 'x' + height + '}';
     }
 }
--- a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
@@ -7,12 +7,15 @@ import com.genymobile.scrcpy.device.DisplayInfo;
 import com.genymobile.scrcpy.device.Size;
 import com.genymobile.scrcpy.util.Ln;
 import com.genymobile.scrcpy.util.LogUtils;
+import com.genymobile.scrcpy.wrappers.DisplayManager;
 import com.genymobile.scrcpy.wrappers.ServiceManager;
 import com.genymobile.scrcpy.wrappers.SurfaceControl;
 
 import android.graphics.Rect;
 import android.hardware.display.VirtualDisplay;
 import android.os.Build;
+import android.os.Handler;
+import android.os.HandlerThread;
 import android.os.IBinder;
 import android.view.IDisplayFoldListener;
 import android.view.IRotationWatcher;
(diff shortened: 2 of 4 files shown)


yume-chan:
LGTM.

I wanted to manually trigger a display changed event (without changing any user-observable properties), so we can detect whether it works proactively. But I can't find a candidate (in [`DisplayInfo.equals`](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/DisplayInfo.java;l=413;drc=a4b3ffea3814d5d82b1aac9969b492d3642dbce4)). Maybe it can be done on a virtual display, but I guess that would take too much time.

Another thing I don't understand is how [f2daa46634f6a1e5e329041f07a27dbc894d71b2](https://android.googlesource.com/platform/frameworks/base.git/+/f2daa46634f6a1e5e329041f07a27dbc894d71b2%5E%21/#F0) fixed the issue. Because it means `com.android.shell` app is not considered dead by Android system?

rom1v:
> Another thing I don't understand is how [f2daa46634f6a1e5e329041f07a27dbc894d71b2](https://android.googlesource.com/platform/frameworks/base.git/+/f2daa46634f6a1e5e329041f07a27dbc894d71b2%5E%21/#F0) fixed the issue. Because it means com.android.shell app is not considered dead by Android system?

I found this commit which seemed related, but this is probably not this one that fixes the issue, since it was introduced in Android 15 but now recent versions of Android 14 work: https://github.com/Genymobile/scrcpy/pull/4469#issuecomment-2427794117

yume-chan:
It actually fixed in [5653c6b5875df599307c3e6bfae32fb2fc17ca1f](https://cs.android.com/android/_/android/platform/frameworks/base/+/5653c6b5875df599307c3e6bfae32fb2fc17ca1f)

yume-chan:
I tested it on a Redmi K70 Pro running Android 14, display changed event doesn't work.

But when changing screen resolution, the screen content is resized to fit the video, unlike Android 13 which clips or leaves blank areas in the video.

When screen aspect ratio changes, the screen content is centered with black bars on sides.

Also, touch inputs are shifted after changing screen resolution in v2.7, but works correctly after #5370, despite the coordinate in generated input events are for the old screen size.

I guess the new screen capture API already makes it much better even without display changed events.

rom1v:
> I tested it on a Redmi K70 Pro running Android 14, display changed event doesn't work.
>
> But when changing screen resolution, the screen content is resized to fit the video, unlike Android 13 which clips or leaves blank areas in the video.
>
> When screen aspect ratio changes, the screen content is centered with black bars on sides.

On `master` (2.7), on current `dev` (including #5370) or on this PR?

> but works correctly after https://github.com/Genymobile/scrcpy/pull/5370, despite the coordinate in generated input events are for the old screen size.

That's unexpected, if the coordinates and the current display size do not match, the event should be discarded with a warning: https://github.com/Genymobile/scrcpy/blob/2c25fd7a8082307da19645a690c31403903fbb1e/server/src/main/java/com/genymobile/scrcpy/control/Controller.java#L358

yume-chan:
> On master (2.7), on current dev (including https://github.com/Genymobile/scrcpy/pull/5370) or on this PR?

All give the same behavior.

> That's unexpected, if the coordinates and the current display size do not match, the event should be discarded with a warning:

Because display change event doesn't work, the server doesn't know the sizes are mismatched.

rom1v:
I fixed an issue with the "session size" (it stored the downscaled size when `-m` was passed, but compared to the source display size, instead it must always use the source display size).

And I mentioned the commit which fixed the issue in AOSP in the commit message.

Issue #4162, opened 2023-07-13
Android14 Beta4 crop not working (Pixel 6 Pro)


dhwz:
- [X] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [X] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Windows 11
 - scrcpy version: 2.1
 - installation method: Win x64 from GitHub
 - device model: Pixel 6 Pro
 - Android version: 14 Beta 4

**Describe the bug**
Tried to crop the black bar from the screen which is always visible on Pixel 6 Pro in Landscape mode. It worked perfectly on Android 13
but crop is completely ignored on Android 14 Beta 4

--crop 1440:2600:0:120 removed the black bar on Android 13 but doesn't do anything on Android 14
Note: Mouse position is still wrong if the crop is applied and also the scrcpy output says it has cropped
INFO: Texture: 1440x2480 vs. INFO: Texture: 1440x2600  if crop is 0

```
scrcpy.exe -w --video-bit-rate 25M --video-codec=h265 --audio-bit-rate=256K --audio-codec=opus --crop 1440:2600:0:120
scrcpy 2.1 <https://github.com/Genymobile/scrcpy>
C:\STORAGE\scrcpy-win64-v2.0\scrcpy-server: 1 file pushed, 0 skipped. 77.6 MB/s (56955 bytes in 0.001s)
[server] INFO: Device: [Google] google Pixel 6 Pro (Android 14)
INFO: Renderer: direct3d
INFO: Texture: 2480x1440
INFO: Texture: 1440x2480
[server] INFO: Device rotation requested: landscape
INFO: Texture: 2480x1440
```

dhwz:
@rom1v installed Beta 5 now, seems it's still broken. Do you have any idea what has changed from Android 13?

rom1v:
Nope, I don't know, they changed things related to display, which also broke `--turn-screen-off` (#3784). To be investigated (once the source code of Android 14 is released)…

SharkWipf:
FWIW: seems like the source is released now: https://android-developers.googleblog.com/2023/10/android-14-is-live-in-aosp.html

I can also confirm crop not working\* on the Pixel 8 Pro, so nothing's changed there since beta it seems.

\* The inputs do get "cropped" though, so any touches go to wrong places.

rom1v:
[...]

PR #5417, opened 2024-10-28
Disable broken options on Android 14

rom1v:
The options `--lock-video-orientation` and `--crop` are broken since Android 14. Hopefully, they will be reimplemented differently.

Meanwhile, when running Android >= 14, fail with an error to prevent incorrect behavior.

Refs #4011
Refs #4162


cutebunnygaming10:
Ok

cutebunnygaming10:
I have android 14 as well

PR #5447, opened 2024-11-03
Add screen off timeout configuration

rom1v:
The Android screen automatically turns off after some delay.

To change this delay while scrcpy is running:

```bash
scrcpy --screen-off-timeout=300  # 300 seconds (5 minutes)
```

The initial value is restored on exit.

This may be useful as an alternative to `--stay-awake` (which only works while the device is plugged in).

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -77,6 +77,7 @@ _scrcpy() {
         --rotation=
         -s --serial=
         -S --turn-screen-off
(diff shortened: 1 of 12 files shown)


rom1v:
(force pushed d3db9c40653d4c1eaafda1ece2abe57acad94906 and eff5b4b219be6043a3baf51149b1d6752569a173 immediately after the merge to add a reference to the PR in the commit messages)

Issue #5444, opened 2024-11-03
OpenGL filter prototype: no frame received


rom1v:
On-device OpenGL video filters will allow to:
 - reimplement `--lock-video-orientation` on Android >= 14 to fix #4011 
 - reimplement `--crop` on Android >= 14 to fix #4162
 - apply a custom capture rotation, useful to mirror a Meta Quest 3 for example (#4658)
 - support virtual display rotation (#5428)
 - rotate the camera capture for V4L2 (#4426)

I wrote a quick&dirty prototype (branch [`opengl_filter`](https://github.com/Genymobile/scrcpy/commits/opengl_filter/)), but I'm blocked because `onNewFrameAvailable()` is never called:

https://github.com/Genymobile/scrcpy/blob/4397dfba89241e764a84c20ec09aca181f22f978/server/src/main/java/com/genymobile/scrcpy/video/VideoFilter.java#L80

I'm pretty sure this is related to threading ([`setOnFrameAvailableListener()`](https://developer.android.com/reference/android/graphics/SurfaceTexture#setOnFrameAvailableListener(android.graphics.SurfaceTexture.OnFrameAvailableListener,%20android.os.Handler)) also accepts a `Handler`), but I did not manage to make it work.

Once this listener is called on every frame, I can implement the shaders to perform the transforms I want, but I need this to work to get started.

Any help welcome.

yume-chan:
https://github.com/Genymobile/scrcpy/compare/opengl_filter...yume-chan:scrcpy:opengl_filter_v2?expand=1

I copied some OpenGL code from https://github.com/google/grafika, and mimicked how https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/cmds/screenrecord/Overlay.cpp initialized it.

rom1v:
Thank you very much! It works perfectly.

I have everything I need to start working on it :tada: :heart:

Issue #5428, opened 2024-10-31
Virtual Display Rotation Listener Problem


mengyanshou:
## Environment

 - **OS:** [Any]
 - **Scrcpy version:** [dev ref]
 - **Installation method:** [git clone and compile]
 - **Device model:**
 - **Android version:** []

## Describe the bug

i notice this code in ScreenCapture.java
```java
    @Override
    public void init() {
        if (displayId == 0) {
            rotationWatcher = new IRotationWatcher.Stub() {
                @Override
                public void onRotationChanged(int rotation) {
                    Ln.i("Display rotated, request reset");
                    requestReset();
                }
            };
            Ln.i("registerRotationWatcher display " + displayId);
            ServiceManager.getWindowManager().registerRotationWatcher(rotationWatcher, displayId);
        }
    }
```
when create Virtual Display use this flag
```java
VIRTUAL_DISPLAY_FLAG_ROTATES_WITH_CONTENT
```
and launch some video player, the display can be rotated actually
but scrcpy screen cannot change render size when virtual display rotated 
then i comment out this part of the code

```java
    @Override
    public void init() {
            rotationWatcher = new IRotationWatcher.Stub() {
                @Override
                public void onRotationChanged(int rotation) {
                    Ln.i("Display rotated, request reset");
                    requestReset();
                }
            };
            Ln.i("registerRotationWatcher display " + displayId);
            ServiceManager.getWindowManager().registerRotationWatcher(rotationWatcher, displayId);
        }
```
every thing work fine~

i think maybe this is a little problem, can i pull a pr to resolve this?
I need to make sure it doesn't cause other problems~

rom1v:
In #5370, a new virtual display could not be rotated.

If you can make it work, sure, please open a new PR :sunglasses: 

Please base your work on #5415 (which I will merge soon).

mengyanshou:
I tested three devices: the Xiaomi 14, Meizu 21, and OPPO foldable screen. After creating a virtual display on each, I opened a video player, and all of them were able to rotate normally. The rotationWatcher also received callbacks as expected.

rom1v:
> Why can't the virtual display be rotated?

Currently, if you run `scrcpy --new-display` on `dev` in portrait mode, then start an app which requires a landscape mode, then the virtual display is not rotated.

> i notice this code in ScreenCapture.java

But `ScreenCapture` is not used when `--new-display` is passed.

What did you test exactly? And what changes did you make so that it works?

[...]

Issue #4426, opened 2023-11-15
Feature request: Allow to rotate the camera.


hellishvictor:
- OS: Win 7 x64 SP1
 - adb v34.0.4 (last version that works with Windows 7)
 - scrcpy version: 1.2.2
 - installation method: scrcpy-win64-v2.2.zip
 - device model: Samsung Galaxy 
 - Android version: 12

Hi, I just updated ScrCpy to the latest version and tested the new camera features, which are awesome, but it only works on horizontal mode, with the top of the device to the left, so it don't rotate at all. 
It would be great if were implemented an option to allow it to rotate along with the device, or just to switch it to vertical mode.

Cheers.

rom1v:
`--rotation` rotates the content on the screen (but not the recording): https://github.com/Genymobile/scrcpy/blob/master/doc/video.md#rotation

hellishvictor:
--rotation=3 did the trick -> Goodbye DroidCam!

tommy8372:
this flag doesn't have an effect for video4linux with camera source and stream always start in horizontal mode

i'm using below for linux webcam
`scrcpy --v4l2-sink=/dev/video10 --video-source=camera --camera-id=1 --no-audio --no-video-playback`

`--rotation=1` do the trick for me for video playback only, any idea how to solve this for camera?

rom1v:
See #4441.

> this flag doesn't have an effect for video4linux with camera source and stream always start in horizontal mode

https://github.com/Genymobile/scrcpy/pull/4441#issuecomment-1820658788

ManuLinares:
Is there any new development on this?

rom1v:
> Is there any new development on this?

What do you try to achieve?

Camera orientation is supported for display and recording, but not for the video4linux feature.

https://github.com/Genymobile/scrcpy/blob/master/doc/video.md#orientation

Cf [highlights of v2.3](https://github.com/Genymobile/scrcpy/releases/tag/v2.3) (but take the latest version, not v2.3).

ManuLinares:
I need to be able to rotate the camera using v4l2loopback.

**How to reproduce:**
1. Connect the phone via USB
2. `scrcpy --video-source=camera --camera-facing=back --v4l2-sink /dev/video1 --no-audio-playback --no-video-playback --camera-fps 30 --camera-size=1280x720`
3. Open a Meet, Teams, Jitsi, etc. My face is upside down.

I expect to be able to rotate the webcam sink.

[...]

PR #5455, opened 2024-11-07
On-device OpenGL video filter prototype

### LINKED ISSUE #4011: Android 14 beta lock orientation doesn't work as expected when the device goes into landscape mode.

### LINKED ISSUE #4162: Android14 Beta4 crop not working (Pixel 6 Pro)

rom1v:
## Purpose

On-device OpenGL video filters will allow to:

 - reimplement `--lock-video-orientation` on Android >= 14 to fix #4011 
 - reimplement `--crop` on Android >= 14 to fix #4162
 - apply a custom capture rotation, useful to mirror a Meta Quest 3 for example (#4658)
 - support virtual display rotation (#5428)
 - rotate the camera capture for V4L2 (#4426)

## Prototype

In this draft PR, I implemented some building blocks that will allow to fix all of these issues:

 - an OpenGL "runner" ([`OpenGLRunner`](https://github.com/Genymobile/scrcpy/blob/glfilter/server/src/main/java/com/genymobile/scrcpy/opengl/OpenGLRunner.java)), which will run a filter to be rendered to a `Surface` from an OpenGL-dedicated thread
 - a simple OpenGL filter API ([`OpenGLFilter`](https://github.com/Genymobile/scrcpy/blob/glfilter/server/src/main/java/com/genymobile/scrcpy/opengl/OpenGLFilter.java))
 - a single "generic" OpenGL filter implementation to apply any 2D [affine transformation](https://en.wikipedia.org/wiki/Affine_transformation) ([`AffineOpenGLFilter`](https://github.com/Genymobile/scrcpy/blob/glfilter/server/src/main/java/com/genymobile/scrcpy/opengl/AffineOpenGLFilter.java))
 - a tool to build affine transforms matrices from "semantic" transformations (rotate, scale…) ([`AffineTransform`](https://github.com/Genymobile/scrcpy/blob/glfilter/server/src/main/java/com/genymobile/scrcpy/opengl/AffineTransform.java))

_Special thanks to @yume-chan for unblocking me in #5444._ :wink:

For now, the filter is added directly in `SurfaceEncoder` (hardcoded). You can edit the code to apply your own transformations for testing.

In the current branch, as an example, it reframes ("crop" but without cutting the pixels outside, which can be made visible with further transforms) to the (80%,80%) top-right area, then rotates by 10 degrees (keeping the same video size):

https://github.com/Genymobile/scrcpy/blob/44a637641bfac9e5c8f326844025d61d3192ebef/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java#L99-L103

I provided another example (to uncomment) to apply a 90° rotation, which changes the target size:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java b/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
index 00e199a99..ff66becac 100644
--- a/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
@@ -80,7 +80,7 @@ public class SurfaceEncoder implements AsyncProcessor {
                 capture.consumeReset(); // If a capture reset was requested, it is implicitly fulfilled
                 capture.prepare();
                 Size inputSize = capture.getSize();
-                Size outputSize = inputSize; //.rotate();
+                Size outputSize = inputSize.rotate();
 
                 if (!headerWritten) {
                     streamer.writeVideoHeader(outputSize);
@@ -98,10 +98,7 @@ public class SurfaceEncoder implements AsyncProcessor {
 
                     // The affine transform describes how we want to transform the input image.
                     // The filter expects the matrix to transform the coordinates, which is the inverse transform.
-                    AffineTransform transform = AffineTransform.reframe(0.2f, 0.2f, 0.8f, 0.8f)
-                            .apply(AffineTransform.rotate(10).fromCenter().preserveAspectRatio(inputSize, outputSize))
-                            .invert();
-                    //AffineTransform transform = AffineTransform.orient(1).fromCenter().invert();
+                    AffineTransform transform = AffineTransform.orient(1).fromCenter().invert();
 
                     Ln.i("=== matrix=" + transform);
                     float[] matrix = transform.to4x4();
```

## TODO

There is still a lot of work to do.

Firstly, I sometimes get an error on device rotation (when the encoding session is destroyed):

```
java.lang.RuntimeException: glError 0x505 out of memory
	at com.genymobile.scrcpy.opengl.GLUtils.checkGlError(GLUtils.java:99)
	at com.genymobile.scrcpy.opengl.AffineOpenGLFilter.draw(AffineOpenGLFilter.java:126)
```

 - [ ] fix (random) bug on destroy

This is not a real "out of memory", but more probably a "use-after-free"/race condition. I don't know why yet.

Then, the clicks must be mapped to their original location (before the transform).

 - [ ] map coordinates for input events

Then, use these building blocks to fix all the related issues (with a consistent design):

 - [ ] reimplement `--lock-video-orientation` (#4011)
 - [ ] reimplement `--crop` (#4162)
 - [ ] expose a command to apply simple custom filters (at least "rotate") (#4658)
 - [ ] automatically rotate virtual displays (#5428)
 - [ ] apply `--lock-video-orientation` to camera video source (#4426)


rom1v:
I've made some progress this week-end:
 - I reimplemented `--crop` for screen capture, which now works on all devices, including Android >= 14
 - I reimplemented `--lock-video-orientation` for screen capture, which now works on all devices, including Android >= 14
 - I implemented virtual display rotation
 - All the events are correctly mapped according to the affine transformations

Please review/test! In particular, make sure that everything works with any crop, locked video orientation, physical device rotation, resize… on devices with and without Android 14. Check that:
 - the video content must matches the configuration
 - all your clicks are correctly mapped

In a virtual display (`scrcpy --new-display`), an app which automatically rotates (or manually, with <kbd>Alt</kbd>+<kbd>r</kbd>) rotates the display.

## TODO

I wanted to add support for `--crop` for camera capture and virtual display capture (because why not!?), but it fact there is a semantic issue: for these sources, `-m`/`--max-size` selects the source size, it does not resize a predefined source like for screen capture. So when we `--crop` and `-m`, it will not behave the same way (it will select another source, then crop). So I think I will not implement `--crop` for camera and virtual display (it's probably useless in practice anyway).

I will:
 - implement `--lock-video-orientation` for camera and virtual displays
 - add `--angle` to specify a custom angle (e.g. 22°, for Meta Quest 3 mirroring), to be applied after cropping, from the center of the resulting region (I suggest `--angle` to avoid confusion with `--rotation`/`--orientation` which rotate by 90° and change the mirrored video size, while `--angle` will rotate the content by a custom angle inside the area.

rom1v:
I submitted a new version where everything works (hopefully). I edited the original post and provided a binary for Windows.

Feedback welcome (please report any issue).

Helaer:
I did a test on Xiaomi's Android 14 model, and after locking the direction, the screen was still incomplete when the phone app was rotated.

Helaer:
> I did a test on Xiaomi's Android 14 model, and after locking the direction, the screen was still incomplete when the phone app was rotated.

6bd62199f0427106661507d0e41554a2
017fae50fbf0d12f8e05487fb7af24e0

Helaer:
I did a test on Xiaomi's Android 14 model, and after locking the direction, the screen was still incomplete when the phone app was rotated.
467b3420-7516-40ff-99a4-f2eed5a0f6af
a73ba98c-9666-4d57-8b94-f3ecad84b56b

rom1v:
Reproducible 100%? Could you retry after reverting https://github.com/Genymobile/scrcpy/pull/5455/commits/239553f7415134709d6f527481ec9c7c0bc35a15 please?

rom1v:
I updated the PR:
 - I fixed the lock orientation issue (https://github.com/Genymobile/scrcpy/pull/5455#issuecomment-2480346770)
 - I applied the given angle also for camera and virtual display
 - I changed the way I workaround the incorrect transform matrix given by Android `getTransformMatrix()` (I replace it by an hardcoded vflip as a transform matrix, instead of using identity and applying a vflip from the filter)
 - I fixed some matrices multiplication order

I published a new binary (`glfilter.55`) in the original post.

Helaer:
I tested post.glfilter.55 on my Xiaomi Android 14 device and it worked fine, showing the screen correctly when rotated after the app.

rom1v:
For consistency with orientation parameters, I changed `--angle` to interpret the value as clockwise (instead of counter-clockwise) (`glfilter.56`).

4nric:
The rotation works on Android 9 although it looked glitchy (window rotates before the "content" does)
https://github.com/user-attachments/assets/781d75c4-de41-4594-adff-7da83f88ce11

I have also tested S10+ running Android 14 custom rom, OneUI 6.1 port. Both main display and vd is broken:

https://github.com/user-attachments/assets/a0f74c58-3654-481e-a7cb-205705e51bed

https://github.com/user-attachments/assets/78e06a8b-de88-4977-b567-6f13a9348c23

rom1v:
@4nric Thank you for your tests!

Could you please re-test with `Do not recreate display on every rotation` reverted? (branch `glfilter.56.revert`)

I built a binary:
- [`scrcpy-win64-glfilter.56.revert.zip`](https://tmp.rom1v.com/scrcpy/5455/3/scrcpy-win64-glfilter.56.revert.zip) <sub>`SHA-256: 3c98ca92d1fca0ae96d05d16ab34f7dfb180ef08686a134600f0d9e4d7e3777`</sub>

rom1v:
> For normal screen capture, I use a fallback by listening to device rotation changes, but for vd I have no solution.

Oh, actually, we could also use a fallback, but the rotation watcher may not be registered immediately, just after a few milliseconds:

```diff
diff --git server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
index 412eb8501..1a87a35ca 100644
--- server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
+++ server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
@@ -20,6 +20,8 @@ import android.hardware.display.VirtualDisplay;
 import android.os.Build;
 import android.os.Handler;
 import android.os.HandlerThread;
+import android.os.SystemClock;
+import android.view.IRotationWatcher;
 import android.view.Surface;
 
 import java.io.IOException;
@@ -191,6 +193,15 @@ public class NewDisplayCapture extends DisplayCapture {
                 }
             }, handler);
 
+            SystemClock.sleep(100); // hack: if we register immediately, the display id is not known by the system
+            IRotationWatcher rotationWatcher = new IRotationWatcher.Stub() {
+                @Override
+                public void onRotationChanged(int rotation) {
+                    Ln.i("NewDisplayCapture: onRotationChanged(" + rotation + ")");
+                    invalidate();
+                }
+            };
+            ServiceManager.getWindowManager().registerRotationWatcher(rotationWatcher, virtualDisplayId);
         } catch (Exception e) {
             Ln.e("Could not create display", e);
             throw new AssertionError("Could not create display");
```

4nric:
@rom1v glad to test things 😊with `scrcpy-win64-glfilter.56.revert.zip`, rotation now works as expected on A14 without the "glitch". on A9 though, the window still rotates before the "content" does

rom1v:
> on A9 though, the window still rotates before the "content" does

How does it behave with scrcpy 2.7?

4nric:
Ohh it was the same behavior on 2.7. Also, i seem to get a lot of this warning messages. I just "lived" with it, it wasn't always like this. I will check if i had these warning messages on previous releases
```
[server] INFO: Device rotation requested: landscape
[server] WARN: Ignore touch event, it was generated for a different device size
[server] WARN: Ignore touch event, it was generated for a different device size
[server] WARN: Ignore touch event, it was generated for a different device size
[server] WARN: Ignore touch event, it was generated for a different device size
[server] WARN: Ignore touch event, it was generated for a different device size
[server] WARN: Ignore touch event, it was generated for a different device size
...
```

https://github.com/user-attachments/assets/cd20f239-8b3e-4d7f-8f63-6b90553f3023

rom1v:
> on A9 though, the window still rotates before the "content" does
> Ohh it was the same behavior on 2.7.

I'm not surprised, the device is rotated, but the Android system and apps may need some time to render the new rotated content.

> i seem to get a lot of this warning messages
>
> WARN: Ignore touch event, it was generated for a different device size

On rotation, it is expected that many successive events are ignored due to size mismatch, when an event was generated from the mirroring window having the old size, but was received on the device with the new size.

> it wasn't always like this.

Probably because now, mouse hover events are forwarded to the device, and they generate a lot of events. I lowered the log level of this message to verbose.

---

Here is a new version (`glfilter.61`):

 - I removed the commit `Do not recreate display on every rotation`, which definitely caused problems
 - I extracted a "display size monitor" to share code between screen capture and virtual display capture to handle display size changes, with workarounds/fallbacks handled at a single place.
 - I lowered the log level of mismatching event size to verbose
 - I added an option `--no-vd-system-decorations` (#5494) (not directly related to OpenGL filters, but it would conflict to much to write it on `dev`)

@4nric  please confirm that it fixes all your issues.

4nric:
Rotation works as expected in Android 9. On Android 14 however, portrait rotations break both in main display and virtual display if coming from landscape orientation.

1. Portrait (good) > Landscape (good) > Portrait (bad)
2. Landscape (good) > Portrait (bad)

https://github.com/user-attachments/assets/3a167ce2-b158-45ba-97dc-33dcb9295bdc

https://github.com/user-attachments/assets/e0c9a9ca-7a65-42b5-a649-452a19374164

https://github.com/user-attachments/assets/bee6c844-b937-4082-a829-f3c528165421

Helaer:
glfilter.61 On Android 14, the screen does not display correctly after the app is rotated.

rom1v:
Thank you very much for your tests and bug report, that's very helpful! :heart: 

I fixed my "display size monitor" when it used fallbacks. The problem only happened on Android 14 >= r1 to Android 14 < r29 (and I have no such device, so I simulated the problem and reproduced).

Fixed in `glfilter.63`.

Helaer:
> Thank you very much for your tests and bug report, that's very helpful! ❤️
> 
> I fixed my "display size monitor" when it used fallbacks. The problem only happened on Android 14 >= r1 to Android 14 < r29 (and I have no such device, so I simulated the problem and reproduced).
> 
> Fixed in `glfilter.63`.

Very good, glfilter. 63 worked very well on my Android 10 and Android 14 devices.

4nric:
The only issue I encountered was vd related. If the window is closed at landscape orientation, then relaunched. The window will show as portrait, but content in landscape also wrong coordinates. If manually rotated, it recovers.

image

I have made a dirty branch from `glfilter.56.revert` to test if it is viable to use IWindowManager's [`registerDisplayWindowListener`](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/view/IWindowManager.aidl#565)

It needs IDisplayWindowListener (added i think on Android 11), so I just imported aidl file to project. [`onDisplayConfigurationChanged`](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/view/IDisplayWindowListener.aidl#46) receives Configuration object unlike DisplayListener's [`onDisplayChanged`](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/hardware/display/DisplayManager.java#1697). Resolution can be extracted from bounds, as well as the display rotation. No need to listen for folding events if vd right?

Also, i removed the workaround because the listener is registered before the virtual display is made. By the time `onDisplayConfigurationChanged` gets called, there should already be the created virtual display. The new config will only be handled if display id matches.

```java
 displayWindowListener = new IDisplayWindowListener(){
                @Override
                public void onDisplayAdded(int displayId) {
                    Ln.d("NewDisplayCapture: onDisplayAdded(" + displayId + ")");
                    if (virtualDisplay.getDisplay().getDisplayId() == displayId){
                        Ln.i("New display: " + displaySize.getWidth() + "x" + displaySize.getHeight() + "/" + dpi + " (id=" + displayId + ")");
                    }
                }

                @Override
                public void onDisplayConfigurationChanged(int displayId, Configuration newConfig) {
                    if (displayId != virtualDisplay.getDisplay().getDisplayId()) {
                        //Ignore configuration changes of other displays
                        return;
                    }
                    Ln.d("NewDisplayCapture: onDisplayConfigurationChanged(" + displayId + ","+newConfig.toString()+")");

                    handleDisplayConfigurationChanged(newConfig);
                }
            };
            ServiceManager.getWindowManager().registerDisplayWindowListener(displayWindowListener);
            virtualDisplay = ServiceManager.getDisplayManager()
                    .createNewVirtualDisplay("scrcpy", displaySize.getWidth(), displaySize.getHeight(), dpi, surface, flags);
```


https://github.com/Genymobile/scrcpy/commit/d971abb3e31345f4aa9cdf76f2d08e4d52e836c4

rom1v:
> The only issue I encountered was virtual display related. If the window is closed at landscape orientation, then relaunched. The window will show as portrait, but content in landscape also wrong coordinates. If manually rotated, it recovers.

On `glfilter.63`, the fallback rotation watcher is registered 50ms after the virtual display is created, to avoid an error:

https://github.com/Genymobile/scrcpy/blob/glfilter.63/server/src/main/java/com/genymobile/scrcpy/video/DisplaySizeMonitor.java#L162-L166

It's hacky, and it didn't handle the race condition properly (it could be fixed).

But instead, using `DisplayWindowListener` for Android 14 simplifies a lot, thank you! I can completely remove the rotation/fold fallbacks (I keep this method for all Android 14 devices). I just use it to get the `onDisplayConfigurationChanged()` event (without using the `Configuration`).

Please test `glfilter.65`. Here is the relevant commit: 108abf486a292dd84734edcb1ac3e15f3b161542.

dhwz:
@rom1v device is not automatically switched on if virtual display is used, can you add an option of behave like the standard way? I need the display for fingerprint.

rom1v:
> device is not automatically switched on if virtual display is used, can you add an option or behave like the standard way? I need the display for fingerprint.

This is very device-specific.

By default, for screen mirroring (not virtual display), the screen is turned on (<kbd>POWER</kbd> is injected if the screen is off). There is an option to disable it (`--no-power-on`).

For a new virtual display, by default, I don't want my device to turn on. Adding yet a new option can be confusing with the previous one.

However, you can press <kbd>POWER</kbd> manually using <kbd>MOD</kbd>+<kbd>p</kbd>.

dhwz:
> > device is not automatically switched on if virtual display is used, can you add an option or behave like the standard way? I need the display for fingerprint.
> 
> This is very device-specific.
> 
> By default, for screen mirroring (not virtual display), the screen is turned on (POWER is injected if the screen is off). There is an option to disable it (`--no-power-on`).
> 
> For a new virtual display, by default, I don't want my device to turn on. Adding yet a new option can be confusing with the previous one.
> 
> However, you can press POWER manually using MOD+p.

Well actually it's not really device specific, I'm running some apps in the business profile on a Pixel 9 Pro (but same behavior is happening on other devices, I'm pretty sure) it requires to use fingerprint or entering PIN on the device display to unlock the apps.
Of course I can press MOD+p but thats not very userfriendly in my opinion and also a but confusing for users as they don't see the lock screen on the switched off display and so the apps clicked on the virtual display don't open and it seems the mouse click is just doing nothing.

rom1v:
> Well actually it's not really device specific, I'm running some apps in the business profile on a Pixel 9 Pro (but same behavior is happening on other devices, I'm pretty sure) it requires to use fingerprint or entering PIN on the device display to unlock the apps.

On my Pixel 8, it does not. I need the fingerprint to unlock the main display, but I can open virtual displays without unlocking.

dhwz:
> > Well actually it's not really device specific, I'm running some apps in the business profile on a Pixel 9 Pro (but same behavior is happening on other devices, I'm pretty sure) it requires to use fingerprint or entering PIN on the device display to unlock the apps.
> 
> On my Pixel 8, it does not. I need the fingerprint to unlock the main display, but I can open virtual displays without unlocking.

Of course I can open a virtial display, but locked apps won't open (like I said any app in a business profile).

4nric:
I can confirm rotation now works on all cases 😊 tested to work on Android 14 both main display and virtual display. Android 9 also. Just noticed that [glfilter.65](https://github.com/Genymobile/scrcpy/commit/108abf486a292dd84734edcb1ac3e15f3b161542) commit is missing the AIDL file. Noticed when I tried to compile https://github.com/Genymobile/scrcpy/pull/5498/commits/9e95811bd3c284ecd4e3e4aeb433bedb58eefe90

rom1v:
> Just noticed that [glfilter.65](https://github.com/Genymobile/scrcpy/commit/108abf486a292dd84734edcb1ac3e15f3b161542) commit is missing the AIDL file. Noticed when I tried to compile https://github.com/Genymobile/scrcpy/commit/9e95811bd3c284ecd4e3e4aeb433bedb58eefe90

Thank you. Fixed in `glfilter.66`.

rom1v:
> I can confirm rotation now works on all cases 😊 tested to work on Android 14 both main display and virtual display.

Did you also test camera? :smile:

4nric:
Camera rotation fails 😅

`scrcpy --video-source=camera --capture-orientation=90 `

```
C:\Users\anric\Downloads\scrcpy-win64-glfilter.65\scrcpy-server: 1 file pushed, 0 skipped. 101.9 MB/s (164641 bytes in 0.002s)
[server] INFO: Device: [samsung] samsung SM-S908B (Android 14)
[server] INFO: Using camera '0'
INFO: Renderer: direct3d
INFO: Texture: 3024x4032
[server] INFO: Retrying with -m2560...
[server] INFO: Retrying...
[server] ERROR: Encoding error: java.lang.IllegalStateException:
[server] ERROR: Exception on thread Thread[OpenGLRunner,5,main]
java.lang.RuntimeException: glError 0x505 out of memory
        at com.genymobile.scrcpy.opengl.GLUtils.checkGlError(GLUtils.java:99)
        at com.genymobile.scrcpy.opengl.AffineOpenGLFilter.draw(AffineOpenGLFilter.java:128)
        at com.genymobile.scrcpy.opengl.OpenGLRunner.render(OpenGLRunner.java:218)
        at com.genymobile.scrcpy.opengl.OpenGLRunner.lambda$run$1$com-genymobile-scrcpy-opengl-OpenGLRunner(OpenGLRunner.java:200)
        at com.genymobile.scrcpy.opengl.OpenGLRunner$$ExternalSyntheticLambda0.onFrameAvailable(D8$$SyntheticClass:0)
        at android.graphics.SurfaceTexture$1.handleMessage(Unknown Source:4)
        at android.os.Handler.dispatchMessage(Unknown Source:19)
        at android.os.Looper.loopOnce(Unknown Source:185)
        at android.os.Looper.loop(Unknown Source:83)
        at android.os.HandlerThread.run(Unknown Source:28)
java.lang.AssertionError
        at com.genymobile.scrcpy.video.CameraCapture.start(CameraCapture.java:256)
        at com.genymobile.scrcpy.video.SurfaceEncoder.streamCapture(SurfaceEncoder.java:95)
        at com.genymobile.scrcpy.video.SurfaceEncoder.lambda$start$0$com-genymobile-scrcpy-video-SurfaceEncoder(SurfaceEncoder.java:293)
        at com.genymobile.scrcpy.video.SurfaceEncoder$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
WARN: Device disconnected
```

rom1v:
Is it better with:

```diff
diff --git server/src/main/java/com/genymobile/scrcpy/video/CameraCapture.java server/src/main/java/com/genymobile/scrcpy/video/CameraCapture.java
index ebccd0356..0e147cb71 100644
--- server/src/main/java/com/genymobile/scrcpy/video/CameraCapture.java
+++ server/src/main/java/com/genymobile/scrcpy/video/CameraCapture.java
@@ -266,10 +266,19 @@ public class CameraCapture extends SurfaceCapture {
             CaptureRequest request = createCaptureRequest(surface);
             setRepeatingRequest(session, request);
         } catch (CameraAccessException | InterruptedException e) {
+            stop();
             throw new IOException(e);
         }
     }
 
+    @Override
+    public void stop() {
+        if (glRunner != null) {
+            glRunner.stopAndRelease();
+            glRunner = null;
+        }
+    }
+
     @Override
     public void release() {
         if (cameraDevice != null) 
```

4nric:
Yep, that fixed the camera rotation issue😊

rom1v:
> Yep, that fixed the camera rotation issue😊

Thank you :+1:

Please test combinations of `--crop=w:h:x:y`, `--angle=XX` and `-m1024`, with pointer locations enabled in Developer Options. I'm sure you will find more bugs with new corner cases :smile:

4nric:
I found no issues 😊Crop, angle, m1024, horizontal flip, for main display, virtual display, and camera. The coordinates map fine as well for main and virtual displays

rom1v:
> Do you have a measure of the performance impact of the OpenGL runner, memory/GPU-wise and especially on mirroring latency? Is it negligible compared to v2.7?

I did not measure. In practice, I haven't noticed an additional latency. I'm pretty sure the internal implementation of `--lock-video-orientation` and `--crop` with `SurfaceControl` in previous versions Android were also using OpenGL.

(And if no transform is needed, OpenGL is not used, the capture is directly encoded as before.)

dudwns9426:
Thank you, I am using scrcpy 3.0 with metaquest3,
I got the left eye of metaquest3 with this command,
scrcpy --crop=1280:1104:588:552 --angle=22 --video-bit-rate=2M --max-fps=15

PR #5505, opened 2024-11-20
Support refresh rates over 60 FPS in NewDisplayCapture

eiyooooo:
When `maxFps` exceeds 60, use the `setRequestedRefreshRate` method to configure the refresh rate for the virtual display.  
Otherwise, the refresh rate will default to a maximum of 60 FPS.  

[Related Android commit](https://android.googlesource.com/platform/frameworks/base/+/6c57176e9a2882eff03c5b3f3cccfd988d38488d)  
[Code limiting the frame rate to 60 FPS](https://android.googlesource.com/platform/frameworks/base/+/6c57176e9a2882eff03c5b3f3cccfd988d38488d/services/core/java/com/android/server/display/VirtualDisplayAdapter.java#562)

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
@@ -14,8 +14,10 @@ import com.genymobile.scrcpy.util.AffineMatrix;
 import com.genymobile.scrcpy.util.Ln;
 import com.genymobile.scrcpy.wrappers.ServiceManager;
 
+import android.annotation.SuppressLint;
 import android.graphics.Rect;
 import android.hardware.display.VirtualDisplay;
+import android.hardware.display.VirtualDisplayConfig;
 import android.os.Build;
 import android.view.Surface;
 
@@ -59,6 +61,7 @@ public class NewDisplayCapture extends SurfaceCapture {
     private Size videoSize;
     private Size displaySize; // the logical size of the display (including rotation)
     private Size physicalSize; // the physical size of the display (without rotation)
+    private final float maxFps;
 
     private int dpi;
 
@@ -67,6 +70,7 @@ public class NewDisplayCapture extends SurfaceCapture {
         this.newDisplay = options.getNewDisplay();
         assert newDisplay != null;
         this.maxSize = options.getMaxSize();
+        this.maxFps = options.getMaxFps();
         this.crop = options.getCrop();
         assert options.getCaptureOrientationLock() != null;
         this.captureOrientationLocked = options.getCaptureOrientationLock() != Orientation.Lock.Unlocked;
@@ -152,6 +156,7 @@ public class NewDisplayCapture extends SurfaceCapture {
         displayTransform = AffineMatrix.multiplyAll(displayRotationMatrix, eventTransform);
     }
 
+    @SuppressLint("WrongConstant")
     public void startNew(Surface surface) {
         int virtualDisplayId;
         try {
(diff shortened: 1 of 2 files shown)


eiyooooo:
I think we could make the refresh rate of the virtual display an adjustable option for users.

rom1v:
Thank you :+1:

However, in practice, I never get more than 60fps:

```
scrcpy --new-display --max-fps=120 -m720
```

Whereas I get 120fps if I mirror the main display.

---

> I think we could make the refresh rate of the virtual display an adjustable option for users.

`--max-fps` is intended to be a "limiter", passing `--max-fps` should never increase the frame rate compared to not passing the option.

Instead, we could add a new `--vd-fps` option, but better, maybe we could specify it directly to `--new-display`:

```
scrcpy --new-display=1920x1080@120
```

eiyooooo:
> However, in practice, I never get more than 60fps:
> 
> ```
> scrcpy --new-display --max-fps=120 -m720
> ```

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
index cadba8f1..f8d9fa05 100644
--- a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
@@ -183,7 +183,7 @@ public class NewDisplayCapture extends SurfaceCapture {
             // If not do so, the refresh rate of the virtual display will be limited to 60.
             // https://android.googlesource.com/platform/frameworks/base/+/6c57176e9a2882eff03c5b3f3cccfd988d38488d
             // https://android.googlesource.com/platform/frameworks/base/+/6c57176e9a2882eff03c5b3f3cccfd988d38488d/services/core/java/com/android/server/display/VirtualDisplayAdapter.java#562
-            if (maxFps > 60 && Build.VERSION.SDK_INT >= AndroidVersions.API_34_ANDROID_14) {
+            if (Build.VERSION.SDK_INT >= AndroidVersions.API_34_ANDROID_14) {
```

Since my monitor doesn’t support 120Hz, I removed the condition and tested with `scrcpy --new-display --max-fps=30 -m720` and `scrcpy --new-display --max-fps=60 -m720`. The difference in refresh rates is clearly noticeable, confirming that the `setRequestedRefreshRate` method is indeed necessary.

---

> Maybe we could specify it directly to `--new-display`:
> 
> ```
> scrcpy --new-display=1920x1080@120
> ```

I agree with this approach. 👍

rom1v:
> The difference in refresh rates is clearly noticeable, confirming that the setRequestedRefreshRate method is indeed necessary.

I'm not sure to understand, even without setting `setRequestedRefreshRate()`, you will see a difference between `--max-fps=30` and `--max-fps=60` (it's a limit applied to the encoder). So `setRequestedRefreshRate()` would be useful to go beyond 60fps.

rom1v:
> > ```
> > scrcpy --new-display=1920x1080@120
> > ```
>
> I agree with this approach. 👍

I implemented that on branch `display_capture_fps`.

However, I will not merge it, because the virtual display rate feature in Android is very unstable/experimental (tested on Pixel 8):
 - setting a value over 60 fps just does not work (I never get more than 60 fps, contrary to the main display)
 - on low fps (not always the same value, but until 12 fps), there are either no frame at all, or a black screen
 - on any other fps value, the target is very unstable: if I request 35 fps, I get 15~22, whereas with `scrcpy --new-display=1920x1080 --max-fps=35` gives a "perfect" 35 fps.

So a scrcpy feature to configure the virtual display rate would be basically broken, it would just "not work" as expected.

I keep the branch for later if the situation improves.

rom1v:
I'm closing for now.

PR #5506, opened 2024-11-20
Dissociate virtual display size and capture size

rom1v:
Allow capturing virtual displays at a lower resolution using `-m`/`--max-size`.

In the original implementation in #5370, the virtual display size was necessarily the same as the capture size. The `--max-size` value was only allowed to determine the virtual display size when no explicit size was provided.

Since the dpi was scaled down accordingly, it is often better to create a virtual display at the target capture size directly. However, not everything is rendered according to the virtual display DPI. For example, a page in Firefox is rendered too big on small virtual displays. Thus, it makes sense to be able create a virtual display at a given size, and capture it at a lower resolution with `--max-size`. This is now possible using OpenGL filters.

Therefore, change the behavior of `--max-size` for virtual displays:
 - it does not impact `--new-display` without size argument anymore (the virtual display size is the main display size);
 - it is used to limit the capture size (whether an explicit size is provided or not).

This new behavior is consistent with main display capture.

@Withoutruless This is the behavior you requested here: https://github.com/Genymobile/scrcpy/pull/5370#issuecomment-2439690605

- [`scrcpy-win64-pr5506.zip`](https://tmp.rom1v.com/scrcpy/5506/1/scrcpy-win64-pr5506.zip) <sub>`SHA-256: 7d4b7451fd56820b64bd33060137b7a44391979318a5c82db67c33ad7354943`</sub>

### DIFF
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -318,14 +318,13 @@ Disable video and audio playback on the computer (equivalent to \fB\-\-no\-video
 
 .TP
 \fB\-\-new\-display\fR[=[\fIwidth\fRx\fIheight\fR][/\fIdpi\fR]]
-Create a new display with the specified resolution and density. If not provided, they default to the main display dimensions and DPI, and \fB\-\-max\-size\fR is considered.
+Create a new display with the specified resolution and density. If not provided, they default to the main display dimensions and DPI.
 
 Examples:
 
     \-\-new\-display=1920x1080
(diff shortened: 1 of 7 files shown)


Withoutruless:
i tried it out and it works just fine, thank you rom1v.

are you going to merge this?

rom1v:
> i tried it out and it works just fine

:+1: Thank you for the test!

> are you going to merge this?

Yes.

PR #5515, opened 2024-11-22
Add release static builds for Linux (and macOS?)

rom1v:
This is the continuation of my work on #5306.

scrcpy-github-actions

This PR make it possible to build a static release for Linux (x86_64).

For that purpose:
 - `release.mk` is replaced by individual release scripts
 - the dependencies script are made more flexible to select native/cross and static/shared
 - the GitHub Actions script is improved (cf screenshot)
 - a Linux build script is added

Please test the binaries generated by GitHub Actions: https://github.com/Genymobile/scrcpy/pull/5515#issuecomment-2495589621

<details><summary>old</summary>
Here is a Linux binary (x86_64) built by GitHub Actions:

- [`scrcpy-linux-ga.tar.gz`](https://tmp.rom1v.com/scrcpy/5515/1/scrcpy-linux-ga.tar.gz) <sub>`SHA-256: f29f4e915f02c77ab3be4b00aaf218a359248463cc2e423c8ee692621041b62`</sub>
</details>

---

Now, I want to do the same for macOS, but I have no mac. So I started a skeleton (the last commit of this PR), I need your help to complete it and make it work. You can read the commit which adds the static build for Linux as a base.

Thank you!

PR #5517, opened 2024-11-23
macos: enable static binaries

muvaf:
Here is the build log, ran on macOS arm64 15.1.1 (24B91).
```bash
release/build_macos.sh 2&>1 > build_log.txt
```
[build_log.txt](https://github.com/user-attachments/files/17880211/build_log.txt)

I saw some warnings about `objc` which might be useful:
```
WARNING: Static library 'objc' not found for dependency 'sdl2', may not be statically linked
```

When I run it against a locally running emulator, I had to copy `scrcpy-server` into the folder the binary exists.
```bash
curl -Lo release/work/build-macos/dist/scrcpy-server https://github.com/Genymobile/scrcpy/releases/download/v2.7/scrcpy-server-v2.7
```
```bash
$ ./scrcpy_bin
scrcpy 2.7 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     --> (tcpip)  emulator-5554                   device  sdk_gphone64_arm64
ERROR: Could not get local file path, using scrcpy-server from current directory
scrcpy-server: 1 file pushed, 0 skipped. 46.9 MB/s (71200 bytes in 0.001s)
[server] INFO: Device: [Google] google sdk_gphone64_arm64 (Android 15)
ERROR: Could not get icon path
WARN: Could not load icon
INFO: Renderer: metal
INFO: Texture: 1080x2400
2024-11-23 18:17:09.632 scrcpy_bin[4677:907645] +[IMKClient subclass]: chose IMKClient_Modern
2024-11-23 18:17:09.632 scrcpy_bin[4677:907645] +[IMKInputSession subclass]: chose IMKInputSession_Modern
WARN: Killing the server...
```

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -173,7 +173,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-            echo TODO
+            brew install meson ninja automake
 
       - name: Build macOS
         run: release/build_macos.sh
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -165,6 +165,33 @@ jobs:
           name: build-win64-intermediate
           path: release/work/build-win64/dist-tar/
 
+  build-macos:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+            echo TODO
+
+      - name: Build macOS
+        run: release/build_macos.sh
+
+      # upload-artifact does not preserve permissions
+      - name: Tar
+        run: |
+            cd release/work/build-macos
+            mkdir dist-tar
+            cd dist-tar
+            tar -C .. -cvf dist.tar.gz dist/
+
+      - name: Upload build-macos artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: build-macos-intermediate
+          path: release/work/build-macos/dist-tar/
+
   package-linux:
     needs:
       - build-scrcpy-server
@@ -272,6 +299,42 @@ jobs:
         with:
           name: release-win64
           path: release/output
+
+  package-macos:
+    needs:
+      - build-scrcpy-server
+      - build-macos
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Download scrcpy-server
+        uses: actions/download-artifact@v4
+        with:
+          name: scrcpy-server
+          path: release/work/build-server/server/
+
+      - name: Download build-macos
+        uses: actions/download-artifact@v4
+        with:
+          name: build-macos-intermediate
+          path: release/work/build-macos/dist-tar/
+
+      # upload-artifact does not preserve permissions
+      - name: Detar
+        run: |
+            cd release/work/build-macos
+            tar xf dist-tar/dist.tar.gz
+
+      - name: Package macos
+        run: release/package_client.sh macos
+
+      - name: Upload macos release
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-macos
+          path: release/output/
 
   release:
     needs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,6 +67,36 @@ jobs:
       - name: Test
         run: release/test_client.sh
 
+  build-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y meson ninja-build nasm ffmpeg libsdl2-2.0-0 \
+             libsdl2-dev libavcodec-dev libavdevice-dev libavformat-dev \
+             libavutil-dev libswresample-dev libusb-1.0-0 libusb-1.0-0-dev
+
+      - name: Build linux
+        run: release/build_linux.sh
+
+      # upload-artifact does not preserve permissions
+      - name: Tar
+        run: |
+            cd release/work/build-linux
+            mkdir dist-tar
+            cd dist-tar
+            tar -C .. -cvf dist.tar.gz dist/
+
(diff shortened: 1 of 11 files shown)


muvaf:
@rom1v When I install those, I see the following in the brew logs:
```
==> Caveats
zlib is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

For compilers to find zlib you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/zlib/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/zlib/include"

For pkg-config to find zlib you may need to set:
  export PKG_CONFIG_PATH="/opt/homebrew/opt/zlib/lib/pkgconfig"
```
```
==> Caveats
libiconv is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have libiconv first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/libiconv/bin:$PATH"' >> ~/.zshrc

For compilers to find libiconv you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/libiconv/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/libiconv/include"
```

So, these are probably expected to be present on all macOS systems, hence not needed to be statically linked. Still, I tried adding the following to before the `meson` command:
```bash
LDFLAGS="$LDFLAGS -L/opt/homebrew/opt/zlib/lib"
CPPFLAGS="$CPPFLAGS -I/opt/homebrew/opt/zlib/include"

LDFLAGS="$LDFLAGS-L/opt/homebrew/opt/libiconv/lib"
CPPFLAGS="$CPPFLAGS -I/opt/homebrew/opt/libiconv/include"
```

Still got the same warnings.
[build_log.txt](https://github.com/user-attachments/files/17880374/build_log.txt)

Adding the following didn't make any difference either:
```bash
PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/opt/homebrew/opt/zlib/lib/pkgconfig"
```

rom1v:
> Still, I tried adding the following to before the meson command:

It should probably be added before the `configure` call for FFmpeg in `app/deps/ffmpeg.sh`.

muvaf:
@rom1v No change. Here is the latest script:
```bash
#!/bin/bash
set -ex
cd "$(dirname ${BASH_SOURCE[0]})"
. build_common
cd .. # root project dir

MACOS_BUILD_DIR="$WORK_DIR/build-macos"
LDFLAGS="$LDFLAGS -L/opt/homebrew/opt/zlib/lib"
CPPFLAGS="$CPPFLAGS -I/opt/homebrew/opt/zlib/include"

LDFLAGS="$LDFLAGS-L/opt/homebrew/opt/libiconv/lib"
CPPFLAGS="$CPPFLAGS -I/opt/homebrew/opt/libiconv/include"

app/deps/adb_macos.sh
app/deps/sdl.sh macos native static
app/deps/ffmpeg.sh macos native static
app/deps/libusb.sh macos native static

DEPS_INSTALL_DIR="$PWD/app/deps/work/install/macos-native-static"
ADB_INSTALL_DIR="$PWD/app/deps/work/install/adb-macos"

rm -rf "$MACOS_BUILD_DIR"
meson setup "$MACOS_BUILD_DIR" \
    --pkg-config-path="$DEPS_INSTALL_DIR/lib/pkgconfig" \
    -Dc_args="-I$DEPS_INSTALL_DIR/include" \
    -Dc_link_args="-L$DEPS_INSTALL_DIR/lib" \
    --buildtype=release \
    --strip \
    -Db_lto=true \
    -Dcompile_server=false \
    -Dportable=true \
    -Dstatic=true
ninja -C "$MACOS_BUILD_DIR"

# Group intermediate outputs into a 'dist' directory
mkdir -p "$MACOS_BUILD_DIR/dist"
cp "$MACOS_BUILD_DIR"/app/scrcpy "$MACOS_BUILD_DIR/dist/scrcpy_bin"
cp app/data/icon.png "$MACOS_BUILD_DIR/dist/"
cp app/data/scrcpy_static_wrapper.sh "$MACOS_BUILD_DIR/dist/scrcpy"
cp -r "$ADB_INSTALL_DIR"/. "$MACOS_BUILD_DIR/dist/"
```

This is the logs: 
[macos_build.3.txt](https://github.com/user-attachments/files/17881810/macos_build.3.txt)

rom1v:
This is not [`app/deps/ffmeg.sh`](https://github.com/Genymobile/scrcpy/blob/macos_build.2/app/deps/ffmpeg.sh#L34) :wink:

muvaf:
OK so I added the following right before the `configure` call of both ffmpeg and sdl:
```bash
    export LDFLAGS="$LDFLAGS -L/opt/homebrew/opt/zlib/lib"
    export CPPFLAGS="$CPPFLAGS -I/opt/homebrew/opt/zlib/include"

    export LDFLAGS="$LDFLAGS-L/opt/homebrew/opt/libiconv/lib"
    export CPPFLAGS="$CPPFLAGS -I/opt/homebrew/opt/libiconv/include"
```

In the logs, I see that the ones warning about `iconv` are gone, so we're down to 4 from 6. 
[macos_build.4.txt](https://github.com/user-attachments/files/17881835/macos_build.4.txt)

@rom1v FWIW, I think you may be able to get faster feedback if you manually trigger the workflow in GH actions after making changes. You may not be able to test it but seeing logs should help with these warnings.

rom1v:
> In the logs, I see that the ones warning about iconv are gone, so we're down to 4 from 6.

With this line in addition:

```
export PKG_CONFIG_PATH="/opt/homebrew/opt/zlib/lib/pkgconfig"
```

the logs about zlib are also gone :+1:

> FWIW, I think you may be able to get faster feedback if you manually trigger the workflow in GH actions after making changes. You may not be able to test it but seeing logs should help with these warnings.

Yes, I'm doing that a lot :+1:

muvaf:
Incorporated into https://github.com/Genymobile/scrcpy/pull/5515

PR #5560, opened 2024-11-28
Implement get_executable_path() for macOS

rom1v:
Implement the function to get the executable path on macOS, so that enabling `-Dportable=true` works on all platforms to find `scrcpy-server` and `icon.png` in the same directory as the executable.

@Genxster1998 I stole this from your branch. Thank you :wink:

Please review/test on macOS.

### DIFF
--- a/app/data/scrcpy_static_wrapper.sh
+++ b/app/data/scrcpy_static_wrapper.sh
@@ -1,6 +1,0 @@
-#!/bin/bash
-cd "$(dirname ${BASH_SOURCE[0]})"
-export ADB="${ADB:-./adb}"
-export SCRCPY_SERVER_PATH="${SCRCPY_SERVER_PATH:-./scrcpy-server}"
-export SCRCPY_ICON_PATH="${SCRCPY_ICON_PATH:-./icon.png}"
-./scrcpy_bin "$@"
--- a/app/meson.build
+++ b/app/meson.build
@@ -46,6 +46,7 @@ src = [
     'src/util/acksync.c',
     'src/util/audiobuf.c',
     'src/util/average.c',
+    'src/util/env.c',
     'src/util/file.c',
     'src/util/intmap.c',
     'src/util/intr.c',
--- a/app/src/adb/adb.c
+++ b/app/src/adb/adb.c
@@ -7,6 +7,7 @@
 
 #include "adb_device.h"
 #include "adb_parser.h"
+#include "util/env.h"
 #include "util/file.h"
 #include "util/log.h"
 #include "util/process_intr.h"
@@ -24,15 +25,31 @@
  */
 #define SC_ADB_COMMAND(...) { sc_adb_get_executable(), __VA_ARGS__, NULL }
 
-static const char *adb_executable;
+static char *adb_executable;
+
+bool
+sc_adb_init(void) {
+    adb_executable = sc_get_env("ADB");
+    if (adb_executable) {
+        return true;
+    }
+
+    adb_executable = strdup("adb");
+    if (!adb_executable) {
+        LOG_OOM();
+        return false;
+    }
+
+    return true;
+}
+
+void
+sc_adb_destroy(void) {
+    free(adb_executable);
+}
 
 const char *
 sc_adb_get_executable(void) {
-    if (!adb_executable) {
-        adb_executable = getenv("ADB");
-        if (!adb_executable)
-            adb_executable = "adb";
-    }
     return adb_executable;
 }
 
--- a/app/src/adb/adb.c
+++ b/app/src/adb/adb.c
@@ -31,6 +31,7 @@ bool
 sc_adb_init(void) {
     adb_executable = sc_get_env("ADB");
     if (adb_executable) {
+        LOGD("Using adb: %s", adb_executable);
         return true;
     }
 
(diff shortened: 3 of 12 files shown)


Genxster1998:
@rom1v I tested the build from `exec_path` branch, it works as expected with both wrapper and directly executing `scrcpy_bin` (with default adb installation) in macOS 14.5 x86_64.

rom1v:
One more thing, the `scrcpy` script does not work if called from a symlink.

For example:

```
$ ln -s scrcpy-linux-x86_64-v3.0/scrcpy aaa
$ ./aaa
./aaa: line 7: ./scrcpy_bin: No such file or directory
```

On Linux, `readlink -f` solves the problem:

```diff
diff --git app/data/scrcpy_static_wrapper.sh app/data/scrcpy_static_wrapper.sh
index b31b4f1f4..9e765ca87 100755
--- app/data/scrcpy_static_wrapper.sh
+++ app/data/scrcpy_static_wrapper.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
-cd "$(dirname ${BASH_SOURCE[0]})"
+cd "$(dirname $(readlink -f ${BASH_SOURCE[0]}))"
 export ADB="${ADB:-./adb}"
 ./scrcpy_bin "$@"
```

also in the case where it's a relative symlink pointing to a relative symlink (which does not work without `-f`).

However, on macOS, I think `readlink` behaves differently (`-f` is not available). There is `realpath`, but it is not installed by default.

Genxster1998:
https://stackoverflow.com/questions/7665/how-to-resolve-symbolic-links-in-a-shell-script

Yeah ,a wrapper posix shell script won't be stable as binary itself, it would have problem resolving links,args,params etc.

rom1v:
I added commits so that scrcpy uses the local adb in portable builds, and I removed the wrapper script completely.

Please test/review :wink:

rom1v:
> Edit: readlink without -f option works in macOS.

Without `-f`, it does not if there are two levels of relative symlinks (on Linux at least).

For example:

```bash
ln -s scrcpy-linux-x86_64-v3.0/scrcpy aaa
mkdir bbb
cd bbb
ln -s ../aaa
./aaa
```

metayan:
Trying to build 39c39213 on macOS 10.15.7 fails with
```
[63/130] Compiling C object app/scrcpy.p/src_sys_unix_file.c.o
FAILED: app/scrcpy.p/src_sys_unix_file.c.o 
cc -Iapp/scrcpy.p -Iapp -I../app -I../app/src -I/usr/local/include -I/usr/local/include/SDL2\
 -I/usr/local/Cellar/libusb/1.0.27/include/libusb-1.0 -fdiagnostics-color=always\
 -Wall -Winvalid-pch -Wextra -std=c11 -O0 -g -Wmissing-prototypes -D_THREAD_SAFE -MD -MQ\
 app/scrcpy.p/src_sys_unix_file.c.o -MF app/scrcpy.p/src_sys_unix_file.c.o.d\
 -o app/scrcpy.p/src_sys_unix_file.c.o -c ../app/src/sys/unix/file.c
../app/src/sys/unix/file.c:66:9: error: call to undeclared function '_NSGetExecutablePath';\
 ISO C99 and later do not support implicit function declarations\
 [-Wimplicit-function-declaration]
   66 |     if (_NSGetExecutablePath(buf, &bufsize) != 0) {
      |         ^
1 error generated.
```

rom1v:
@metayan Does this help:

```diff
diff --git app/src/sys/unix/file.c app/src/sys/unix/file.c
index 60c806bb2..6123c788c 100644
--- app/src/sys/unix/file.c
+++ app/src/sys/unix/file.c
@@ -6,6 +6,9 @@
 #include <string.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#ifdef __APPLE__
+# include <mach-o/dyld.h> // for _NSGetExecutablePath()
+#endif
 
 #include "util/log.h"
 
```

Genxster1998:
> > Edit: readlink without -f option works in macOS.
> 
> Without `-f`, it does not if there are two levels of relative symlinks (on Linux at least).
> 
> For example:
> 
> ```shell
> ln -s scrcpy-linux-x86_64-v3.0/scrcpy aaa
> mkdir bbb
> cd bbb
> ln -s ../aaa ccc
> ./ccc
> ```

Yeah , I tried that, It won't resolve all symbolic links. I linked the stackflow query discussing the same.

metayan:
@rom1v
The include solved it. Builds and works fine.

By the way, the release scripts are also excellent, taking care of the dependencies.
Very useful.

Noticed this, though, unrelated to this PR, but appears when building on macOS:
```
ld: warning: could not create compact unwind for _kbd_window_init:
 stack subq instruction is too different from dwarf stack size
```

rom1v:
Merged into `dev`.

I will probably publish a release 3.0.1 soon with the current `dev` branch, which fixes several issues from 3.0.

Genxster1998:
> I added commits so that scrcpy uses the local adb in portable builds, and I removed the wrapper script completely.
You mean it checks for system wide  `adb` env var or not before using local adb?

rom1v:
By default, it uses the `adb` the local directory.

This can be changzd by setting the ADB envvar.

For example:

```bash
ADB=adb scrcpy
```

It will use `adb` (like in non-portable builds), which will be found in the PATH.

Genxster1998:
@metayan I would strongly suggest you to move to later macOS releases preferably latest 2 releases.Catalina ceased to get security updates in oct 2022 susceptible to easy root exploits like dirtycow patched in dec 2022.

PR #5572, opened 2024-11-30
List and launch apps faster, add option to launch new instance, show app suggestions

4nric:
Show exact matches + other app labels with the keyword
```
scrcpy --start-app="?Camera"    
```
```
[server] ERROR: No unique launchable app named "Camera" found from app drawer
Found 2 exact matches:
 - Camera                         com.samsung.android.scan3d
 - Camera                         com.sec.android.app.camera
Found 2 other potential matches:
 - IP Camera                      com.shenyaocn.android.WebCam
 - Open Camera                    net.sourceforge.opencamera
```

Look for the app from `pm.getInstalledApplications()` (may take time) if not found in drawer. Also if string provided are with spaces, "?" may be dropped. If without spaces, "?" is still required to differentiate with package names. Package names can never have spaces.
```
scrcpy --start-app="?Device Care"
```
```
scrcpy --start-app="Device Care" 
```
```
[server] ERROR: No unique launchable app named "Device Care" found from app drawer
[server] WARN: Trying to find from list of all apps
[server] INFO: Starting app "Device care" [com.samsung.android.lool] on display 0...
```

Recognize system apps but ignore if they don't have a launch intent.
```
scrcpy --start-app=?Shell
...
[server] ERROR: No unique launchable app named "Shell" found from app drawer
[server] WARN: Trying to find from list of all apps
[server] WARN: Ignoring Shell [com.android.shell] which has no launch intent
[server] ERROR: No unique launchable app named "Shell" found from list of all apps
Found 1 other potential match:
 - Remote ADB Shell               com.cgutman.androidremotedebugger
```
```
scrcpy --start-app=com.android.shell
...
[server] ERROR: No launchable app with package name "com.android.shell" found from app drawer
[server] WARN: Trying to find from list of all apps
[server] ERROR: No launch intent for Shell [com.android.shell]
```

Try launch a new instance. App must support like Chrome, Samsung Gallery & Calculator, etc
```
scrcpy --start-app=+com.sec.android.app.popupcalculator --new-display
```
```
scrcpy --start-app=+?Gallery --new-display
```
image

Force stopping an app changed to `-` to accommodate `+` for new instance feature.
```
scrcpy --start-app=-com.android.chrome
```

`--list-apps` list drawer apps and does so almost twice as fast than previous. Also launching apps in main display or new display are faster, as long as the apps are found from app drawer. 

Edit: document changes instead


rom1v:
Thank you for the PR.

I didn't read the details, but could you explain a bit why this new method would be preferable (benefits/drawbacks)?

Here are some differences I observed between current scrcpy and the current version of this PR.

System apps are not listed separately.

Some items do not appear with this PR. For example:

```
 * Google Play Store              com.android.vending
```

Some items have a different name. With current scrcpy:

```
- Breezy Weather                 org.breezyweather
- VLC                            org.videolan.vlc
```

With this PR:

```
  VLC reporter                   org.videolan.vlc
  Weather                        org.breezyweather
```

The packages having the same label appear only once. With current scrcpy:

```
 * Calculator                     com.google.android.calculator
 - Calculator                     com.darkempire78.opencalculator
 - Gamepad Tester                 com.dylan.gamepad.tester.pro
 - Gamepad Tester                 ru.elron.gamepadtester
```

With this PR:

```
 Calculator                     com.darkempire78.opencalculator
 Gamepad Tester                 ru.elron.gamepadtester
```

4nric:
Hello😊 This will instead list the apps available to launchers alphabetically.

https://github.com/Genymobile/scrcpy/blob/6f323a2e3e084231fd594e51950e97ae13b55b77/server/src/main/java/com/genymobile/scrcpy/device/Device.java#L244-L257

The missing labels if there are duplicates are now fixed in this commit: https://github.com/Genymobile/scrcpy/pull/5572/commits/f8c32cff8d2b5111e3dc8fb50facdd0179e7d937

In my device, the label for `com.android.vending` is `Playstore`. In my installed launcher, I removed all folders and sorted alphabetically. The list produced in this PR is exactly the same as in my launcher.

The benefit with this approach is that time can be shaved off because every time trying to launch an app or just build list, we no longer need to go through every installed application (which some system apps are not launchable), and filter it ourselves:
https://github.com/Genymobile/scrcpy/blob/45382e3f017ea17925243b5cdcdeb3b1a5a44a37/server/src/main/java/com/genymobile/scrcpy/device/Device.java#L258-L268


This mainly is for user-convenience (for me at least). I am planning on adding an option to specifically launch a specific activity, some activities of apps are exported. Even system ones that can't normally be launched from main default launcher. For this, I propose another list type, maybe "expert?" --list-apps=expert (unimplemented). This will instead go through all installed apps, check every activity if exported or not, then add to list. This one may take a long time to build the list.

I have been using [Activity Launcher](https://github.com/butzist/ActivityLauncher) myself to directly launch to specific activities.

`--start-app` can then be modified easily to launch if for example, given this format:
`--start-app='com.android.settings/com.android.settings.Settings$DevelopmentSettingsDashboardActivity'` (must be surrounded with single quote). Here's a proof of concept: https://github.com/Genymobile/scrcpy/commit/574c425390a63ca3ddf1d00a183ab4d370d70abd

4nric:
Hello @rom1v, I have made possible to show suggestions if there's no exact match for labels/package names, exact names and alike. The list refers to the same apps that are listed in launcher app drawers. For example, if there are multiple apps installed with "Camera" in its label, it will show this:

```
[server] ERROR: No unique app found named "Camera"
Found 2 exact matches:
 - Camera                         com.samsung.android.scan3d
 - Camera                         com.sec.android.app.camera
Found 2 potential matches:
 - IP Camera                      com.shenyaocn.android.WebCam
 - Open Camera                    net.sourceforge.opencamera
```

Same goes for packages names:
```
[server] ERROR: No app found with package name "com.android"
Found 3 potential matches:
 - Chrome                         com.android.chrome
 - Play Store                     com.android.vending
 - Settings                       com.android.settings
```

Package names can never contain spaces. This PR now also allows treating string passed to `--start-app` as app name or label instead as a package name if it contains a space.

For example,  scrcpy --start-app="Play Store" is allowed, treated as label instead as package name though "?" was not indicated.

Previously, even if full app label is not entered, app with the closest label and is unique will still launch. I don't know if that is by design, but this PR requires full unique label/package name (no ambiguity).


In the original version, several queries need to be made before the app gets launched.

Mainly, this method gets called [`for all installed packages`](https://github.com/Genymobile/scrcpy/blob/45382e3f017ea17925243b5cdcdeb3b1a5a44a37/server/src/main/java/com/genymobile/scrcpy/device/Device.java#L233)
https://github.com/Genymobile/scrcpy/blob/45382e3f017ea17925243b5cdcdeb3b1a5a44a37/server/src/main/java/com/genymobile/scrcpy/device/Device.java#L242-L249

And when finally launching the app, because only package name is provided to Device.startApp,
launch intent needs to be queried again, which i think is redundant.
https://github.com/Genymobile/scrcpy/blob/45382e3f017ea17925243b5cdcdeb3b1a5a44a37/server/src/main/java/com/genymobile/scrcpy/device/Device.java#L290

Approach in this PR instead gets the already filtered intents. The system does it for us. And since this only limit to the list of apps that are visible in default launcher, there'd be no need to get the launch intent for all packages in the system.

4nric:
I have noticed that some apps are missing from each other, even have different label as mentioned (Play Store vs Google Play Store). Current scrcpy drops also some items that are normally listed in app drawer. Drive, Facebook, and Maps for example. 

Screen recording above includes the approximate duration at the bottom. Current scrcpy takes almost twice as long, if not more, compared to this PR. The difference can also be felt when launching apps in main display and virtual display.

4nric:
Hello @rom1v ignore this mess 😅 If resolving apps available to "launchers," label may differ `ResolveInfo.activityInfo` Usually have a branch not pushed yet that also take into account that + the label from `ApplicationInfo` (current implentation)

server/src/main/java/com/genymobile/scrcpy/util/LogUtils.java
```diff
@@ -191,54 +194,43 @@ private static SortedSet<Integer> getUniqueSet(Range<Integer>[] ranges) {
...
-        //  3. package name
-        // Comparator.comparing() was introduced in API 24, so it cannot be used here to simplify the code
-        Collections.sort(apps, (thisApp, otherApp) -> {
-            // System apps first
-            int cmp = -Boolean.compare(thisApp.isSystem(), otherApp.isSystem());
-            if (cmp != 0) {
-                return cmp;
-            }
+        for (ResolveInfo app : drawerApps) {
+            appMap.put(app.activityInfo.packageName, app.loadLabel(context.getPackageManager()).toString());
```
krishtoautomate, line 206: can you add version of the app as well

PR #5574, opened 2024-11-30
Add ability to generate .ico file from app icon for shortcut files

4nric:
Currently, .ico files are saved at /data/local/tmp/ico

These .ico files can be used to replace the icon of a shortcut (.lnk), like scrcpy.exe launched with args "--start-app=com.android.settings --new-display --window-title=Settings"

Opening the .lnk shortcut will show another entry in taskbar with the replaced icon


krishtoautomate:
good option but need controller mapping as well to finish this task

PR #5614, opened 2024-12-05
Inject main display events to the original display

### LINKED ISSUE #5545: Navigation gestures no longer work using a mouse in fullscreen apps
## Environment

- OS: Windows 10
- Scrcpy version: 3.0
- Installation method: Windows release
- Device model: S22 Ultra
- Android version: Android 14

## Describe the bug

Can no longer use navigation gestures (back/home app switch) in fullscreen apps with immersion mode through mouse swiping since upgrading to Scrcpy 3.0 (it's still fine in 2.7).

### LINKED ISSUE #5605: `scrcpy` v >= 3.0.0: UI elements (e.g. brightness slider) not reacting to mouse when downsizing video (--max-size/-m)
## Environment

 - **OS:** [e.g. Debian, Windows, macOS...]
 - **Scrcpy version:** 3.0.0, 3.0.1, 3.0.2
 - **Installation method:** Windows x64 release zip from Github releases page
 - **Device model:** Samsung Galaxy S23 Ultra (SM-S918B, EUX CSC)
 - **Android version:** 14. One UI 6.1

## Describe the bug

 - When I try to use/run scrcpy to control my primary phone (S23 Ultra as stated above), although I experience the exact same issue with some of my other Samsung devices (S22 Ultra, Galaxy Tab S9+...):
 - Connection method: via tcpip + wifi, using bundled `adb` binary
 - I've been using scrcpy for *at least* 1.5 years now, for all of my android devices, never experiencing this problem that I recall.
 - It specifically appeared since 3.0.0 release
 - Every time I downsize the rendered video stream (with --max-size/-m), to anything lower than my SM-S918B native resolution (1440x3088) (e.g. --max-size=2000), and after ensuring that the SDL window is automatically resized to match the video surface (i.e. ModifierKey + G), it does what's expected (in my case, height drops from 3088 to 2000), but ** certain ui elements (seems to be happening with SystemUI-related elements but not sure if it's limited to just that), do not react to mouse clicks/button down/ups (best example is the brightness slider shown inside the quicksettings/notification panel pulled down from top border) **

- I have reproduced this multiples times, with different combinations of switches, encoders, video-renderers, and with my other devices. The same cmd. line, with the same device, but with `scrcpy` older than 3.0.0, does not reproduce this bug.

Let me know if you need a snapshot/other info. for further troubleshooting.

Thanks!

rom1v:
When mirroring a secondary display, touch and scroll events must be sent to the mirroring virtual display id (with coordinates relative to the virtual display size), rather than to the original display (with coordinates relative to the original display size). (see #4598 #5137)

This behavior, introduced by d19396718ee0c0ba7fb578f595a6553c0458da59 (#5370), was also applied for the main display for consistency. However, it has been found to cause some UI elements to become unclickable.
    
To minimize inconveniences, restore the previous behavior when mirroring the main display: send all events to the original display id (0) with coordinates relative to the original display size.

Fixes #5545
Fixes #5605

Here is a binary to replace in scrcpy 3.0.2:

- [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/5614/1/scrcpy-server) <sub>`SHA-256: 350f96085b68f1bf5fd33cbf521ecec0c1b697048728a55d5ca402ab804a48b`</sub>

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
@@ -124,23 +124,9 @@ public class ScreenCapture extends SurfaceCapture {
             inputSize = videoSize;
         }
 
-        int virtualDisplayId;
-        PositionMapper positionMapper;
         try {
             virtualDisplay = ServiceManager.getDisplayManager()
                     .createVirtualDisplay("scrcpy", inputSize.getWidth(), inputSize.getHeight(), displayId, surface);
-
-
-            if (displayId == 0) {
-                // Main display: send all events to the original display, relative to the device size
-                Size deviceSize = displayInfo.getSize();
-                positionMapper = PositionMapper.create(videoSize, transform, deviceSize);
-                virtualDisplayId = 0;
-            } else {
-                // The positions are relative to the virtual display, not the original display (so use inputSize, not deviceSize!)
-                positionMapper = PositionMapper.create(videoSize, transform, inputSize);
-                virtualDisplayId = virtualDisplay.getDisplay().getDisplayId();
-            }
             Ln.d("Display: using DisplayManager API");
         } catch (Exception displayManagerException) {
             try {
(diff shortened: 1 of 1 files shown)


rodrigofd:
Cannot approve as I lack permissions, but manually tested it on behalf my #5605 and it fixed the issue completely 👍🏻

PR #5615, opened 2024-12-05
Add --no-vd-destroy-content

rom1v:
Add an option to disable the following flag for virtual displays:

    DisplayManager.VIRTUAL_DISPLAY_FLAG_DESTROY_CONTENT_ON_REMOVAL
   
With this option, when the virtual display is closd, the running apps are moved to the main display rather than being destroyed.


liuxing10:
aa

amitbra2021:
scrcpy --new-display --no-vd-destroy-content

PR #5647, opened 2024-12-10
Improve manual rotation reliability

rom1v:
Manually requesting a device orientation change using MOD+r often fails.

Introducing a small delay dramatically improves reliability.


rom1v:
I cannot reproduce reliably. Currently, without that change, it always works… :confused:

PR #5658, opened 2024-12-12
Set icon and server env paths for meson devenv

ColinKinloch:
This allows users to compile and run the project without root.

```sh
meson setup x
meson compile -C x
meson devenv -C x
scrcpy
```

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -279,3 +279,9 @@ if get_option('buildtype') == 'debug'
         test(t[0], exe)
     endforeach
 endif
+
+if meson.version().version_compare('>= 0.58.0')
+       devenv = environment()
+       devenv.set('SCRCPY_ICON_PATH', join_paths(meson.current_source_dir(), 'data/icon.png'))
+       meson.add_devenv(devenv)
+endif
--- a/server/meson.build
+++ b/server/meson.build
@@ -23,3 +23,9 @@ else
                   install: true,
                   install_dir: 'share/scrcpy')
 endif
+
+if meson.version().version_compare('>= 0.58.0')
+       devenv = environment()
+       devenv.set('SCRCPY_SERVER_PATH', join_paths(meson.current_build_dir(), 'scrcpy-server'))
+       meson.add_devenv(devenv)
+endif


rom1v:
Great, I didn't know about `meson devenv`!

I also discovered that since 0.49, you can use the `/` operator for paths instead of `join_paths()`:

```diff
diff --git app/meson.build app/meson.build
index c74ee1021..abef527f5 100644
--- app/meson.build
+++ app/meson.build
@@ -282,6 +282,6 @@ endif
 
 if meson.version().version_compare('>= 0.58.0')
        devenv = environment()
-       devenv.set('SCRCPY_ICON_PATH', join_paths(meson.current_source_dir(), 'data/icon.png'))
+       devenv.set('SCRCPY_ICON_PATH', meson.current_source_dir() / 'data/icon.png'))
        meson.add_devenv(devenv)
 endif
diff --git server/meson.build server/meson.build
index 2e5fa9d2a..8a53656cf 100644
--- server/meson.build
+++ server/meson.build
@@ -26,6 +26,6 @@ endif
 
 if meson.version().version_compare('>= 0.58.0')
        devenv = environment()
-       devenv.set('SCRCPY_SERVER_PATH', join_paths(meson.current_build_dir(), 'scrcpy-server'))
+       devenv.set('SCRCPY_SERVER_PATH', meson.current_build_dir() / 'scrcpy-server'))
        meson.add_devenv(devenv)
 endif
```

- https://mesonbuild.com/Reference-manual_functions.html#join_paths
- https://mesonbuild.com/Syntax.html#string-path-building

rom1v:
Oh, I thought it was ok without bumping because it is in a block checking version >= 0.5.8.

But in fact, it will probably still fail due to a syntax error on 0.48.

So it's probably better to keep it like it is.

rom1v:
> I'm not sure what the 0.48 minimum is based on.

It was bumped by eb8f7a1f288b43e333db12e0f329e7706da5a164.

ColinKinloch:
The commit is from 2020 when Stretch went out of support.
Buster has `0.56.2` and went out of support in 2022.
Maybe it's time 🤷

ColinKinloch:
It looks fine and runs fine. However I didn't run an `meson install`.
There are a couple of unrelated commits on it though.

rom1v:
Thank you. Merged into `dev` :rocket:

Issue #4027, opened 2023-05-23
Faulty symlink when using WinGet


headquarter8302:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Windows 10
 - scrcpy version: 2.0
 - installation method: winget
 - device model: Lenovo P1M
 - Android version: 5.1

**Describe the bug**
After installing through winget, a symlink is generated at `%LOCALAPPDATA%\Microsoft\WinGet\Links` as a mean of getting the executable into Path. However, when executing the symlink through the terminal (PowerShell 7.3.4 Core on Win10), it doesn't execute anything, just outputting nothing and returning to prompt. Even though manually going through the directory using File Explorer and opening it by double-clicking will launch it as normal.

```ps
 ASUS  PS  Monday 09:20:08 PM
┌ C:\Users\ASUS
└ ❯ scrcpy

 ASUS  PS  Monday 09:20:12 PM
┌ C:\Users\ASUS
└ ❯
```

And no, the custom PS profile isn't interfering with anything. Heck, running with `-noprofile` will return an error instead, as if there was no symlink at all.

```ps
PowerShell 7.3.4
PS C:\Users\ASUS> scrcpy
scrcpy: The term 'scrcpy' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
PS C:\Users\ASUS>
```

However, executing the command via cmd.exe results in a stream of error prompts instead, reading: "The code execution cannot proceed because [multiple DLLs] was not found. Reinstalling the program may fix this problem."

rom1v:
Please report to WinGet.

headquarter8302:
WinGet repo confirmed that this is an upstream issue with their app not being able to handle portable executables with DLLs

rom1v:
> their app not being able to handle portable executables with DLLs

Then they should probably build a non-portable version and use the dependencies (ffmpeg, sdl, libusb) from their repo: https://github.com/Genymobile/scrcpy/blob/958f22490b3ab86677bd8b6126ddaa4de4082798/meson_options.txt#L4

BurtHarris:
Yup, just had the same problem.   You might consider removing the mention of winget from 
/windows.md

[...]

PR #5686, opened 2024-12-19
Revert c27ab46, re-adding WinGet as an installation option

headquarter8302:
see #4027, fixed in https://github.com/microsoft/winget-pkgs/pull/196442

### DIFF
--- a/doc/windows.md
+++ b/doc/windows.md
@@ -20,6 +20,11 @@ and extract it.
 
 ### From a package manager
 
+From [WinGet], ADB and other dependencies will be installed alongside scrcpy:
+```bash
+winget install --exact Genymobile.scrcpy
+```
+
 From [Chocolatey]:
 
 ```bash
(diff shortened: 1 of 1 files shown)


rom1v:
:+1:

In the previous version, it was documented as `winget install scrcpy`.

Now, it is documented as `winget install --exact Genymobile.scrcpy`. Could you explain what this changes?

yume-chan:
WinGet packages have a unique id (`Genymobile.scrcpy`) so apps with the same name can co-exist. They can also have aliases/monikers (`scrcpy`) so apps with unique names can be installed using a shorter name.

The `--exact`/`-e` argument specifies the unique id. In official documentations, and because it can be copy-pasted, I think it should use the `--exact` version.

rom1v:
Merged into `master`: 5ae01749bf4fad9fb4d8bf7c879dc60f479c1013 :rocket:

PR #5697, opened 2024-12-22
Support older macOS versions in CI build

### LINKED ISSUE #5649: is there a version for macos 10.15?
3.x seems only support macos 13, and 2.x only has windows version.

yume-chan:
Fixes #5649

These older OS versions are no longer supported by Apple, but the latest macOS SDK still supports building binaries for 10.13, and I see no downsides doing that.

I tested x86-64 build in a virtual machine running 10.15.7. I can't test the arm64 build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -206,6 +206,13 @@ jobs:
                 libtool
 
       - name: Build
+        env:
+          # the default Xcode (and macOS SDK) version can be found at
+          # <https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#xcode>
+          #
+          # then the minimal supported deployment target of that macOS SDK can be found at
+          # <https://developer.apple.com/support/xcode/#minimum-requirements>
+          MACOSX_DEPLOYMENT_TARGET: 10.13
         run: release/build_macos.sh aarch64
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -206,6 +206,13 @@ jobs:
                 libtool
 
       - name: Build
+        env:
+          # the default Xcode (and macOS SDK) version can be found at
+          # <https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#xcode>
+          #
+          # then the minimal supported deployment target of that macOS SDK can be found at
+          # <https://developer.apple.com/support/xcode/#minimum-requirements>
```
rom1v, line 214: OK for the change, but why these comments related to xcode?

yume-chan: Because macOS SDK is part of Xcode. 

In the first link, you can see GitHub installed multiple versions of Xcode in the runner image, and each macOS SDK version is contained in one or more Xcode versions.

In fact, neither Xcode nor macOS SDK version matter now, because all recent macOS SDK versions support building for OS X 10.13. I just want to document where this 10.13 comes from.

rom1v:
Merged: 5b1229a55f8e89facaeb0d3757e37c49d62e88fb

PR #5703, opened 2024-12-25
Sets the policy for how the display should show IME.

Mr-JingShi:
Hi，bro. 
 Scrcpy is very good, I like it.

 I want to display IME in virtualdisplay。

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -94,6 +94,7 @@ enum {
     OPT_CAMERA_FPS,
     OPT_CAMERA_HIGH_SPEED,
     OPT_DISPLAY_ORIENTATION,
+    OPT_DISPLAY_IME_POLICY,
     OPT_RECORD_ORIENTATION,
     OPT_ORIENTATION,
     OPT_KEYBOARD,
@@ -365,6 +366,20 @@ static const struct sc_option options[] = {
                 "The available display ids can be listed by:\n"
                 "    scrcpy --list-displays\n"
                 "Default is 0.",
+    },
+    {
+        .longopt_id = OPT_DISPLAY_IME_POLICY,
+        .longopt = "display-ime-policy",
+        .argdesc = "value",
+        .text = "Sets the policy for how the display should show IME.\n"
+                "Possible values are \"local\", \"fallback_display\" and "
+                "\"hide\".\n"
+                "The \"local\" policy means that the IME should appear on the "
+                "local display.\n"
+                "The \"fallback_display\" policy means that the IME should "
+                "appear on a fallback display. The fallback display is always "
+                "DEFAULT DISPLAY.\n"
+                "The \"hide\" policy means that the IME should be hidden.\n",
     },
     {
         .longopt_id = OPT_DISPLAY_ORIENTATION,
@@ -1616,6 +1631,24 @@ parse_audio_output_buffer(const char *s, sc_tick *tick) {
 }
 
 static bool
+parse_display_ime_policy(const char *optarg, enum sc_display_ime_policy *policy) {
+    if (!strcmp(optarg, "local")) {
+        *policy = SC_DISPLAY_IME_POLICY_LOCAL;
+        return true;
+    }
+    if (!strcmp(optarg, "fallback_display")) {
+        *policy = SC_DISPLAY_IME_POLICY_FALLBACK_DISPLAY;
+        return true;
+    }
+    if (!strcmp(optarg, "hide")) {
+        *policy = SC_DISPLAY_IME_POLICY_HIDE;
+        return true;
+    }
+    LOGE("Unsupported display IME policy: %s (expected local, fallback_display, hide)", optarg);
+    return false;
+}
+
+static bool
 parse_orientation(const char *s, enum sc_orientation *orientation) {
     if (!strcmp(s, "0")) {
         *orientation = SC_ORIENTATION_0;
@@ -2723,6 +2756,11 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             case OPT_NO_VD_SYSTEM_DECORATIONS:
                 opts->vd_system_decorations = false;
                 break;
+            case OPT_DISPLAY_IME_POLICY:
+                if (!parse_display_ime_policy(optarg, &opts->display_ime_policy)) {
+                    return false;
+                }
+                break;
             default:
                 // getopt prints the error message on stderr
                 return false;
@@ -2978,6 +3016,11 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
             return false;
         }
 
+        if (opts->display_ime_policy != SC_DISPLAY_IME_POLICY_UNDEFINDED) {
+            LOGE("--display-ime_policy is only available with --video-source=display");
+            return false;
+        }
+
         if (opts->camera_id && opts->camera_facing != SC_CAMERA_FACING_ANY) {
             LOGE("Cannot specify both --camera-id and --camera-facing");
             return false;
@@ -3014,9 +3057,17 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
         return false;
     }
 
-    if (opts->display_id != 0 && opts->new_display) {
-        LOGE("Cannot specify both --display-id and --new-display");
-        return false;
+    if (opts->display_id != 0) {
+        if (opts->new_display) {
+            LOGE("Cannot specify both --display-id and --new-display");
+            return false;
+        }
+    } else {
+        if (!opts->new_display
+             && opts->display_ime_policy != SC_DISPLAY_IME_POLICY_UNDEFINDED) {
+            LOGE("--display-ime-policy not supported if display_id == 0 and new_display == NULL");
+            return false;
+        }
     }
 
     if (opts->audio && opts->audio_source == SC_AUDIO_SOURCE_AUTO) {
@@ -3198,6 +3249,10 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
         }
         if (opts->display_id) {
             LOGE("OTG mode: could not select display");
+            return false;
+        }
+        if (opts->display_ime_policy != SC_DISPLAY_IME_POLICY_UNDEFINDED) {
+            LOGE("OTG mode: could not set the policy for how the display should show IME.");
             return false;
         }
         if (v4l2) {
--- a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
@@ -4,6 +4,7 @@ import com.genymobile.scrcpy.device.Device;
 import com.genymobile.scrcpy.util.Ln;
 import com.genymobile.scrcpy.util.Settings;
 import com.genymobile.scrcpy.util.SettingsException;
+import com.genymobile.scrcpy.wrappers.ServiceManager;
 
 import android.os.BatteryManager;
 import android.system.ErrnoException;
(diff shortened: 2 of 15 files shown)


server/src/main/java/com/genymobile/scrcpy/CleanUp.java
```diff
@@ -100,15 +101,27 @@ private void runCleanUp(Options options) {
         boolean powerOffScreen = options.getPowerOffScreenOnClose();
         int displayId = options.getDisplayId();
 
+        int restoreDisplayImePolicy = -1;
+        if (displayId > 0) {
```
rom1v, line 105: This is only for secondary displays? (not the main display?)

server/src/main/java/com/genymobile/scrcpy/wrappers/WindowManager.java
```diff
@@ -198,4 +210,64 @@ public void unregisterDisplayWindowListener(IDisplayWindowListener listener) {
...
+            return DISPLAY_IME_POLICY_FALLBACK_DISPLAY;
+        }
+    }
+
+    @TargetApi(AndroidVersions.API_29_ANDROID_10)
+    private Method getSetDisplayImePolicyMethod() throws NoSuchMethodException {
+        if (setDisplayImePolicyMethod == null) {
+            try {
+                setDisplayImePolicyMethod = manager.getClass().getMethod("setDisplayImePolicy", int.class, int.class);
+            } catch (NoSuchMethodException e) {
```
rom1v, line 248: The change between `shouldShowIme()` and `getDisplayImePolicy()` was performed by [this commit](https://android.googlesource.com/platform/frameworks/base/+/acd7a80462a17111d04d2a5eeab154f33f0bfa54%5E%21/) introduced in Android 12, so it can be checked directly with `Build.VERSION.SDK_INT < AndroidVersions.API_31_ANDROID_12`.

rom1v (commented):
Thank you for your PR. After a quick glance, it looks very good and clean (with all details implemented) :+1:

As a French speaker, I have never needed to use an IME, so I'm not very familiar with IME policies. I'd like to get opinions from IME users :slightly_smiling_face: 

There are 3 options (local, fallback, hide), is it absolutely necessary to expose them? (Maybe yes, I'm just asking.)
Or isn't there an "obvious" default option to set without asking the user?

For example, are there cases where you need to use an IME on your main display if you're mirroring a virtual display?

Wouldn't it be sufficient to call unconditionnally in `NewDisplayCapture`:

```
ServiceManager.getWindowManager().setDisplayImePolicy(virtualDisplayId, DISPLAY_IME_POLICY_LOCAL)`
```

Mr-JingShi:
> Thank you for your PR. After a quick glance, it looks very good and clean (with all details implemented) 👍
> 
> As a French speaker, I have never needed to use an IME, so I'm not very familiar with IME policies. I'd like to get opinions from IME users 🙂
> 
> There are 3 options (local, fallback, hide), is it absolutely necessary to expose them? (Maybe yes, I'm just asking.) Or isn't there an "obvious" default option to set without asking the user?
> 
> For example, are there cases where you need to use an IME on your main display if you're mirroring a virtual display?
> 
> Wouldn't it be sufficient to call unconditionnally in `NewDisplayCapture`:
> 
> ```
> ServiceManager.getWindowManager().setDisplayImePolicy(virtualDisplayId, DISPLAY_IME_POLICY_LOCAL)`
> ```

Thanks for your replay.

I have a project about Huawei electric car，the project sometimes needs to input text then search on secondary display, and then notify the main driver（main display）.Huawei electric cars are very expensive, and not every developer can have one.
so we developed an app instead of Huawei electric cars.

The secondary display is TYPE_HDMI on Huawei electric car, IME show in secondary display.
Android Studio-Add seconday display is TYPE_VIRTUAL, IME show in secondary display.
Developer options-Simulate secondary displays is TYPE_OVERLAY, IME show in main display.
we create a virtualdisplay in my app, the virtualdisplay is TYPE_VIRTUAL,  But IME show in main display.

I'm very confused, then I find Developer options-Force desktop mode, it can show IME in secondary display.
But Some phones don't have this feature.

We know，the main display must be able to display IME normally，so，I just want to display IME in secondary display when I touch in secondary display such as Huawei electric car and Android Studio‘s secondary display。

```
/**
 * Virtual display flag: Indicates that the display should support system decorations. Virtual
 * displays without this flag shouldn't show home, navigation bar or wallpaper.
 * <p>This flag doesn't work without {@link #VIRTUAL_DISPLAY_FLAG_TRUSTED}</p>
 *
 * @see #createVirtualDisplay
 * @see #VIRTUAL_DISPLAY_FLAG_TRUSTED
 * @hide
 */
@TestApi
public static final int VIRTUAL_DISPLAY_FLAG_SHOULD_SHOW_SYSTEM_DECORATIONS = 1 << 9;

/**
 * Virtual display flags: Indicates that the display is trusted to show system decorations and
 * receive inputs without users' touch.
 *
 * @see #createVirtualDisplay
 * @see #VIRTUAL_DISPLAY_FLAG_SHOULD_SHOW_SYSTEM_DECORATIONS
 * @hide
 */
@SystemApi
public static final int VIRTUAL_DISPLAY_FLAG_TRUSTED = 1 << 10;
```
If the secondary screen wants to display IME, it must has the VIRTUAL_DISPLAY_FLAG_TRUSTED and VIRTUAL_DISPLAY_FLAG_SHOULD_SHOW_SYSTEM_DECORATIONS flag, otherwise it will report a permission problem.
```
java.lang.SecurityException: Attempted to set IME flag to an untrusted virtual display: 1
```
The virtualdisplay that created by my app does not have VIRTUAL_DISPLAY_FLAG_TRUSTED in Android 11 、12. 
```
/**
 * @return {@code true} if the display is non-system created virtual display.
 */
boolean isUntrustedVirtualDisplay() {
    return mDisplay.getType() == Display.TYPE_VIRTUAL
            && mDisplay.getOwnerUid() != Process.SYSTEM_UID;
}
```
Although Android 10 does not have a VIRTUAL_DISPLAY_FLAG_TRUSTED flag, But Android 10 will check OwnerUid.
So, Not all virtual display can be set up successfully. then the default value will be meaningless（local、fallback or hide）.

server/src/main/java/com/genymobile/scrcpy/CleanUp.java
Mr-JingShi: Yes, the main display must be able to display IME normally.

server/src/main/java/com/genymobile/scrcpy/wrappers/WindowManager.java
Mr-JingShi: Thank you.

app/src/cli.c
```diff
@@ -3014,9 +3057,17 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
...
-        return false;
+    if (opts->display_id != 0) {
+        if (opts->new_display) {
+            LOGE("Cannot specify both --display-id and --new-display");
+            return false;
+        }
+    } else {
+        if (!opts->new_display
+             && opts->display_ime_policy != SC_DISPLAY_IME_POLICY_UNDEFINDED) {
+            LOGE("--display-ime-policy not supported if display_id == 0 and new_display == NULL");
```
rom1v, line 3068: `--display-ime-policy=hide` would not make sense for the main display?

server/src/main/java/com/genymobile/scrcpy/CleanUp.java
rom1v: (nitpick: maybe we should still allow `--display-ime-policy` for display 0, and let the user not pass the option for the main display if he does not want to?)

app/src/cli.c
Mr-JingShi: I try "setDisplayImePolicy(0, 2)", it doest not work, the main display always show IME. I don't know why until now.
Testing Machine List：
| brand | model | api-level | abi |
| :----: | :----: | :----: | :----: |
| HUAWEI | AGS5-W00 | 31 | arm64-v8a |
| google | Pixel 3a | 32 | arm64-v8a |
| Lenovo | TB331FC | 33 | arm64-v8a |

Mr-JingShi:
> One more question:
> 
> The display IME policy is set at two different places:
> 
> 1. one for the virtual display created by `--new-display`
> 2. one for the case `--display=<secondary_display>`
> 
> For the second case, does it work if you apply the IME policy to the mirroring virtual display instead of using the "clean up" mechanism?
> 
> Something like:
> 
> ```diff
> diff --git server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
> index 5f4e1803f..e9ab32534 100644
> --- server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
> +++ server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
> @@ -127,6 +127,8 @@ public class ScreenCapture extends SurfaceCapture {
>          try {
>              virtualDisplay = ServiceManager.getDisplayManager()
>                      .createVirtualDisplay("scrcpy", inputSize.getWidth(), inputSize.getHeight(), displayId, surface);
> +            int virtualDisplayId = virtualDisplay.getDisplay().getDisplayId();
> +            ServiceManager.getWindowManager().setDisplayImePolicy(virtualDisplayId, displayImePolicy);
>              Ln.d("Display: using DisplayManager API");
>          } catch (Exception displayManagerException) {
>              try {
> ```

The virtualdiplay created by this way is private and untrusted.
```
mBaseDisplayInfo=DisplayInfo{"scrcpy", displayId 4, displayGroupId 0, FLAG_PRIVATE, real 1920 x 1080, largest app 1920 x 1080, smallest app 1920 x 1080, appVsyncOff 0, presDeadline 16666666, mode 860.0, defaultMode 8, modes [{id=8, width=1920, height=1080, fps=60.0, alternativeRefreshRates=[], supportedHdrTypes=[]}], hdrCapabilities null, userDisabledHdrTypes [], minimalPostProcessingSupported false, rotation 0, state ON, committedState UNKNOWN, type VIRTUAL, uniqueId "virtual:com.android.shell,2000,scrcpy,0", app 1920 x 1080, density 1 (1.0 x 1.0) dpi, layerStack 4, colorMode 0, supportedColorModes [0], deviceProductInfo null, owner com.android.shell (uid 2000), removeMode 1, refreshRateOverride 0.0, brightnessMinimum 0.0, brightnessMaximum 0.0, brightnessDefault 0.0, installOrientation ROTATION_0, layoutLimitedRefreshRate null, hdrSdrRatio not_available, thermalRefreshRateThrottling {}, thermalBrightnessThrottlingDataId default}
```
```
Caused by: java.lang.SecurityException: Attempted to set IME policy to an untrusted virtual display: 4
```

app/src/cli.c
Mr-JingShi: my Code:
```
int displayImePolicy = ServiceManager.getWindowManager().getDisplayImePolicy(0);
Ln.d("displayImePolicy:" + displayImePolicy + " before setDisplayImePolicy");
ServiceManager.getWindowManager().setDisplayImePolicy(0, 2);
displayImePolicy = ServiceManager.getWindowManager().getDisplayImePolicy(0);
Ln.d("displayImePolicy:" + displayImePolicy + " after setDisplayImePolicy");
```
print:
[server] DEBUG: displayImePolicy:0 before setDisplayImePolicy
[server] DEBUG: displayImePolicy:0 after setDisplayImePolicy

rom1v:
Sorry for the delay.

I made minor changes:
 - rename the exposed values from "fallback_display" to "fallback" (more consistent with "local" and "hide", even if it's not the case internally in Android)
 - pass the explicit string `"local"` `"fallback"` or `"hide"` as a server argument (instead of an implicit `int` value)
 - reword some documentation parts

Please review and test: e6e8c8a128b170fc966f410ee1a52b503d043227 (branch `pr5703`)

Mr-JingShi:
> Sorry for the delay.
> 
> I made minor changes:
> 
> * rename the exposed values from "fallback_display" to "fallback" (more consistent with "local" and "hide", even if it's not the case internally in Android)
> * pass the explicit string `"local"` `"fallback"` or `"hide"` as a server argument (instead of an implicit `int` value)
> * reword some documentation parts
> 
> Please review and test: [e6e8c8a](https://github.com/Genymobile/scrcpy/commit/e6e8c8a128b170fc966f410ee1a52b503d043227) (branch `pr5703`)

https://github.com/Genymobile/scrcpy/blob/e6e8c8a128b170fc966f410ee1a52b503d043227/server/src/main/java/com/genymobile/scrcpy/wrappers/WindowManager.java#L254C12-L262C14

```
Method method = getSetDisplayImePolicyMethod();
if (Build.VERSION.SDK_INT >= AndroidVersions.API_31_ANDROID_12) {
    method.invoke(manager, displayId, displayImePolicy);
    return;
}
if (displayImePolicy != DISPLAY_IME_POLICY_HIDE) {
    method.invoke(manager, displayId, displayImePolicy == DISPLAY_IME_POLICY_LOCAL);
} else {
    Ln.w("DISPLAY_IME_POLICY_HIDE is not supported before Android 12");
}
```
or
```
Method method = getSetDisplayImePolicyMethod();
if (Build.VERSION.SDK_INT >= AndroidVersions.API_31_ANDROID_12) {
    method.invoke(manager, displayId, displayImePolicy);
} else if (displayImePolicy != DISPLAY_IME_POLICY_HIDE) {
    method.invoke(manager, displayId, displayImePolicy == DISPLAY_IME_POLICY_LOCAL);
} else {
    Ln.w("DISPLAY_IME_POLICY_HIDE is not supported before Android 12");
}
```

rom1v:
Good catch :+1: Fixed up on `pr5703`.

Mr-JingShi:
> Good catch 👍 Fixed up on `pr5703`.

test OK.

rom1v:
Merged into `dev`: fd8bef68b794442d9f3bbf47d85632f9a16e504a

PR #5731, opened 2025-01-03
Error: `brew cask` is no longer a `brew` command. Use `brew <command>…

zhaonian:
Error: `brew cask` is no longer a `brew` command. Use `brew <command> --cask` instead.

### DIFF
--- a/doc/macos.md
+++ b/doc/macos.md
@@ -34,7 +34,7 @@ brew install scrcpy
 You need `adb`, accessible from your `PATH`. If you don't have it yet:
 
 ```bash
-brew install --cask android-platform-tools
+brew install android-platform-tools --cask 
 ```
 
 Alternatively, Scrcpy is also available in [MacPorts], which sets up `adb` for you:


rom1v:
> Error: `brew cask` is no longer a brew command.

You're just moving `--cask` to the end, there was no `brew cask` in the documentation. Does it change anything?

zhaonian:
brew cask is not a command anymore, it's a param now. `--cask` makes it a command. 

With the latest brew, `brew cask install android-platform-tools` will throw an error now .

image

rom1v:
> brew cask is not a command anymore. We will have to use it as a param now. --cask makes it a parameter.

That's already what the documentation suggests: https://github.com/Genymobile/scrcpy/blob/master/doc/macos.md (`brew --cask` instead of `brew cask`).

There is nothing to change.

zhaonian:
Ooops. Apologies, I overlooked that. Closing

PR #5762, opened 2025-01-16
Use matrix build in GitHub Actions

Biswa96:


Biswa96:
This will help to de-duplicate similar build steps with x86_64 and aarch64 architectures. I am planning to add Linux aarch64 build. Before that, I would like to ask if build matrix would simplify the yaml file in this repository. I will convert other build jobs to matrix style if this get green light.

rom1v:
Thank you for the PR.

> I would like to ask if build matrix would simplify the yaml file in this repository.

Yes, probably :+1: I will check in more details later, but in principle it is better (as long as the steps are exactly the same).

PR #5764, opened 2025-01-17
allow requesting IDR frames through the control channel

aluedeke:
I am using scrcpy server to capture the screen and then broadcast the stream to multiple web clients. Todo so i need an new IDR frame whenever a new client connects.

The change has of course its flaws, as accessing the the MediaCodec isn't really possible from another thread atm. Thus an IDR gets only requested when a new frame was generated.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/Server.java
+++ b/server/src/main/java/com/genymobile/scrcpy/Server.java
@@ -156,6 +156,7 @@ public final class Server {
 
                 if (controller != null) {
                     controller.setSurfaceCapture(surfaceCapture);
+                    controller.setSurfaceEncoder(surfaceEncoder);
                 }
             }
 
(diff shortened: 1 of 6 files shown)


aluedeke:
Closing this PR as it isn't possible to request a key frame and force frame creation, so the reset control message is actually the way to go

rom1v:
> Closing this PR as it isn't possible to request a key frame and force frame creation

I have never played with this, but why isn't it possible to request a key frame?

(it's not really necessary in scrcpy itself, but for your use case it would be useful)

aluedeke:
Requesting the key frame worked fine. But only a change on the screen caused a new (key) frame to be encoded. I couldn't find a way to get a new frame without changes on the screen. Like this new clients had to wait until something changed on the screen when connection which can take up to a minute (an the clock changes). Resetting the connection does exactly this: genereate a new key frame and i couln't observ any negative impact on existing connections.

PR #5765, opened 2025-01-17
feat: handle WM_LBUTTONDOWN WM_LBUTTONUP in Win32

H999:
Recently, I have been using your repo and noticed that it does not support WM_LBUTTONDOWN and WM_LBUTTONUP events in Win32. While searching through the SDL library, I found that these events can be supported by enabling SDL_EventState(SDL_SYSWMEVENT, SDL_ENABLE). I tested this on my device, and it worked stably and met my needs. However, I'm not sure if there could be any negative consequences from using it more widely. Could you consider enhancing the functionality in your repo to include this, as well as potentially other related features?

Here my tested code and I tested with SamSung touch screen diagnostic (SamSung M21):
```py
hwnd = win32gui.FindWindow(None, "")
x=400
y=400
win32gui.PostMessage(hwnd, win32con.WM_LBUTTONDOWN, 0, (y << 16) | x)
win32gui.PostMessage(hwnd, win32con.WM_LBUTTONUP, None, (y << 16) | x)
```

### DIFF
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -2,7 +2,9 @@
 
 #include <assert.h>
 #include <string.h>
+#include <windowsx.h>
 #include <SDL2/SDL.h>
+#include <SDL2/SDL_syswm.h>
 
 #include "events.h"
 #include "icon.h"
@@ -325,6 +327,7 @@ sc_screen_frame_sink_push(struct sc_frame_sink *sink, const AVFrame *frame) {
 bool
 sc_screen_init(struct sc_screen *screen,
                const struct sc_screen_params *params) {
+    SDL_EventState(SDL_SYSWMEVENT, SDL_ENABLE);
     screen->resize_pending = false;
     screen->has_frame = false;
     screen->fullscreen = false;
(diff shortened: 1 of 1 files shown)


dcq01 (commented):
Hi! I'm a grad student working on a research project about using large language models to automate code review. Based on your commit 40305fc9b789f551aa96a27c4251dfe7ce8b0ad0 and the changes in app/src/screen.c, my tool generated this comment:
1. Consider checking if the `sdlEvent.button.x` and `sdlEvent.button.y` values are within the bounds of the window or screen.
2. 2. It may be beneficial to add error handling or logging in case `event->syswm.msg` is `NULL`.
3. 3. Ensure that all parameters passed to `sc_screen_init` and other functions are validated. Check that `params->window_width` and `params->window_height` are non-negative and within acceptable limits to prevent any potential buffer overflows or invalid memory access.
4. 4. The new code does not include any error handling for the SDL event pushing. It is generally good practice to check the return value of `SDL_PushEvent` to ensure that the event was successfully pushed.
5. 5. Ensure that all resources are properly released in every error path in `sc_screen_init`, including checking if `screen->frame` is allocated and freeing it if necessary.
6. 6. The use of `assert(title);` assumes that `title` is never null. Consider adding a proper error handling mechanism instead of relying solely on assertions.
7. 7. Create a helper function to handle the creation and pushing of the SDL event based on the button state to avoid duplication.
8. 8. There are inconsistent spaces around the `->` operator in the new code. For example, `event -> syswm.msg` should be `event->syswm.msg` for consistency with common C coding conventions.
9. 9. Ensure that there are adequate tests covering the new functionality introduced by this patch. Having a solid suite of tests will help ensure that future changes do not introduce regressions.

As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:
1. Does this comment provide suggestions from a dimension you hadn’t considered?
2. 2. Do you find this comment helpful?

Thanks a lot for your time and feedback! And sorry again if this message is a bother.

Issue #5704, opened 2024-12-26
It is recommended to send the current rotation direction of the device when the device is rotated.


sbfkcel:
- I have checked that a similar [Listen for rotation](https://github.com/Genymobile/scrcpy/issues?q=Listen+for+rotation) does not already exist.

**Is your feature request related to a problem? Please describe.**
Monitors the device's rotation direction.

**Describe the solution you'd like**
When the screen size changes, the screen rotation direction is obtained and sent out through the socket along with the key frame data before or after SPS.

**Describe alternatives you've considered**
This is useful if you need to draw the device appearance correctly. Although it is also possible to obtain it through adb, there are delays or instabilities.

**Additional context**
Related Discussions

https://github.com/Genymobile/scrcpy/issues/3605

https://github.com/Genymobile/scrcpy/issues/925

rom1v:
In scrcpy, the video frame size (known once the frame is decoded) is sufficient, so it does not transmit an additional rotation.

Moreover, it is not always obvious what the rotation value should be in all cases:
 - `scrcpy --new-display=1920x1080`
 - `scrcpy --new-display=1080x1920`
 - `scrcpy --capture-orientation=90` (when the device is initially portrait)
 - `scrcpy --capture-orientation=90` (when the device is initially landscape)
 - `scrcpy --capture-orientation=flip90` (…)
 - `scrcpy --video-source=camera …`

sbfkcel:
@rom1v I submitted a merge request to fix this issue.

Issue #3605, opened 2022-12-02
How to listen on the change of the screen when the screen is rotated


iTestinglab:
If I only use the scrcpy-server, and then use the jmuxer to decode on frontend, how will the frontend listen when the screen is rotated. Only when it listens, can the next step, such as reset jmuxer to make sure the screen display is normal

I refer to this [issue](https://github.com/Genymobile/scrcpy/issues/925), but didn't find the answer.

Looking forward to your reply.  @rom1v

rom1v:
> how will the frontend listen when the screen is rotated

On the client side, `scrcpy` does not detect rotation per-se, but only video frame size changes. So when a new frame has a different size from the previous one, it resizes the window accordingly.

iTestinglab:
> but only video frame size changes

That is to say, my frontend needs to detect the size of each frame in real time, and reset jmuxer if it is different from the last time?

sbfkcel:
It is recommended to call the getCurrentRotation method to get the device's orientation when the device sends a rotation. The orientation information can be sent with the video in the keyframe header or in the control socket connection.

Issue #925, opened 2019-11-13
how to listen on the change of the screen ?rotation


ShiverZm:
(maybe i know  i look like  i am fool)
i am viewing your code ,i know the aidl listens the phone rotation,but how the pc knows phone is roating?

rom1v:
> how the pc knows phone is roating

The PC doesn't know the device rotation.

When the device rotates, it just receives a frame having another dimension, so it resizes itself.

(Note that it would not be sufficient if we would need to draw a device around the video on the client side)

iTestinglab:
> > how the pc knows phone is roating
> 
> The PC doesn't know the device rotation.
> 
> When the device rotates, it just receives a frame having another dimension, so it resizes itself.
> 
> (Note that it would not be sufficient if we would need to draw a device around the video on the client side)

you means client auto resize itself when  it receives a frame having another dimension.
But f I only use the scrcpy server, and then use the jmuxer to decode on frontend, how will the frontend listen when the screen is rotated. Only when it is monitored can the jmuxer be reset

Issue #1645, opened 2020-08-07
Ignore touch event, it was generated for a different device size


swamikamal:
Ignore touch event, it was generated for a different device size
-v 1.15 not recognizing my click as touch on screen
please help me 
![Screenshot (423)](https://user-images.githubusercontent.com/49489982/89632868-8f6fa300-d8c0-11ea-87cc-f60940423de8.png)

rom1v:
Ah, great, it was a good idea to add the log a973757fd141fc07ef81314d350bd4f4de64f59c
See https://github.com/Genymobile/scrcpy/issues/1518#issuecomment-655517010

The reason is that on some devices, the video encoder on the device does not respect the requested size, and we can't detect this change on the server side.

To avoid the problem, check the first lines in scrcpy output:

```
INFO: Initial texture: 1080x2336
INFO: New texture: 888x1920
```

And limit the size to what the encoder actually produces. Here:

```
scrcpy -m1920
```

(take the greatest value of the second line)

swamikamal:
![Screenshot (424)](https://user-images.githubusercontent.com/49489982/89634952-df039e00-d8c3-11ea-8ffe-da54ab0ba4f0.png)

rom1v:
_(please post output as text instead of a screenshot)_

Then just run `scrcpy -m1920` as I explained above.

swamikamal:
INFO: Initial texture: 1080x2336
INFO: New texture: 888x1920 
...................................................................
now what i have to do

swamikamal:
Sorry!! And Thank you very much Sir.

xrobb30:
thank you so much, master. it works for my oppo F7 hihi

skivox52:
Same issue on MI9SE.
I tried using different  "-m XXXX" option, but still doesn't work.

How to know which value to use please ? 
Here is the output while launching scrcpy without "-m" option : 

[server] INFO: Device: Xiaomi Mi 9 SE (Android 10)INFO: Renderer: opengl
INFO: OpenGL version: 3.0 Mesa 20.0.8
INFO: Trilinear filtering enabled
INFO: Initial texture: 1080x2336

Merci

skivox52:
I do not have these two lines.
Only the lines I shown you.

I do not get any error message, but the mouse event doesn't work.
if i open app touching phisically my phone, i can see it on scrcpy but i cannot interact with it using the mouse.

Sorry if I posted in the wrong thread.
Thanks

[...]

PR #5894, opened 2025-03-03
Add device orientation information to header information

sbfkcel:
This is useful if need to draw the device appearance correctly.

Related Discussions

https://github.com/Genymobile/scrcpy/issues/5704
https://github.com/Genymobile/scrcpy/issues/3605
https://github.com/Genymobile/scrcpy/issues/925

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build/
 /x/
 local.properties
 /scrcpy-server
+.DS_Store
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -23,6 +23,7 @@ _scrcpy() {
         -d --select-usb
         --disable-screensaver
         --display-id=
+        --display-ime-policy=
         --display-orientation=
         -e --select-tcpip
         -f --fullscreen
@@ -146,6 +147,10 @@ _scrcpy() {
             ;;
         --orientation|--display-orientation)
             COMPREPLY=($(compgen -W '0 90 180 270 flip0 flip90 flip180 flip270' -- "$cur"))
+            return
+            ;;
+        --display-ime-policy)
+            COMPREPLY=($(compgen -W 'local fallback hide' -- "$cur"))
             return
             ;;
         --record-orientation)
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -30,6 +30,7 @@ arguments=(
     {-d,--select-usb}'[Use USB device]'
     '--disable-screensaver[Disable screensaver while scrcpy is running]'
     '--display-id=[Specify the display id to mirror]'
+    '--display-ime-policy[Set the policy for selecting where the IME should be displayed]'
     '--display-orientation=[Set the initial display orientation]:orientation values:(0 90 180 270 flip0 flip90 flip180 flip270)'
     {-e,--select-tcpip}'[Use TCP/IP device]'
     {-f,--fullscreen}'[Start in fullscreen]'
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -389,7 +389,7 @@ Disable video playback on the computer.
 
 .TP
 .B \-\-no\-window
-Disable scrcpy window. Implies --no-video-playback and --no-control.
+Disable scrcpy window. Implies --no-video-playback.
 
 .TP
 .BI "\-\-orientation " value
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -162,6 +162,19 @@ The available display ids can be listed by \fB\-\-list\-displays\fR.
 Default is 0.
 
 .TP
+.BI "\-\-display\-ime\-policy " value
+Set the policy for selecting where the IME should be displayed.
+
+Possible values are "local", "fallback" and "hide":
+
+ - "local" means that the IME should appear on the local display.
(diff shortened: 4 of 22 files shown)


rom1v:
Hi,

Thanks for your work.

However, as I said in https://github.com/Genymobile/scrcpy/issues/5704#issuecomment-2562463440, there is no really need to know the device orientation, since it can be inferred from the frame size in practice, and there are cases where the "orientation" to write is not clear.

You replied:

> But in some cases, it would be good to have the correct device rotation direction. For example, it is necessary to correctly draw the appearance of a device

You can do this based on the actual frame size.

In theory, it could make sense to pass the device orientation (at least in some cases) in packets headers (even if it could be argued that it's overkill to pass it for each packet/frame).

But then, the orientation must exactly match the frame it describes, which is not guaranteed here. There is no mechanism to make sure that the frame produced by MediaCodec is in the current device orientation (as returned by `Device.getCurrentRotation(this.displayId);`), it may have changed, so a few frames may be tagged with an incorrect orientation. On the contrary, the frame size is guaranteed to be in sync.

sbfkcel:
The size of the screen can only be used to determine whether it is displayed horizontally or vertically.

If it is just a pure screen display, then it is meaningless to obtain the device direction.

However, if you need to draw the hole position of the front camera in the screen normally, or if you need to draw the position of the device appearance button (power button) as discussed above, then you need an accurate device direction.

I used to monitor the change of the screen size, and when the size changes, I used ADB to obtain the screen direction, which will cause a very large delay.

From the current test, the implementation in the code is relatively ideal. The human eye cannot clearly distinguish that the obtained value is abnormal.

Or is it feasible for us to make this item an optional method?

rom1v:
>
> However, if you need to draw the hole position of the front camera in the screen normally, or if you need to draw the position of the device appearance button (power button) as discussed above, then you need an accurate device direction.

OK.

What is the client on your screenshot? Is it a modified scrcpy client, or an alternative client you wrote?

> The human eye cannot clearly distinguish that the obtained value is abnormal.

If the orientation is sent "inline", it must be in sync with the encoded frames. Btw, it could allow to print a better error message for #1645.

The current [protocol](https://github.com/Genymobile/scrcpy/blob/master/doc/develop.md#video-and-audio) has two "packet types":
 - a codec packet; for video: codec id (`u32`), width (`u32`), height (`u32`)
 - a media packet for each frame (with a 12 bytes header)

To support your use case, I suggest to add a new packet type:
 - a codec packet, with only the codec id (`u32`, i.e. 4 bytes)
 - a session packet for video (12 bytes): width (`u32`), height (`u32`), orientation (`u3`, there a 8 possible orientation with `--capture-orientation`), sent every time the encoding is restarted (e.g. on rotation)
 - a media packet for each frame (with a 12 bytes header)

The session packet could be distinguished from the media packet with a single bit, and for simplicity both headers would have 12 bytes.

It would be sent when a new encoding session starts, with the values in sync to what is passed to the encoder.

sbfkcel:
Yes, I wrote an alternative client.

I will try to make some adjustments based on your suggestions and then submit a new implementation.

sbfkcel:
OK, I did overlook factors like the camera. I'll submit a new implementation plan later.

sbfkcel:
Through testing and verification, it is found that the changes in each direction of the device cannot be captured in ScreenCapture. For example: 0->2, 1->3, when the screen size does not change, the logic will not pass here, but the physical device direction has changed.

Rotating from 0->1->2->3 in sequence will trigger re-capture because the size has changed.

Therefore, the best way may still need to be implemented in SurfaceEncoder. By adding a parameter to SurfaceEncoder to distinguish whether the device is a display,.

rom1v:
> it is found that the changes in each device direction cannot be captured in ScreenCapture. For example: 0->2, 1->3, when the screen size does not change, the logic will not be called, but the physical device direction has changed.

Because currently the encoding session is "reset" only if the size changes:

https://github.com/Genymobile/scrcpy/blob/7998811fa55a4f1610dd07680a387251ead2dd1f/server/src/main/java/com/genymobile/scrcpy/video/DisplaySizeMonitor.java#L107-L138

rom1v:
With this change, it resets on rotation even if the size is the same: a98e50078afd25c2b6070d25f56c7bb72f1e6771

rom1v:
On a PR, please never include merge commits, always rebase on the latest `dev` branch.

sbfkcel:
OK, wrong operation

rom1v:
> With this change, it resets on rotation even if the size is the same: https://github.com/Genymobile/scrcpy/commit/a98e50078afd25c2b6070d25f56c7bb72f1e6771

On the other hand, it causes an unnecessary reset (since it works without it). This is just a corner case, but in theory we should NOT reset the encoding, but still send a new session packet to signal the new rotation.

sbfkcel:
> a session packet for video (12 bytes): width (u32), height (u32), orientation (u3, there a 8 possible orientation with --capture-orientation), sent every time the encoding is restarted (e.g. on rotation)

Is the session packet treated as a special header packet without any data (11 bits + 1 area for type distinction)?

Or is it the same as the media packet with header information + session packet itself?

rom1v:
> Is the session packet treated as a special header packet without any data (11 bytes + 1 byte for type distinction)?

Yes, I would do something like that. The very first bit allows to distinguish between session and media packet.

sbfkcel:
OK

PR #5905, opened 2025-03-06
Arm build (Github Actions)

Tianming-Wu:
You can notice the user to replace the adb with the one on there computer first.


Tianming-Wu:
What about getting from the ubuntu repository? There is an arm version that runs perfectly.

PR #5918, opened 2025-03-13
Add device orientation information to header information

sbfkcel:
This is useful if need to draw the device appearance correctly.

Related Discussions

https://github.com/Genymobile/scrcpy/issues/5704
https://github.com/Genymobile/scrcpy/issues/3605
https://github.com/Genymobile/scrcpy/issues/925

https://github.com/Genymobile/scrcpy/pull/5894

---

```
CODEC    SESSION packet                         SESSION packet
[codec] [type|width|height|orient] […] […] […] [type|width|height|orient] […] […]
                                              ^
                                              rotation occurs
```
**CODEC packet：**
- `codecId` 4byte

**SESSION packet:**
- `type` 1byte, The SESSION packet type is `0xff`
- `width` 4byte
- `hegiht` 4byte
- `orient` 3byte 
  - `isFlip` 1byte
  - `180°` 1byte
  - `90°` 1byte

**Orient format reference source**
https://github.com/Genymobile/scrcpy/blob/7998811fa55a4f1610dd07680a387251ead2dd1f/app/src/options.h#L72-L85

Test run

```
meson setup x --buildtype=release --strip -Db_lto=true &&  ninja -Cx && ./run x
```

Output

```
INFO: Width=1920, Height=864, Flip=False, Direction=1
```


sbfkcel:
@rom1v Can this PR be merged in the next version? If it can be confirmed, I can't wait to update the Web client protocol parsing method.

rom1v:
I just submitted #6159.

It introduces the protocol change to add a "session packet" (you're mentioned as co-author of that commit).

The session packet only includes the video resolution (since scrcpy currently does not use the device rotation), but the architecture is here, you can add the rotation on your branch with minimal changes I guess.

I adapted the client so that everything works correctly (as mentioned here: https://github.com/Genymobile/scrcpy/pull/5894#issuecomment-2713850575).

sbfkcel:
Okay, future versions will align with the new protocol.
Adding different flags to distinguish data will not pollute the project.
look forward to merging it into the official branch after implementation.

PR #6008, opened 2025-04-16
Make inject mode configurable

bartsaintgermain:
For some use-cases we would like to avoid the async behaviour of the key inject event.

To do so the MR does make the inject mode configurable on starting ScrCpy server.


rom1v:
> For some use-cases

Could you give examples? I have never managed to observe a difference.

rom1v:
Any concrete use case?

Issue #5523, opened 2024-11-24
ERROR: Could not create display


Frogperson:
## Environment

 - **OS:** Windows 11
 - **Scrcpy version:** 3.0
 - **Installation method:** Windows release
 - **Device model:** Pixel 9 Pro XL
 - **Android version:** 15 QPR2 Beta 1 (BP11.241025.006)

## Describe the bug

Running `scrcpy --new-display` results in the following error:
```
[server] INFO: Device: [Google] google Pixel 9 Pro XL (Android 15)
[server] ERROR: Could not create display
java.lang.SecurityException: Requires ADD_TRUSTED_DISPLAY permission to create a trusted virtual display.
        at android.os.Parcel.createExceptionOrNull(Parcel.java:3231)
        at android.os.Parcel.createException(Parcel.java:3215)
        at android.os.Parcel.readException(Parcel.java:3198)
        at android.os.Parcel.readException(Parcel.java:3140)
        at android.hardware.display.IDisplayManager$Stub$Proxy.createVirtualDisplay(IDisplayManager.java:1570)
        at android.hardware.display.DisplayManagerGlobal.createVirtualDisplay(DisplayManagerGlobal.java:757)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1186)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1146)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1094)
        at android.hardware.display.DisplayManager.createVirtualDisplay(DisplayManager.java:1033)
        at com.genymobile.scrcpy.wrappers.DisplayManager.createNewVirtualDisplay(DisplayManager.java:164)
        at com.genymobile.scrcpy.video.NewDisplayCapture.startNew(NewDisplayCapture.java:186)
        at com.genymobile.scrcpy.video.NewDisplayCapture.start(NewDisplayCapture.java:207)
        at com.genymobile.scrcpy.video.SurfaceEncoder.streamCapture(SurfaceEncoder.java:95)
        at com.genymobile.scrcpy.video.SurfaceEncoder.lambda$start$0$com-genymobile-scrcpy-video-SurfaceEncoder(SurfaceEncoder.java:292)
        at com.genymobile.scrcpy.video.SurfaceEncoder$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
        at java.lang.Thread.run(Thread.java:1012)
[...]

Issue #4074, opened 2023-06-09
[Android 14 Beta3] Can't control the device screen


DerekWuYD:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: debian
 - scrcpy version: v2.0
 - installation method: manual built
 - device model: Pixel 7 Pro
 - Android version: Android 14 Beta 3

**Describe the bug**
In Android 14 beta 3, mouse and keyboard input events are no longer valid. 
Screencasting is working, but remote control has failed.

```
Start arguments : --force-adb-forward --no-audio

scrcpy 2.0 <https://github.com/Genymobile/scrcpy>
x/server/scrcpy-server: 1 file pushed, 0 skipped. 74.6 MB/s (52911 bytes in 0.001s)
[server] INFO: Device: Google Pixel 7 Pro (Android 14)
[server] ERROR: Could not invoke method
java.lang.NoSuchMethodException: android.content.IClipboard$Stub$Proxy.addPrimaryClipChangedListener [interface android.content.IOnPrimaryClipChangedListener, class java.lang.String, class java.lang.String, int]
	at java.lang.Class.getMethod(Class.java:2937)
	at java.lang.Class.getMethod(Class.java:2449)
	at com.genymobile.scrcpy.wrappers.ClipboardManager.getAddPrimaryClipChangedListener(ClipboardManager.java:129)
	at com.genymobile.scrcpy.wrappers.ClipboardManager.addPrimaryClipChangedListener(ClipboardManager.java:139)
	at com.genymobile.scrcpy.Device.<init>(Device.java:100)
	at com.genymobile.scrcpy.Server.scrcpy(Server.java:64)
	at com.genymobile.scrcpy.Server.main(Server.java:381)
	at com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method)
	at com.android.internal.os.RuntimeInit.main(RuntimeInit.java:359)
[server] ERROR: Exception on thread Thread[Thread-2,5,main]
java.lang.AssertionError: java.lang.reflect.InvocationTargetException
	at com.genymobile.scrcpy.wrappers.ServiceManager.getInputManager(ServiceManager.java:72)
	at com.genymobile.scrcpy.Device.injectEvent(Device.java:192)
[...]

PR #6009, opened 2025-04-16
Associate UHID devices to displays on Android 15

### LINKED ISSUE #4829: Unable to control Scrcpy because of the absence of mouse cursor's functionality, when interacting with experimental desktop from android developer settings

**Environment**
 - OS: [e.g. Debian, Windows, macOS...]
 - scrcpy version: [e.g. 1.12.1]
 - installation method: [e.g. manual build, apt, snap, brew, Windows release...]
 - device model:
 - Android version: [e.g. 10]

**Describe the bug**

On errors, please provide the output of the console (and `adb logcat` if relevant).

```
Please paste terminal output in a code block.
```

Please do not post screenshots of your terminal, just post the content as text instead.

### LINKED ISSUE #5547: Is there a setting that allows me to use a mouse thats plugged into the PC on the virtual display instead of the actual phone?
Hello, I was wondering if it was possible to use the mouse plugged into the PC using scrcpy for a virtual display. When I try to make a new virtual display, everything works fine, except I can't use the mouse on the display. Instead, it shows up on my actual phone. If it helps, I use a Samsung Galaxy A21. I don't know if I'm just not seeing something, or its some sort of bug, but if I could get any sort of help, that would be really appreciated. Thank you.

### LINKED ISSUE #5557: Add virtual mouse and keyboard support for Virtual Display
I can't use --keyboard=uhid and --mouse=uhid in virtual Display because it's showing in real Display

yume-chan:
Fixes #4829
Fixes #5547
Fixes #5557

Only works on Android 15 and newer because it requires `ASSOCIATE_INPUT_DEVICE_TO_DISPLAY` permission.

When running multiple instances of `scrcpy --display-id` or `scrcpy --new-display`, each display will have its own mouse pointer.

Has no effect on keyboard, the whole system still only has one input focus.

I used `FakeContext.get().getSystemService(Context.INPUT_SERVICE)` to get `InputManager`, because initially I was testing `addPortAssociation`, which doesn't exist on `InputManagerGlobal`. This simplifies the code, especially for `ClipboardManager`, where we can use the stable public API (https://github.com/yume-chan/leap-scrcpy/blob/f9aaf1b05118261d75b82ba88b462f08e37eecdc/server/app/src/main/java/leap/scrcpy/server/FakeContext.kt#L62-L72)

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
@@ -1,8 +1,10 @@
 package com.genymobile.scrcpy.control;
 
 import com.genymobile.scrcpy.AndroidVersions;
+import com.genymobile.scrcpy.device.DisplayInfo;
 import com.genymobile.scrcpy.util.Ln;
 import com.genymobile.scrcpy.util.StringUtils;
+import com.genymobile.scrcpy.wrappers.ServiceManager;
 
 import android.os.Build;
 import android.os.HandlerThread;
@@ -31,6 +33,8 @@ public final class UhidManager {
 
     private static final int SIZE_OF_UHID_EVENT = 4380; // sizeof(struct uhid_event)
 
+    private static final String inputPort = "scrcpy:" + Os.getpid();
+
     private final ArrayMap<Integer, FileDescriptor> fds = new ArrayMap<>();
     private final ByteBuffer buffer = ByteBuffer.allocate(SIZE_OF_UHID_EVENT).order(ByteOrder.nativeOrder());
 
@@ -45,6 +49,13 @@ public final class UhidManager {
             queue = thread.getLooper().getQueue();
         } else {
             queue = null;
+        }
+    }
+
+    public static void setDisplayId(int displayId) {
+        if (Build.VERSION.SDK_INT >= AndroidVersions.API_34_ANDROID_14 && displayId != 0) {
+            DisplayInfo displayInfo = ServiceManager.getDisplayManager().getDisplayInfo(displayId);
+            ServiceManager.getInputManager().addUniqueIdAssociationByPort(inputPort, displayInfo.getUniqueId());
         }
     }
 
@@ -170,16 +181,21 @@ public final class UhidManager {
          * } __attribute__((__packed__));
          */
 
-        byte[] empty = new byte[256];
         ByteBuffer buf = ByteBuffer.allocate(280 + reportDesc.length).order(ByteOrder.nativeOrder());
         buf.putInt(UHID_CREATE2);
 
         String actualName = name.isEmpty() ? "scrcpy" : name;
-        byte[] utf8Name = actualName.getBytes(StandardCharsets.UTF_8);
-        int len = StringUtils.getUtf8TruncationIndex(utf8Name, 127);
-        assert len <= 127;
-        buf.put(utf8Name, 0, len);
-        buf.put(empty, 0, 256 - len);
+        byte[] nameBytes = actualName.getBytes(StandardCharsets.UTF_8);
+        int nameLen = StringUtils.getUtf8TruncationIndex(nameBytes, 127);
+        buf.put(nameBytes, 0, nameLen);
+        buf.position(buf.position() + 128 - nameLen);
+
+        byte[] physBytes = inputPort.getBytes(StandardCharsets.UTF_8);
+        int physLen = StringUtils.getUtf8TruncationIndex(physBytes, 63);
+        buf.put(physBytes, 0, physLen);
+        buf.position(buf.position() + 64 - physLen);
+
+        buf.position(buf.position() + 64); // uniq
 
         buf.putShort((short) reportDesc.length);
         buf.putShort(BUS_VIRTUAL);
--- a/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
@@ -53,7 +53,7 @@ public final class UhidManager {
     }
 
     public static void setDisplayId(int displayId) {
-        if (Build.VERSION.SDK_INT >= AndroidVersions.API_34_ANDROID_14 && displayId != 0) {
+        if (Build.VERSION.SDK_INT >= AndroidVersions.API_35_ANDROID_15 && displayId != 0) {
             DisplayInfo displayInfo = ServiceManager.getDisplayManager().getDisplayInfo(displayId);
             ServiceManager.getInputManager().addUniqueIdAssociationByPort(inputPort, displayInfo.getUniqueId());
         }
@@ -241,6 +241,10 @@ public final class UhidManager {
     }
 
     public void closeAll() {
+        if (Build.VERSION.SDK_INT >= AndroidVersions.API_35_ANDROID_15) {
+            ServiceManager.getInputManager().removeUniqueIdAssociationByPort(inputPort);
+        }
+
         for (FileDescriptor fd : fds.values()) {
             close(fd);
         }
--- a/server/src/main/java/com/genymobile/scrcpy/control/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/control/Controller.java
@@ -148,6 +148,7 @@ public class Controller implements AsyncProcessor, VirtualDisplayListener {
                 displayDataAvailable.notify();
             }
         }
+        UhidManager.setDisplayId(virtualDisplayId);
     }
 
     public void setSurfaceCapture(SurfaceCapture surfaceCapture) {
--- a/server/src/main/java/com/genymobile/scrcpy/device/DisplayInfo.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/DisplayInfo.java
@@ -7,16 +7,18 @@ public final class DisplayInfo {
     private final int layerStack;
     private final int flags;
     private final int dpi;
+    private final String uniqueId;
 
     public static final int FLAG_SUPPORTS_PROTECTED_BUFFERS = 0x00000001;
 
-    public DisplayInfo(int displayId, Size size, int rotation, int layerStack, int flags, int dpi) {
+    public DisplayInfo(int displayId, Size size, int rotation, int layerStack, int flags, int dpi, String uniqueId) {
         this.displayId = displayId;
         this.size = size;
         this.rotation = rotation;
         this.layerStack = layerStack;
         this.flags = flags;
         this.dpi = dpi;
+        this.uniqueId = uniqueId;
     }
 
     public int getDisplayId() {
@@ -41,6 +43,10 @@ public final class DisplayInfo {
 
     public int getDpi() {
         return dpi;
+    }
+
+    public String getUniqueId() {
+        return uniqueId;
     }
 }
 
--- a/server/src/main/java/com/genymobile/scrcpy/wrappers/DisplayManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/wrappers/DisplayManager.java
@@ -81,7 +81,7 @@ public final class DisplayManager {
         int density = Integer.parseInt(m.group(5));
         int layerStack = Integer.parseInt(m.group(6));
 
-        return new DisplayInfo(displayId, new Size(width, height), rotation, layerStack, flags, density);
+        return new DisplayInfo(displayId, new Size(width, height), rotation, layerStack, flags, density, "");
     }
 
     private static DisplayInfo getDisplayInfoFromDumpsysDisplay(int displayId) {
@@ -129,7 +129,8 @@ public final class DisplayManager {
             int layerStack = cls.getDeclaredField("layerStack").getInt(displayInfo);
             int flags = cls.getDeclaredField("flags").getInt(displayInfo);
             int dpi = cls.getDeclaredField("logicalDensityDpi").getInt(displayInfo);
-            return new DisplayInfo(displayId, new Size(width, height), rotation, layerStack, flags, dpi);
+            String uniqueId = (String)cls.getDeclaredField("uniqueId").get(displayInfo);
+            return new DisplayInfo(displayId, new Size(width, height), rotation, layerStack, flags, dpi, uniqueId);
         } catch (ReflectiveOperationException e) {
             throw new AssertionError(e);
         }
--- a/server/src/main/java/com/genymobile/scrcpy/wrappers/InputManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/wrappers/InputManager.java
@@ -89,13 +89,7 @@ public final class InputManager {
 
     private static Method getAddUniqueIdAssociationByPortMethod() throws NoSuchMethodException {
         if (addUniqueIdAssociationByPortMethod == null) {
-            try {
-                // Android 15
-                addUniqueIdAssociationByPortMethod = android.hardware.input.InputManager.class.getMethod("addUniqueIdAssociationByPort", String.class, String.class);
-            } catch (NoSuchMethodException ignored) {
-                // Android 12 to 14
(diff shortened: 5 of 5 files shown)


rom1v:
Thank you :+1: I will review later.

Just a quick question:

> I used `FakeContext.get().getSystemService(Context.INPUT_SERVICE) `to get `InputManager`, because initially I was testing `addPortAssociation`, which doesn't exist on `InputManagerGlobal`.

Doesn't it cause a regression for an issue fixed by 5bd75148716496755dc32daab21a322f51dc58ca on some Android versions?

yume-chan:
> Doesn't it cause a regression for an issue fixed by https://github.com/Genymobile/scrcpy/commit/5bd75148716496755dc32daab21a322f51dc58ca on some Android versions?

I see #4074 is caused by `ActivityThread.currentApplication()` being `null`, but now there is a polyfilled one from

https://github.com/Genymobile/scrcpy/blob/dc2fcc46f516588f4575c1fe8cfeca3e57a1653c/server/src/main/java/com/genymobile/scrcpy/Workarounds.java#L119-L134

So If I change `com.genymobile.scrcpy.wrappers.InputManager#create` back to

```java
    static InputManager create() {
        try {
            Object im = android.hardware.input.InputManager.class.getDeclaredMethod("getInstance").invoke(null);
            return new InputManager(im);
        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
            throw new RuntimeException(e);
        }
    }
```

It still works (and calls `FakeContext.get().getSystemService(Context.INPUT_SERVICE)` through several layers of indirections)

rom1v:
> This simplifies the code, especially for ClipboardManager, where we can use the stable public API

I just tested to simplify `ClipboardManager` by using `FakeContext.get().getSystemService(FakeContext.CLIPBOARD_SERVICE)` as follow: 363c9a5b3f09d20bcbe5d178cadf9f92a091a544

But then, when I paste:

```
java.lang.SecurityException: Package android does not belong to 2000
	at android.os.Parcel.createExceptionOrNull(Parcel.java:3261)
	at android.os.Parcel.createException(Parcel.java:3245)
	at android.os.Parcel.readException(Parcel.java:3228)
	at android.os.Parcel.readException(Parcel.java:3170)
	at android.content.IClipboard$Stub$Proxy.getPrimaryClip(IClipboard.java:459)
	at android.content.ClipboardManager.getPrimaryClip(ClipboardManager.java:253)
	at com.genymobile.scrcpy.wrappers.ClipboardManager.getText(ClipboardManager.java:27)
	at com.genymobile.scrcpy.device.Device.getClipboardText(Device.java:106)
	at com.genymobile.scrcpy.device.Device.setClipboardText(Device.java:119)
	at com.genymobile.scrcpy.control.Controller.setClipboard(Controller.java:568)
	at com.genymobile.scrcpy.control.Controller.handleEvent(Controller.java:279)
	at com.genymobile.scrcpy.control.Controller.control(Controller.java:192)
	at com.genymobile.scrcpy.control.Controller.lambda$start$1$com-genymobile-scrcpy-control-Controller(Controller.java:200)
	at com.genymobile.scrcpy.control.Controller$$ExternalSyntheticLambda1.run(D8$$SyntheticClass:0)
	at android.app.AppOpsManager.checkPackage(AppOpsManager.java:9636)
	at com.android.server.clipboard.ClipboardService.clipboardAccessAllowed(ClipboardService.java:1329)
	at com.android.server.clipboard.ClipboardService.clipboardAccessAllowed(ClipboardService.java:1313)
	at com.android.server.clipboard.ClipboardService.-$$Nest$mclipboardAccessAllowed(ClipboardService.java:0)
	at com.android.server.clipboard.ClipboardService$ClipboardImpl.getPrimaryClip(ClipboardService.java:663)
```

rom1v:
Indeed, thank you, it works: a8997525b2c59661f3965514c7e4f27b0922f75c

server/src/main/java/com/genymobile/scrcpy/control/UhidManager.java
```diff
@@ -52,13 +82,15 @@ public void open(int id, int vendorId, int productId, String name, byte[] report
         try {
             FileDescriptor fd = Os.open("/dev/uhid", OsConstants.O_RDWR, 0);
             try {
-                FileDescriptor old = fds.put(id, fd);
+                // Must be unique across the system
+                String inputPort = "scrcpy:" + Os.getpid() + ":" + id;
```
rom1v, line 86: There is only one virtual display per scrcpy session, do we need one input port per UHID device?

I think we could just use `String inputPort = "scrcpy:" + Os.getpid();` and call `addUniqueIdAssociationByPort()` only once (and all UHID devices would use the same input port associated to the single display id).

I'm not asking you to make any change, I would just like to know what you think about it.

(Currently, I cannot test virtual displays because of #5523 on my current device.)

yume-chan: I never thought about that. Initially I saw Android Studio is using different `phys` values for each device: https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:streaming/screen-sharing-agent/app/src/main/cpp/virtual_input_device.cc;l=66-68;drc=fc06f05ca98ff7d3c8558976fd7eaf1b5480e2e8

But I believe multiple devices can have the same `phys` value, it means they are in one physical device, but with multiple input features. And `addUniqueIdAssociationByPort` still works.

The original commit has been backed up at https://github.com/yume-chan/scrcpy/tree/feat/associate-inpt-port-2

rom1v:
> (Currently, I cannot test virtual displays because of https://github.com/Genymobile/scrcpy/issues/5523 on my current device.)

I tested on two devices:
 - Motorola Edge 30 Neo with Android 14
 - Redmi Note 13 with Android 14

On both of them, running either your initial branch or your new one, the UHID mouse pointer does not appear on the virtual display (but it correctly disappears from the main display).

I run with:

```bash
scrcpy --new-display -M
scrcpy --new-display -M --start-app=org.videolan.vlc  # to get an app running
```

yume-chan:
I see. So Android 14 still only has one mouse pointer across all displays. We need not only associate the UHID device to the display, but also move the mouse cursor to the display, which is impossible because the API is not available outside system server (same as my original comment https://github.com/Genymobile/scrcpy/issues/4829#issuecomment-2047051252).

Only Android 15 added seperated mouse cursors for each display (https://cs.android.com/android/_/android/platform/frameworks/native/+/da10dd34cb41f0f117fe87f568f16f6025ab3d72)

rom1v:
Thank you.

Here is a branch: [`pr6009.3`](https://github.com/Genymobile/scrcpy/commits/pr6009.3/) (not tested, because I currently have no device with Android 15 not impacted by #5523), with the following changes:

 - extract the change for using the public `InputManager` API to a separate commit
 - pass the display unique id in the "synchronized" `DisplayData` instance (`DisplayManager` can only be called from the "display" thread currently, as there is no synchronization, so pass the display unique id along with the display id and the position mapper)
 - wait for the virtual display to be available before creating the UhidManager (with a timeout of 1 second), using the same mechanism as for `--start-app`
 - handle the case where the unique id is `null` (in that case, do not associate an input port)
 - (as a side effect) also associate in case `displayId == 0` (should we make a special case?)

TODO:
 - retrieve the display unique id also with the fallback mechanism in `DisplayManager.parseDisplayInfo()`

rom1v:
> TODO:
>  - retrieve the display unique id also with the fallback mechanism in `DisplayManager.parseDisplayInfo()`

It would require to rework the parsing, as a single regex is not sufficient (depending on the device, the values are not always in the same order). Since it's just a fallback for some old devices, it is probably useless, so I will not do it until necessary.

So I think the branch can be considered _ready_.

Could you please review/test? I managed to test it successfully on a Pixel 6a.

yume-chan:
The timing between adding port association and creating UHID devices doesn't matter, so I think it could be simplified a little (just adding the association when initializing UHID manager and remove it on exit).

The association will NOT be removed by Android OS when Scrcpy process exits, although it won't cause any harm, if we want to make sure it's 100% clean, maybe we need to add it to the CleanUp process.

Otherwise looks good to me and tests OK on my Redmi K70 Pro running Android 15.

rom1v:
Thank you for your review.

> The timing between adding port association and creating UHID devices doesn't matter, so I think it could be simplified a little (just adding the association when initializing UHID manager and remove it on exit).

When scrcpy runs without UHID device, I want to avoid adding a useless association. So it must be done only when at least one UHID device is created (and only once the virtual display id is known, whichever happens first).

Also, I think that my branch only add the association for "new virtual displays" (not when `--display-id=XX` is passed).

What makes it more complicated is that currently, `DisplayManager` may not be called from the controller thread (and I don't want to add unnecessary synchronization, but maybe I'll have to so that I can get the `displayUniqueId` from a `displayId` in the controller thread, I will think about it).

rom1v:
> Also, I think that my branch only adds the association for "new virtual displays" (not when --display-id=XX is passed).

Fixed, now it should be called both for virtual displays and non-main display (I disabled the association for the main display, since it works by default).

> What makes it more complicated is that currently, `DisplayManager` may not be called from the controller thread (and I don't want to add unnecessary synchronization, but maybe I'll have to so that I can get the `displayUniqueId` from a `displayId` in the controller thread, I will think about it).

I added synchronization and now call `getDisplayInfo()` from the controller thread.

> The association will NOT be removed by Android OS when Scrcpy process exits, although it won't cause any harm, if we want to make sure it's 100% clean, maybe we need to add it to the CleanUp process.

OK. Since it is harmless, for simplicity I just remove the association when it closes normally, but not if the process is killed.

New version: [`pr6009.7`](https://github.com/Genymobile/scrcpy/commits/pr6009.7/)

rom1v:
I rebased: `pr6009.8`.

@yume-chan Could you please test and confirm that it works correctly before merging? I was able to test with an older version of the branch, but the device I used has been updated and is now impacted by #5523).

yume-chan:
I tested `pr6009.8` works on my Redmi K70 Pro running Android 15.

rom1v:
@yume-chan Thank you for your tests :+1:

I just noticed that 3b36f29be94785ba9ddf96ff6afb259a0d4815a1 breaks copy-paste from the Android device to the computer: the callback passed to `clipboardManager.addPrimaryClipChangedListener()` is never called.

anotheruserofgithub:
@rom1v Just a side remark in `Controller.waitDisplayData(long)` called from [`Controller.getUhidManager()`](https://github.com/Genymobile/scrcpy/commit/8314cecd35cb764fc617457ff01d8da0528c5da3#diff-369ec79aac5f1da9cc5c6880ca1ed60c68ab7f8f2f14ab0363bceb9de5206d40R162-R163): https://github.com/Genymobile/scrcpy/blob/e0f37f834bb2e9371c0ca893757b60eb36e759af/server/src/main/java/com/genymobile/scrcpy/control/Controller.java#L698-L702 It should test `if (timeout <= 0)` because if `timeout` is zero then[`Object.wait(long)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#wait-long-) waits for a notification (or a spurious wakeup).

Other classes in the `java.util.concurrent` package behave differently when `timeout` is zero or negative, though you often have to delve into the implementation to be sure how they work. For instance, [`ReentrantLock.tryLock(long, TimeUnit)`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html#tryLock-long-java.util.concurrent.TimeUnit-) ([source](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/concurrent/locks/AbstractQueuedSynchronizer.java#L915-L916), called from [here](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/concurrent/locks/ReentrantLock.java#L442)) will not synchronize and not wait at all, while if you create a `Condition` from a `ReentrantLock` then [`Condition.await(long, TimeUnit)`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Condition.html#await-long-java.util.concurrent.TimeUnit-) ([source](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/concurrent/locks/AbstractQueuedSynchronizer.java#L2153-L2161), `ConditionObject` created [here](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/concurrent/locks/ReentrantLock.java#L500)) will still yield the lock. It was more clearly documented [e.g. in JDK 11](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/concurrent/locks/AbstractQueuedSynchronizer.java#L2198-L2200) but comments disappeared in more recent versions.

rom1v:
@anotheruserofgithub Indeed, we should not call `wait(0)`:

> If `timeout` is zero, however, then real time is not taken into consideration and the thread simply waits until notified. 

Since this is on `dev` (not related to this PR), could you open a PR (so that the commit has your authorship)?

Even if it does not change anything in practice, I would not test `timeout <= 0` because it feels wrong to not read the value if `timeout` is 0. I'd prefer something like:

```diff
diff --git server/src/main/java/com/genymobile/scrcpy/control/Controller.java server/src/main/java/com/genymobile/scrcpy/control/Controller.java
index 5e64a4c5b..24d827fda 100644
--- server/src/main/java/com/genymobile/scrcpy/control/Controller.java
+++ server/src/main/java/com/genymobile/scrcpy/control/Controller.java
@@ -699,7 +699,9 @@ public class Controller implements AsyncProcessor, VirtualDisplayListener {
                 if (timeout < 0) {
                     return null;
                 }
-                displayDataAvailable.wait(timeout);
+                if (timeout > 0) {
+                    displayDataAvailable.wait(timeout);
+                }
                 data = displayData.get();
             }
 
```

anotheruserofgithub:
> could you open a PR (so that the commit has your authorship)?

I don't wish to have authorship, this is such a minor thing.

> I would not test `timeout <= 0` because it feels wrong to not read the value if `timeout` is 0.

You're right! So it's definitely your authorship. ;)

rom1v:
@anotheruserofgithub OK, thanks. Merged: 52f5d08d1fab9600e78b21c71fa4a6c106d3783f

yume-chan:
> I just noticed that [3b36f29](https://github.com/Genymobile/scrcpy/commit/3b36f29be94785ba9ddf96ff6afb259a0d4815a1) breaks copy-paste from the Android device to the computer: the callback passed to `clipboardManager.addPrimaryClipChangedListener()` is never called.

I see. `ClipboardManager` uses main looper to post the callback (https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/content/ClipboardManager.java;l=102;drc=61197364367c9e404c7da6900658f1b16c42d0da), so the main looper needs to be running. That's pretty annoying.

I can either

1. Manually run the main looper on a new thread like https://github.com/yume-chan/scrcpy/commit/1044101029a5db2529e4460b5425788390be8911 (looks little bit hacky)
2. Let the main thread run the main looper (as normal Android app would do), and run the main `scrcpy` method on another thread
3. While doing the first half of 2, make the main method event-based using the main looper, so it doesn't block on `completion.await()`
 
Still simpler than old `ClipboardService` wrapper I would say.

(While the main looper is running, it can also be used elsewhere so we don't need to create so many `HandlerThread`s, for example with camera: https://github.com/yume-chan/scrcpy/commit/08de0b679832819d6d7bf6339edca7ec6a696d02 (although camera might be a bad example because it shouldn't be blocked by other tasks))

rom1v:
> Let the main thread run the main looper […] make the main method event-based using the main looper

Yes, I think it would be a good solution (even for other Android features that might use the main looper).

Since this is a separate issue, I will probably merge this PR without the clipboard refactor, and this can be done afterwards.

rom1v:
Merged on `dev` (without the `ClipboardManager` refactor).

I will open a new PR with a looper.

PR #6017, opened 2025-04-18
Considering change licence to license for consistency

NicholasWilsonDEV:
Although "licence" is correct in British English, the rest of the statement uses "license," so it might be better to change it for consistency.


rom1v:
Thank you. Merged: c5ed2cfc28ee7c7b59b11eb4db1258ac1c633bff :+1:

(I never know which one to use, I'm always confused because in French it's "licence")

PR #6050, opened 2025-05-02
fix ubuntu version

Forensax:
ubuntu-20.04 is no longer available

See:
https://github.com/actions/runner-images/issues/11101


rom1v:
We should revert 69858c6f437b1bfece96bc291c607de842837d36 instead, to build on the latest available version.

PR #6107, opened 2025-05-22
Add app name SDL hint

ColinKinloch:
This allows pulseaudio to label the audio stream "scrcpy" rather than "SDL Application".

| before | after |
| - | - |
| Screenshot From 2025-05-22 18-55-09 | Screenshot From 2025-05-22 18-39-28 |


app/src/compat.h
```diff
@@ -75,6 +75,10 @@
 # define SCRCPY_SDL_HAS_THREAD_PRIORITY_TIME_CRITICAL
 #endif
 
+#if SDL_VERSION_ATLEAST(2, 0, 22)
```
rom1v, line 78: Shouldn't it be `(2, 0, 18)`?

https://github.com/libsdl-org/SDL/blob/d0c2d8bc40a90cc1a763f6cf5397e8c9958a33d8/WhatsNew.txt#L210

rom1v (commented):
Otherwise, LGTM.

app/src/compat.h
ColinKinloch: Apparently so.

rom1v:
Thank you. Merged: d2cc930975a8c7a2a073a9bffd9c2576d58cff7b

PR #6129, opened 2025-06-05
Run a main looper

rom1v:
In particular, this allows to use the public `ClipboardManager` (which simplifies the implementation).

This will probably be useful for other use cases  too.

cc @yume-chan https://github.com/Genymobile/scrcpy/pull/6009#issuecomment-2940810736

Maybe there's a proper way (without hacking `quitAllowed`).

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/wrappers/ClipboardManager.java
+++ b/server/src/main/java/com/genymobile/scrcpy/wrappers/ClipboardManager.java
@@ -1,270 +1,43 @@
 package com.genymobile.scrcpy.wrappers;
 
-import com.genymobile.scrcpy.AndroidVersions;
 import com.genymobile.scrcpy.FakeContext;
-import com.genymobile.scrcpy.util.Ln;
 
 import android.content.ClipData;
-import android.content.IOnPrimaryClipChangedListener;
-import android.os.Build;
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/genymobile/scrcpy/wrappers/ClipboardManager.java
```diff
@@ -1,270 +1,43 @@
...
-    private int setMethodVersion;
-    private int addListenerMethodVersion;
+    private final android.content.ClipboardManager manager;
 
     static ClipboardManager create() {
-        IInterface clipboard = ServiceManager.getService("clipboard", "android.content.IClipboard");
-        if (clipboard == null) {
+        android.content.ClipboardManager manager = (android.content.ClipboardManager) FakeContext.get()
+                .getSystemService(FakeContext.CLIPBOARD_SERVICE);
+        if (manager == null) {
```
yume-chan, line 13: `ClipboardManager`'s constructor will throw when `ClipboardService` is not found: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/content/ClipboardManager.java;l=134-135;drc=16f7cdf2e01e158efe8d8334a15e9e20e59de929

Issue #5984, opened 2025-04-08
How to unlock my Android phone from SCRCPY?


luiz-eduardo-ramirez:
I have a A52s Android phone, I have ALS desease, I am trying to control my phone on my laptop using TOBII EYE TRACKER 5.

I cant move my hands so it is impossible for me to unlock my phone from SCRCPY, but on pin screen I am getting black screen on the application.

Help me please, thank you in advance

rugulous:
This works for my phone, hopefully it works the same for you. I couldn't find anything in the documentation to allow automatically unlocking (or sending arbitrary commands), but if you start `scrcpy` and then open a new terminal window, the following commands should work?

```
adb shell input keyevent 82
adb shell input touchscreen swipe 530 1420 530 1120
adb shell input text <your pin>
```

If you have to press enter after entering your pin then you'll also need to run `adb shell input keyevent 66`

SpeedOfSpin:
This worked for me
adb shell input keyevent 26
adb shell input swipe 500 1800 500 300
adb shell input text <your pin>
adb shell input keyevent 66

psython123:
> This worked for me adb shell input keyevent 26 adb shell input swipe 500 1800 500 300 adb shell input text adb shell input keyevent 66

This works for me too, but howd i do it in case of an in-app login?

PR #6159, opened 2025-06-15
Add session metadata for video stream

rom1v:
Introduce a new packet type, a "session" packet, containing metadata about the encoding session. It is used only for the video stream, and currently includes the video resolution.

For illustration, here is a sequence of packets on the video stream:

                                        device rotation
                                        v
    CODEC | SESSION | MEDIA | MEDIA | … | SESSION | MEDIA | MEDIA | …
           1920x1080 <-----------------> 1080x1920 <------------------
                      encoding session 1            encoding session 2

This metadata is not strictly necessary, since the video resolution can be determined after decoding. However, it allows detection of cases where the encoder does not respect the requested size (and logs a warning), even without decoding (e.g., when there is no video playback).

Additional metadata could be added later if necessary, for example the actual device rotation.
    
Refs #5918
Refs #5984

### DIFF
--- a/app/src/audio_player.c
+++ b/app/src/audio_player.c
@@ -30,7 +30,10 @@ sc_audio_player_frame_sink_push(struct sc_frame_sink *sink,
 
 static bool
 sc_audio_player_frame_sink_open(struct sc_frame_sink *sink,
-                                const AVCodecContext *ctx) {
+                                const AVCodecContext *ctx,
+                                const struct sc_stream_session *session) {
+    (void) session;
+
     struct sc_audio_player *ap = DOWNCAST(sink);
 
 #ifdef SCRCPY_LAVU_HAS_CHLAYOUT
--- a/app/src/decoder.c
+++ b/app/src/decoder.c
@@ -10,14 +10,15 @@
 #define DOWNCAST(SINK) container_of(SINK, struct sc_decoder, packet_sink)
 
 static bool
-sc_decoder_open(struct sc_decoder *decoder, AVCodecContext *ctx) {
+sc_decoder_open(struct sc_decoder *decoder, AVCodecContext *ctx,
+                const struct sc_stream_session *session) {
     decoder->frame = av_frame_alloc();
     if (!decoder->frame) {
         LOG_OOM();
         return false;
     }
 
-    if (!sc_frame_source_sinks_open(&decoder->frame_source, ctx)) {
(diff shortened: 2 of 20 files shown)


sbfkcel:
To make the session data more compact, additional metadata is reserved. Would it be better to put the flag and width together in the first 4 bytes?
```
    public void writeSessionMeta(int width, int height) throws IOException {
        writeSessionMeta(width, height, 0);
    }

    public void writeSessionMeta(int width, int height, int extra) throws IOException {
        if (sendStreamMeta) {
            headerBuffer.clear();

            int flagAndWidth = (int)(PACKET_FLAG_SESSION | (width & ((1 << 31) - 1)));
            headerBuffer.putInt(flagAndWidth);
            headerBuffer.putInt(height);
            headerBuffer.putInt(extra);
            headerBuffer.flip();
            IO.writeFully(fd, headerBuffer);
        }
    }

```
https://github.com/Genymobile/scrcpy/blob/02047ff102ac50b5823aa6cefb5b5087a11d4834/server/src/main/java/com/genymobile/scrcpy/device/Streamer.java#L92-L102

rom1v:
That way, the packet header is always 12 bytes: https://github.com/Genymobile/scrcpy/blob/e5e58b1b307d92cbe3432431a9e22cd648f8d4d1/app/src/demuxer.c#L103

If it was variable, it would require the client to buffer the data (more copies) and "packetize" (more complexity). As a tradeoff, it uses 2 syscalls (one to read the header, one to read the full payload).

sbfkcel:
Sorry, I may not have explained it clearly enough. The Session Meta data is indeed always 12 bytes. However, since the flag, width, and height each occupy 4 bytes, there’s no remaining space for expansion.

Therefore, my proposal is to combine the flag and width into 4 bytes, keep height as 4 bytes, and reserve the remaining 4 bytes for future expansion.

rom1v:
Note that there is no backward or forward compatibility (the protocol is always used between matching client and server), so there is nothing to "reserve", the protocol can be changed completely for any version.

If we want to add something taking 31 bits or less, you can store it just after the flag. If we need a lot more data, we can add a payload. We can also move the width/height fields as needed.

rom1v:
Now that #6216 is merged, the next scrcpy version will be 4.0.

This is a good time to merge this protocol change.

Issue #3800, opened 2023-03-12
Add H264/H265/AV1 GPU accelerated video decode support


Ciancy28:
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
Currently, scrcpy doesen't use the GPU's hardware accelerated video decoding capabilities, which increases resource usage/decode delay, this doesen't impact most modern PCs much, but on systems with weaker processors (especially with AV1), it should make a big difference, as far as I know, there is no downside to using video decoding acceleration, and most applications already use it as expected (web browsers, telegram/whatsapp, video editing software, every media player in existance... scrcpy is the only application I've ever found that doesen't take advantage of this).   

**Describe the solution you'd like**
Hardware accelerated video decode should be either an option, or enabled by default and togglable with a parameter.   

**Describe alternatives you've considered**

**Additional context**
![H264 scrcpy](https://user-images.githubusercontent.com/77066322/224558842-d505a786-22fc-4d73-be95-883a1cf9ccb0.png)
![H264 movie](https://user-images.githubusercontent.com/77066322/224558646-8a52e5df-96cf-4620-8725-9bf47acddc90.png)
![H265 scrcpy](https://user-images.githubusercontent.com/77066322/224558651-51992ece-b405-4607-bfa0-dfee18dbf1e1.png)
![H265 movie](https://user-images.githubusercontent.com/77066322/224558932-eff6ab1f-96a4-43a2-a1fc-61a7e4d20ef7.png)

I observed the lack of gpu video decoding capabilities on a Windows 11 system with an AMD RX 580 GPU running scrcpy v2.0, I'm not sure if hardware decoding is limited to Linux or Nvidia/Intel GPUs, I couldn't find anything regarding this subject in the documentation, so I think it's missing on every system.
I couldn't test the presence of AV1 video decoding capabilities as I don't have an Android device with an AV1 encoder and I don't have a GPU with an AV1 decoder.

rom1v:
Yes, that would be great. But that's not trivial to keep the decoded image on the GPU until display for all video output. On VLC, we implement interops manually for OpenGL, DirectX, etc. I have not the resource to develop and maintain all video outputs manually. I would like SDL to support it: https://github.com/libsdl-org/SDL/issues/5405

[...]

Issue #3460, opened 2022-08-22
Option to lock aspect ratio of window to screen


rp1231:
**Is your feature request related to a problem? Please describe.**
I dont like it when there are extra black borders around the screen area and I think it would be nice to have a mode in which you can lock the aspect ratio of the screen to the phone screen being displayed.

rom1v:
I tried to force the other dimension on resize with SDL, but it didn't work as expected.

Note that you can remove the black bars afterwards either with Alt+w or by double-clicking inside the bars.

In theory, keeping the aspect ratio while resizing is possible (mpv supports it), but not via SDL.

rom1v:
@rp1231 https://github.com/Genymobile/scrcpy/issues/2317#issuecomment-841692150

burtvadr:
* * idea. something like ` --lock-window-aspect-ratio ` sound the most predictable option to me but other options could be

* `--window-aspect-ratio=lock`
* `--lock-window-size-ratio`
* `--keep-window-aspect-ratio`
* `--force-window-aspect-ratio`

Issue #55, opened 2018-03-12
Screen is skewed incredibly far to the side


nonnymoose:
![screenshot from 2018-03-12 14-45-32](https://user-images.githubusercontent.com/19860432/37305998-4c7c8688-2604-11e8-8e6e-2a3d948665fa.png)

This is the settings screen.

This is really crazy! What can I do to provide some more useful information? (Given that this issue is probably not easily reproduced :)

rom1v:
It seems similar to https://github.com/Genymobile/scrcpy/issues/33. I have no clue.

rom1v:
Could you try to record your screen using `screenrecord`:

    adb screenrecord /sdcard/file.mp4

And pull the file:

    adb pull /sdcard/file.mp4

Is the video ok?

nonnymoose:
That works fine, although it does give a message that is unclear whether it's a warning or just info:

    The max width/height supported by codec is 1280x720

[mtp.mp4.gz](https://github.com/Genymobile/scrcpy/files/1805131/mtp.mp4.gz)

nonnymoose:
The only thing I can think of that's similar between the two is we both have MTK devices.

rom1v:
Could you try with a lower size?

    scrcpy -m 800

nonnymoose:
Sorry for the delay. It still exhibits the same problem if I use a lower resolution:
![screenshot from 2018-03-26 17-25-18](https://user-images.githubusercontent.com/19860432/37936332-c360c5d6-311a-11e8-893f-0deb7e494bea.png)

rom1v:
Could you please execute this to list video codec details on your device: [`CheckVideoCodecs`](https://github.com/rom1v/CheckVideoCodecs)?

rom1v:
@nonnymoose Also, on [this line](https://github.com/Genymobile/scrcpy/blob/82efff34e8bcb9d5c4089dbee7f58af965a61778/app/src/screen.c#L140), please try replacing `SDL_PIXELFORMAT_YV12` by a [different value](https://wiki.libsdl.org/SDL_CreateTexture#Remarks). Especially:
 - `SDL_PIXELFORMAT_IYUV`
 - `SDL_PIXELFORMAT_YUY2`
 - `SDL_PIXELFORMAT_UYVY`
 - `SDL_PIXELFORMAT_YVYU`
 - `SDL_PIXELFORMAT_NV12`
 - `SDL_PIXELFORMAT_NV21`

[...]

PR #6216, opened 2025-07-11
Migrate from SDL2 to SDL3

rom1v:
First, this PR contains some fixes that could be merged directly into `dev`.

Then, there are some refactors to make the transition easier, especially the handling of integer scrolling (https://github.com/Genymobile/scrcpy/issues/6156#issuecomment-3016836940) and the configuration of the color space at texture creation.

Finally, there's a "big" migration commit, followed by the changes for the build script and the GitHub Actions script.

Links:
 - SDL3 migration docs: https://wiki.libsdl.org/SDL3/README-migration
 - Relevant discussion about the new SDL3 audio API for scrcpy: https://github.com/libsdl-org/SDL/issues/13319
 - Release binaries from GitHub Actions (download `scrcpy-release-sdl3`): https://github.com/rom1v/scrcpy/actions/runs/16226120027

----

Tests and reviews welcome.

Remaining questions:
 - check if the "continuous resizing" workaround is still necessary, or if it can be removed (0b1e59186f1fb75d74f2929f636128958b067ab4) (answer: :x: no https://github.com/Genymobile/scrcpy/pull/6216#issuecomment-3094411968)
 - is it possible with SDL3 to keep the window aspect ratio while resizing (instead of adding black bars)? (#3460)

---

I don't plan to merge this immediately, SDL3 is not available yet on many distros (not even in current Debian stable).

I'd like to implement hardware decoding first (#3800 | https://github.com/libsdl-org/SDL/issues/5405#issuecomment-1754956025 | https://github.com/libsdl-org/SDL/pull/8366), and if everything works as expected, release a new major version (scrcpy 4.0).

EDIT: hardware decoding will not be for 4.0. Flex display was implemented instead (#6772).


yume-chan:
Quickly tried it on Windows (using the prebuilt binary) and it keeps printing `Unexpected SDL audio behavior: too much data requested`

```txt
┖[~\Downloads\scrcpy-win64-sdl3]> .\scrcpy.exe
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
C:\Users\simon\Downloads\scrcpy-win64-sdl3\scrcpy-server: ...file pushed, 0 skipped. 112.9 MB/s (90788 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
INFO: Renderer: direct3d11
INFO: Texture: 1440x3200
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
```

Enabling `-Vverbose` shows more `[Audio] Buffering threshold exceeded` messages

```txt
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 240 samples
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 240 samples
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: Unexpected SDL audio behavior: too much data requested
ERROR: Unexpected SDL audio behavior: too much data requested
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
DEBUG: [Audio] Buffering threshold exceeded, skipping 960 samples
```

rom1v:
Thank you for your tests!

I assumed (on purpose, to detect early if that was false) that the `additional_amount` requested by SDL3 never exceeds the "audio output buffer" hint. It is true on my machine, but apparently not everywhere (I only tested on Linux).

How much data does it require:

```diff
diff --git app/src/audio_player.c app/src/audio_player.c
index e1a4134e7..6ca13a1e0 100644
--- app/src/audio_player.c
+++ app/src/audio_player.c
@@ -21,7 +21,8 @@ sc_audio_player_stream_callback(void *userdata, SDL_AudioStream *stream,
         assert(len <= ap->aout_buffer_size);
         if (len > ap->aout_buffer_size) {
             // Just in case for release builds
-            LOGE("Unexpected SDL audio behavior: too much data requested");
+            LOGE("Unexpected SDL audio behavior: too much data requested: %d",
+                 additional_amount);
             len = ap->aout_buffer_size;
         }
 
```

?

Is it different with `--audio-output-buffer=10` or `--audio-output-buffer=20`? (default is 5)

On SDL2, do you need to use `--audio-output-buffer=xx` to get audio working correctly?

yume-chan:
> How much data does it require:

`additional_amount` is always `3840`, and audio sounds like a robot

> Is it different with `--audio-output-buffer=10` or `--audio-output-buffer=20`? 

There is no error with either `--audio-output-buffer=10` or `--audio-output-buffer=20`, `additional_amount` is still `3840`, and audio is working

> On SDL2, do you need to use --audio-output-buffer=xx to get audio working correctly?

No

rom1v:
> There is no error with either `--audio-output-buffer=10` or `--audio-output-buffer=20`, and audio is working

What is the value of `additional_amount` in these cases?

@icculus Sorry to ping you here, but do you know why setting `SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES` to 240 (5 milliseconds of samples, 1920 bytes for 2 channels of 4 bytes each) can result in the callback being called with `additional_amount = 3840` (instead of 1920) on Windows?

Apparently, on SDL2, the callback was correctly called with 1920 bytes. @yume-chan, this can be confirmed on scrcpy 3.3.1 with:

```diff
diff --git app/src/audio_player.c app/src/audio_player.c
index 9413c2ea8..07979b31b 100644
--- app/src/audio_player.c
+++ app/src/audio_player.c
@@ -11,6 +11,8 @@ static void SDLCALL
 sc_audio_player_sdl_callback(void *userdata, uint8_t *stream, int len_int) {
     struct sc_audio_player *ap = userdata;
 
+    LOGI("=== SDL2 audio callback: %d bytes (%d samples)", len_int, len_int / 8);
+
     assert(len_int > 0);
     size_t len = len_int;
 
```

icculus:
Not awake yet, but I'll try the Windows build today and see what happens.

metayan:
`scrcpy` with `SDL3` can even be built on macOS Catalina with MacOSX_SDK11 using
```
SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk ./release/build_macos.sh x86_64
```
and it works splendidly, both with OpenGL and Metal.
I have an older Android phone, so can't test audio.

app/deps/sdl.sh
```diff
@@ -35,45 +35,48 @@ else
...
-        --prefix="$INSTALL_DIR/$DIRNAME"
+        -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR/$DIRNAME"
     )
 
     if [[ "$HOST" == linux ]]
     then
         conf+=(
-            --enable-video-wayland
-            --enable-video-x11
+            -DSDL_WAYLOAD=ON
```
metayan, line 44: You probably mean SDL_WAYLAND?

metayan:
```diff
diff --git a/app/deps/sdl.sh b/app/deps/sdl.sh
index ae79ae9c..a308ae9e 100755
--- a/app/deps/sdl.sh
+++ b/app/deps/sdl.sh
@@ -28,7 +28,7 @@ export CXXFLAGS="$CFLAGS"
 
 if [[ -d "$DIRNAME" ]]
 then
-    echo "'$PWD/$HDIRNAME' already exists, not reconfigured"
+    echo "'$PWD/$DIRNAME' already exists, not reconfigured"
     cd "$DIRNAME"
 else
     mkdir "$DIRNAME"
@@ -41,7 +41,7 @@ else
     if [[ "$HOST" == linux ]]
     then
         conf+=(
-            -DSDL_WAYLOAD=ON
+            -DSDL_WAYLAND=ON
             -DSDL_X11=ON
         )
     fi
```

metayan:
Also tested with SDL 3.2.18 and it works fine on macOS. Audio not tested.
```diff
diff --git a/app/deps/sdl.sh b/app/deps/sdl.sh
index ae79ae9c..970fcbfa 100755
--- a/app/deps/sdl.sh
+++ b/app/deps/sdl.sh
@@ -5,10 +5,10 @@ cd "$DEPS_DIR"
 . common
 process_args "$@"
 
-VERSION=3.2.16
+VERSION=3.2.18
 FILENAME=SDL-$VERSION.tar.gz
 PROJECT_DIR=SDL-release-$VERSION
-SHA256SUM=a1af401fb49b824ba721683862439e014498f79753602827e174f3613357aff0
+SHA256SUM=51539fa13e546bc50c632beed3f34257de2baa38a4c642048de56377903b4265
 
 cd "$SOURCES_DIR"
```

yume-chan:
> What is the value of `additional_amount` in these cases?

It's still `3840`

```diff
diff --git a/app/src/audio_player.c b/app/src/audio_player.c
index e1a4134e..71fbef2a 100644
--- a/app/src/audio_player.c
+++ b/app/src/audio_player.c
@@ -16,6 +16,8 @@ sc_audio_player_stream_callback(void *userdata, SDL_AudioStream *stream,
     struct sc_audio_player *ap = userdata;
 
     if (additional_amount > 0) {
+        LOGE("additional_amount: %d", additional_amount);
+
         size_t len = additional_amount;
 
         assert(len <= ap->aout_buffer_size);
@@ -82,6 +84,9 @@ sc_audio_player_frame_sink_open(struct sc_frame_sink *sink,
     int r = snprintf(str, sizeof(str), "%" PRIu16, (uint16_t) aout_samples);
     assert(r >= 0 && (size_t) r < sizeof(str));
     (void) r;
+
+    LOGE("setting SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES to %s", str);
+
     if (!SDL_SetHint(SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES, str)) {
         LOGE("Could not set audio output buffer");
         sc_audio_regulator_destroy(&ap->audioreg);
```

```txt
> ./scrcpy --audio-output-buffer=10
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\o...file pushed, 0 skipped. 102.8 MB/s (91004 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
INFO: Renderer: direct3d11
ERROR: setting SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES to 480
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
```

```txt
> ./scrcpy --audio-output-buffer=20
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\o...file pushed, 0 skipped. 107.7 MB/s (91004 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
INFO: Renderer: direct3d11
ERROR: setting SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES to 960
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
INFO: Texture: 1440x3200
ERROR: additional_amount: 3840
ERROR: additional_amount: 3840
```

---

> this can be confirmed on scrcpy 3.3.1

```txt
> ./scrcpy
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\o... file pushed, 0 skipped. 99.5 MB/s (91004 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
INFO: Renderer: direct3d
INFO: Texture: 1440x3200
ERROR: === SDL2 audio callback: 1920 bytes (240 samples)
ERROR: === SDL2 audio callback: 1920 bytes (240 samples)
ERROR: === SDL2 audio callback: 1920 bytes (240 samples)
ERROR: === SDL2 audio callback: 1920 bytes (240 samples)
ERROR: === SDL2 audio callback: 1920 bytes (240 samples)
```

```txt
> ./scrcpy --audio-output-buffer=10
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\o...file pushed, 0 skipped. 104.3 MB/s (91004 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
INFO: Renderer: direct3d
INFO: Texture: 1440x3200
ERROR: === SDL2 audio callback: 3840 bytes (480 samples)
ERROR: === SDL2 audio callback: 3840 bytes (480 samples)
ERROR: === SDL2 audio callback: 3840 bytes (480 samples)
ERROR: === SDL2 audio callback: 3840 bytes (480 samples)
ERROR: === SDL2 audio callback: 3840 bytes (480 samples)
ERROR: === SDL2 audio callback: 3840 bytes (480 samples)
```

```txt
> ./scrcpy --audio-output-buffer=20
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\o...file pushed, 0 skipped. 103.7 MB/s (91004 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
INFO: Renderer: direct3d
INFO: Texture: 1440x3200
ERROR: === SDL2 audio callback: 7680 bytes (960 samples)
ERROR: === SDL2 audio callback: 7680 bytes (960 samples)
ERROR: === SDL2 audio callback: 7680 bytes (960 samples)
ERROR: === SDL2 audio callback: 7680 bytes (960 samples)
ERROR: === SDL2 audio callback: 7680 bytes (960 samples)
```

app/deps/sdl.sh
rom1v: Good catch :+1:

Indeed, in the logs:

```
 CMake Warning:
-- Generating done (0.1s)
  Manually-specified variables were not used by the project:

    SDL_WAYLOAD
```

(it should be an error instead of a warning)

metayan:
Numbers and punctuation now only work with `--raw-key-events`.

yume-chan:
https://github.com/libsdl-org/SDL/blob/bc5c9a686ca01c7cffa45e223efcd0cb7ef1efb7/src/audio/wasapi/SDL_wasapi.c#L759-L760

SDL3 calls [IAudioClient3::GetSharedModeEnginePeriod](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient3-getsharedmodeengineperiod) and clamps `device->sample_frames`. `min_period_in_frames` and `max_period_in_frames` are both `480`, so `additional_amount` is always `3840`. Not sure what it means, nor how does SDL2 works.

---

`directsound` audio driver doesn't work with the default `--audio-output-buffer=5` neither, `additional_amount` is `1920` as expected, but audio played is broken:

<details>

<summary>log</summary>

```txt
> ./scrcpy --audio-output-buffer=5
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  73294bba                        device  M2011K2C
DEBUG: Device serial: 73294bba
DEBUG: Using SCRCPY_SERVER_PATH: D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\outputs\apk\release\server-release-unsigned.apk
D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\server\build\o...file pushed, 0 skipped. 104.2 MB/s (91004 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi M2011K2C (Android 13)
[server] DEBUG: Using audio encoder: 'c2.android.opus.encoder'
[server] DEBUG: Using video encoder: 'c2.qti.avc.encoder'
[server] DEBUG: Display: using SurfaceControl API
DEBUG: Server connected
DEBUG: Starting controller thread
DEBUG: Starting receiver thread
DEBUG: Using icon (portable): D:\dev\yume-chan\scrcpy-devcontainer\scrcpy\release\work\build-win64\dist\icon.png
INFO: Renderer: direct3d11
DEBUG: Trilinear filtering disabled (not an OpenGL renderer)
DEBUG: Demuxer 'video': starting thread
DEBUG: Demuxer 'audio': starting thread
INFO: Texture: 1440x3200
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
DEBUG: [Audio] Buffering threshold exceeded, skipping 240 samples
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
DEBUG: [Audio] Buffering threshold exceeded, skipping 480 samples
ERROR: additional_amount: 1920
ERROR: additional_amount: 1920
DEBUG: [Audio] Buffering threshold exceeded, skipping 240 samples
```

</details>

rom1v:
> Numbers and punctuation now only work with `--raw-key-events`.

Oh, `SDL_EVENT_TEXT_INPUT` is never received: https://github.com/Genymobile/scrcpy/blob/3615c75ad6074f088cddb0616410022ce7c0249b/app/src/input_manager.c#L1019

Reported here: https://github.com/libsdl-org/SDL/issues/13380

rom1v:
> `directsound` audio driver doesn't work with the default `--audio-output-buffer=5` neither, `additional_amount` is 1920 as expected, but audio played is broken

Is it also the case on SDL2?

`--audio-output-buffer` was added because some audio outputs on Windows did not support 5ms, so if the behavior was the same on SDL2 (and respects the `sample_frames` value), it's ok I guess: https://github.com/Genymobile/scrcpy/issues/3793

rom1v:
> Oh, SDL_EVENT_TEXT_INPUT is never received

Should be fixed on the new branch.

I added:

```diff
diff --git app/src/screen.c app/src/screen.c
index 27259e830..6ecb64167 100644
--- app/src/screen.c
+++ app/src/screen.c
@@ -394,6 +394,12 @@ sc_screen_init(struct sc_screen *screen,
         goto error_destroy_fps_counter;
     }
 
+    ok = SDL_StartTextInput(screen->window);
+    if (!ok) {
+        LOGE("Could not enable text input: %s", SDL_GetError());
+        goto error_destroy_window;
+    }
+
     SDL_Surface *icon = scrcpy_icon_load();
     if (icon) {
         SDL_SetWindowIcon(screen->window, icon);
```

(cf https://github.com/libsdl-org/SDL/issues/13380)

As a consequence, it might open an IME on scrcpy start (https://github.com/libsdl-org/SDL/issues/13380#issuecomment-3083668746). If that's the case (to be tested), we need a workaround.

yume-chan:
> > `directsound` audio driver doesn't work with the default `--audio-output-buffer=5` neither, `additional_amount` is 1920 as expected, but audio played is broken
> 
> Is it also the case on SDL2?

Yes, with SDL2, directsound driver, and `--audio-output-buffer=5`, audio playback is also broken. `wasapi` driver works.

yume-chan:
> check if the "continuous resizing" workaround is still necessary, or if it can be removed (https://github.com/Genymobile/scrcpy/commit/0b1e59186f1fb75d74f2929f636128958b067ab4)

It still doesn't work, video stream pauses when moving or resizing window. 

I think https://github.com/libsdl-org/SDL/issues/1059#issuecomment-1802723797 only applies automatically when using [main callbacks](https://wiki.libsdl.org/SDL3/README-main-functions#how-to-use-main-callbacks-in-sdl3), otherwise the `SC_EVENT_NEW_FRAME ` custom events are still not delivered (https://github.com/Genymobile/scrcpy/issues/3458#issuecomment-2768988367)

rom1v:
Thank you :+1:

Maybe it can make sense to use SDL main callbacks (to be tested). This will require to rework the way the scrcpy-server is started, but in the end in might simplify (or not).

rom1v:
@yume-chan

> === SDL2 audio callback: 1920 bytes (240 samples)

Could you please also add the timestamp of the call, to see if the callback is called twice in a row: https://github.com/libsdl-org/SDL/issues/13397

szmarczak:
> I don't plan to merge this immediately, SDL3 is not available yet on many distros (not even in current Debian stable).

Why not use a statically linked lib? Unless it's lots of megabytes in size but I'd still accept the tradeoff for better color accuracy.

rom1v:
> Why not use a statically linked lib

This is one of the possibilities: https://github.com/Genymobile/scrcpy/blob/master/doc/linux.md#from-the-official-release

But I would like to avoid a dependency "too recent", which is not even packaged (Debian stable is my reference, if it's in stable, it's not "too recent").

Anyway, it will take some time to resolve the remaining issues with SDL3.

adamponi:
Hi,

I compiled the SDL3 branch in Termux, and it works well with the default settings. However, when using the --render=vulkan option, the colors appear inverted, although scrcpy remains usable.

In contrast, the [precompiled build linked in the first post](https://github.com/rom1v/scrcpy/actions/runs/16226120027), which I ran on Windows, performs much worse. When launched with the --render=vulkan parameter, the screen displays only glitches, and the image is completely unreadable.
```
Downloads\scrcpy-release-sdl3\scrcpy-win64-sdl3> ./scrcpy --render=vulkan --no-audio
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  ec8faaa3                        device  xxxxxxxxxxx
C:\Users\xxxxxxx\Downloads\scrcpy-release-sdl3\scrcpy-... file pushed, 0 skipped. 27.7 MB/s (90788 bytes in 0.003s)
[server] INFO: Device: [Xiaomi] Redmi xxxxxxxxxxxxxxxxxx (Android 13)
INFO: Renderer: vulkan
INFO: Texture: 1224x2712
```

image


If there’s anything I can do to help diagnose or resolve this issue, please let me know. I’m happy to provide logs or any other information needed.

For reference, on Windows 11 I have the latest graphics drivers installed for the Intel® Core™ Ultra 5 Processor 135U.

Thanks!

rom1v:
Looks like there is an alignment issue. Maybe because in 1224x2712, 1224 and 2712 are not multiple of 16. (see #55)
Try `-m1024` or something else until you get both dimensions multiple of 16.

Can you reproduce with scrcpy + SDL2?

adamponi:
I can't reproduce this using SDL2 because the version built on github doesn't offer a Vulkan renderer.

adamponi:
```
PS C:\Users\xxxxxxxxxxxx\Downloads\scrcpy-release-sdl3\scrcpy-win64-sdl3> ./scrcpy --render=vulkan --no-audio -m1024
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  ec8faaa3                        device  xxxxxxxxxxxx
C:\Users\xxxxxxxxxxxx\Downloads\scrcpy-release-sdl3\scrcpy-... file pushed, 0 skipped. 28.7 MB/s (90788 bytes in 0.003s)
[server] INFO: Device: [Xiaomi] Redmi xxxxxxxxxxxx (Android 13)
INFO: Renderer: vulkan
INFO: Texture: 464x1024
```

image

This parameter didn't help. The problem is the same. I tested several other renderers, and the ones I tested were:
-direct3d12
-direct3d11
-opengl

The problem doesn't occur. The problem only occurs with the Vulkan renderer.

adamponi:
I also just tested the "direct3d" renderer and everything works fine there. The problem is only with the "vulkan" renderer.

rom1v:
Indeed, I can reproduce on Linux. Even with dimensions which are a multiple of 16 (or even 32).

Thank you for you report.

I created an issue here: https://github.com/libsdl-org/SDL/issues/13734

DienoX:
Thank you very much for your commitment and project.
I have a question about SDL3.
Does it bring any other benefits besides hardware acceleration of video decoding and DirectX 11, 12, Vulkan rendering?

DienoX:
I'm asking in the context of scrcpy

metayan:
Paste doesn't work on macOS.
Tried SDL 3.2.16 3.2.18 3.2.20.

SDL issue filed:
https://github.com/libsdl-org/SDL/issues/13737

Copy works and sometimes Control-Shift-V works.

rom1v:
I updated the PR to not assume that the SDL SAMPLES_FRAMES hint is honored (refs https://github.com/libsdl-org/SDL/issues/13397). @yume-chan this should fix https://github.com/Genymobile/scrcpy/pull/6216#issuecomment-3072768615.

@adamponi The vulkan driver issue is fixed: https://github.com/libsdl-org/SDL/issues/13734

> Does it bring any other benefits besides hardware acceleration of video decoding and DirectX 11, 12, Vulkan rendering?

Besides hardware acceleration (which requires some work), the main benefit is to use a version of SDL that will continue to evolve (not become abandoned/deprecated).

rom1v:
I added some changes on this branch.

Firstly, I removed a workaround for macOS (e59d40ebc7507a2eee75a1ea6c350768b4f713b5), because the underlying issue should be fixed in SDL3 (https://github.com/libsdl-org/SDL/issues/5340, https://github.com/libsdl-org/SDL/issues/5976). Please verify on macOS.

Then, I noticed a very annoying change from SDL2 to SDL3: many previously infallible functions have been made fallible without good reason (in my opinion). See the discussion in https://github.com/libsdl-org/SDL/issues/14223 for details.

For example, `SDL_GetWindowSize()`, which returned `void` in SDL2, now returns `bool` in SDL3. In general, it's a good thing that a function reports an error when an error can occur. However, in these changes, the functions only report programming errors, basically when SDL is not initialized or an invalid argument is passed (at least for cases they can detect):

https://github.com/libsdl-org/SDL/blob/5dab2c73f0537bd121d2311b6949a654c5692af9/src/video/SDL_video.c#L3203-L3213
https://github.com/libsdl-org/SDL/blob/5dab2c73f0537bd121d2311b6949a654c5692af9/src/video/SDL_video.c#L159-L178

Handling such errors would have cascading consequences in scrcpy. For instance, if a function calls `SDL_GetWindowSize()`, it should report the error to its caller, which would also have be made fallible, and so on… adding a lot of complexity for no reason, since in practice, `SDL_GetWindowSize()` cannot fail except due to programming errors.

The SDL maintainers disagree with the statement that "these functions only fails if the object is invalid", and suggest that these functions may fail for other reasons in the future.

In practice, I still assume that they may not fail, and that they never will (at least for `SDL_GetWindowXXX()`):
 - The current implementation cannot fail when the functions are called correctly.
 - If they add genuine failure cases in the future, every call to `SDL_GetWindowSize()` that doesn't check the error in a project would effectively be a bug (leading to undefined behavior if the output variables are not initialized beforehand).
 - The migration script from SDL2 to SDL3 does not add these error checks, so this would basically introduce undefined behavior in every migrated project.

Therefore, I added a wrapper in scrcpy that check the reported values, log them when failing is not fatal, and `abort()` (think "assertion enabled in release mode") to avoid undefined behavior for window size and position getters: a9a8fb190cb8dd9b2db1021c9bf0fbe3cf2a78f4.

Error handling could be added later if my assumptions prove to be incorrect.

What do you think?

EDIT: see #6802

metayan:
Tested on macOS and the only thing not working here is still the pasting.
But I'm using macOS Catalina (10.15.7), so not sure if that's causing the issue.
I have reported it to SDL, but not investigated thoroughly.

Your solution to the new SDL error reporting looks reasonable. Nice.

rom1v:
I think it's going to be time to migrate to SDL3. I plan to merge this branch into `dev` soon.

If you want to test, a release of this branch is available here (built by Github Actions): https://github.com/rom1v/scrcpy/actions/runs/20572157206

app/src/icon.c
```diff
@@ -248,17 +236,35 @@ load_from_path(const char *path) {
...
+    if (!props) {
+        LOGE("Could not get surface properties: %s", SDL_GetError());
+        SDL_DestroySurface(surface);
+        goto error;
+    }
+
+    // frame owns the data
+    bool ok = SDL_SetPointerProperty(props, "sc_frame", frame);
+    if (!ok) {
+        LOGE("Could not get surface properties: %s", SDL_GetError());
```
anotheruserofgithub:
```suggestion
        LOGE("Could not set pointer property: %s", SDL_GetError());
```

app/src/input_manager.c
```diff
@@ -911,31 +906,31 @@ sc_input_manager_process_mouse_wheel(struct sc_input_manager *im,
...
         }
 
-        SDL_Joystick *joystick = SDL_GameControllerGetJoystick(gc);
+        SDL_Joystick *joystick = SDL_GetGamepadJoystick(gc);
         if (!joystick) {
             LOGW("Could not get controller joystick");
-            SDL_GameControllerClose(gc);
+            SDL_CloseGamepad(gc);
             return;
         }
```
anotheruserofgithub: Same things as in previous comments: reword "game controller" and "controller" to "gamepad" and find a better name for the `gc` pointer.

anotheruserofgithub (commented):
Hi, thanks for your great work! Here are a few things I spotted in commit c7a4717d9862e3cd2f4bd38c25341eb4b442c79b (the others are neat to me), though I didn't check the lines that aren't modified.

Unfortunately I don't have any device to test. :( Sorry! Only got a very old phone and USB debugging isn't detected apparently (the USB socket is in a very bad state and is barely capable of charging the battery, might be the issue).

anotheruserofgithub (commented):
Comments, especially in file `input_events.h` (but also in other files, see https://github.com/Genymobile/scrcpy/pull/6216#discussion_r2651778910), should be reviewed too.

I'm referring to lines in [`app/src/input_events.h`](https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/app/src/input_events.h) on the current master branch (commit db2f72a58aaedfe3384b0dea838175e9237a3324):
- L17 there is a `sc_gamepad_processor` too.
- L326 the migration guide has been moved to https://wiki.libsdl.org/SDL3/README-migration (hyphen instead of slash). But maybe you should now point to the actual documentation / wiki page? No longer need to refer to the SDL2 -> SDL3 transition.
- L416 wrongly points to SDL2 wiki.
- L505 should be `SDL_GAMEPAD_AXIS_*`.
- L515 should be `SDL_GAMEPAD_BUTTON_*`.

Additionally in [`app/src/mouse_capture.c`](https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/app/src/mouse_capture.c#L63):
- L63 the event new name is `SDL_EVENT_MOUSE_BUTTON_UP`.

Also links to SDL2 header files should be changed to SDL3, here in [`app/src/hid/hid_keyboard.h`](https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/app/src/hid/hid_keyboard.h#L12): https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/app/src/hid/hid_keyboard.h#L12

And library versions should be adapted (**two** occurrences in each block): https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/doc/linux.md?plain=1#L42-L43 https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/.github/workflows/release.yml#L78-L79 https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/.github/workflows/release.yml#L104-L105 https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/.github/workflows/release.yml#L135-L136 https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/.github/workflows/release.yml#L166-L167

I may have missed other details.

rom1v:
@anotheruserofgithub Thank you very much for your review. :+1:

I fixed most of them. I will investigate the remaining ones (especially https://github.com/Genymobile/scrcpy/pull/6216#discussion_r2651810005).

> L17 there is a `sc_gamepad_processor` too.

I added a separate commit.

> L326 the migration guide has been moved to https://wiki.libsdl.org/SDL3/README-migration (hyphen instead of slash). But maybe you should now point to the actual documentation / wiki page? No longer need to refer to the SDL2 -> SDL3 transition.

I removed the comment, it was only necessary when the named differred from SDL.

> (changes in `scrcpy/.github/workflows/release.yml`)

These ones must still refer to SDL2:
 - the old Ubuntu version used does not package SDL3
 - this is necessary (at least, this is an easy way) to install the SDL dependencies required to build a working static binary

app/src/input_manager.c
```diff
@@ -911,31 +906,31 @@ sc_input_manager_process_mouse_wheel(struct sc_input_manager *im,
...
 sc_input_manager_process_gamepad_device(struct sc_input_manager *im,
-                                       const SDL_ControllerDeviceEvent *event) {
-    if (event->type == SDL_CONTROLLERDEVICEADDED) {
-        SDL_GameController *gc = SDL_GameControllerOpen(event->which);
-        if (!gc) {
+                                       const SDL_GamepadDeviceEvent *event) {
+    if (event->type == SDL_EVENT_GAMEPAD_ADDED) {
+        SDL_Gamepad *sdl_gamepad = SDL_OpenGamepad(event->which);
+        if (!sdl_gamepad) {
             LOGW("Could not open game controller");
```
anotheruserofgithub:
```suggestion
            LOGW("Could not open gamepad");
```

app/src/input_manager.c
```diff
@@ -911,31 +906,31 @@ sc_input_manager_process_mouse_wheel(struct sc_input_manager *im,
...
             return;
         }
 
-        SDL_Joystick *joystick = SDL_GameControllerGetJoystick(gc);
+        SDL_Joystick *joystick = SDL_GetGamepadJoystick(sdl_gamepad);
         if (!joystick) {
             LOGW("Could not get controller joystick");
```
anotheruserofgithub:
```suggestion
            LOGW("Could not get gamepad joystick");
```

anotheruserofgithub (commented):
Leftovers from https://github.com/Genymobile/scrcpy/pull/6216#discussion_r2651850473.

anotheruserofgithub (commented):
Typo when fixing https://github.com/Genymobile/scrcpy/pull/6216#pullrequestreview-3616928335.

app/src/main.c
```diff
@@ -5,8 +5,9 @@
 #ifdef HAVE_V4L2
 # include <libavdevice/avdevice.h>
 #endif
-#define SDL_MAIN_HANDLED // avoid link error on Linux Windows Subsystem
-#include <SDL2/SDL.h>
+#define SDL_FUNCTION_POINTER_IS_VOID_POINTER
+#include <SDL3/SDL_stdinc.h>
```
anotheruserofgithub:
```suggestion
```

I scratched my head when reading https://wiki.libsdl.org/SDL3/README-main-functions#the-main-entry-point-in-sdl3 but now I think I understand the global picture.

In SDL2:
- `SDL.h` includes `SDL_main.h` [here](https://github.com/libsdl-org/SDL/blob/SDL2/include/SDL.h#L31).
- `SDL_main.h` checks for a defined symbol `SDL_MAIN_HANDLED` to avoid redefining the `main` function [here](https://github.com/libsdl-org/SDL/blob/SDL2/include/SDL_main.h#L33).
- Thus scrcpy has to define `SDL_MAIN_HANDLED` because it includes `SDL.h` which includes `SDL_main.h`.
- This is important when it is linked to the `SDLmain` static library which provides the real `main` entry point.
- `SDL.h` includes `SDL_stdinc.h` [here](https://github.com/libsdl-org/SDL/blob/SDL2/include/SDL.h#L32).
- `SDL_stdinc.h` does not check `SDL_FUNCTION_POINTER_IS_VOID_POINTER` nor does it define `SDL_FunctionPointer`.

In SDL3:
- `SDL.h` does *not* include `SDL_main.h` [here](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL.h#L33).
- `SDL_main.h` checks for a defined symbol `SDL_MAIN_HANDLED` to avoid redefining the `main` function [here](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_main.h#L147).
- Thus scrcpy does not need to define `SDL_MAIN_HANDLED` because it includes `SDL.h` but which does not include `SDL_main.h`.
- This is on purpose since the `SDLmain` static library has been replaced by a single-header library (`SDL_main.h` does the whole work now).
- `SDL.h` includes `SDL_stdinc.h` [here](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL.h#L34).
- `SDL_stdinc.h` checks `SDL_FUNCTION_POINTER_IS_VOID_POINTER` [here](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_stdinc.h#L6156) in order to define `SDL_FunctionPointer`.
- Hence scrcpy has to define `SDL_FUNCTION_POINTER_IS_VOID_POINTER` so as to work around a compiler error somewhere (right?).
- However scrcpy does not need to explicitly include `SDL_stdinc.h` because it is already included by `SDL.h`.

anotheruserofgithub (commented):
As a consequence I expect you can also remove `SDL_MAIN_HANDLED` for tests in `app/meson.build` (see commit ba1b36758e8a6cbebe8215d9bb0936d4598d6045): https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/app/meson.build#L278

But do they need `SDL_FUNCTION_POINTER_IS_VOID_POINTER`?

anotheruserofgithub:
From https://github.com/Genymobile/scrcpy/pull/6216#issuecomment-3697831083:

> I will investigate the remaining ones (especially [#6216 (comment)](https://github.com/Genymobile/scrcpy/pull/6216#discussion_r2651810005)).

This one seems tricky indeed. What would the migration script do with these asynchronous requests for window operations? [EDIT: Apparently nothing relevant, it just [adapts the result checking](https://github.com/libsdl-org/SDL/blob/e1a623f129e75ad532315852d656fb26c80382a6/build-scripts/SDL_migration.cocci#L3739); see why [here](https://github.com/libsdl-org/SDL/blob/e1a623f129e75ad532315852d656fb26c80382a6/build-scripts/SDL_migration.cocci#L4-L6).] I hope you won't have to refactor event handling too much.

> > (changes in `scrcpy/.github/workflows/release.yml`)
> 
> These ones must still refer to SDL2

Fine!

By the way, from https://github.com/Genymobile/scrcpy/pull/6216#issuecomment-3399353701:

> Then, I noticed a very annoying change from SDL2 to SDL3: many previously infallible functions have been made fallible without good reason (in my opinion). See the discussion in https://github.com/libsdl-org/SDL/issues/14223 for details.
> 
> [...]
> 
> Therefore, I added a wrapper in scrcpy that check the reported values, log them when failing is not fatal, and abort() (think "assertion enabled in release mode") to avoid undefined behavior for window size and position getters: https://github.com/Genymobile/scrcpy/commit/a9a8fb190cb8dd9b2db1021c9bf0fbe3cf2a78f4.
> 
> Error handling could be added later if my assumptions prove to be incorrect.
> 
> What do you think?

I think your solution is very nice, I like how you did it.

I don't know how much effort it would take to refactor the code and propagate all errors (over and over) to the callers. I still think it would be great to avoid such wrappers around SDL functions (otherwise, why not wrap them all, for "instrumentation purposes" let's say?), but that could be done in the future if it's not too much work.

anotheruserofgithub (commented):
An occurrence of "controller" that I missed.

The last one is in `README.md`: https://github.com/Genymobile/scrcpy/blob/db2f72a58aaedfe3384b0dea838175e9237a3324/README.md?plain=1#L132 Since there is already the verb "Control" perhaps you want to change `gamepad controllers` to `gamepads`? (More generically these are "game controllers", not "gamepad controllers".)

app/src/screen.c
```diff
@@ -712,8 +712,8 @@ void
 sc_screen_toggle_fullscreen(struct sc_screen *screen) {
     assert(screen->video);
 
-    uint32_t new_mode = screen->fullscreen ? 0 : SDL_WINDOW_FULLSCREEN_DESKTOP;
-    if (SDL_SetWindowFullscreen(screen->window, new_mode)) {
+    bool ok = SDL_SetWindowFullscreen(screen->window, !screen->fullscreen);
+    if (!ok) {
         LOGW("Could not switch fullscreen mode: %s", SDL_GetError());
         return;
     }
```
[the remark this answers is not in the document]
rom1v: I fixed it in a separate commit for clarity: e24018b4b7fdc3e18eb40d0f958916536358c989

I completely removed the local fields. Even if they were only updated on SDL events, their state was fundamentally inconistent. For example:
 1. start in windowed mode
 2. maximize the window → MAXIMIZED event
 3. enable fullscreen → RESTORED event then ENTER_FULLSCREEN event
 4. disable fullscreen → LEAVE_FULLSCREEN event then MAXIMIZED event

If the `maximized` field was updated on `MAXIMIZED`, then its value would be `false` in `LEAVE_FULLSCREEN`. But `SDL_GetWindowFlags()` still returns a `MAXIMIZED` state (the internal state is set before the event is posted, but the boolean field can only be updated when the event is handled).

So far, I only tested on Linux+X11. The relevant use case is:
 1. start in windowed mode
 2. maximize the window
 3. enable fullscreen
 4. rotate the device (switch portrait/landscape)
 5. disable fullscreen
 6. unmaximize

The resulting window should be resized to the new rotated dimensions.

rom1v:
> I don't know how much effort it would take to refactor the code and propagate all errors (over and over) to the callers.

If the errors are really just assertions, I don't want them to propagate as runtime errors (regardless of the exposed SDL API).

> I still think it would be great to avoid such wrappers around SDL functions

Yes, but IMO the correct solution would be to fix it upstream. This would be a breaking change (so not before SDL4), and it was rejected anyway.

anotheruserofgithub (commented):
A few questions about the new commit e24018b4b7fdc3e18eb40d0f958916536358c989.

app/src/screen.c
```diff
@@ -799,26 +785,18 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
+        case SDL_EVENT_WINDOW_ENTER_FULLSCREEN:
+            LOGD("Switched to fullscreen mode");
+            return true;
+        case SDL_EVENT_WINDOW_LEAVE_FULLSCREEN:
+            LOGD("Switched to windowed mode");
+            if (is_windowed(screen)) {
                 apply_pending_resize(screen);
-                sc_screen_render(screen, true);
             }
             return true;
```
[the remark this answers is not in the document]
rom1v: > On `RESTORED` you resize only if it is a video window, but on `LEAVE_FULLSCREEN` you resize even if there is no video, is it because you allow fullscreen only if `assert(screen->video)`?

Yes, exactly, without video (`--no-video`), there is no way to enable fullscreen. I just added some asserts.

> You removed `sc_screen_render(screen, true)`, is it because it will be done on `PIXEL_SIZE_CHANGED` after `sc_sdl_set_window_size()` is called?

Yes, that was the reason. But I finally reverted that changed, because it seems it's important in some cases (see this old commit 7343b233e457076d9419fb0f6e587ba13a7e79c8). And the change should not belong to this refactor anyway.

app/src/screen.c
```diff
@@ -759,15 +750,10 @@ void
...
 
-    if (screen->fullscreen || screen->minimized) {
+    if (!is_windowed(screen)) {
         return;
     }
 
-    if (screen->maximized) {
-        sc_sdl_restore_window(screen->window);
-        screen->maximized = false;
-    }
```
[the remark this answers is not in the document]
rom1v: Done.

app/src/screen.c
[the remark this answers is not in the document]
anotheruserofgithub: I agree with the latest changes. ;)

rom1v:
You can test the latest version of this PR (branch `sdl3.36`) built here: https://github.com/rom1v/scrcpy/actions/runs/20615612335

rom1v:
> > (changes in `scrcpy/.github/workflows/release.yml`)
> 
> These ones must still refer to SDL2:
> 
>    * the old Ubuntu version used does not package SDL3
> 
>    * this is necessary (at least, this is an easy way) to install the SDL dependencies required to build a working static binary

In fact, the list of dependencies is specified here: https://wiki.libsdl.org/SDL3/README-linux#build-dependencies

So I added commit 33bf7587e4c27dc7b0675a5767267dc178742f82.

anotheruserofgithub:
> In fact, the list of dependencies is specified here: https://wiki.libsdl.org/SDL3/README-linux#build-dependencies
> 
> So I added commit [33bf758](https://github.com/Genymobile/scrcpy/commit/33bf7587e4c27dc7b0675a5767267dc178742f82).

To cross-build for Windows (32 & 64) do you have to keep installing libsdl2 or can you also install each dependency?

rom1v:
> To cross-build for Windows (32 & 64) do you have to keep installing libsdl2 or can you also install each dependency?

We don't need the package at all. I added 1c6ffd4085c0ff567c3ea7fe2eb3102d59101586.

rom1v:
This PR is ready to be merged. :rocket: 

Here are binaries for the latest version: https://github.com/rom1v/scrcpy/actions/runs/20693631961

doc/build.md
anotheruserofgithub, line 1: There are still some references to SDL2 in this file: https://github.com/Genymobile/scrcpy/blob/e3fa95387483635182962fc1fbdda207bfea5cb7/doc/build.md?plain=1#L33 https://github.com/Genymobile/scrcpy/blob/e3fa95387483635182962fc1fbdda207bfea5cb7/doc/build.md?plain=1#L39 https://github.com/Genymobile/scrcpy/blob/e3fa95387483635182962fc1fbdda207bfea5cb7/doc/build.md?plain=1#L80 https://github.com/Genymobile/scrcpy/blob/e3fa95387483635182962fc1fbdda207bfea5cb7/doc/build.md?plain=1#L124 https://github.com/Genymobile/scrcpy/blob/e3fa95387483635182962fc1fbdda207bfea5cb7/doc/build.md?plain=1#L139

rom1v: Thank you, fixed.

rom1v:
:rocket:

Issue #5195, opened 2024-08-14
【HELP】The mp4 file recorded through scrcpy cannot be opened, the file is corrupted.


codematrixer:
I use the following Python script to start screen recording:

```
process = subprocess.Popen(['scrcpy', '-s', 'a3c1fab5', '--no-display', '-b', '300000', '--record', '/tmp/20240814141706907.mp4', '--max-size', '1280'],
                                   stdout=subprocess.PIPE,
                                   stderr=subprocess.PIPE,
                                   encoding="utf-8",
                                   shell=False)
```

And then use the following Python script to stop screen recording:

```
process.terminate()
try:
    process.wait(timeout=30)
except subprocess.TimeoutExpired:
    logger.warning("Scrcpy process did not terminate within timeout period.")
```

It always throws `subprocess.TimeoutExpired` error,  It seems that the resources of scrcpy are not being released, causing a timeout.

and then the file `/tmp/20240814141706907.mp4` cannot be opened.

Is there an issue with the way I am closing the scrcpy process? Looking forward to your reply.

------
p.s.  I directly execute the command `scrcpy -s a3c1fab5 --no-display -b 300000 --record /tmp/20240814141706907.mp4 --max-size 1280` through the command line. After pressing Ctrl+C, the file can be opened.

rom1v:
> on Windows `TerminateProcess()` is used. Note that exit handlers and finally clauses, etc., will not be executed.


So the recording cannot complete properly.

rom1v:
For Windows, you could try: 

```python
import signal
os.kill(self.p.pid, signal.CTRL_C_EVENT)
```

https://stackoverflow.com/a/62222396/1987178

For Linux, it should probably be:

```python
os.kill(pid, signal.SIGINT)
```

(not tested)

codematrixer:
> For Windows, you could try:
> 
> ```python
> import signal
> os.kill(self.p.pid, signal.CTRL_C_EVENT)
> ```
> 
> https://stackoverflow.com/a/62222396/1987178
> 
> For Linux, it should probably be:
> 
> ```python
> os.kill(pid, signal.SIGINT)
> ```
> 
> (not tested)
> 
> Also ref #5122

On macOS, I tried  `os.kill(pid, signal.SIGINT)`

but it doesn't work.

rom1v:
When you press Ctrl+C in a terminal on macOS, it sends SIGINT, so it should work.

https://stackoverflow.com/a/11570297/1987178

[...]

PR #6244, opened 2025-07-21
Add catching of CTRL_BREAK_EVENT signal on Windows OS

filipbuda:
If a scrcpy process is spawned using **Python 3.8.10** with **scrcpy v3.3.1**  on **Windows 10 OS** as follows:

```python
proc = subprocess.Popen(
        "scrcpy --no-window --record=file.mp4",
        creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
    )
```

and the process is terminated as follows:

```python
proc.send_signal(signal.CTRL_BREAK_EVENT)
```

Then the process will not terminate gracefully and the recording file will be corrupted.

This can be fixed by adding the catching of the CTRL_BREAK_EVENT signal to the `windows_ctrl_handler` function in scrcpy.c
The fix was tested on a Samsung Galaxy S6 Edge+. The following script:

```python
import signal
import subprocess
import time


if __name__ == "__main__":
    proc = subprocess.Popen(
        "scrcpy --no-window --record=file.mp4",
        creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
    )
    time.sleep(10)
    proc.send_signal(signal.CTRL_BREAK_EVENT)
    proc.wait()
```
was run using **scrcpy v3.3.1** and the resulting recording file was corrupted. The same script was run using the PR build and the resulting recording file was not corrupted.

Possibly fixes the below issue:
https://github.com/Genymobile/scrcpy/issues/5195

### DIFF
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -93,7 +93,7 @@ struct scrcpy {
 
 #ifdef _WIN32
 static BOOL WINAPI windows_ctrl_handler(DWORD ctrl_type) {
-    if (ctrl_type == CTRL_C_EVENT) {
+    if (ctrl_type == CTRL_C_EVENT || ctrl_type == CTRL_BREAK_EVENT) {
         sc_push_event(SDL_QUIT);
         return TRUE;
     }


rom1v:
Thank you.

> and the process is terminated as follows:
>
> ```python
> proc.send_signal(signal.CTRL_BREAK_EVENT)
> ```

What if you send `CTRL_C_EVENT` instead:

```c
proc.send_signal(signal.CTRL_C_EVENT)
```

instead?

filipbuda:
If there is no `creationflags=subprocess.CREATE_NEW_PROCESS_GROUP` argument, as in 
```python
    proc = subprocess.Popen(
        "scrcpy --no-window --record=file.mp4"
    )
```
then
```python
proc.send_signal(signal.CTRL_C_EVENT)
```
will terminate the process gracefully and the recording file will not be corrupt. This issue is that without `creationflags=subprocess.CREATE_NEW_PROCESS_GROUP`, sending `CTRL_C_EVENT` will cause Python to raise `KeyboardInterrupt`, which may not be desired. Also, if the process is spawned as follows
```python
proc = subprocess.Popen(
        "scrcpy --no-window --record=file.mp4",
        creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
    )
```
then sending `CTRL_C_EVENT` does not terminate the process at all.

rom1v:
OK, thanks. LGTM I think. :+1:

I will merge later.

filipbuda:
Sounds good, thank you

rom1v:
Merged into `dev`: 8057835a0d2913a45d7982bc4710a82b0a7f4fed :rocket:

PR #6247, opened 2025-07-21
Add architect documentation for scrcpy

nguyendkn:


rom1v:
It is generated by an LLM, right?

nguyendkn:
> It is generated by an LLM, right?

yes, I don't see any architecture document for the current project so I created a document file to understand the whole project

PR #6334, opened 2025-09-06
Add Windows VBS script for automatic scrcpy launch on device connection

Nicolas8aa:
This PR introduces a new Windows helper script (`scrcpy-noconsole-auto.vbs`) that automatically launches scrcpy when an Android device is connected via USB.

## How it works

- The script runs adb wait-for-device && scrcpy in a hidden loop.
- When a device is plugged in and authorized, scrcpy opens automatically.
- When the session ends (device unplugged or scrcpy closed), the script waits again for the next connection.
- Runs silently in the background with negligible resource usage.

## Usage

- Place scrcpy_auto.vbs in the same folder as scrcpy.exe (or adjust paths).
- Double-click to run, or create a shortcut in the Startup folder (shell:startup) to launch automatically at Windows login.
- From then on, simply connect your Android device to start mirroring without manually launching scrcpy.

## Benefits

- Eliminates the need to manually run scrcpy each time.
- Provides a lightweight, zero-configuration “plug-and-play” experience on Windows.
- Useful for users who frequently mirror their device and want automation.


EmeperorsMod:
i just write this script 
[scrcpy_helper.zip](https://github.com/user-attachments/files/22303663/scrcpy_helper.zip)
work for window only because using cls command

rom1v:

PR #6335, opened 2025-09-06
Add --otg-automatic-screenshot

pierrezurek:
With this option, scrcpy takes a screenshot automatically each time mouse capture is exited in OTG mode.
This is particularly useful when OTG mode is used to enable USB debugging on a phone with a broken screen.
The screenshot will display the mouse pointer that will indicate where to move next.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -114,6 +114,7 @@ enum {
     OPT_NO_VD_SYSTEM_DECORATIONS,
     OPT_NO_VD_DESTROY_CONTENT,
     OPT_DISPLAY_IME_POLICY,
+    OPT_OTG_AUTOMATIC_SCREENSHOT,
 };
 
 struct sc_option {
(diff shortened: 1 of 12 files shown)


computerscibro:
Hello there,

I just tried out your commit and it works great! I'm still thinking about how this could be used in a real-world scenario. How do you usually view the screenshots from a computer? Is it through the cloud or some other method?

I think this is a really good idea, but we need to make it more user-friendly. Can you imagine a way to transmit the screenshots to a computer without using USB tethering? I'm not sure if that's even possible, but I’d love to hear your thoughts on it.

Also, I was thinking we could implement a system that continuously captures screenshots and then deletes them right after. That way, we could simulate a kind of "stream" using a sequence of JPEG images.

pierrezurek:
Hi, I used this idea to help a friend whose phone screen is broken but hadn’t enabled developer options or USB debugging. With scrcpy --otg, he was able to use his keyboard to enter his PIN and enable MTP file transfer. Once MTP transfer was activated, screenshots were visible almost in real-time in the computer’s file explorer. This way, we could enable developer options and USB debugging.

computerscibro:
Hi again,
I tried out MTP and it works great with two scripts: one for taking screenshots and another for automatically displaying them. However, I couldn't manage to create a "usable" version of this prototype.
How many screenshots per second did you take?

Best regards,

PR #6338, opened 2025-09-08
Feature get app icon

rsanjuan87:
# Get app icon (`--get-app-icon`)

The `--get-app-icon` option allows you to extract the icon of one or more installed applications from the Android device.

## Syntax

```
scrcpy --get-app-icon=<packageID>[:path]
```

- `<packageID>`: Required. One or more package IDs, separated by a comma (`,`) (e.g., `com.android.settings,com.android.chrome`). If the special value `all` is used, icons for all installed applications are extracted.
- `[:path]`: Optional. Destination path to save the extracted icons.

## Path behavior

- If `path` is an existing folder, each icon is saved as `<packageID>.png` inside that folder.
- If `path` is an existing file, the extracted icon will overwrite that file, so be careful.
- If `path` does not exist, the containing folder is created and the file is named with the last segment of the path (regardless of extension).
- If `path` is `tmpDir`, the files are stored in the user's temporary folder: `scrcpy/icons/<deviceId>`.

## Examples

- `scrcpy --get-app-icon=com.android.settings`
- `scrcpy --get-app-icon=com.android.settings:~/icon.png`
- `scrcpy --get-app-icon=com.android.settings,com.android.chrome:$HOME/icons/`
- `scrcpy --get-app-icon=all:/home/user/all_icons/`
- `scrcpy --get-app-icon=com.android.chrome,com.android.settings:tmpDir`

## Notes

- If multiple `packageID`s are specified, all indicated icons are extracted.
- For `all`, the `--list-apps` functionality is reused to obtain all package IDs and extract their icons.
- If `:path` is not specified, icons are saved in the current directory.

Tested on MacOS Tahoe + Xiaomi HyperOS 1.0.12 (Android 13)

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -515,8 +515,25 @@ static const struct sc_option options[] = {
     {
         .longopt_id = OPT_GET_APP_ICON,
         .longopt = "get-app-icon",
-        .argdesc = "package_name",
-        .text = "Get the icon of an Android app installed on the device.",
+        .argdesc = "package_names[:path]",
+        .text = "Get the icon of an Android app installed on the device."
+                "The required argument is a comma-separated list of package names"
+                " (e.g. \"com.android.chrome,org.mozilla.firefox\").\n"
+                "Optionally, each package name may be followed by ':path' to "
+                "specify where to save the icon (default is the current "
+                "directory).\n"
+                "The icon is saved as a PNG file, named <package_name>.png by "
+                "default, or <path> if specified.\n"
+                "If 'all' is provided instead of a package name, then the "
+                "icon of all installed apps is extracted.\n"
+                "Examples:\n"
+                "    scrcpy --get-app-icon=com.android.chrome\n"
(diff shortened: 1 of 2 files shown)


rom1v:
Hi,

Thank you for the contribution.

However, I think this feature does not really belong in scrcpy (it could be a separate app to retrieve the icon of an app).

I don't want to add the complexity of extracting, storing, retrieving the icons for a feature which is not really the purpose of scrcpy. It could be useful to retrieve the icons internally to expose it in a client UI if scrcpy had such an UI, but currently it's not the case.

Also refs https://github.com/Genymobile/scrcpy/pull/5574

rsanjuan87:
this is a first step of a internal roadmap to get more gui and behavior to the app
feat: get icon
feat: exit on close [PR](https://github.com/Genymobile/scrcpy/pull/6343)
feat: launch as app (scrcpy icon and name, exit on close)

PR #6350, opened 2025-09-13
Resizable virtual display

rsanjuan87:
feat: allow to resize display as the scrcpy window size

examples:
```
--new-display=1920x1080/420:r
```
```
--new-display=:r
```
```
--new-display=/420:r
```

**_:r_** to indicate resizable screen

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -621,14 +621,17 @@ static const struct sc_option options[] = {
     {
         .longopt_id = OPT_NEW_DISPLAY,
         .longopt = "new-display",
-        .argdesc = "[<width>x<height>][/<dpi>]",
+        .argdesc = "[<width>x<height>][/<dpi>][:r]",
         .optional_arg = true,
         .text = "Create a new display with the specified resolution and "
                 "density. If not provided, they default to the main display "
-                "dimensions and DPI.\n"
+                "dimensions and DPI. Add ':r' to make the display resizable.\n"
                 "Examples:\n"
                 "    --new-display=1920x1080\n"
                 "    --new-display=1920x1080/420  # force 420 dpi\n"
+                "    --new-display=1920x1080/420:r  # resizable display\n"
+                "    --new-display=:r    # resizable with main display size\n"
+                "    --new-display=/420:r  # resizable with main display size and 420 dpi\n"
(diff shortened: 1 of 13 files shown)


rom1v:

Here are some general remarks after a quick glance.

Firstly, it kinda works :+1: I think there is maybe too much complexity (e.g., many fields added in `sc_screen`), but maybe it's necessary, it would require to review more deeply.

Quick question: How much AI has been used to write the code?

The VD size changes when the user resizes the window, but it must also handle the case where the app is rotated (decided by the app or as a result of MOD+R for example).

Small issue: the VD size is exactly the size of the window, but scrcpy always rounds to the nearest multiple of 8 for encoding constraints, causing bad quality:

```
[server] INFO: Resizing display to 1654x757
[server] INFO: Resized display to: 1654x757/280
INFO: Texture: 1648x760
```

Some resize code is done in `event_watcher()`, which is just a workaround for Windows, it should not be there.

rsanjuan87:
> Quick question: How much AI has been used to write the code? 

A lot :)

after some test this is failing, I'll try to fix

rom1v:
> > Quick question: How much AI has been used to write the code?
>
> A lot :)

Thank you for your honesty.

Semantically, one clear clue that it was done by AI is the whole "content-driven resize handling to avoid feedback loop". This is indeed a concern in many contexts, but here it does not apply. The resize is initiated by the client but applied on the server, which then notifies to the client asynchronously, so a feedback loop cannot occur. A human implementing this feature would have necessarily known it, and no observation would have led them to implement this handling that way.

rsanjuan87:
> > > Quick question: How much AI has been used to write the code?
> > 
> > 
> > A lot :)
> 
> Thank you for your honesty.
> 
> Semantically, one clear clue that it was done by AI is the whole "content-driven resize handling to avoid feedback loop". This is indeed a concern in many contexts, but here it does not apply. The resize is initiated by the client but applied on the server, which then notifies to the client asynchronously, so a feedback loop cannot occur. A human implementing this feature would have necessarily known it, and no observation would have led them to implement this handling that way.

In facts, I had this problem in some other attempts, once I check the feature, I saw whats was happening and I "prompted" to avoid the bucle

PR #6351, opened 2025-09-14
Resizable virtual display with factor

rsanjuan87:
feat: allow to resize virual display as the scrcpy window size
imagen
left: no resizable virtual display after resize window
rigth: resizable virtual display after resize window

imagen

left: --new-display=:r6 ( window size * 6 )
center: --new-display (no reszizable)
right: --new-display=:r  (window size * autofactor: calculated from device screen and default initial window size, this case autofactor = 2)

The resizable factor allows you to scale the size to avoid losing image quality.

usage example:
```
--new-display=1920x1080/420:r2
```
```
--new-display=:r
```
```
--new-display=/420:r
```

**_:rN_** 

r: to indicate resizable screen
N: factor to multiply

tested on: MacOS Tahoe & Redmi Note 11 Pro 5G, & Android Studio Emulator (Android 16)

fixing https://github.com/Genymobile/scrcpy/pull/6350

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -632,9 +632,9 @@ static const struct sc_option options[] = {
                 "    --new-display=1920x1080\n"
                 "    --new-display=1920x1080/420  # force 420 dpi\n"
                 "    --new-display=1920x1080/420:r  # resizable display\n"
-                "    --new-display=1920x1080/420:r2.0  # resizable with 2x resolution\n"
-                "    --new-display=:r0.5  # resizable with main display size and half resolution\n"
-                "    --new-display=/420:r1.5  # resizable with main display size, 420 dpi and 1.5x resolution\n"
+                "    --new-display=1920x1080/420:r2.0  # resizable with 2x factor\n"
+                "    --new-display=:r0.5  # resizable with main display size and 0.5x factor\n"
+                "    --new-display=/420:r1.5  # resizable with main display size, 420 dpi and 1.5x factor\n"
                 "    --new-display         # main display size and density\n"
                 "    --new-display=/240    # main display size and 240 dpi",
     },
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -806,12 +806,11 @@ aoa_complete:
         bool resizable_new_display = false;
         float resolution_factor = 1.0f;
         bool auto_resolution_factor = false;
-        
+
         if (options->new_display) {
             const char *r_pos = strstr(options->new_display, ":r");
             if (r_pos) {
                 resizable_new_display = true;
-                
                 // Check if there's a resolution factor after ":r"
                 const char *factor_start = r_pos + 2; // Skip ":r"
                 if (*factor_start && (*factor_start >= '0' && *factor_start <= '9')) {
@@ -833,41 +832,32 @@ aoa_complete:
             }
         }
 
-        // Si se solicitó cálculo automático del factor de resolución, calcularlo basado en el tamaño de la pantalla principal
+        // If automatic resolution factor calculation was requested, calculate it based on the main display size
         if (auto_resolution_factor && resizable_new_display) {
-            // Obtener el tamaño de la pantalla principal
+            // Get the main display size
             SDL_Rect display_bounds;
             if (SDL_GetDisplayUsableBounds(0, &display_bounds) == 0) {
-                // Calcular el factor de resolución basado en la relación entre el tamaño de la pantalla
-                // y el tamaño de la ventana solicitado
+                // Calculate the resolution factor based on the ratio between the display size and the requested window size
                 int display_width = display_bounds.w;
                 int display_height = display_bounds.h;
-                
-                // Usar el tamaño de ventana solicitado o un valor predeterminado
                 int window_width = options->window_width ? options->window_width : display_width / 2;
                 int window_height = options->window_height ? options->window_height : display_height / 2;
-                
-                // Calcular el factor como la relación entre el tamaño de la pantalla y el tamaño de la ventana
-                // Usar el menor de los factores para evitar que la ventana sea más grande que la pantalla
                 float width_factor = (float)display_width / window_width;
                 float height_factor = (float)display_height / window_height;
                 resolution_factor = width_factor < height_factor ? width_factor : height_factor;
-                
-                // Limitar el factor entre 0.1 y 10.0
                 if (resolution_factor < 0.1f) {
                     resolution_factor = 0.1f;
                 } else if (resolution_factor > 10.0f) {
                     resolution_factor = 10.0f;
                 }
-                
-                LOGI("Calculated automatic resolution factor: %.2f (display: %dx%d, window: %dx%d)", 
+                LOGI("Calculated automatic resolution factor: %.2f (display: %dx%d, window: %dx%d)",
                      resolution_factor, display_width, display_height, window_width, window_height);
             } else {
                 LOGW("Could not get display bounds, using default resolution factor: 1.0");
                 resolution_factor = 1.0f;
             }
         }
-        
+
         struct sc_screen_params screen_params = {
             .video = options->video_playback,
             .controller = controller,
--- a/doc/virtual_display.md
+++ b/doc/virtual_display.md
@@ -15,9 +15,9 @@ You can make the display resizable by adding `:r` to the option. Additionally, y
 
 ```bash
 scrcpy --new-display=1920x1080:r           # resizable display
-scrcpy --new-display=1920x1080:r2.0        # resizable with 2x resolution
-scrcpy --new-display=:r0.5                 # resizable with half resolution
-scrcpy --new-display=/420:r1.5             # resizable with 1.5x resolution
+scrcpy --new-display=1920x1080:r2.0        # resizable with 2x factor
+scrcpy --new-display=:r0.5                 # resizable with 0.5x factor
+scrcpy --new-display=/420:r1.5             # resizable with 1.5x factor
 ```
 
 The resolution factor must be between 0.1 and 10.0.


notchyves:
This is an actually great commit, and would make it much easier to use on wider displays such as mine!

VuAnhIT (approved):

rom1v:
I don't really understand the meaning of "resolution factor", in particular how it provides additional information not already encoded into width×height/dpi.

EDIT: I understand now, it can be useful to have the virtual display DPI not correlated to the client resolution.

rsanjuan87:
> I don't really understand the meaning of "resolution factor", in particular how it provides additional information not already encoded into width×height/dpi.

(no resolution) "resizable" factor is to multiply the real window size to send to the virtual display resolution
ex: window size 100x50 with factor 2, will resize the virtual display to 200x100
this will avoid small screen resolution (as same window size)
the width×height/dpi is not applicable as the window will resize the display, you may indicate the initial resolution but after resizing the scrcpy window will resize the display

the autofactor (:r with out N factor) will calculate the factor with default window size and default resolution

cusanarecords-art:
Wie kann ich das beweisen? think it's good to have a screen that can adapt to what I need and not just the one that's already created or the actual screen on the device.

ivonnefbr:
Has Chrome switched to desktop mode?

> feat: allow to resize virual display as the scrcpy window size imagen left: no resizable virtual display after resize window rigth: resizable virtual display after resize window
> imagen
> 
> left: --new-display=:r6 ( window size * 6 ) center: --new-display (no reszizable) right: --new-display=:r (window size * autofactor: calculated from device screen and default initial window size, this case autofactor = 2)
> 
> The resizable factor allows you to scale the size to avoid losing image quality.
> 
> usage example:
> 
> ```
> --new-display=1920x1080/420:r2
> ```
> 
> ```
> --new-display=:r
> ```
> 
> ```
> --new-display=/420:r
> ```
> 
> **_:rN_**
> 
> r: to indicate resizable screen N: factor to multiply
> 
> tested on: MacOS Tahoe & Redmi Note 11 Pro 5G, & Android Studio Emulator (Android 16)
> 
> fixing #6350

miltongg:
> feat: allow to resize virual display as the scrcpy window size imagen left: no resizable virtual display after resize window rigth: resizable virtual display after resize window
> 
> imagen
> left: --new-display=:r6 ( window size * 6 ) center: --new-display (no reszizable) right: --new-display=:r (window size * autofactor: calculated from device screen and default initial window size, this case autofactor = 2)
> 
> The resizable factor allows you to scale the size to avoid losing image quality.
> 
> usage example:
> 
> ```
> --new-display=1920x1080/420:r2
> ```
> 
> ```
> --new-display=:r
> ```
> 
> ```
> --new-display=/420:r
> ```
> 
> **_:rN_**
> 
> r: to indicate resizable screen N: factor to multiply
> 
> tested on: MacOS Tahoe & Redmi Note 11 Pro 5G, & Android Studio Emulator (Android 16)
> 
> fixing #6350

Good addition! The resizable virtual display makes scaling much more flexible while preserving image quality — really improves usability and customization

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -264,4 +270,53 @@ private static int scaleDpi(Size initialSize, int initialDpi, Size size) {
...
+
+        try {
+            // Calculate new DPI based on the size change
+            int newDpi = dpi;
+            if (displaySize != null) {
+                // Scale DPI proportionally to maintain similar pixel density
+                int oldMax = Math.max(displaySize.getWidth(), displaySize.getHeight());
+                int newMax = Math.max(newWidth, newHeight);
+                newDpi = (dpi * newMax) / oldMax;
+            }
```
LHLaurini, line 293: Nice work!

Just a suggestion though: could you make the DPI scaling optional? Otherwise maximizing the scrcpy window makes apps look HUGE. I personally find that not changing the DPI feels more natural.

app/src/cli.c
```diff
@@ -621,14 +621,20 @@ static const struct sc_option options[] = {
...
         .longopt = "new-display",
-        .argdesc = "[<width>x<height>][/<dpi>]",
+        .argdesc = "[<width>x<height>][/<dpi>][:r[<factor>]]",
         .optional_arg = true,
         .text = "Create a new display with the specified resolution and "
                 "density. If not provided, they default to the main display "
-                "dimensions and DPI.\n"
+                "dimensions and DPI. Add ':r' to make the display resizable. "
+                "Optionally add a resolution factor after ':r' to scale the "
+                "display size (between 0.1 and 10.0).\n"
```
LHLaurini, line 630: Perhaps this could explain a bit better what the resolution factor does.

app/src/scrcpy.c
```diff
@@ -802,6 +802,62 @@ scrcpy(struct scrcpy_options *options) {
...
+        if (options->new_display) {
+            const char *r_pos = strstr(options->new_display, ":r");
+            if (r_pos) {
+                resizable_new_display = true;
+                // Check if there's a resolution factor after ":r"
+                const char *factor_start = r_pos + 2; // Skip ":r"
+                if (*factor_start && (*factor_start >= '0' && *factor_start <= '9')) {
+                    // Parse the resolution factor
+                    char *end;
+                    resolution_factor = strtof(factor_start, &end);
```
LHLaurini, line 819: This should also check whether `end` matches the actual end of the string and complain otherwise.

app/src/scrcpy.c
```diff
@@ -802,6 +802,62 @@ scrcpy(struct scrcpy_options *options) {
...
+                        // Prevent too small factors
+                        resolution_factor = 0.1f;
+                    } else if (resolution_factor > 10.0f) {
+                        // Prevent too large factors
+                        resolution_factor = 10.0f;
+                    }
+                } else {
+                    // No explicit resolution factor provided, will calculate automatically
+                    auto_resolution_factor = true;
+                    LOGI("No resolution factor specified, will calculate automatically");
```
LHLaurini, line 830: Maybe show a different message if the user provided an invalid number

app/src/screen.c
```diff
@@ -11,8 +11,18 @@
...
+static Uint32 resize_timer_callback(Uint32 interval, void *param);
+
 #define DOWNCAST(SINK) container_of(SINK, struct sc_screen, frame_sink)
 
+// Prototipo para evitar error de declaración implícita
+static bool sizes_are_close(int a, int b, int tolerance);
+
+static void
+sc_screen_send_resize_display(struct sc_screen *screen, int width, int height);
+
```
LHLaurini, line 25: It's not necessary to forward-declare these functions, just make sure you define them before they're used.

app/src/screen.c
```diff
@@ -945,3 +1077,26 @@ sc_screen_hidpi_scale_coords(struct sc_screen *screen, int32_t *x, int32_t *y) {
...
+        return 0;
+    }
+    // Round to multiple of 8 to match server-side rounding and avoid quality degradation
+    int rounded_w = (width + 4) & ~7;  // Round to nearest multiple of 8
+    int rounded_h = (height + 4) & ~7;  // Round to nearest multiple of 8
+    
+    LOGD("[RESIZE_FINISHED] Notifying display resize: %dx%d (rounded from %dx%d)", rounded_w, rounded_h, width, height);
+    sc_screen_send_resize_display(screen, rounded_w, rounded_h);
+    return 0;
+}
```
LHLaurini, line 1102: Documentation for `SDL_AddTimer` says

> The callback is run on a separate thread.

It should be safe to call `sc_controller_push_msg` here, but you're also modifying `pending_resize_width` and `pending_resize_height` which could lead to race conditions.

It would probably be better to push an event with `sc_push_event` and do this from there (it seems you tried something like that on line 840...).

app/src/screen.c
```diff
@@ -871,9 +949,63 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
+            float scale = 1.0f;
+            if (adjusted_width > adjusted_height) {
+                scale = (float)MAX_DIMENSION / adjusted_width;
+            } else {
+                scale = (float)MAX_DIMENSION / adjusted_height;
+            }
+            adjusted_width = (int)(adjusted_width * scale);
+            adjusted_height = (int)(adjusted_height * scale);
+            LOGW("Limiting dimensions to %dx%d to avoid encoder issues", adjusted_width, adjusted_height);
+        }
```
LHLaurini, line 990: This should be done server-side, if necessary

app/src/screen.c
```diff
@@ -333,6 +341,9 @@ sc_screen_init(struct sc_screen *screen,
     screen->paused = false;
     screen->resume_frame = NULL;
     screen->orientation = SC_ORIENTATION_0;
+    screen->initial_setup = true;
+    screen->content_driven_resize = false;
+    screen->last_resize_time = 0;
```
LHLaurini, line 346: `last_resize_time` is unused

app/src/screen.c
```diff
@@ -800,6 +836,17 @@ sc_screen_resize_to_pixel_perfect(struct sc_screen *screen) {
...
+    if (event->type == SDL_USEREVENT && event->user.code == 0x5343525A && event->user.data1 == screen) {
+        // Timer expirado: notificar resize
+        if (screen->pending_resize_width > 0 && screen->pending_resize_height > 0) {
+            LOGD("[RESIZE_FINISHED] Notifying display resize: %dx%d", screen->pending_resize_width, screen->pending_resize_height);
+            sc_screen_send_resize_display(screen, screen->pending_resize_width, screen->pending_resize_height);
+        }
+        screen->resize_timer = 0;
+        return true;
+    }
+    // --- Fin manejo especial ---
```
LHLaurini, line 849: I don't think anything uses this.

Right now you're doing this from the timer callback, though doing it here is probably better (but move this to inside the switch, please).

app/src/screen.c
LHLaurini, line 1: Make sure to keep comments in English, please

app/src/screen.c
```diff
@@ -835,6 +882,37 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
+                            screen->last_window_width = w;
+                            screen->last_window_height = h;
+                            // --- Resize diferido ---
+                            // Round to multiple of 8 to match server-side rounding and avoid quality degradation
+                            int rounded_w = (w + 4) & ~7;  // Round to nearest multiple of 8
+                            int rounded_h = (h + 4) & ~7;  // Round to nearest multiple of 8
+                            
+                            // Resize window to rounded dimensions to maintain quality
+                            if (rounded_w != w || rounded_h != h) {
+                                SDL_SetWindowSize(screen->window, rounded_w, rounded_h);
```
LHLaurini, line 899: This makes resizing kinda jerky. It would probably be best to do this only after the timer elapses.

app/src/screen.c
```diff
@@ -333,6 +341,9 @@ sc_screen_init(struct sc_screen *screen,
     screen->paused = false;
     screen->resume_frame = NULL;
     screen->orientation = SC_ORIENTATION_0;
+    screen->initial_setup = true;
+    screen->content_driven_resize = false;
```
LHLaurini, line 345: I don't think the `content_driven_resize` flag is doing anything -- see comments below

app/src/screen.c
```diff
@@ -550,7 +579,11 @@ resize_for_content(struct sc_screen *screen, struct sc_size old_content_size,
                 / old_content_size.height,
     };
     target_size = get_optimal_size(target_size, new_content_size, true);
+
+    // Mark that we're doing a content-driven resize to avoid feedback loop
+    screen->content_driven_resize = true;
     set_window_size(screen, target_size);
+    screen->content_driven_resize = false;
```
LHLaurini, line 586: `set_window_size` will produce an event, but we won't deal with it until we return to the event loop, so this flag does nothing

app/src/screen.c
```diff
@@ -577,8 +610,11 @@ apply_pending_resize(struct sc_screen *screen) {
     assert(!screen->maximized);
     assert(!screen->minimized);
     if (screen->resize_pending) {
+        // Mark that we're doing a content-driven resize to avoid feedback loop
+        screen->content_driven_resize = true;
         resize_for_content(screen, screen->windowed_content_size,
                                    screen->content_size);
+        screen->content_driven_resize = false;
```
LHLaurini, line 617: Same as above, but in addition `resize_for_content` is a no-op if `resizable_new_display` is `true`

app/src/screen.c
```diff
@@ -835,6 +882,37 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
                     sc_screen_render(screen, true);
                     break;
                 case SDL_WINDOWEVENT_SIZE_CHANGED:
+                    if (screen->resizable_new_display) {
+                        int w, h;
+                        SDL_GetWindowSize(screen->window, &w, &h);
+                        if ((!sizes_are_close(w, screen->last_window_width, 2) || !sizes_are_close(h, screen->last_window_height, 2)) &&
+                            !screen->initial_setup && !screen->content_driven_resize) {
```
LHLaurini, line 889: This check probably wouldn't be necessary if you waited to resize the window when the timer elapses.

app/src/screen.h
```diff
@@ -59,7 +59,21 @@ struct sc_screen {
...
     enum sc_orientation orientation;
     // rectangle of the content (excluding black borders)
-    struct SDL_Rect rect;
+    SDL_Rect rect;
+
+    // --- Añadido para evitar bucles y zonas negras en displays redimensionables ---
+    int last_window_width;
+    int last_window_height;
+    int last_display_width;
+    int last_display_height;
```
LHLaurini, line 68: Unused

doc/virtual_display.md
```diff
@@ -11,6 +11,17 @@ scrcpy --new-display         # use the main display size and density
...
+
+```bash
+scrcpy --new-display=1920x1080:r           # resizable display
+scrcpy --new-display=1920x1080:r2.0        # resizable with 2x factor
+scrcpy --new-display=:r0.5                 # resizable with 0.5x factor
+scrcpy --new-display=/420:r1.5             # resizable with 1.5x factor
+```
+
+The resolution factor must be between 0.1 and 10.0.
+
```
LHLaurini, line 24: Same as other comment, explain what "resolution factor" means

server/src/main/java/com/genymobile/scrcpy/control/ControlMessage.java
```diff
@@ -25,12 +25,12 @@ public final class ControlMessage {
     public static final int TYPE_OPEN_HARD_KEYBOARD_SETTINGS = 15;
     public static final int TYPE_START_APP = 16;
     public static final int TYPE_RESET_VIDEO = 17;
+    public static final int TYPE_RESIZE_DISPLAY = 18;
 
     public static final long SEQUENCE_INVALID = 0;
 
     public static final int COPY_KEY_NONE = 0;
     public static final int COPY_KEY_COPY = 1;
-    public static final int COPY_KEY_CUT = 2;
```
LHLaurini, line 33: Removed by accident?

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -197,7 +199,11 @@ public void startNew(Surface surface) {
                 ServiceManager.getWindowManager().setDisplayImePolicy(virtualDisplayId, displayImePolicy);
             }
 
-            displaySizeMonitor.start(virtualDisplayId, this::invalidate);
+            // Only start display size monitoring for non-resizable displays
+            // For resizable displays, we control the resizing manually from the client
+            if (!resizable) {
+                displaySizeMonitor.start(virtualDisplayId, this::invalidate);
+            }
```
LHLaurini, line 206: If feel like maybe we should still resize the window if the display is resized by something else in the device...

Then you also wouldn't have to worry about rounding and resizing the window in screen.c

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
rsanjuan87: scaling is optional ( without :r option)

app/src/cli.c
rsanjuan87: resolution factor is, if your scrcpy window is 200x200 it will multiply with the screen resolution factor 
:r2 => 200x200 * 2 = 400x400

app/src/scrcpy.c
rsanjuan87: example: 
const char *input = ":r2.3";
const char *r_pos = strchr(input, 'r'); // points to 'r'
const char *factor_start = r_pos + 2;   // points to '2'

char *end;
float resolution_factor = strtof(factor_start, &end);

// resolution_factor = 2.3
// end -> points to '\0'

app/src/scrcpy.c
rsanjuan87: if the user sends :r without number or sends a not number ":rasd"
 it will catch this case

PR #6408, opened 2025-10-05
Clean handling of ANDROID_PLATFORM containing decimal, like 36.1

metayan:
Removes the following error:
./server/build_without_gradle.sh: line 89:
[[: 36.1: syntax error: invalid arithmetic operator (error token is ".1")

Tested with
```sh
ANDROID_PLATFORM=36.1 ANDROID_BUILD_TOOLS=36.1.0 ./server/build_without_gradle.sh
```

### DIFF
--- a/server/build_without_gradle.sh
+++ b/server/build_without_gradle.sh
@@ -86,7 +86,7 @@ javac -encoding UTF-8 -bootclasspath "$ANDROID_JAR" \
 echo "Dexing..."
 cd "$CLASSES_DIR"
 
-if [[ $PLATFORM -lt 31 ]]
+if [[ ${PLATFORM%.*} -lt 31 ]]
 then
     # use dx
     "$BUILD_TOOLS_DIR/dx" --dex --output "$BUILD_DIR/classes.dex" \


rom1v:
I have not checked (I'm replying from my phone), but I think ANDROID_PLATFORM is expected to be an integer. Why do you need to pass 36.1 instead of 36?

metayan:
The latest available through `sdkmanager` is
platforms;android-36.1 - Android SDK Platform 36.1
and needs to be specified as `36.1`, because it gets installed in
`/usr/local/share/android-commandlinetools/platforms/android-36.1/`

Otherwise:
```
ANDROID_PLATFORM=36 ANDROID_BUILD_TOOLS=36.1.0 ./server/build_without_gradle.sh 
Platform: android-36
Build-tools: 36.1.0
Build dir: /Users/Shared/src/scm/android/scrcpy/build_manual
Generating java from aidl...
ERROR: /usr/local/share/android-commandlinetools/platforms/android-36/framework.aidl:
  Error while opening file for parsing
```

With this patch (and 36.1):
```
ANDROID_PLATFORM=36.1 ANDROID_BUILD_TOOLS=36.1.0 ./server/build_without_gradle.sh 
Platform: android-36.1
Build-tools: 36.1.0
Build dir: /Users/Shared/src/scm/android/scrcpy/build_manual
Generating java from aidl...
Compiling java sources...
Dexing...
Server generated in /Users/Shared/src/scm/android/scrcpy/build_manual/scrcpy-server
```

rom1v:
:+1:

I replaced `${A%.*}` with `${A%%.*}` so that it will also work with 36.1.1. Merged into `dev`: f3d4fde15b51fd85a59c835d0310aac19d429737

metayan:
Great. Good catch.

PR #6487, opened 2025-11-18
Fix: when calling `Throwable.printStackTrace`, logs from other threads may be inserted.

valord577:
Fix: when calling `Throwable.printStackTrace`, logs from other threads may be inserted.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/util/Ln.java
+++ b/server/src/main/java/com/genymobile/scrcpy/util/Ln.java
@@ -75,11 +75,9 @@ public final class Ln {
         if (isEnabled(Level.WARN)) {
             Log.w(TAG, message, throwable);
             synchronized (CONSOLE_ERR) {
-                CONSOLE_ERR.print(PREFIX + "ERROR: " + message + '\n');
+                CONSOLE_ERR.print(PREFIX + "WARN: " + message + '\n');
                 if (throwable != null) {
-                    StringWriter sw = new StringWriter();
-                    throwable.printStackTrace(new PrintWriter(sw));
-                    CONSOLE_ERR.print(sw.toString());
+                    throwable.printStackTrace(CONSOLE_ERR);
                 }
             }
         }
@@ -95,9 +93,7 @@ public final class Ln {
             synchronized (CONSOLE_ERR) {
                 CONSOLE_ERR.print(PREFIX + "ERROR: " + message + '\n');
                 if (throwable != null) {
-                    StringWriter sw = new StringWriter();
-                    throwable.printStackTrace(new PrintWriter(sw));
-                    CONSOLE_ERR.print(sw.toString());
+                    throwable.printStackTrace(CONSOLE_ERR);
                 }
             }
         }


valord577:
@rom1v Hi, I am new here. Do you have time to review this PR?

rom1v:
Thank you for the PR. :+1:

Indeed, I have noticed that some bug reports contain interleaved error messages (but I could never reproduced).

I am wondering, is it really necessary to use an intermediate `StringWriter`?

I think this should also work:

```diff
diff --git server/src/main/java/com/genymobile/scrcpy/util/Ln.java server/src/main/java/com/genymobile/scrcpy/util/Ln.java
index 46057b4d3..b9b548afc 100644
--- server/src/main/java/com/genymobile/scrcpy/util/Ln.java
+++ server/src/main/java/com/genymobile/scrcpy/util/Ln.java
@@ -77,9 +77,7 @@ public final class Ln {
             synchronized (CONSOLE_ERR) {
                 CONSOLE_ERR.print(PREFIX + "ERROR: " + message + '\n');
                 if (throwable != null) {
-                    StringWriter sw = new StringWriter();
-                    throwable.printStackTrace(new PrintWriter(sw));
-                    CONSOLE_ERR.print(sw.toString());
+                    throwable.printStackTrace(CONSOLE_ERR);
                 }
             }
         }
@@ -95,9 +93,7 @@ public final class Ln {
             synchronized (CONSOLE_ERR) {
                 CONSOLE_ERR.print(PREFIX + "ERROR: " + message + '\n');
                 if (throwable != null) {
-                    StringWriter sw = new StringWriter();
-                    throwable.printStackTrace(new PrintWriter(sw));
-                    CONSOLE_ERR.print(sw.toString());
+                    throwable.printStackTrace(CONSOLE_ERR);
                 }
             }
         }
```

rom1v:
> Indeed, I have noticed that some bug reports contain interleaved error messages (but I could never reproduced).

In fact, this should be a different problem, since `throwable.printStackTrace(…)` _does_ lock the stream/writer:

```java
    private void printStackTrace(PrintStreamOrWriter s) {
        // Guard against malicious overrides of Throwable.equals by
        // using a Set with identity equality semantics.
        Set<Throwable> dejaVu = Collections.newSetFromMap(new IdentityHashMap<>());
        dejaVu.add(this);

        synchronized (s.lock()) {
            // Print our stack trace
            s.println(this);
            StackTraceElement[] trace = getOurStackTrace();
            for (StackTraceElement traceElement : trace)
                s.println("\tat " + traceElement);

            // Print suppressed exceptions, if any
            for (Throwable se : getSuppressed())
                se.printEnclosedStackTrace(s, trace, SUPPRESSED_CAPTION, "\t", dejaVu);

            // Print cause, if any
            Throwable ourCause = getCause();
            if (ourCause != null)
                ourCause.printEnclosedStackTrace(s, trace, CAUSE_CAPTION, "", dejaVu);
        }
    }
```

Adding synchronization would only avoid another message inserted between the `print()` and `printStackTrace()` calls.

valord577:
> `throwable.printStackTrace(…)` does lock the stream/writer

As I understand it, it locks not `CONSOLE_ERR`, but `new WrappedPrintStream(s)`. 

Maybe I've misunderstood?

```java
// File: java/lang/Throwable.java

    public void printStackTrace(PrintStream s) {
        printStackTrace(new WrappedPrintStream(s));
                     // ^ Locks this object.
    }

    private void printStackTrace(PrintStreamOrWriter s) {
        ...
        ...
    }
```

------

> Indeed, I have noticed that some bug reports contain interleaved error messages (but I could never reproduced).

Screenshot 2025-11-18 at 16 55 53

------

> I am wondering, is it really necessary to use an intermediate StringWriter?

> Adding synchronization would only avoid another message inserted between the print() and printStackTrace() calls.

Perhaps we can emulate Android's logcat to implement a custom printStackTrace method?

See the line contains `lbbw.flush();`.

```java
// File: android/util/Log.java

    public static int printlns(int bufID, int priority, @Nullable String tag, @NonNull String msg,
            @Nullable Throwable tr) {
        ImmediateLogWriter logWriter = new ImmediateLogWriter(bufID, priority, tag);
        // Acceptable buffer size. Get the native buffer size, subtract two zero terminators,
        // and the length of the tag.
        // Note: we implicitly accept possible truncation for Modified-UTF8 differences. It
        //       is too expensive to compute that ahead of time.
        int bufferSize = PreloadHolder.LOGGER_ENTRY_MAX_PAYLOAD    // Base.
                - 2                                                // Two terminators.
                - (tag != null ? tag.length() : 0)                 // Tag length.
                - 32;                                              // Some slack.
        // At least assume you can print *some* characters (tag is not too large).
        bufferSize = Math.max(bufferSize, 100);

        LineBreakBufferedWriter lbbw = new LineBreakBufferedWriter(logWriter, bufferSize);

        lbbw.println(msg);

        if (tr != null) {
            // This is to reduce the amount of log spew that apps do in the non-error
            // condition of the network being unavailable.
            Throwable t = tr;
            while (t != null) {
                if (t instanceof UnknownHostException) {
                    break;
                }
                if (t instanceof DeadSystemException) {
                    lbbw.println("DeadSystemException: The system died; "
                            + "earlier logs will point to the root cause");
                    break;
                }
                t = t.getCause();
            }
            if (t == null) {
                tr.printStackTrace(lbbw);
            }
        }

        lbbw.flush();
          // ^^^ Only flush logs here ^^^

        return logWriter.getWritten();
    }
```

valord577:
Synchronization locks only occur when errors are reported. Such situations should be avoided whenever possible. Therefore, the impact on performance should be minimal.

valord577:
> I am wondering, is it really necessary to use an intermediate `StringWriter`?

It is necessary to use a string concatenation class to collect stack trace information. Otherwise, it will be printed directly to stderr/stdout.

```java
    private void printStackTrace(PrintStreamOrWriter s) {

        Set<Throwable> dejaVu = Collections.newSetFromMap(new IdentityHashMap<>());
        dejaVu.add(this);

        synchronized (s.lock()) {
            s.println(this);
            StackTraceElement[] trace = getOurStackTrace();
            for (StackTraceElement traceElement : trace)
                s.println("\tat " + traceElement);
               // ^^^ Print message directly to stderr/stdout

            for (Throwable se : getSuppressed())
                se.printEnclosedStackTrace(s, trace, SUPPRESSED_CAPTION, "\t", dejaVu);
                // ^^^ Print message directly to stderr/stdout
            
            Throwable ourCause = getCause();
            if (ourCause != null)
                ourCause.printEnclosedStackTrace(s, trace, CAUSE_CAPTION, "", dejaVu);
                      // ^^^ Print message directly to stderr/stdout
        }
    }
```

rom1v:
> As I understand it, it locks not `CONSOLE_ERR`, but `new WrappedPrintStream(s)`.

It locks `s.lock()`:

```java
synchronized (s.lock()) {
    // …
}
```

And `WrappedPrintStream` is defined as follow:

```java
    private static class WrappedPrintStream extends PrintStreamOrWriter {
        private final PrintStream printStream;

        WrappedPrintStream(PrintStream printStream) {
            this.printStream = printStream;
        }

        Object lock() {
            return printStream;
        }

        void println(Object o) {
            printStream.println(o);
        }
    }
```

So it locks `CONSOLE_ERR`.

---

Screenshot 2025-11-18 at 16 55 53

:+1:

Yes, I think adding `synchronized (CONSOLE_ERR)` around the two calls will fix this issue.

> Synchronization locks only occur when errors are reported. Such situations should be avoided whenever possible. Therefore, the impact on performance should be minimal.

Sure.

> It is necessary to use a string concatenation class to collect stack trace information. Otherwise, it will be printed directly to stderr/stdout.

In any case, it is necessary to call `printStackTrace()` with a `Writer` argument.

But that `Writer` can be `CONSOLE_ERR` directly, without using an additional intermediate `StringWriter`.

valord577:
LGTM. I change the code! 👍

rom1v:
I squashed your two commits and updated the commit message: 7dd9bcaf606eca7f785ab71423aa113c214522a5

Does it look ok to you?

valord577:
LGTM. No problem.

rom1v:
Merged into `dev`.

PR #6569, opened 2025-12-29
Prevent division-by-zero when mouse moves at start

metayan:
A simple solution by checking that the frame is created before handling input events.


metayan:
Ha, this time I managed to correct the base branch... ;)

PR #6597, opened 2026-01-13
fix desktop file for linux 

normaluser-nobody:


normaluser-nobody:
did NOT work because:
    * .desktop files do not interpret shell-style escaping like \$
    * $SHELL expands incorrectly or becomes empty
    * /bin/sh -c ends up with no valid command  
- Working Exec lines:
    Exec=/bin/bash -ic "scrcpy --pause-on-exit=if-error"
    Exec=/bin/bash -ic "scrcpy"
    because:
    * bash -i loads .bashrc properly
    * .desktop can parse this line correctly
    * No confusing escaping is needed

rom1v:
The current version works for me. What's your config?

If you run explicitly with bash, then environment variables (like `ADB`) set for the default shell (for example zsh) will not work.

> ```bash
> # For some users, the PATH or ADB environment variables are set from the shell
> # startup file, like .bashrc or .zshrc… Run an interactive shell to get
> # environment correctly initialized.
> ```

normaluser-nobody:
for me $SHELL is not expanding "Invalid environment variable name evaluates to an empty string: SHELL -i -c 'scrcpy --pause-on-exit=if-error'
/bin/sh: -c: option requires an argument"

PR #6637, opened 2026-01-28
Allow macOS to get better OpenGL version again

### LINKED ISSUE #6636: Comment to WIP: renderer.6 branch doesn't build on macOS
The tweak to get a better OpenGL version than 2.1 on macOS doesn't work on [`renderer.6`.](https://github.com/Genymobile/scrcpy/blob/renderer.6)

Edit:
Fix is in #6637 .

~~My attempts at fixing it got me too tangled up in the dependencies - window, display, screen, texture... ;)~~

~~An option is to just disable it for now and wait to see if something is done with https://github.com/libsdl-org/SDL/issues/12788 , which would probably allow for a much easier solution.~~

~~Otherwise, if it's too much of a hassle to get it to work again, then maybe it's alright to just remove it and rely on `metal` for macOS.~~

The relevant places:
https://github.com/Genymobile/scrcpy/blob/c2e9aa85e3ff117cba096dea8bdcd1fd3d9380b7/app/src/texture.h#L14-L17
https://github.com/Genymobile/scrcpy/blob/c2e9aa85e3ff117cba096dea8bdcd1fd3d9380b7/app/src/texture.c#L17-L20
https://github.com/Genymobile/scrcpy/blob/c2e9aa85e3ff117cba096dea8bdcd1fd3d9380b7/app/src/texture.c#L25-L41
https://github.com/Genymobile/scrcpy/blob/c2e9aa85e3ff117cba096dea8bdcd1fd3d9380b7/app/src/texture.c#L72-L74

Feel free to let me know if there is a better place to put comments such as this.

metayan:
Figured out how to fix #6636 .
Getting OpenGL 4.1 instead of 2.1 again.

### DIFF
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -417,6 +417,31 @@ sc_screen_init(struct sc_screen *screen,
         goto error_destroy_window;
     }
 
+#ifdef SC_DISPLAY_FORCE_OPENGL_CORE_PROFILE
+    screen->gl_context = NULL;
+    // starts with "opengl"
+    const char *renderer_name = SDL_GetRendererName(screen->renderer);
+    bool use_opengl = renderer_name && !strncmp(renderer_name, "opengl", 6);
+    if (use_opengl) {
+
+        LOGD("Attempting to get better OpenGL version");
+        // Persuade macOS to give us something better than OpenGL 2.1.
+        // If we create a Core Profile context, we get the best OpenGL version.
+        bool ok = SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK,
+				      SDL_GL_CONTEXT_PROFILE_CORE);
+	if (!ok) {
+	    LOGW("Could not set a GL Core Profile Context");
+	}
+
+	LOGD("Creating OpenGL Core Profile context");
+	screen->gl_context = SDL_GL_CreateContext(screen->window);
+	if (!screen->gl_context) {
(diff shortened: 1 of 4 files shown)


rom1v:
Perfect, thank you, I will squash the fix into my branch.

rom1v:
Please test the `renderer.11` branch.

metayan:
`renderer.11` branch works fine for me.
(Only the SDL3 copy/paste issue still on macOS Catalina.)
Something special that needs testing?

By the way, the build procedure you have put together is a joy to use.
Both the client and server. Exceptional.

rom1v:
> `renderer.11` branch works fine for me.

:tada:

> Something special that needs testing?

Not for the moment, thanks.

> By the way, the build procedure you have put together is a joy to use.
> Both the client and server. Exceptional.

Thank you :heart:

Do you build only the client+server, or also the dependencies (SDL, FFmpeg, etc.)?

The code to build the deps is a bit ad-hoc/hacky and was painful to write, but at least it works :smile:

metayan:
> Do you build only the client+server, or also the dependencies (SDL, FFmpeg, etc.)?

I use the dependency-building magic. It's superb.


> The code to build the deps is a bit ad-hoc/hacky and was painful to write, but at least it works 😄

It's really good and makes it simple to get the right things built.
I also like that it's easy to, for example, swap in different SDL versions by just editing `app/deps/sdl.sh`.

rom1v:
#6651

Issue #2701, opened 2021-10-12
Software renderer on macOS only displays top left quadrant


Atemu:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: macOS 11.6
 - scrcpy version: 
 ```
scrcpy 1.19

dependencies:
 - SDL 2.0.14
 - libavcodec 58.134.100
 - libavformat 58.76.100
 - libavutil 56.70.100
```
 - installation method: NixOS/nixpkgs@ee084c02040e864eeeb4cf4f8538d92f7c675671
 - device model: SM-A505FN
 - Android version: 11

**Describe the bug**

scrcpy only renders the top left quadrant in software renderer mode. This does not happen with opengl or metal.

It'd probably be best to use one of those by default instead of software rendering.

**Comment by lalyos** [comment]
I had the same issue: only the top-left corner was visible.
My solution was:
```
scrcpy --render-driver metal
```

I've tried all drivers and 2 of them worked: `metal` and `opengl`.

Atemu:
Why doesn't scrcpy default to metal or OGL?

rom1v:
> Why doesn't scrcpy default to metal or OGL?

Scrcpy leave SDL choose the render driver.

If you explicitly set `--render-driver=`, then it just sets an hint ([`SDL_HINT_RENDER_DRIVER`](https://wiki.libsdl.org/SDL_HINT_RENDER_DRIVER) for SDL:
https://github.com/Genymobile/scrcpy/blob/b4caa483dd6682279170b5ec61963e6239f15072/app/src/scrcpy.c#L95-L97

So maybe a SDL issue?

rom1v:
Oh, maybe it's related to nix? (both of you use nix)

Atemu:
Could certainly be a packaging bug on our side. How is this list determined?

mccartykim:
Sorry to necro but I made a pr I thought might be useful for everyone? If this is clearly a nixpkgs quirk that people don't repro outside of nixpkgs, I can close my pr and try just patching the nixpkgs derivation @rom1v

PR #6639, opened 2026-01-29
Fix software renderer visual corruption on macOS with ARM64 emulators

mccartykim:
Hi! This contribution is meant to fix this issue that's been bugging me: https://github.com/Genymobile/scrcpy/issues/2701, where I was getting an ugly mosaic still instead of video with the default scrcpy command.

(Truthfully, I never thought to change the renderer, so I think this is a major barrier to people adopting a great tool)

## Process

I had an AI agent investigate why scrcpy stopped working on my works' tablets quite a while ago. It found there's a bug with the SDL cpu library this repo uses by default, but that Metal or OpenGL renderers on hardware work fine.

I'm not sure if the team feels strongly about having CPU rendering be the default, but since Macs have had Metal a long time, it feels quite safe to enable it for the platform.

To test, I built the pr within the nixpkgs' development environment on my M1 Macbook Pro since I don't have homebrew, and the result worked as expected with perfect video streaming over USB.

Following is the AI agent's description of the patch it wrote. It was Claude Sonnet 4.5 in Windsurf Cascade if that's of interest:

## AI Description

The SDL software renderer displays visual static/noise when rendering YV12 textures from ARM64 Android emulators and some tablets on macOS. This is due to a bug in SDL's software renderer YUV texture handling on macOS.

Hardware renderers (Metal, OpenGL) work correctly, so this patch adds automatic fallback to Metal renderer (preferred by Apple) then OpenGL on macOS when the default accelerated renderer fails, preventing users from encountering the buggy software renderer.

Root cause: SDL_UpdateYUVTexture with SDL_PIXELFORMAT_YV12 produces corrupted output in SDL's software renderer on macOS with ARM64 video sources.

Workaround: Automatically use Metal/OpenGL renderer on macOS instead of falling back to software renderer. Metal is tried first as it's Apple's modern, actively maintained graphics API.

## Screenshots:

Before:
Untitled

After:
image

### DIFF
--- a/app/src/display.c
+++ b/app/src/display.c
@@ -34,8 +34,34 @@ sc_display_init(struct sc_display *display, SDL_Window *window,
     display->renderer =
         SDL_CreateRenderer(window, -1, SDL_RENDERER_ACCELERATED);
     if (!display->renderer) {
+#ifdef __APPLE__
+        // Workaround for SDL software renderer bug on macOS
+        // The software renderer has known issues with YV12 texture updates
+        // from ARM64 Android emulators, displaying visual corruption.
+        // Force Metal or OpenGL renderer instead of falling back to software.
+        LOGW("Could not create accelerated renderer: %s", SDL_GetError());
+        LOGI("Trying Metal renderer (software renderer has known issues on macOS)");
+        
+        SDL_SetHint(SDL_HINT_RENDER_DRIVER, "metal");
+        display->renderer = SDL_CreateRenderer(window, -1, 0);
+        
+        if (!display->renderer) {
+            LOGW("Could not create Metal renderer: %s", SDL_GetError());
+            LOGI("Trying OpenGL renderer");
+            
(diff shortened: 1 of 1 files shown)


rom1v:
IIUC, the software renderer is broken on macOS.

First, you should try with SDL3 to see if the error still exists: https://github.com/Genymobile/scrcpy/pull/6216#issuecomment-3708095567 (explicitly use `--render-driver=software`).

Anyway, I don't really understand the patch, because currently (when you get noise with software renderer), `display->renderer` is not `NULL` (otherwise scrcpy would have exited). So unless I'm mistaken, the code added in that block is never called, so the behavior should not change.

mccartykim:
I just realized what's happening, building master from scratch on darwin works just fine. Thanks for bearing with me and pointing out my fix wouldn't really do anything! I'll see what nix is doing different and bug nixpkgs instead. Tysm!

image

PR #6661, opened 2026-02-09
Add config file support for persisting default options

### LINKED ISSUE #2715: Configuration file

As a user, I would like to be able to execute scrcpy without having to specify my configuration flags every time.

There should be a config file where you can specify options for scrcpy. Ideally a plain text file where the long command line options can be assigned in a UNIX config file format like this:

```
max-fps = 15 # Comment
bit-rate = 2M
```

(Or another standardised format like TOML, JSON, YAML, INI, Dhall etc.)

The config file should live in the OS' default configuration directory. This means `$XDG_CONFIG_HOME` on Freedesktop systems, `$XDG_CONFIG_HOME` or `~/Library/Application Support/` on macOS systems and on Windows... `%AppData%` I guess?

The only alternative would be specifying the wanted options in one or multiple environment variables or asking of every user to write their own wrappers.

**Additional context**

This is especially important for running scrcpy as a desktop app where command line flags cannot trivially be modified.

Persistent configuration via GUIs depends on this being implemented.

### LINKED ISSUE #5266: Config file for default parameters

I have to type a number of flags/parameters to pass to scrcpy each time I run it. It's hard to remember and type them accurately

A configuration file (e.g. ini file) in the SCRCPY installation folder that is empty by default, but can be filled with parameters that are appended onto any execution of `scrcpy.exe`.

This file could be edited manually, or it might be nice to include two additional commands, "setdefault" and "getdefault"

1. `scrcpy.exe --getdefault` would return the contents of the config file as StdOut 
2. `scrcpy.exe --setdefault --foo --bar` would replace the contents of the config file with "--foo --bar" (aka valid scrcpy parameters)

Additionally, it might be valuable to have a parameter like `--nodefault`, so the user can pass other parameters to the specific execution of scrcpy WITHOUT appending the config file's contents

Making a Windows shortcut or environmental variable

savked:
## Summary

- Adds support for reading default options from a config file (`$XDG_CONFIG_HOME/scrcpy/config.ini` or `~/.config/scrcpy/config.ini` on Unix, `%APPDATA%\scrcpy\config.ini` on Windows)
- Config file uses a simple `key=value` format where keys correspond to long option names (e.g. `render-driver=opengl`)
- Supports named profiles such as `[phone]`, selected with `scrcpy phone`
- CLI arguments always override config file values (config args are prepended, so "last wins" applies naturally)
- Adds `--no-config` flag to skip loading any config file
- Adds `--config-file=<path>` flag and `SCRCPY_CONFIG_FILE` env var to specify a custom config file path
- Supports the portable build layout via `sc_file_get_local_path`

## Motivation

My original motivation for this was the macOS rendering issue (#2701) where the software renderer only displays the top-left quadrant, requiring `--render-driver=opengl` on every invocation. I had been using a shell alias as a workaround, but that felt fragile and doesn't carry over across terminals or scripts easily. A config file is a cleaner solution and benefits anyone who regularly passes the same set of flags.

## Config file format

```ini
# ~/.config/scrcpy/config.ini
render-driver=opengl
max-fps=60
no-audio
```

## Named profiles

Thanks to @mytagssma for suggesting named per-device profiles in [this PR comment](https://github.com/Genymobile/scrcpy/pull/6661#issuecomment-5600443727).

Configuration files may contain named sections:

```ini
max-fps=30

[phone]
serial=192.168.1.10:5555
max-size=1080
```

`scrcpy phone` loads the global options followed by the `[phone]` options. Command-line options are processed last and therefore take precedence.

The profile name uses scrcpy's previously unused positional argument. This is backward-compatible with valid existing commands, since positional arguments were previously rejected, but it intentionally reserves that syntax for configuration profiles. This can be changed to an explicit `--profile=phone` option if preferred.

## Test plan

- [ ] Verify config file is loaded from default path when present
- [ ] Verify `--no-config` skips loading
- [ ] Verify `--config-file=<path>` overrides the default location
- [ ] Verify CLI args override config file values
- [ ] Verify comments and blank lines are skipped
- [ ] Verify missing config file is silently ignored
- [ ] Build on Linux, macOS, and Windows

---

Closes: https://github.com/Genymobile/scrcpy/issues/2715
Closes: https://github.com/Genymobile/scrcpy/issues/5266

> This PR was developed with the assistance of an AI coding tool (Codex and Claude Code).


Impeta:
@rom1v any chance we could see this merged soon? Would be REALLY useful for us who fickle up with very specific configs to scrcpy.

mytagssma:
> ⚠️ **AI-assisted:** This comment was drafted with the help of AI.

Could the config file eventually support **per-device profiles**, similar to `~/.ssh/config`?

For example:

```ini
[phone]
serial=192.168.1.10:5555
max-size=1080
video-bit-rate=4M

[tablet]
serial=192.168.1.20:5555
max-size=1600
video-bit-rate=8M
```

Then `scrcpy phone` could automatically use the device and its configured options.

This would be especially useful for users who regularly use multiple devices.

savked:
@mytagssma great suggestion, will add it!

PR #6662, opened 2026-02-09
Display disconnected icon before closing

rom1v:
When the connection to the device is lost while mirroring, the window closed immediately, suggesting scrcpy had crashed.

To make it clear that a disconnection occurred, display a disconnected icon for 2 seconds before closing the window.

The icon replaces the screen content immediately:

scrcpy_disconnected_small

On purpose, the last frame is immediately removed. For example, during a presentation, the device might inadvertently show some private content on screen, the user should be able to immediately disconnects the USB cable to remove the content.

---

I intentionally did not add an option to configure the timeout delay (it's hardcoded to 2 seconds), there are already a lot of options…

The implementation is a bit "ad hoc", but hey it works.

I replaced `SCRCPY_ICON_PATH` (which defined the unique scrcpy icon) with `SCRCPY_ICON_DIR` now that there are several icons.

---

I get the (temporary) disconnected icon from a free icons website (I can't remember where).

@varlesh As the author of the main icon (https://github.com/Genymobile/scrcpy/pull/1987#issuecomment-949684080), may I ask you to create an icon for this use case?

This is an "Android disconnection"
 - in normal mode, it's the adb socket (either over USB or wifi)
 - in OTG mode, it's the AOA connection (over USB)

so this is not specifically for USB or Wifi (and even less wall socket). If it needs to represent something physical, USB is probably fine though.


rom1v:
I fixed some issues and updated the icon.

The new icon is derived from [this one](https://uxwing.com/mobile-data-icon/).

I think it makes it clear that the communication with the device is lost:

scrcpy_disconnected_2_small

app/src/screen.c
```diff
@@ -903,6 +974,46 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
 
+void
+sc_screen_handle_disconnection(struct sc_screen *screen) {
+    if (!screen->window_shown) {
+        // No window open, quit immediately
+        return;
+    }
+
+    SDL_Event event;
+    while (SDL_WaitEvent(&event)) {
```
anotheruserofgithub, line 1003: I think there is an issue here in case `sc_disconnect_start()` failed to initialize the mutex/cond/thread. I would add a check like this above the while-loop:
```c
    if (!screen->disconnect_started) {
        // If sc_disconnect_start() failed, quit immediately
        return;
    }
```

app/src/disconnect.c
```diff
@@ -0,0 +1,89 @@
...
+
+    if (d->deadline != SC_TICK_NONE) {
+        sc_mutex_lock(&d->mutex);
+        bool timed_out = false;
+        while (!d->interrupted && !timed_out) {
+            timed_out = !sc_cond_timedwait(&d->cond, &d->mutex, d->deadline);
+        }
+        sc_mutex_unlock(&d->mutex);
+
+        if (!d->interrupted) {
```
anotheruserofgithub: `d->interrupted` is guarded by the mutex in `sc_disconnect_interrupt()` thus I would copy its value while the mutex is still locked in order to avoid a data race:
```suggestion
        bool interrupted = d->interrupted;
        sc_mutex_unlock(&d->mutex);

        if (!interrupted) {
```

But maybe you can simply check `timed_out` here?
```suggestion
        sc_mutex_unlock(&d->mutex);

        if (timed_out) {
```

Even if both `d->interrupted` and  `timed_out` are true, which can happen when the user quits the app at the same moment the timeout expires, posting a superfluous `SC_EVENT_DISCONNECTED_TIMEOUT` event won't do any harm as `sc_screen_handle_disconnection()` will have already returned by `SDL_EVENT_QUIT`. Whichever of these two events happens first on the event queue will win, so there will be no extra logging. Am I missing a critical point?

app/src/disconnect.c
```diff
@@ -0,0 +1,89 @@
...
+    struct sc_disconnect *d = userdata;
+
+    SDL_Surface *icon = sc_icon_load(SC_ICON_FILENAME_DISCONNECTED);
+    if (icon) {
+        d->cbs->on_icon_loaded(d, icon, d->cbs_userdata);
+    } else {
+        LOGE("Could not load disconnected icon");
+    }
+
+    if (d->deadline != SC_TICK_NONE) {
```
anotheruserofgithub: I assume `SC_TICK_NONE` means immediate timeout, not infinite, right? So there is a missing `else` branch for the timeout callback:
```c
    if (d->deadline != SC_TICK_NONE) {
        ...
    } else {
        d->cbs->on_timeout(d, d->cbs_userdata);
    }
```

Otherwise `sc_screen_handle_disconnection()` would wait forever.

app/src/disconnect.c
```diff
@@ -0,0 +1,89 @@
...
+        bool timed_out = false;
+        while (!d->interrupted && !timed_out) {
+            timed_out = !sc_cond_timedwait(&d->cond, &d->mutex, d->deadline);
+        }
+        sc_mutex_unlock(&d->mutex);
+
+        if (!d->interrupted) {
+            d->cbs->on_timeout(d, d->cbs_userdata);
+        }
+    }
```
anotheruserofgithub: Then it should be possible to rearrange the code slightly such as to have a single occurrence of `d->cbs->on_timeout()`. Assuming my previous comments are correct, something along these lines perhaps:
```suggestion
    bool timed_out;

    if (d->deadline == SC_TICK_NONE) {
        timed_out = true;
    } else {
        timed_out = false;
        sc_mutex_lock(&d->mutex);
        while (!d->interrupted && !timed_out) {
            timed_out = !sc_cond_timedwait(&d->cond, &d->mutex, d->deadline);
        }
        sc_mutex_unlock(&d->mutex);
    }

    if (timed_out) {
        d->cbs->on_timeout(d, d->cbs_userdata);
    }
```

anotheruserofgithub (commented):
Hi, first I noticed a few possible issues.

app/src/disconnect.c
```diff
@@ -0,0 +1,89 @@
...
+    return 0;
+}
+
+bool
+sc_disconnect_start(struct sc_disconnect *d, sc_tick deadline,
+                    const struct sc_disconnect_callbacks *cbs,
+                    void *cbs_userdata) {
+    bool ok = sc_mutex_init(&d->mutex);
+    if (!ok) {
+        return false;
```
anotheruserofgithub, line 40: What would you think of treating `sc_mutex_init()` failure like `sc_cond_init()` failure with a `goto` statement?
```suggestion
        goto error_return;
```
And adding the `error_return` label before the final return of this function.

app/src/util/tick.h
```diff
@@ -6,6 +6,7 @@
 #include <stdint.h>
 
 typedef int64_t sc_tick;
+#define SC_TICK_NONE INT64_MIN
 #define PRItick PRIi64
 #define SC_TICK_FREQ 1000000 // microsecond
```
anotheruserofgithub, line 10: Would it look better by grouping both `SC_TICK_****` values?
```suggestion
typedef int64_t sc_tick;
#define PRItick PRIi64
#define SC_TICK_NONE INT64_MIN
#define SC_TICK_FREQ 1000000 // microsecond
```

anotheruserofgithub (commented):
Then some possible (debatable) improvements.

app/src/disconnect.c
rom1v: `SC_TICK_NONE` was intended to mean "no timeout" (keep the window open until the user closes).

But in the current version, it's not used, so I completely removed `SC_TICK_NONE`).

app/src/screen.c
rom1v: Thank you, done.

app/src/disconnect.c
rom1v: Ah yes, thank you, fixed :+1:

I prefer the first version (semantically, if we know the interruption already occurred, we don't want to call the callback), but in this case it's basically equivalent.

app/src/disconnect.c
rom1v: (not necessary since there is no `else` branch)

app/src/util/tick.h
rom1v: `SC_TICK_NONE` is removed.

app/src/disconnect.c
rom1v: I usually return directly for the first error, but both would be ok.

app/src/screen.c
```diff
@@ -903,6 +974,46 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
+                LOGD("Closing after device disconnection");
+                goto end;
+            case SDL_EVENT_QUIT:
+                LOGD("User requested to quit");
+                goto end;
+        }
+    }
+
+end:
+    sc_screen_interrupt_disconnect(screen);
```
anotheruserofgithub:
```suggestion
            case SC_EVENT_DISCONNECTED_TIMEOUT:
                LOGD("Closing after device disconnection");
                return;
            case SDL_EVENT_QUIT:
                LOGD("User requested to quit");
                sc_screen_interrupt_disconnect(screen);
                return;
        }
    }
```

I think it's semantically wrong to mark the disconnection process as interrupted in case it was actually ended by a timeout. When the timeout expires, the thread will complete cleanly, there is no need for signaling.

Imagine you want to inspect the `screen`'s `disconnect` struct for debugging an abnormal closing of the app. Assume `SDL_EVENT_QUIT` never happens. When the timeout expires, the "scrcpy-dis" thread will push a `SC_EVENT_DISCONNECTED_TIMEOUT` event and terminate. Then this event triggers a call to `sc_disconnect_interrupt()`, which will mark `screen->disconnect.interrupted` as true even though the "scrcpy-dis" thread already terminated. From now on you could believe that this thread was interrupted in the middle of its `run()`, which is wrong, and you can't distinguish this case from an actual `SDL_EVENT_QUIT` anymore.

Of course you could also have a data race in the sense that both the timeout and the interruption would be triggered around the same time. In this case `screen->disconnect.interrupted` would be true even though the timeout already expired, but it wouldn't be semantically wrong since the user actually requested to quit. And if you really wanted to store the two states you could add `timed_out` in `struct sc_disconnect` alongside `interrupted`.

anotheruserofgithub (commented):
One last suggested improvement from my side.

app/src/screen.c
rom1v: Yes, that makes sense.

The purpose of `sc_screen_interrupt_disconnect()` is just to make sure the thread is not in a blocking call when it is joined, so I wanted to call it unconditionally when returning from `sc_screen_handle_disconnection()` (even when not necessary).

But indeed, since there are only two cases where the function returns, one of them where `sc_screen_interrupt_disconnect()` is unnecessary, we can simplify. I applied the change.

Thank you for your in-depth review.

rom1v:
When the disconnected icon is shown, we should still handle events (to leave fullscreen mode for example).

I will need to rework that before merging.

rom1v:
> I will need to rework that before merging.

Done. Ready to be merged.

(Please review and test if you can)

app/src/events.h
```diff
@@ -14,18 +14,20 @@ enum {
...
     SC_EVENT_TIME_LIMIT_REACHED,
     SC_EVENT_CONTROLLER_ERROR,
     SC_EVENT_AOA_OPEN_ERROR,
+    SC_EVENT_DISCONNECTED_ICON_LOADED,
+    SC_EVENT_DISCONNECTED_TIMEOUT,
 };
 
 bool
-sc_push_event_impl(uint32_t type, const char *name);
+sc_push_event_impl(uint32_t type, void* ptr, const char *name);
```
anotheruserofgithub: This is your convention in the `events` module:
```suggestion
sc_push_event_impl(uint32_t type, void *ptr, const char *name);
```

app/src/events.c
```diff
@@ -6,9 +6,13 @@
 #include "util/thread.h"
 
 bool
-sc_push_event_impl(uint32_t type, const char *name) {
-    SDL_Event event;
-    event.type = type;
+sc_push_event_impl(uint32_t type, void* ptr, const char *name) {
```
anotheruserofgithub:
```suggestion
sc_push_event_impl(uint32_t type, void *ptr, const char *name) {
```

app/src/screen.c
```diff
@@ -645,6 +660,17 @@ sc_screen_destroy(struct sc_screen *screen) {
     SDL_DestroyWindow(screen->window);
     sc_fps_counter_destroy(&screen->fps_counter);
     sc_frame_buffer_destroy(&screen->fb);
+
+    SDL_Event event;
+    int nevents = SDL_PeepEvents(&event, 1, SDL_GETEVENT,
+                                 SC_EVENT_DISCONNECTED_ICON_LOADED,
+                                 SC_EVENT_DISCONNECTED_ICON_LOADED);
+    if (nevents == 1) {
```
anotheruserofgithub, line 668: In `terminate_runnables_on_event_loop()` (`scrcpy.c`) you're doing it in a while-loop without temporary `nevents` variable, thus here even though it's an if-condition maybe `nevents` could be removed just for symmetry? But maybe it's just less readable after all.
```suggestion
    if (SDL_PeepEvents(&event, 1, SDL_GETEVENT,
                          SC_EVENT_DISCONNECTED_ICON_LOADED,
                          SC_EVENT_DISCONNECTED_ICON_LOADED) == 1) {
```

app/src/screen.c
```diff
@@ -915,6 +987,54 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
+    }
+
+    SDL_Event event;
+    while (SDL_WaitEvent(&event)) {
+        switch (event.type) {
+            case SC_EVENT_DISCONNECTED_ICON_LOADED: {
+                SDL_Surface *icon_disconnected = event.user.data1;
+                assert(icon_disconnected);
+
+                bool ok = sc_texture_set_from_surface(&screen->tex, icon_disconnected);
```
anotheruserofgithub: I think this line goes beyond your line length limit, but I don't know where to cut it.
```suggestion
                bool ok =
                  sc_texture_set_from_surface(&screen->tex, icon_disconnected);
```

app/src/screen.c
```diff
@@ -915,6 +987,54 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
...
+                    // not fatal
+                    LOGE("Could not set disconnected icon");
+                }
+
+                sc_icon_destroy(icon_disconnected);
+                break;
+            }
+            case SDL_EVENT_WINDOW_EXPOSED:
+                sc_screen_render(screen, true);
+                break;
```
anotheruserofgithub: What do you think about moving this case up at the top of the switch? Then you would have the two `DISCONNECTED` events one after the other.

anotheruserofgithub (commented):
Small details of style (just to nitpick).

anotheruserofgithub:
I still couldn't test for real, I'd need to buy a new phone… one day.

app/src/screen.c
rom1v: Sure :+1:

app/src/screen.c
rom1v: :+1:

app/src/screen.c
rom1v: I prefer the version with an explicit variable `nevents` (so that we immediately know what the result represents), than can be tested in an if-condition.

But in a while-condition, setting an explicit variable then testing it would not be as straightforward.

app/src/events.c
rom1v: Good catch. :+1:

Did you review "manually", or with some tool, to catch such mistakes?

app/src/events.h
rom1v: :+1:

app/src/events.c
anotheruserofgithub: Manually, though I looked at it multiple times without seeing it. Juste un coup de chance en l'occurrence.

app/src/screen.c
anotheruserofgithub: Agreed!

anotheruserofgithub:
I just noticed that since you renamed the icon in 8a4955ec57c92b0655f44c703310a45068e25d0e now the SVG image in `README.md` is broken (oops): https://github.com/Genymobile/scrcpy/blob/a4000a14879831bd61f935fccc5af896e3098dfb/README.md?plain=1#L7

rom1v:
@anotheruserofgithub Thank you, fixed: c2e1cf08848c8c8e640ed8c0c27eb12d53d0c773

rodrigofd:
> I replaced `SCRCPY_ICON_PATH` (which defined the unique scrcpy icon) with `SCRCPY_ICON_DIR` now that there are several icons.

Now that the `SCRCPY_ICON_PATH` env. var. is not used anymore; is there still a way to customize the main app icon when running a new instance?

rom1v:
> Now that the `SCRCPY_ICON_PATH` env. var. is not used anymore; is there still a way to customize the main app icon when running a new instance?

You can no longer cutomize a single icon. Instead, use `SCRCPY_ICON_DIR` to customize all the icons at once by passing a directory. If you want to change only one icon, copy all the others in that custom directory.

lbbboy:
3.3.4 This issue doesn't occur; however, if I position my phone vertically, this prompt appears and the app crashes. Also, with full-screen video, scrcpy automatically corrects the video orientation.

Issue #6248, opened 2025-07-21
Mistakes TCP as USB


szmarczak:
_Please read the [prerequisites] to run scrcpy._

[prerequisites]: https://github.com/Genymobile/scrcpy#prerequisites

_Also read the [FAQ] and check if your [issue][issues] already exists._

[FAQ]: https://github.com/Genymobile/scrcpy/blob/master/FAQ.md
[issues]: https://github.com/Genymobile/scrcpy/issues

## Environment

 - **OS:** Windows
 - **Scrcpy version:** 3.3.1
 - **Installation method:** From GitHub releases
 - **Device model:** Nothing Phone 2 (A065)
 - **Android version:** 15

## Describe the bug

`scrcpy` mistakes TCP connection as USB, resulting in UHID keyboard not working. If I connect the device via USB or specify `-e` it works. However, if I use the `adb-fc29a614-i3szUK._adb-tls-connect._tcp` connection (which is also over TCP) it mistakes it as USB.

```
>scrcpy.exe -d --new-display=1920x1080/192 -K --start-app=org.mozilla.firefox
scrcpy 3.3.1 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:         (tcpip)  192.168.1.113:39243             device  A065
INFO:     -->   (usb)  adb-fc29a614-i3szUK._adb-tls-connect._tcp            device  A065

>adb devices -l
List of devices attached
192.168.1.113:39243    device product:PongEEA model:A065 device:Pong transport_id:82
adb-fc29a614-i3szUK._adb-tls-connect._tcp device product:PongEEA model:A065 device:Pong transport_id:74
```

szmarczak:
https://github.com/Genymobile/scrcpy/blob/f01231dff8294fe2c99045a4f9a14b233a71bb86/app/src/adb/adb_device.c#L42

Looks like this is the fault

szmarczak:
https://android.googlesource.com/platform/packages/modules/adb/+/refs/tags/android-16.0.0_r2/adb_mdns.h#31

rom1v:
This might be sufficient to fix the problem:

```diff
diff --git app/src/adb/adb_device.c app/src/adb/adb_device.c
index 5ea8eb44a..c63c38cad 100644
--- app/src/adb/adb_device.c
+++ app/src/adb/adb_device.c
@@ -39,5 +39,9 @@ sc_adb_device_get_type(const char *serial) {
[...]

Issue #3537, opened 2022-10-14
Can not connect to device with leading space in device serial number


CameronBanga:
- [x] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [x] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: macOS
 - scrcpy version: [e.g. 1.24]
 - installation method: brew
 - device model: Glorystar GAD-101
 - Android version: 5.1

**Describe the bug**
See issue https://github.com/Genymobile/scrcpy/issues/3534#issuecomment-1279373227 for further information.

While trying to connect using scrcpy, I get the following:

```
scrcpy 1.24 <https://github.com/Genymobile/scrcpy>
2022-10-14 14:33:40.022 scrcpy[12626:145516] ERROR: Could not find any ADB device
2022-10-14 14:33:40.415 scrcpy[12626:145515] ERROR: Server connection failed
```

As determined by @rom1v, this is due to my device serial number starting with a leading space:

```
Downloads|⇒ adb -s GS101200400611 shell echo ok
adb: device 'GS101200400611' not found

Downloads|⇒ adb -s ' GS101200400611' shell echo ok
ok
```

rom1v:
Thank you for the report.

I started to work on supporting serial with spaces. I thought it was easy: `adb devices` uses a tab separator between the serial and the state, but `adb devices -l` uses spaces, so a space in the serial is theoretically indistinguishable from a separator. We could make a special case for leading spaces, but it seems odd :/

CameronBanga:
If it helps anyone who happens to stumble upon this, as an interim fix, I have the following script that resolves issue. Have to run again after a power cycle though:

```
#!/bin/bash

adb root
adb shell 'echo -n "WhateverSerialNumberYouWantHere" > /sys/class/android_usb/android0/iSerial'
adb kill-server && adb start-server

# After Completing, you need to unplug and then re-plug in your USB cable to device.
```

rom1v:
Fixed by #6664 (see https://github.com/Genymobile/scrcpy/issues/6248#issuecomment-3874097635).

PR #6663, opened 2026-02-09
Properly serialize command line arguments for Windows

rom1v:
Scrcpy exposes a process API to execute commands:

https://github.com/Genymobile/scrcpy/blob/c8c1316db9adefca6aae054be936049c2f8e02f1/app/src/util/process.h#L72-L95

In practice, it is only used to execute simple `adb` commands, so the Windows implementation was very simple, but very incomplete:

https://github.com/Genymobile/scrcpy/blob/c8c1316db9adefca6aae054be936049c2f8e02f1/app/src/sys/win/process.c#L12-L24

In order to support device serials containing spaces (#3537, #6248), properly quote all arguments so that a parser that follows [the rules](https://learn.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments) correctly splits the arguments.

Please test/review. In particular, the implementation must not be vulnerable to command injection (make a valid parser incorrectly parse the serialized data).

(cc @yume-chan maybe?)

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -75,6 +75,7 @@ conf.set('_GNU_SOURCE', true)
 if host_machine.system() == 'windows'
     windows = import('windows')
     src += [
+        'src/util/command.c',
         'src/sys/win/file.c',
         'src/sys/win/process.c',
         windows.compile_resources('scrcpy-windows.rc'),
@@ -237,6 +238,12 @@ if get_option('buildtype') == 'debug'
             'src/util/str.c',
             'src/util/strbuf.c',
             'src/util/term.c',
+        ]],
+        ['test_command_windows', [
+          'tests/test_command_windows.c',
+          'src/util/command.c',
+          'src/util/str.c',
+          'src/util/strbuf.c',
         ]],
         ['test_control_msg_serialize', [
             'tests/test_control_msg_serialize.c',
--- a/app/src/adb/adb.c
+++ b/app/src/adb/adb.c
@@ -331,30 +331,11 @@ sc_adb_reverse_remove(struct sc_intr *intr, const char *serial,
 bool
 sc_adb_push(struct sc_intr *intr, const char *serial, const char *local,
             const char *remote, unsigned flags) {
-#ifdef _WIN32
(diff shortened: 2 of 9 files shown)


yume-chan:
Looks like I also need #6664 to test devices with spaces in serial?

Everything works except `LOGD` can't print non-ASCII characters:

```text
 ./scrcpy -Vverbose --no-audio --no-control
scrcpy 3.3.4 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  adb-bd9d00cc-irloe7 (2)._adb-tls-connect._tcp            device  23117RK66C
DEBUG: Device serial: adb-bd9d00cc-irloe7 (2)._adb-tls-connect._tcp
DEBUG: Using server (portable): C:\Users\simon\Downloads\test testµ╡ïΦ»ò\scrcpy-server
C:\Users\simon\Downloads\test test测试\scrcpy-server: 1 file pushed, 0 skipped. 44.9 MB/s (93312 bytes in 0.002s)       
[server] INFO: Device: [Xiaomi] Redmi 23117RK66C (Android 16)
```

(Notice scrcpy prints `Downloads\test testµ╡ïΦ»ò\` and ADB prints `Downloads\test test测试\`)

Googled and this fixes the issue:

```patch
diff --git a/app/src/util/log.c b/app/src/util/log.c
index c583b8fb..470f1e67 100644
--- a/app/src/util/log.c
+++ b/app/src/util/log.c
@@ -151,6 +151,10 @@ sc_sdl_log_print(void *userdata, int category, SDL_LogPriority priority,
 
 void
 sc_log_configure(void) {
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     SDL_SetLogOutputFunction(sc_sdl_log_print, NULL);
     // Redirect FFmpeg logs to SDL logs
     av_log_set_callback(sc_av_log_callback);

```

rom1v:
Thank you :+1: bb6d1c634834d6262d8daee6a162ce53311c976f

PR #6664, opened 2026-02-09
Improve adb devices parsing

### LINKED ISSUE #3537: Can not connect to device with leading space in device serial number

rom1v:
`adb devices -l` prints one device per line, containing, separated by spaces:
 - the device serial,
 - the device state,
 - a list of key:value pairs.

However, the device serial itself may contain spaces, making a simple split ambiguous.

To avoid ambiguity, parse the string backwards:
 - first, parse all the trailing values until the device state, identified using a list of well-known values;
 - finally, treat the remaining leading token as the device serial.

Refs #6248
Fixes #3537

### DIFF
--- a/app/src/adb/adb_parser.c
+++ b/app/src/adb/adb_parser.c
@@ -8,6 +8,49 @@
 #include "util/log.h"
 #include "util/str.h"
 
+static size_t
+rstrip_len(const char *s, size_t len) {
+    size_t i = len;
+
+    // Ignore trailing whitespaces
+    while (i > 0 && (s[i-1] == ' ' || s[i-1] == '\t')) {
+        --i;
+    }
+
+    return i;
+}
+
+static void
+locate_last_token(const char *s, size_t len, size_t *start, size_t *end) {
+    size_t i = rstrip_len(s, len);
+    *end = i; // excluded
+
+    // The token contains non-whitespace chars
+    while (i > 0 && (s[i-1] != ' ' && s[i-1] != '\t')) {
+        --i;
+    }
+
+    *start = i; // included
+}
+
+static bool
+is_device_state(const char *s) {
+    // <https://android.googlesource.com/platform/packages/modules/adb/+/1cf2f017d312f73b3dc53bda85ef2610e35a80e9/adb.cpp#144>
+    // "device", "unauthorized" and "offline" are the most common states, so
+    // check them first.
+    return !strcmp(s, "device")
+        || !strcmp(s, "unauthorized")
+        || !strcmp(s, "offline")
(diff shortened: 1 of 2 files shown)


app/src/adb/adb_parser.c
```diff
@@ -25,64 +50,54 @@ sc_adb_parse_device(char *line, struct sc_adb_device *device) {
...
+        }
+
+        const char *token = &line[start];
+        line[end] = '\0';
+
+        if (!strncmp("model:", token, sizeof("model:") - 1)) {
+            model = &token[sizeof("model:") - 1];
+            // We only need the model
+        } else if (!strchr(token, ':')) {
+            // The first non-key:value token, it's the device state
```
yume-chan: When using libusb backend (`$ADB_LIBUSB=1` or default on some [older versions](https://developer.android.com/tools/releases/platform-tools#3500_february_2024)) the devpath doesn't have a colon:

```txt
$ adb devices -l
List of devices attached
bd9d00cc               device 3-1 product:manet model:23117RK66C device:manet transport_id:2
```

I check if the token is one of the known states:

https://github.com/yume-chan/ya-webadb/blob/c0d98c6421677552cfeb0d194ddd41ba1f150499/libraries/adb/src/server/client.ts#L117-L123

rom1v: > When using libusb backend (`$ADB_LIBUSB=1` or default on some [older versions](https://developer.android.com/tools/releases/platform-tools#3500_february_2024)) the devpath doesn't have a colon:

Thank you!

> I check if the token is one of the known states:

Yes, this is what I had in mind if more edge cases came up later, but I didn't think it was already a problem in practice.

Fixed the PR.

PR #6667, opened 2026-02-10
Keep Windows terminal open on error by default

rom1v:
If scrcpy is launched by double-clicking `scrcpy.exe` in Windows Explorer, automatically set `--pause-on-exit=if-error`.

Without this, the terminal would close immediately, preventing the user from seeing the error.

Also remove `scrcpy-console.bat`, which is now useless.

---

Detecting whether it is launched by double-clicking relies on a heuristic (it is not 100% reliable). Basically it checks how many processes share the console. If there is only one, then it's probably launched by double-clicking from Windows Explorer. In practice, it works:
 - running by double-clicking on `scrcpy.exe` implicitly sets `--pause-on-exit=if-error`
 - running from a terminal or a `.bat` does not implicit pause on exit.

Like #6662, the goal is to avoid getting a window that disappears without knowing why, which could suggest that a crash occurred (even though there is no crash).

### DIFF
--- a/doc/windows.md
+++ b/doc/windows.md
@@ -72,18 +72,6 @@ Documentation for command line arguments is available:
  - `scrcpy --help`
  - on [github](/README.md)
 
-To start scrcpy directly without opening a terminal, double-click on one of
-these files:
- - `scrcpy-console.bat`: start with a terminal open (it will close when scrcpy
-   terminates, unless an error occurs);
- - `scrcpy-noconsole.vbs`: start without a terminal (but you won't see any error
-   message).
-
-_Avoid double-clicking on `scrcpy.exe` directly: on error, the terminal would
-close immediately and you won't have time to read any error message (this
-executable is intended to be run from the terminal). Use `scrcpy-console.bat`
-instead._
-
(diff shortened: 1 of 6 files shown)


doc/windows.md
```diff
@@ -92,9 +80,17 @@ example:
...
 [show file extensions]: https://www.howtogeek.com/205086/beginner-how-to-make-windows-show-file-extensions/
 
-Then just double-click on that file.
+Then just double-click on that file to run it.
 
-You could also edit (a copy of) `scrcpy-console.bat` or `scrcpy-noconsole.vbs`
-to add some arguments.
+To start scrcpy without opening a terminal, double-click `scrcpy-noconsole.vbs`
+(note that errors won't be shown). To pass arguments, edit (a copy of)
+`scrcpy-noconsole.vbs` add and the desired arguments.
```
anotheruserofgithub, line 96:
```suggestion
`scrcpy-noconsole.vbs` and add the desired arguments.
```

rom1v: Thank you :+1: Merged 10e0c3226c3a3a9078c86a42c931d9b288e035f0

PR #6694, opened 2026-02-25
Open the scrcpy window earlier

### LINKED ISSUE #6546: The program doesn't work at all when used with the '--no-power-on' on Android 16
Computer OS: Debian 14 (AKA the testing repository) with Plasma 6.5.4 (Wayland session)
Phone OS: LineageOS 23.0 (AKA Android 16)
Program: scrcpy-linux-x86_64-v3.3.4

Tested on a Samsung and a Xiaomi phone, both running the same LineageOS 23.0 version from 2-3 months ago.
On both, running scrcpy with the '--no-power-on' argument, the scrcpy execution blocks, until I double tap the phones' screen to turn on their screens.

Immediately as I do these double taps and the screen turns on, then the scrcpy's window (mirroring) appears and I can proceed with unlocking them.
Same thing happening if I press once the power button instead of double tapping the screen.

I noticed this problem since v3.3.3 of scrcpy, but pretty much in the same time interval I also upgraded from Lineage 22 (Android 15) to 23 (Android 16) so I don't know which one is the culprit for this new problem.
As I have been using this argument for probably more than a year.

I've searched and looked for a while at other issues, but I could not tell if any of them are exactly what I'm experiencing.

BTW, after the server info showing the device's name and Android version these are the last lines before blocking and waiting for the phone to be touched and manually turn on the screen:
```
INFO: Renderer: opengl
INFO: OpenGL version: 4.6 (Compatibility Profile) Mesa 25.2.8-2+b3
INFO: Texture: 720x1520

```
Until this is fixed somewhere, it would be great to at least show a message like:
"Waiting for the screen to be manually turned on... (try to double tap it or press the power button)"

Thank you!

rom1v:
Previously, the window was opened only once the first frame was available. Instead, open it as soon as the frame dimensions are known.
    
This is especially useful when the encoder does not produce an initial frame while the phone is turned off (e.g., with `--no-power-on`).
    
Fixes #6546


app/src/screen.c
```diff
@@ -198,19 +198,31 @@ sc_screen_update_content_rect(struct sc_screen *screen) {
...
     sc_sdl_render_clear(renderer);
 
     bool ok = false;
     SDL_Texture *texture = screen->tex.texture;
     if (!texture) {
-        LOGW("No texture to render");
+        // Draw a dark 10x10 square in the top-right corner to distinguish a
+        // black frame from the absence of a frame
+        struct sc_size render_size =
+            sc_sdl_get_render_output_size(screen->renderer);
```
anotheruserofgithub:
```suggestion
        struct sc_size render_size = sc_sdl_get_render_output_size(renderer);
```

app/src/screen.c
```diff
@@ -292,6 +304,19 @@ sc_screen_frame_sink_open(struct sc_frame_sink *sink,
         return false;
     }
 
+    // content_size can be written from this thread, because it is never read
+    // from the main thread before handling SC_EVENT_OPEN_WINDOW (which acts as
+    // a synchronization point) when video is enabled
+    screen->frame_size.width = session->video.width;
```
anotheruserofgithub, line 306: Casts to `(void)` for unused variables can be removed in this function since `ctx`, `session` and `screen` are now used.

app/src/screen.c
```diff
@@ -292,6 +304,19 @@ sc_screen_frame_sink_open(struct sc_frame_sink *sink,
...
+    screen->frame_size.height = session->video.height;
+    screen->content_size = get_oriented_size(screen->frame_size,
+                                             screen->orientation);
+
+    bool ok = sc_push_event(SC_EVENT_OPEN_WINDOW);
+    if (!ok) {
+        return false;
+    }
```
anotheruserofgithub, line 314: What if pushing the event fails? Will the app quit? Else, will the first frame (and next ones) be able to render or will there be just no window at all?

app/src/screen.c
```diff
@@ -854,9 +863,18 @@ sc_screen_resize_to_pixel_perfect(struct sc_screen *screen) {
...
+        case SC_EVENT_OPEN_WINDOW: {
+            sc_screen_show_initial_window(screen);
+
+            if (sc_screen_is_relative_mode(screen)) {
+                // Capture mouse on start
+                sc_mouse_capture_set_active(&screen->mc, true);
+            }
+
+            sc_screen_render(screen, false);
+            break;
```
anotheruserofgithub:
```suggestion
            return;
```
Don't know if input events should be handled after this; all other cases return from this switch.

anotheruserofgithub (commented):
Just a quick code review, I didn't test anything.

app/src/screen.c
anotheruserofgithub: Also notice that braces shouldn't be needed around this case.

app/src/screen.c
rom1v: >What if pushing the event fails? Will the app quit?

Yes. Tested with:

```diff
diff --git a/app/src/screen.c b/app/src/screen.c
index d47b96338..f1b9ff2b8 100644
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -309,7 +309,7 @@ sc_screen_frame_sink_open(struct sc_frame_sink *sink,
                                              screen->orientation);
 
     bool ok = sc_push_event(SC_EVENT_OPEN_WINDOW);
-    if (!ok) {
+    if (true||!ok) {
         return false;
     }
 
```

rom1v:
@anotheruserofgithub Thank you for your review, I applied your suggestions/fixes. :+1:

PR #6705, opened 2026-03-04
Add adaptive virtual display that adjusts to window size.

TinuraD:
**I noticed some developers have made GUI or companion apps using scrcpy, so being able to open apps in resizable windows would be cool. SO I’ve just added an adaptive virtual display option.** 

- Instead of using `--new-display`, you can now also use `--adaptive-new-display` too, which provides an adaptive virtual display that resizes according to your current window.
- It also includes `--adaptive-scale`, which allows you to change the scale (default is 1).  [Eg: 0.5, 2.0]
- This basically changes the virtual display's resolution and calculates an adaptive DPI based on your current PC screen resolution.  
- All existing functions remain untouched, so `--new-display` still works as before.  
- I attached two video clips to show how it works.
- But i didn't update documentations.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

https://github.com/user-attachments/assets/73a1cd9a-b85e-497d-a2f2-ebcf42aba408

https://github.com/user-attachments/assets/150c0489-d886-4e4f-9fad-87fb0c6ed386


TinuraD:
P.S. I noticed that someone implemented the same feature in [#6351]. Since it hasn't been merged yet, I decided to reopen my PR too. I hope one of these will be added to scrcpy soon.

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -266,4 +278,33 @@ private static int scaleDpi(Size initialSize, int initialDpi, Size size) {
...
+            try {
+                virtualDisplay.resize(width, height, newDpi);
+                displaySizeMonitor.setSessionDisplaySize(requestedDisplaySize);
+                Ln.i("Virtual display resized in-place to " + width + "x" + height + "/" + newDpi);
+            } catch (Exception e) {
+                Ln.w("Virtual display resize failed, fallback to recreate", e);
+                displaySizeMonitor.stopAndRelease();
+                virtualDisplay.release();
+                virtualDisplay = null;
+            }
```
LHLaurini, line 302: The whole point of `resize` is to "to allow applications using virtual displays to adapt to changing conditions without having to tear down and recreate the display" (from the docs). Is this really necessary?

IMO unless this happens in some devices, you should just show a warning and keep going

app/src/screen.c
```diff
@@ -39,6 +72,60 @@ get_window_size(const struct sc_screen *screen) {
...
+    if (!screen->im.controller) {
+        return;
+    }
+    struct sc_size size = get_window_size(screen);
+    if (size.width == 0 || size.height == 0) {
+        return;
+    }
+    if (size.width < 64 || size.height < 64) {
+        return;
+    }
```
LHLaurini: Redundant...

app/src/screen.c
```diff
@@ -811,6 +934,7 @@ sc_screen_handle_event(struct sc_screen *screen, const SDL_Event *event) {
             return true;
         }
         case SC_EVENT_NEW_FRAME: {
+            sc_screen_maybe_send_vd_resize(screen);
```
LHLaurini: You probably only need one `sc_screen_maybe_send_vd_resize` here. I'd keep the one in `SC_EVENT_NEW_FRAME`.

app/src/cli.c
```diff
@@ -632,6 +634,22 @@ static const struct sc_option options[] = {
                 "    --new-display         # main display size and density\n"
                 "    --new-display=/240    # main display size and 240 dpi",
     },
+    {
+        .longopt_id = OPT_ADAPTIVE_NEW_DISPLAY,
+        .longopt = "adaptive-new-display",
+        .argdesc = "[<width>x<height>][/<dpi>]",
```
LHLaurini, line 640: This dpi value is almost immediately ignored in favor of the one from `adaptive-scale`

app/src/screen.c
```diff
@@ -26,6 +27,38 @@ get_oriented_size(struct sc_size size, enum sc_orientation orientation) {
     return oriented_size;
 }
 
+static uint16_t
+get_window_dpi(const struct sc_screen *screen) {
+    if (screen->adaptive_new_display) {
```
LHLaurini, line 36: Maybe you meant to check `vd_scale` here? This is always `true`.

app/src/screen.c
```diff
@@ -39,6 +72,60 @@ get_window_size(const struct sc_screen *screen) {
...
+    struct sc_size size = get_window_size(screen);
+    if (size.width == 0 || size.height == 0) {
+        return;
+    }
+    if (size.width < 64 || size.height < 64) {
+        return;
+    }
+    screen->vd_resize_size = size;
+    screen->vd_resize_pending = true;
+    screen->vd_resize_deadline_ms = SDL_GetTicks() + 250;
```
LHLaurini: Use a constant for this value

app/src/screen.c
```diff
@@ -491,9 +587,22 @@ sc_screen_show_initial_window(struct sc_screen *screen) {
...
+        } else {
+            window_size = get_initial_optimal_size(screen->content_size,
+                                                   screen->req.width,
+                                                   screen->req.height);
+        }
+    } else {
+        window_size = get_initial_optimal_size(screen->content_size,
+                                               screen->req.width,
+                                               screen->req.height);
+    }
```
LHLaurini, line 608: Could you explain the goal of this change?

app/src/screen.c
```diff
@@ -557,6 +666,14 @@ static void
...
     assert(screen->video);
 
+    if (screen->new_display) {
+        // For new-display, keep the window size controlled by the user.
+        // Do not auto-resize the window to the content size to avoid
+        // feedback loops with virtual display resizing.
+        screen->content_size = new_content_size;
+        return;
+    }
+
```
LHLaurini, line 679: Is this really a problem? Unless I'm mistaken, shouldn't the check in lines 108-110 avoid that?

server/src/main/java/com/genymobile/scrcpy/control/Controller.java
```diff
@@ -146,6 +149,11 @@ public void onNewVirtualDisplay(int virtualDisplayId, PositionMapper positionMap
                 displayDataAvailable.notify();
             }
         }
+        if (pendingRelaunchOnResize && lastStartedAppPackage != null) {
+            pendingRelaunchOnResize = false;
+            Ln.i("Relaunching app \"" + lastStartedAppPackage + "\" on resized display " + virtualDisplayId + "...");
+            Device.startApp(lastStartedAppPackage, virtualDisplayId, lastStartedAppForceStop);
+        }
```
LHLaurini: You shouldn't relaunch the app if you were able to resize the display successfully, as that causes some apps to restart (particularly PWAs)

LHLaurini (changes requested):
Thank you for your PR!

I left you some comments. I personally find your implementation cleaner than the other one, although that one works a bit better at the moment.

Your implementation makes the video look a bit blurry, probably because the server does some rounding to the dimensions.

app/src/screen.c
TinuraD: Basically, the resizing function isn’t tied to the initial startup, as it’s mostly based on normal new-display behavior. This allows the virtual display to have adaptive window size on initial launch following px to dpi conversion or using custom arguments.

app/src/screen.c
TinuraD: This is to avoid issues when Android's rotation causes the virtual display to resize. Some Android apps, like video players, change the rotation while the app is in use, such as when opening a video. Even though the window appears the same, it still makes a very small size change as i noticed which can create loops.

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
TinuraD: Removed and added the warning.

app/src/screen.c
TinuraD: Removed

app/src/screen.c
TinuraD: Changed. Kept the one you did recommend.

app/src/cli.c
TinuraD: Since scaling can basically handle dpi I thought it wasn’t needed. But it makes sense to add a known dpi value. So I fixed it.

app/src/screen.c
TinuraD: Removed

app/src/screen.c
TinuraD: Done

server/src/main/java/com/genymobile/scrcpy/control/Controller.java
TinuraD: Changed to not restart

TinuraD:
Hey, thanks for reviewing the PR. I’ve checked your comments and updated the required files.

TinuraD:
Superseded by PR #6772

PR #6723, opened 2026-03-15
Created scrcpy-console-easy-wireless.bat

Sidtropy:
Added a .bat file to allow easy wireless connection to Android devices. 

The batch script prompts for an [IP: port] and remembers the last used [IP: port] for simplifying re-connections (just hit 'Enter' to skip through the prompts to use the last recorded values). 

NOTE: It seems like Android devices are updating the connection port more frequently and only connecting via the nominated port in the wireless debugging settings. I have observed this behaviour on more recent Android versions (Android 16, Build number CP1A.260305.018 onwards).

### DIFF
--- a/app/data/scrcpy-console-easy-wireless.bat
+++ b/app/data/scrcpy-console-easy-wireless.bat
@@ -1,0 +1,164 @@
+@echo off
+setlocal ENABLEDELAYEDEXPANSION
+
+:: =============
+:: PAIR FIRST
+:: =============
+:: If the connection fails, pair this computer to the android device first:
+:: On the android device:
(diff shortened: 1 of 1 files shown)


rom1v:

Sidtropy:
That's a fair response. 

Scrcpy already ships with a few scripts by default, I just thought I'd share a somewhat user-friendly one, since a lot of users may not be devs. 

The reason I submitted this: There is a significant difficulty for most users when first learning how to use Scrcpy. 

This script alleviates that to some extent. I wanted to provide a low friction path: run the script, follow the instructions, and connect to Android phone/device.

PR #6725, opened 2026-03-15
doc/v4l2: add an example for how to create a loopback device

Managor:


doc/v4l2.md
```diff
@@ -28,6 +28,12 @@ If you encounter problems detecting your device with Chrome/WebRTC, you can try
 sudo modprobe v4l2loopback exclusive_caps=1
 ```
 
+If a loopback device was not created automatically, you can make a new one:
```
rom1v, line 31: Why would it not be created automatically?

Managor: I'm guessing what is happening on my laptop is that Iriun Webcam hogs the default loopback device that is automatically created and names it after itself. After that there was no generically named loopback device on the list, hence my misunderstanding

```
$ sudo modprobe v4l2loopback
$ v4l2-ctl -A
Iriun Webcam (platform:v4l2loopback-000):
        /dev/video0

Integrated Camera: Integrated C (usb-0000:00:14.0-8):
        /dev/video1
        /dev/video2
        /dev/media0
```

I'll leave this up to you to decide if this note should be included

rom1v:
Thank you, merged: 158cae323787661ac7563e5bf5ce40be3e096445

doc/v4l2.md
```diff
@@ -28,6 +28,12 @@ If you encounter problems detecting your device with Chrome/WebRTC, you can try
...
+
+```
```
anotheruserofgithub, line 33:
```suggestion
```bash
```
Code block language identifier for the comment to be shown in gray: https://github.com/Genymobile/scrcpy/blob/158cae323787661ac7563e5bf5ce40be3e096445/doc/v4l2.md?plain=1#L43-L46 Missing here too: https://github.com/Genymobile/scrcpy/blob/158cae323787661ac7563e5bf5ce40be3e096445/doc/v4l2.md?plain=1#L27-L29

rom1v: Thank you, fixed: 4671927c34ce390e3f3c0044187b31c910f1c11e :+1:

PR #6746, opened 2026-03-31
Use optimal size alignment

### LINKED ISSUE #4949: The resolution goes from 1366x768 to 1360x768

**Environment**
 - OS: Debian
 - scrcpy version: 2.4
 - installation method: manual build
 - device model: SM-918B
 - Android version: 14

**Describe the bug**

The screen resolution is 1366x768, but the texture resolution is set to 1360x768? This makes the screen look really blurry with no way to fix it.

The screen is a virtual screen created from an option called 'Simulate secondary displays' in 'Developer options` with 1080p resolution. Then the resolution is changed through an app called 'SecondScreen'  from the play store(I could give more info on the app if needed).

I looked through other people's logs and they also seem to be doing the same thing (only on this resolution).

Is there a way to set it to the correct resolution?

Also thank you for the awesome software!

Output of `scrcpy --list-displays`:
```
scrcpy 2.4 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  R5CWC05MXTX                     device  SM_S918B
/usr/local/share/scrcpy/scrcpy-server: 1 file pushed. 4.5 MB/s (69007 bytes in 0.015s)
[server] INFO: Device: [samsung] samsung SM-S918B (Android 14)
[server] INFO: List of displays:
    --display-id=0    (1440x3088)
    --display-id=27    (1366x768)

```

Terminal log:
```
scrcpy 2.4 <https://github.com/Genymobile/scrcpy>
INFO: ADB device found:
INFO:     -->   (usb)  R5CWC05MXTX                     device  SM_S918B
/usr/local/share/scrcpy/scrcpy-server: 1 file pushed. 7.4 MB/s (69007 bytes in 0.009s)
[server] INFO: Device: [samsung] samsung SM-S918B (Android 14)
[server] WARN: Display doesn't have FLAG_SUPPORTS_PROTECTED_BUFFERS flag, mirroring can be restricted
INFO: Renderer: opengl
INFO: OpenGL version: 4.6 (Compatibility Profile) Mesa 24.0.3-1pop1~1711635559~22.04~7a9f319
INFO: Texture: 1360x768
[server] WARN: Could not get initial audio timestamp

```
image
Im not sure if it is visible from the screenshot :P But you can tell that it is not the right resolution from the tiny 6 pixels worth of black bars around the screenshot.

### LINKED ISSUE #6236: Resolution recognized by scrcpy is not the same as the actual resolution
OS: [Win 11]
Scrcpy version: [3.3.1]
**Device model:Redmagic 10 air
Android version: [15]

My phone's resolution is 2480*1116, but scrcpy recognizes it as 2480*1120, which results in two extra 2 pixel wide black borders on either side of the window. It doesn't affect actual use much, just wondering why.

rom1v:
The video was always constrained to use a size that is a multiple of 8.

This was sometimes not necessary (recent codecs only require a video size that is a multiple of 2 or even 1) and sometimes insufficient (some codecs require a size that is multiple of 16).

Use the size alignment required by the codec.

Also provide an option (`--min-size-alignment`) to force a minimal size alignment. This is a power-of-2 value (1, 2, 4, 8 or 16) that the video width and height must be a multiple of.

The actual alignment will be the maximum of this value and the video codec's alignment requirement.

---

Fixes #4949
Fixes #6236

This also fixes cases where the encoder has a multiple-of-16 alignment requirement, and avoids unnecessary constraints when the alignment requirements are lower.

_Spoiler: this will also be useful for resizable virtual displays._

### DIFF
--- a/doc/video.md
+++ b/doc/video.md
@@ -30,6 +30,17 @@ If encoding fails, scrcpy automatically tries again with a lower definition
 For camera mirroring, the `--max-size` value is used to select the camera source
 size instead (among the available resolutions).
 
+The size is rounded to a multiple of the _alignment_ required by the encoder, a
+power-of-2 value (1, 2, 4, 8 or 16) that the video width and height must be
+multiples of.
+
+The alignment can be forced to a minimum value. For instance, to force the width
+and height to be multiples of 8:
+
+```bash
+scrcpy --min-size-alignment=8
+```
+
 
 ## Bit rate
 
--- a/server/src/main/java/com/genymobile/scrcpy/device/Size.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Size.java
@@ -31,7 +31,6 @@ public final class Size {
 
     public Size limit(int maxSize) {
         assert maxSize >= 0 : "Max size may not be negative";
-        assert maxSize % 8 == 0 : "Max size must be a multiple of 8";
 
         if (maxSize == 0) {
             // No limit
@@ -55,13 +54,14 @@ public final class Size {
     }
 
     /**
-     * Round both dimensions of this size to be a multiple of 8 (as required by many encoders).
+     * Round both dimensions of this size to be multiples of {@code alignment}.
      *
-     * @return The current size rounded.
+     * @param alignment the required alignment
+     * @return the current size rounded
      */
-    public Size round8() {
-        if (isMultipleOf8()) {
-            // Already a multiple of 8
+    public Size round(int alignment) {
+        if (isMultipleOf(alignment)) {
+            // Already aligned
             return this;
         }
 
@@ -69,8 +69,8 @@ public final class Size {
         int major = portrait ? height : width;
         int minor = portrait ? width : height;
 
-        major &= ~7; // round down to not exceed the initial size
-        minor = (minor + 4) & ~7; // round to the nearest to minimize aspect ratio distortion
+        major = major / alignment * alignment; // round down to not exceed the initial size
+        minor = (minor + (alignment / 2)) / alignment * alignment; // round to the nearest to minimize aspect ratio distortion
         if (minor > major) {
             minor = major;
         }
@@ -80,8 +80,8 @@ public final class Size {
         return new Size(w, h);
     }
 
-    public boolean isMultipleOf8() {
-        return (width & 7) == 0 && (height & 7) == 0;
+    public boolean isMultipleOf(int alignment) {
+        return width % alignment == 0 && height % alignment == 0;
     }
 
     public Rect toRect() {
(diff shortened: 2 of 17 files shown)


server/src/main/java/com/genymobile/scrcpy/device/Size.java
```diff
@@ -59,18 +58,18 @@ public Size limit(int maxSize) {
      *
      * @return The current size rounded.
      */
-    public Size round8() {
-        if (isMultipleOf8()) {
-            // Already a multiple of 8
+    public Size round(int alignment) {
```
anotheruserofgithub, line 62: Documentation should be updated, e.g.:
```diff
    /**
-    * Round both dimensions of this size to be a multiple of 8 (as required by many encoders).
+    * Round both dimensions of this size to be a multiple of {@code alignment} (as required by many encoders).
     *
+    * @param alignment The required alignment.
     * @return The current size rounded.
     */
     public Size round(int alignment) {
```

Apart from that it looks great. :)

rom1v: Thanks :+1: Fixed, and I also added a documentation commit.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceCapture.java
```diff
@@ -86,4 +88,15 @@ public void stop() {
     public boolean isClosed() {
         return false;
     }
+
+    /**
+     * Return the video alignment
+     * <p>
+     * This a power-of-2 value that the video width and height must be a multiple of.
```
anotheruserofgithub: To follow your recent changes:
```suggestion
     * This a power-of-2 value that the video width and height must be multiples of.
```

doc/video.md
```diff
@@ -30,6 +30,17 @@ If encoding fails, scrcpy automatically tries again with a lower definition
...
 size instead (among the available resolutions).
 
+The size is rounded to a multiple of the _alignment_ required by the encoder, a
+power-of-2 value (1, 2, 4, 8 or 16) that the video width and height must be
+multiples of.
+
+The alignment can be forced to a minimum value. For instance, to force the width
+and height to be multiples of 8:
+
+```
```
anotheruserofgithub: Like other code blocks:
```suggestion
```bash
```

server/src/main/java/com/genymobile/scrcpy/video/SurfaceCapture.java
rom1v: :+1:

doc/video.md
rom1v: :+1:

PR #6753, opened 2026-04-05
doc : fix one message throwed

nameearly:


rom1v:
The message is for an assertion:
 - `pointers.size() >= MAX_POINTERS` is checked just above
 - if `pointers.size() < maxFingers`, then a local id should be available (but it is not, something is very wrong, this is a programming error/assertion)

This is not just a runtime error signaling to the user that all ids are in used, so the current error message is more appropriate.

Issue #5099, opened 2024-07-16
Flag to prevent scrcpy from blocking sleep on PC


MenacingPerson:
- [x] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**Is your feature request related to a problem? Please describe.**
I just use scrcpy to channel audio to my pc. I still want to put my PC to sleep.

**Describe the solution you'd like**
Add a flag to prevent scrcpy from blocking sleep.

**Describe alternatives you've considered**
It becomes default when only audio is being transferred

**Additional context**

![Screenshot_20240716_204631](https://github.com/user-attachments/assets/649bee7a-52f4-4317-bf24-68ffdeb15fd4)

This is because of scrcpy

rom1v:
Weird: https://github.com/Genymobile/scrcpy/blob/master/doc/window.md#disable-screensaver

PR #6754, opened 2026-04-05
Fix screensaver disabled unexpectedly

ayooh:
If the value of `video_playback` is `false` (e.g., when `--no-video`, `--no-window`, or `no-video-playback` is passed), the screensaver is disabled by default, according to [this ref](https://wiki.libsdl.org/SDL3/SDL_DisableScreenSaver).

By default, scrcpy does not prevent the screensaver from running on the computer unless the `--disable-screensaver` option is passed.

Related: #5099

### DIFF
--- a/app/src/scrcpy.c
+++ b/app/src/scrcpy.c
@@ -140,7 +140,7 @@ sdl_set_hints(const char *render_driver) {
 }
 
 static void
-sdl_configure(bool video_playback, bool disable_screensaver) {
+sdl_configure(bool disable_screensaver) {
 #ifdef _WIN32
     // Clean up properly on Ctrl+C on Windows
     bool ok = SetConsoleCtrlHandler(windows_ctrl_handler, TRUE);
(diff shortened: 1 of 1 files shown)


rom1v:
Thank you :+1:

I agree with the change. I noticed that the screensaver can also be set using a [hint](https://wiki.libsdl.org/SDL3/SDL_HINT_VIDEO_ALLOW_SCREENSAVER), so I made some refactors.

Please review the entire branch and confirm that it works as expected: [`fix-screensaver.2`](https://github.com/Genymobile/scrcpy/commits/fix-screensaver.2/)

ayooh:
Thanks for the refactor! I’ve reviewed the branch and confirmed that it works as expected.

rom1v:
Merged into `dev`:
 - 758d0a1cc03732cbdea764be096136d909cbee05 (your commit)
 - 003616bd2c0807bb5171682a3b7bb64d007d4569
 - c527bc76de3c0fe8d75a72ec400dc406c3bf2e7b
 - 07f8eb0ec66c8a878874fcd095fe302a344c99fa

:rocket:

PR #2947, opened 2022-01-15
Automatically try with a lower definition on MediaCodec error

rom1v:
Some devices are not able to encode at the device screen definition.
    
Instead of just failing, automatically try with a lower definition (fallback) on any MediaCodec error.

This is a recurrent issue for new users. The solution is explained in the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md#exception) (and sometimes in the error message, depending on the error), but it will help a lot if scrcpy automatically retries with different "max size" values.

An option `--no-downsize-on-error` is also added to disable this behavior.

Here are binaries (replace in your v1.21 release):

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/2947/3/scrcpy.exe) _sha256:bdf50503f7f0f2771758abd877344e10a7cd37d3b3539bef2d80c7323807a41e_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/2947/3/scrcpy-server) _sha256:c2edcdb775776d5bca93fe04ca37159c7dc2cd36a6b7619e583c702e75eaa855_

<details><summary>v2 (obsolete)</summary>

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/2947/2/scrcpy.exe) _sha256:9f43f1ef5452b0a826cce78740fd2b93582b36accd5b9295796322c1ee02c5f8_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/2947/2/scrcpy-server) _sha256:86d9c13a88f9523150017c50061bf04b5ad234808139df31b337c2701e6b3791_

</details>

<details><summary>v1 (obsolete)</summary>

 - [`scrcpy.exe`](https://tmp.rom1v.com/scrcpy/2947/1/scrcpy.exe) _sha256:d5043a52eb572b445ffc4410d8327bb363258d0303dd645bc56423805296694d_
 - [`scrcpy-server`](https://tmp.rom1v.com/scrcpy/2947/1/scrcpy-server) _sha256:86d9c13a88f9523150017c50061bf04b5ad234808139df31b337c2701e6b3791_

</details>

I don't have any device which actually fails, so tests and feedback are welcome.

PR #6766, opened 2026-04-11
Replace retry-with-lower-resolution mechanism with video capabilities constraints

rom1v:
A mechanism was introduced to retry capture at a lower resolution to support devices unable to encode at the device screen resolution (#2947).
    
While useful, this approach is inherently limited and will not be able to handle the dynamic resizing required for resizable virtual displays.

Disable this mechanism entirely, and adjust the size in advance according to video encoder capababilities.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/device/Size.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Size.java
@@ -29,59 +29,35 @@ public final class Size {
         return new Size(height, width);
     }
 
-    public Size limit(int maxSize) {
+    public Size constrain(int maxSize, int alignment) {
         assert maxSize >= 0 : "Max size may not be negative";
+        assert alignment > 0 : "Alignment must be positive";
+        assert (alignment & (alignment - 1)) == 0 : "Alignment must be a power-of-two";
 
-        if (maxSize == 0) {
-            // No limit
-            return this;
+        int alignedMaxSize = maxSize / alignment * alignment; // round to a multiple of alignment
+        int w, h;
+
+        if (maxSize > 0 && (width > alignedMaxSize || height > alignedMaxSize)) {
+            if (width > height) {
+                w = alignedMaxSize;
+                h = round(height * alignedMaxSize / width, alignment);
+            } else {
+                w = round(width * alignedMaxSize / height, alignment);
+                h = alignedMaxSize;
+            }
+        } else {
+            w = round(width, alignment);
+            h = round(height, alignment);
         }
 
-        boolean portrait = height > width;
-        int major = portrait ? height : width;
-        if (major <= maxSize) {
-            return this;
-        }
+        assert maxSize == 0 || w <= maxSize : "The width cannot exceed maxSize if maxSize is aligned";
+        assert maxSize == 0 || h <= maxSize : "The height cannot exceed maxSize if maxSize is aligned";
 
-        int minor = portrait ? width : height;
-
-        int newMajor = maxSize;
-        int newMinor = maxSize * minor / major;
-
-        int w = portrait ? newMinor : newMajor;
-        int h = portrait ? newMajor : newMinor;
         return new Size(w, h);
     }
 
-    /**
-     * Round both dimensions of this size to be multiples of {@code alignment}.
-     *
-     * @param alignment the required alignment
-     * @return the current size rounded
-     */
-    public Size round(int alignment) {
-        if (isMultipleOf(alignment)) {
-            // Already aligned
-            return this;
-        }
-
-        boolean portrait = height > width;
-        int major = portrait ? height : width;
-        int minor = portrait ? width : height;
-
-        major = major / alignment * alignment; // round down to not exceed the initial size
-        minor = (minor + (alignment / 2)) / alignment * alignment; // round to the nearest to minimize aspect ratio distortion
-        if (minor > major) {
-            minor = major;
-        }
-
-        int w = portrait ? minor : major;
-        int h = portrait ? major : minor;
-        return new Size(w, h);
-    }
-
-    public boolean isMultipleOf(int alignment) {
-        return width % alignment == 0 && height % alignment == 0;
+    private static int round(int value, int alignment) {
+        return (value + (alignment / 2)) / alignment * alignment;
     }
 
     public Rect toRect() {
--- a/server/src/main/java/com/genymobile/scrcpy/device/Size.java
+++ b/server/src/main/java/com/genymobile/scrcpy/device/Size.java
@@ -1,5 +1,7 @@
 package com.genymobile.scrcpy.device;
 
+import com.genymobile.scrcpy.video.VideoConstraints;
+
 import android.graphics.Rect;
 
 import java.util.Objects;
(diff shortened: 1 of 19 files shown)


server/src/main/java/com/genymobile/scrcpy/video/SurfaceCapture.java
```diff
@@ -83,13 +83,13 @@ public boolean isClosed() {
     }
 
     /**
-     * Return the video alignment
+     * Return the video constraints.
      * <p>
      * This a power-of-2 value that the video width and height must be multiples of.
```
anotheruserofgithub:
```suggestion
```

server/src/main/java/com/genymobile/scrcpy/device/Size.java
```diff
@@ -39,24 +39,38 @@ public Size constrain(VideoConstraints constraints) {
...
             }
         } else {
             w = round(width, alignment);
             h = round(height, alignment);
         }
 
-        assert maxSize == 0 || w <= maxSize : "The width cannot exceed maxSize if maxSize is aligned";
-        assert maxSize == 0 || h <= maxSize : "The height cannot exceed maxSize if maxSize is aligned";
+        assert w <= maxWidth : "The width cannot exceed maxWidth";
+        assert h <= maxHeight : "The height cannot exceed maxSize";
```
anotheruserofgithub:
```suggestion
        assert h <= maxHeight : "The height cannot exceed maxHeight";
```

server/src/main/java/com/genymobile/scrcpy/device/Size.java
```diff
@@ -72,6 +72,15 @@ public Size constrain(VideoConstraints constraints) {
...
         assert h <= maxHeight : "The height cannot exceed maxSize";
 
+        // Minimum codec size must be respected (regardless of requested maxSize)
+        int minCodecSize = constraints.getMinCodecSize();
+        if (w < minCodecSize) {
+            w = minCodecSize;
+        }
+        if (h < minCodecSize) {
+            h = minCodecSize;
+        }
```
anotheruserofgithub, line 82: This breaks the aspect ratio, is it desired?

server/src/main/java/com/genymobile/scrcpy/video/SurfaceCapture.java
rom1v: :+1:

server/src/main/java/com/genymobile/scrcpy/device/Size.java
rom1v: Yes, the minimum size is like 32×32, the encoder cannot support less, so we accept breaking the aspect ratio (in practice, the encoder adds black bars).

In practice, nothing is really visible at this size, the goal is just to avoid failure when we resize the virtual display when resizing the window (pending work).

server/src/main/java/com/genymobile/scrcpy/device/Size.java
rom1v: :+1:

server/src/main/java/com/genymobile/scrcpy/device/Size.java
anotheruserofgithub: Great!

server/src/main/java/com/genymobile/scrcpy/device/Size.java
```diff
@@ -72,6 +72,15 @@ public Size constrain(VideoConstraints constraints) {
...
         assert h <= maxHeight : "The height cannot exceed maxHeight";
 
+        // Minimum codec size must be respected (regardless of requested maxSize)
+        int minCodecSize = constraints.getMinCodecSize();
+        if (w < minCodecSize) {
+            w = minCodecSize;
+        }
+        if (h < minCodecSize) {
+            h = minCodecSize;
+        }
```
anotheruserofgithub, line 82: There is an issue: `minCodecSize` must also be a multiple of `alignment`. However, while the max must be floored, the min must be ceiled (not just rounded). Of course this is assuming that the final `minSize` is not greater than `maxCodecSize.width` and `maxCodecSize.height`; maybe it would be good to assert it? It would also be good to assert the proper alignment of `w` and `h` at the end, since I ran into this issue.

Suggested change so that you see what I mean:
```suggestion
        // Minimum codec size must be respected (regardless of requested maxSize)
        int minCodecSize = constraints.getMinCodecSize();
        int minSize = (minCodecSize + (alignment - 1)) / alignment * alignment;

        assert minSize <= maxCodecSize.width : "The minimum width cannot exceed the maximum codec width";
        assert minSize <= maxCodecSize.height : "The minimum height cannot exceed the maximum codec height";

        if (w < minSize) {
            w = minSize;
        }
        if (h < minSize) {
            h = minSize;
        }

        assert w % alignment == 0 : "The width must be a multiple of alignment";
        assert h % alignment == 0 : "The height must be a multiple of alignment";
```

rom1v: I think that it can be assumed that the min size is a multiple of alignment (otherwise the video capabilities are inconsistent).

anotheruserofgithub: That's true but I think it's not complete. The question is whether you want to enforce respecting `--min-size-alignment` or not, since [it can be higher than the codec alignment](https://github.com/Genymobile/scrcpy/blob/175134c0ca1d9018c14d10b467dbfcdff6d1c3be/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java#L61-L66) (and minimum size).

With my emulator (OK, you can say it's buggy, but still, it would be great to avoid such crashes), `minCodecSize` is reported as 1 but the size can't actually be less than 16. I can set `--min-size-alignment=16`, however it would still crash with a display size < 16 if we don't enforce the alignment of `minCodecSize`.

Even if, as you say we can assume, the video capabilities are consistent (i.e. correctly reported codec limits and alignment), it is possible that the user sets a `--min-size-alignment` higher than the codec alignment.

rom1v: > That's true but I think it's not complete. The question is whether you want to enforce respecting `--min-size-alignment` or not

In practice, if it's not respected due to the min size constraint, I did not care. But yes, in theory it's better to respect it (the aspect ratio is not respected either with the min size constraint).

> `minCodecSize` is reported as 1 but the size can't actually be less than 16. I can set `--min-size-alignment=16`, however it would still crash with a display size < 16 if we don't enforce the alignment of `minCodecSize`.

Ok so even the minSize is not correctly declared by the codec. In that case, indeed it's better to make the min alignment a workaround for this case.

rom1v: Fixed in #6772 (commit `Ensure minimum codec size respects alignment`, I don't link its sha1 because it changes on every update).

_I did not assert that minSize is lower than maxSize after alignment: if the constraints have no solution (this is not absolutely impossible, so no assertion), then maxSize will not be respected, and the encoding will fail. In practice, this will never happen._

anotheruserofgithub: That makes sense, thanks!

anotheruserofgithub: @rom1v Could you please mark this review as resolved? I don't have the rights to do it.

PR #6770, opened 2026-04-14
Reset capture on rotation (fix square displays)

rom1v:
`DisplayMonitor` previously only triggered a capture reset when the display size changed. In most cases, rotation also changes dimensions, so the behavior was correct… except for square displays where width and height remain unchanged.
    
However, rotation still requires a capture reset even when dimensions do not change, to ensure the orientation filter is applied so virtual displays are rendered correctly.
    
To reproduce the issue:
    
    scrcpy --new-display=600x600 --start-app=com.android.settings
    
Then press <kbd>Alt</kbd>+<kbd>r</kbd> to rotate the Settings app.

(Note: the shortcut was broken on scrcpy 3.3.4, it was fixed on `dev` by 5fedc79530672ae2b0df6ac7654d75a6f2e18ce3)

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
+++ b/server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
@@ -2,7 +2,6 @@ package com.genymobile.scrcpy.display;
 
 import com.genymobile.scrcpy.AndroidVersions;
 import com.genymobile.scrcpy.device.Device;
-import com.genymobile.scrcpy.model.Size;
 import com.genymobile.scrcpy.util.Ln;
 import com.genymobile.scrcpy.wrappers.DisplayManager;
 import com.genymobile.scrcpy.wrappers.DisplayWindowListener;
@@ -14,10 +13,10 @@ import android.os.Handler;
 import android.os.HandlerThread;
 import android.view.IDisplayWindowListener;
 
-public class DisplaySizeMonitor {
+public class DisplayMonitor {
 
     public interface Listener {
-        void onDisplaySizeChanged();
+        void onDisplayPropertiesChanged();
     }
 
     // On Android 14, DisplayListener may be broken (it never sends events). This is fixed in recent Android 14 upgrades, but we can't really
@@ -33,7 +32,7 @@ public class DisplaySizeMonitor {
 
     private int displayId = Device.DISPLAY_ID_NONE;
 
-    private Size sessionDisplaySize;
+    private DisplayProperties props;
 
     private Listener listener;
 
@@ -51,11 +50,11 @@ public class DisplaySizeMonitor {
             Handler handler = new Handler(handlerThread.getLooper());
             displayListenerHandle = ServiceManager.getDisplayManager().registerDisplayListener(eventDisplayId -> {
                 if (Ln.isEnabled(Ln.Level.VERBOSE)) {
-                    Ln.v("DisplaySizeMonitor: onDisplayChanged(" + eventDisplayId + ")");
+                    Ln.v("DisplayMonitor: onDisplayChanged(" + eventDisplayId + ")");
                 }
 
                 if (eventDisplayId == displayId) {
-                    checkDisplaySizeChanged();
+                    checkDisplayPropertiesChanged();
                 }
             }, handler);
         } else {
@@ -63,11 +62,11 @@ public class DisplaySizeMonitor {
                 @Override
                 public void onDisplayConfigurationChanged(int eventDisplayId, Configuration newConfig) {
                     if (Ln.isEnabled(Ln.Level.VERBOSE)) {
-                        Ln.v("DisplaySizeMonitor: onDisplayConfigurationChanged(" + eventDisplayId + ")");
+                        Ln.v("DisplayMonitor: onDisplayConfigurationChanged(" + eventDisplayId + ")");
                     }
 
                     if (eventDisplayId == displayId) {
-                        checkDisplaySizeChanged();
+                        checkDisplayPropertiesChanged();
                     }
                 }
             };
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
anotheruserofgithub, line 39: Two `Ln.v("DisplaySizeMonitor:` in `start()` that should be replaced with `DisplayMonitor`.

server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
```diff
@@ -97,38 +96,38 @@ public void stopAndRelease() {
...
+        return oldProps;
     }
 
-    public synchronized void setSessionDisplaySize(Size sessionDisplaySize) {
-        this.sessionDisplaySize = sessionDisplaySize;
+    public synchronized void setSessionDisplayProperties(DisplayProperties props) {
+        this.props = props;
     }
 
     private void checkDisplaySizeChanged() {
```
anotheruserofgithub:
```suggestion
    private void checkDisplayPropertiesChanged() {
```

server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
```diff
@@ -14,7 +13,7 @@
 import android.os.HandlerThread;
 import android.view.IDisplayWindowListener;
 
-public class DisplaySizeMonitor {
+public class DisplayMonitor {
 
     public interface Listener {
         void onDisplaySizeChanged();
```
anotheruserofgithub:
```suggestion
        void onDisplayPropertiesChanged();
```

server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
```diff
@@ -97,38 +96,38 @@ public void stopAndRelease() {
...
-        this.sessionDisplaySize = sessionDisplaySize;
+    public synchronized void setSessionDisplayProperties(DisplayProperties props) {
+        this.props = props;
     }
 
     private void checkDisplaySizeChanged() {
         DisplayInfo di = ServiceManager.getDisplayManager().getDisplayInfo(displayId);
         if (di == null) {
             Ln.w("DisplayInfo for " + displayId + " cannot be retrieved");
             // We can't compare with the current size, so reset unconditionally
```
anotheruserofgithub:
```suggestion
            // We can't compare with the current properties, so reset unconditionally
```

server/src/main/java/com/genymobile/scrcpy/display/DisplayMonitor.java
```diff
@@ -97,38 +96,38 @@ public void stopAndRelease() {
...
             listener.onDisplaySizeChanged();
         } else {
-            Size size = di.getSize();
+            DisplayProperties newProps = new DisplayProperties(di.getSize(), di.getRotation());
 
-            Size oldDisplaySize = getAndSetSessionDisplaySize(size); // exchange with synchronization
-            if (!size.equals(oldDisplaySize)) {
+            DisplayProperties oldProps = getAndSetDisplayProperties(newProps); // exchange with synchronization
+            if (!newProps.equals(oldProps)) {
                 // Reset only if the size is different
```
anotheruserofgithub:
```suggestion
                // Reset only if the properties are different
```

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -113,7 +114,7 @@ public void prepare() {
 
             displayRotation = 0;
             // Set the current display size to avoid an unnecessary call to invalidate()
```
anotheruserofgithub:
```suggestion
            // Set the current display properties to avoid an unnecessary call to invalidate()
```

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -42,7 +43,7 @@ public class NewDisplayCapture extends SurfaceCapture {
     private final VirtualDisplayListener vdListener;
     private final NewDisplay newDisplay;
 
-    private final DisplaySizeMonitor displaySizeMonitor = new DisplaySizeMonitor();
+    private final DisplayMonitor displaySizeMonitor = new DisplayMonitor();
```
anotheruserofgithub:
```suggestion
    private final DisplayMonitor displayMonitor = new DisplayMonitor();
```

server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
```diff
@@ -38,7 +39,7 @@ public class ScreenCapture extends SurfaceCapture {
     private DisplayInfo displayInfo;
     private Size videoSize;
 
-    private final DisplaySizeMonitor displaySizeMonitor = new DisplaySizeMonitor();
+    private final DisplayMonitor displaySizeMonitor = new DisplayMonitor();
```
anotheruserofgithub:
```suggestion
    private final DisplayMonitor displayMonitor = new DisplayMonitor();
```

server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
```diff
@@ -76,7 +77,7 @@ public void prepare() throws ConfigurationException {
         }
 
         Size displaySize = displayInfo.getSize();
-        displaySizeMonitor.setSessionDisplaySize(displaySize);
+        displaySizeMonitor.setSessionDisplayProperties(new DisplayProperties(displaySize, displayInfo.getRotation()));
```
anotheruserofgithub:
```suggestion
        int displayRotation = displayInfo.getRotation();
        displayMonitor.setSessionDisplayProperties(new DisplayProperties(displaySize, displayRotation));
```
Then there are three more occurrences of `displayInfo.getRotation()` that can be replaced with `displayRotation` below in `prepare()`.

anotheruserofgithub (commented):
Note: You should first reorder imports after your last commit on `dev` branch.
Plus, tests should also be moved to new corresponding modules.

server/src/main/java/com/genymobile/scrcpy/display/DisplayProperties.java
```diff
@@ -0,0 +1,36 @@
+package com.genymobile.scrcpy.display;
+
+import com.genymobile.scrcpy.model.Size;
+
+public final class DisplayProperties {
+    private Size size;
+    private int rotation;
```
anotheruserofgithub, line 9: I think you have to override [`Object.equals()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)) (used in `DisplayMonitor`), right? And `Object.hashCode()` too for good measure, like you did in `Size`.

rom1v:
> Note: You should first reorder imports after your last commit on dev branch.
> Plus, tests should also be moved to new corresponding modules.

:+1:

Oops, I just pushed-force `dev` to fix it, sorry for that.

server/src/main/java/com/genymobile/scrcpy/display/DisplayProperties.java
rom1v: Oh yes, I thought I generated it with Android Studio at the same time as `toString()`, but I didn't. Thank you :+1:

server/src/main/java/com/genymobile/scrcpy/display/DisplayProperties.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(size, rotation);
+    }
+
+    @Override
+    public String toString() {
+        return size + " [rotation=" + rotation + ']';
```
anotheruserofgithub:
```suggestion
        return size + " [rotation=" + rotation + "]";
```
I don't know if using a char was intentional? For instance in `Size.toString()` you make a string literal for the character in the middle: `"x"`

rom1v: A `char` is ok (it was generated by Android studio when I generated to `toString()` method), I just edited to use `[rotation=]`). But for consistency, I will change to `"`.

anotheruserofgithub: > For instance in Size.toString() you make a string literal for the character in the middle

I think in this case you have no choice otherwise it wouldn't concatenate the ints into a string.

But yeah, double quote seems more consistent.

ea136917-art (approved):

PR #6771, opened 2026-04-16
Align the virtual display size

rom1v:
When the virtual display size is not aligned, a resize filter may be inserted to compensate for small dimension mismatches, increasing processing and resulting in a blurry image.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
@@ -108,6 +108,10 @@ public class NewDisplayCapture extends SurfaceCapture {
             if (!newDisplay.hasExplicitSize()) {
                 displaySize = mainDisplaySize;
             }
+
+            // Align the physical display size to avoid unnecessary mismatches with the output size
+            displaySize = displaySize.align(getVideoConstraints().getAlignment());
+
             if (!newDisplay.hasExplicitDpi()) {
                 dpi = scaleDpi(mainDisplaySize, mainDisplayDpi, displaySize);
             }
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -117,9 +121,11 @@ public void prepare() {
             displayMonitor.setSessionDisplayProperties(new DisplayProperties(displaySize, displayRotation));
         } else {
             DisplayInfo displayInfo = ServiceManager.getDisplayManager().getDisplayInfo(virtualDisplay.getDisplay().getDisplayId());
-            displaySize = displayInfo.getSize();
             dpi = displayInfo.getDpi();
             displayRotation = displayInfo.getRotation();
+
+            // Align the physical display size to avoid unnecessary mismatches with the output size
+            displaySize = displayInfo.getSize().align(getVideoConstraints().getAlignment());
```
anotheruserofgithub, line 128: If I understand correctly, you want to avoid `filter.addResize(filteredSize)` below? If yes, then is there a reason not to do `displaySize = displayInfo.getSize().constrain(getVideoConstraints())` in order to match min/max constraints in addition to alignment?

rom1v: It's the filtered size, not the initial display size, that must be constrained:

```java
Size filteredSize = outputSize.constrain(getVideoConstraints());
```

anotheruserofgithub: OK, I get it now.

anotheruserofgithub: @rom1v Could you please mark this review as resolved? I don't have the rights to do it.

PR #6761, opened 2026-04-08
Add option to enforce window aspect ratio to always match video

Tech-Tac:
Adds a `--window-preserve-aspect-ratio` CLI option (defaults to false) that uses SDL3's new [SDL_SetWindowAspectRatio](https://wiki.libsdl.org/SDL3/SDL_SetWindowAspectRatio) functionality to preserve the window's aspect ratio when resized so that it always matches the content, eliminating the black bars.

Tested on KWin Wayland compositor version 6.6.4

PR #6774, opened 2026-04-17
Fix aspect ratio accuracy

rom1v:
The window aspect ratio was previously enforced by calling `SDL_SetWindowAspectRatio()` before explicitly setting the window size, using the width/height ratio of the new size.

However, this size may already be a scaled version of the actual content size, subject to integer rounding. Using this derived size instead of the actual content size could result in an inaccurate ratio, causing a mismatch between the content and window aspect ratios, which may result in a thin row or column of black pixels.

To avoid this issue, always set the aspect ratio based on the content size.

Refs 1b4fd672869dcaa7f0ba8f11b722b731488d690d
Refs #6761

### DIFF
--- a/app/src/screen.c
+++ b/app/src/screen.c
@@ -15,18 +15,16 @@
 #define DOWNCAST(SINK) container_of(SINK, struct sc_screen, frame_sink)
 
 static void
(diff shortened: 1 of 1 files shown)


Tech-Tac:
I didn't notice this issue initially, this looks good.

PR #6776, opened 2026-04-17
Update protocol documentation

rom1v:
The protocol changed in commit 78cba1b7c29fe45057d180a183ce7efcd7044106.

Update the documentation accordingly.

Refs #6159

### DIFF
--- a/doc/develop.md
+++ b/doc/develop.md
@@ -113,10 +113,10 @@ commands:
 ```bash
 adb push scrcpy-server /data/local/tmp/scrcpy-server.jar
 adb forward tcp:27183 localabstract:scrcpy
-adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 2.1
+adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 4.0
 ```
 
-The first argument (`2.1` in the example) is the client scrcpy version. The
+The first argument (`4.0` in the example) is the client scrcpy version. The
 server fails if the client and the server do not have the exact same version.
 The protocol between the client and the server may change from version to
 version (see [protocol](#protocol) below), and there is no backward or forward
@@ -136,7 +136,7 @@ execution will look like this:
 
 ```bash
 # scid is a random number to identify different clients running on the same device
-adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 2.1 scid=12345678 log_level=info audio=false max_size=1920
+adb shell CLASSPATH=/data/local/tmp/scrcpy-server.jar app_process / com.genymobile.scrcpy.Server 4.0 scid=12345678 log_level=info audio=false max_size=1920
 ```
 
 ### Components
@@ -303,7 +303,7 @@ number of sockets, the order in which the sockets must be opened, the data
 format on the wire…) from version to version. A client must always be run with a
 matching server version.
 
-This section documents the current protocol in scrcpy v2.1.
+This section documents the current protocol in scrcpy 4.0.
 
 ### Connection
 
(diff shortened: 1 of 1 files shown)


doc/develop.md
```diff
@@ -303,7 +303,7 @@ number of sockets, the order in which the sockets must be opened, the data
 format on the wire…) from version to version. A client must always be run with a
 matching server version.
 
-This section documents the current protocol in scrcpy v2.1.
+This section documents the current protocol in scrcpy 4.0.
```
anotheruserofgithub, line 306: If you want to be pedantic, maybe you should also change `scrcpy v2.0` to `scrcpy 2.0` at line 32?
(The blog post writes it "scrcpy 2.0" so this would match.)

doc/develop.md
```diff
@@ -352,19 +352,41 @@ Then each socket is used for its intended purpose.
...
+ - `"h265"` (`0x68323635`)
+ - `"av1"` (`0x00617631`)
+
+Valid codec ids for _audio_ streams are:
+ - `"opus"` (`0x6F707573`)
+ - `"aac"` (`0x00616163`)
+ - `"flac"` (`0x666C6163`)
+ - `"raw"` (`0x00726177`)
 
 [codec metadata]: https://github.com/Genymobile/scrcpy/blob/a3cdf1a6b86ea22786e1f7d09b9c202feabc6949/server/src/main/java/com/genymobile/scrcpy/Streamer.java#L33-L51
```
anotheruserofgithub:
```suggestion
```
`[codec metadata]` isn't referred to anymore but I don't know if you'd like to reinstate it somehow?

doc/develop.md
```diff
@@ -352,19 +352,41 @@ Then each socket is used for its intended purpose.
...
+                video width                         video height
+```
+
+For the _audio_ stream, there are no _session packets_.
+
+Then _media packets_ are sent, each containing the payload produced by
+`MediaCodec`, prefixed by a 12-byte [frame header]:
  - config packet flag (`u1`)
  - key frame flag (`u1`)
  - PTS (`u62`)
```
anotheruserofgithub:
```suggestion
- media packet flag (`u1`)
- config packet flag (`u1`)
- key frame flag (`u1`)
- PTS (`u61`)
```
New session/media packet flag and thus PTS is only 61 bits.

doc/develop.md
```diff
@@ -382,14 +404,24 @@ Here is a schema describing the frame header:
...
+    ........ ........ ........ ........ ........ ........ . . .
+    <---------------------------------> <---------------- . . .
+               packet size                       raw packet
+
 ```
 
+_Session packets_ and _media packets_ are distinguished by their first bit (the
+MSB).
+
 [frame header]: https://github.com/Genymobile/scrcpy/blob/a3cdf1a6b86ea22786e1f7d09b9c202feabc6949/server/src/main/java/com/genymobile/scrcpy/Streamer.java#L83
```
anotheruserofgithub, line 424: I think it would be great to update all the commit references in this file (and check if all the links still work), especially because the protocol implementation changed since v2.1.

doc/develop.md
rom1v: Done :+1:

doc/develop.md
rom1v: In fact, it had never been used :sweat_smile: Removed.

doc/develop.md
rom1v: Good catch, fixed :+1:

doc/develop.md
rom1v: Yes, I'll do that after the merge of the "flex display" feature (which impacts the protocol a bit) in a separate change. :+1:

doc/develop.md
anotheruserofgithub: I think it was: https://github.com/Genymobile/scrcpy/blob/1015b42e53462b8537a635eb27f3a53c31c8a96b/doc/develop.md#video-and-audio
Thus commit 2b2896a6cd9707b3b67e09c5df5d6033132b08c4 breaks the link: https://github.com/Genymobile/scrcpy/blob/2b2896a6cd9707b3b67e09c5df5d6033132b08c4/doc/develop.md#video-and-audio

rom1v: Oh, good catch, I'm so stupid, I searched `codec metadata` but it was split into two lines :facepalm: 

I removed the link anyway (not necessary).

PR #6772, opened 2026-04-16
Add flex display support (resizable virtual display)

rom1v:
```bash
# Start Android Settings in a window
scrcpy --new-display=1024x768 --start-app=com.android.settings --flex-display

# -x is equivalent to --flex-display
scrcpy --new-display=1024x768 --start-app=com.android.settings -x

# By default, the display size/dpi is 1280x960/160
scrcpy --new-display --start-app=com.android.settings --flex-display
```

Use `--keep-active` to prevent the screen from turning off:

```bash
scrcpy --new-display --flex-display --keep-active
```

Increase the bit rate (and/or change the codec) to maintain good quality even with large windows:

```bash
scrcpy --new-display -x --video-codec=h265 -b16M
```

## Demo

Here is Firefox for Android running in a "flex" virtual display, ran as follow:

```bash
scrcpy --new=/192 -x --start-app=org.mozilla.firefox --keep-active --no-vd-system-decorations
```

https://github.com/user-attachments/assets/0822bcf7-58c1-4106-87f6-c8089c08a777

<details><summary>previous video</summary>

```bash
scrcpy --new-display=800x600 -x --start-app=org.mozilla.firefox
```

https://github.com/user-attachments/assets/0a86f0e1-9911-45b5-9a25-70d14355ab82

</details>

## Download binaries

Here are binaries built by Github Actions (for `flex-display.16`): https://github.com/rom1v/scrcpy/actions/runs/25569279750

_(download the artifact `release-XXX` where `XXX` is your target platform)_

<details><summary>old versions</summary>

 - `flex-display.1`: https://github.com/rom1v/scrcpy/actions/runs/24530523954
 - `flex-display.4`: https://github.com/rom1v/scrcpy/actions/runs/24731426812
 - `flex-display.5`: https://github.com/rom1v/scrcpy/actions/runs/24899984361
 - `flex-display.6`: https://github.com/rom1v/scrcpy/actions/runs/25006302461
 - `flex-display.7`: https://github.com/rom1v/scrcpy/actions/runs/25009172508
 - `flex-display.10`: https://github.com/rom1v/scrcpy/actions/runs/25065382694
 - `flex-display.13`: https://github.com/rom1v/scrcpy/actions/runs/25282088772
 - `flex-display.15`: https://github.com/rom1v/scrcpy/actions/runs/25405231193
 - `flex-display.16`: https://github.com/rom1v/scrcpy/actions/runs/25508659976
</details>

## Preparation

[...]

PR #6789, opened 2026-04-25
[Draft] Keep awake by simulating user activity

### LINKED ISSUE #6787: Virtual displays go black after being idle for 10sec when screen is off
## Environment

 - **OS:** Arch Linux
 - **Scrcpy version:** 3.3.4 (release), #6772
 - **Installation method:** pacman, git
 - **Device model:** Realme C51 (RMX3830)
 - **Android version:** 15
 - **Connection method:** TCP/IP

## Describe the bug
Virtual displays created with `--new-display` all automatically go black after not interacting with any of them for 10 seconds, the countdown seems to start only once the physical screen turns off.

This doesn't occur with `--stay-awake`, but that requires the device to be plugged in which might not always be practical. The only workaround I found to prevent this behavior when connecting wirelessly is to keep the physical screen awake manually, either by tapping it or through third party utilities (Caffeinate, etc), which is still not ideal since it increases battery usage.

One weird thing about this blackout behavior, that is only noticeable with #6772, is that the black out seems to be just an opaque black surface that takes the size of the virtual display the moment it appears, and stays the same size even if the display is resized, revealing that the app below is still rendering and still playing animations, albeit non-intractable.

Example of a resized virtual display with this issue

The blackout doesn't disappear once the physical screen is waked up again, but the virtual display regains input interactivity, but only in the areas not covered by the blackout surface.

I also tried `--no-vd-system-decorations` but to no avail.

rom1v:
To fix #6787, call `PowerManager.userActivity()` every 4 seconds to keep the virtual display awake.

It is implemented on top of #6772 and #6788.

This is just a PoC. Please test.

Are there any side effects?

Binaries are available here (download `release-XXX` where `XXX` is your platform): https://github.com/rom1v/scrcpy/actions/runs/24926398341

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -80,6 +80,7 @@ _scrcpy() {
         --record-format=
         --record-orientation=
         --render-driver=
+        --render-fit=
         --require-audio
         -s --serial=
         -S --turn-screen-off
@@ -174,6 +175,10 @@ _scrcpy() {
             ;;
         --render-driver)
             COMPREPLY=($(compgen -W 'direct3d opengl opengles2 opengles metal software' -- "$cur"))
+            return
+            ;;
+        --render-fit)
+            COMPREPLY=($(compgen -W 'letterbox disabled' -- "$cur"))
             return
             ;;
         --shortcut-mod)
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -179,7 +179,7 @@ _scrcpy() {
             return
             ;;
         --render-fit)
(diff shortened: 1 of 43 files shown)


Tech-Tac:
This works great for me, thanks!

rom1v:
Thank you for your feedback. I will add a `--keep-active` option (that will work both for physical and virtual displays).

Something like:

```bash
scrcpy --keep-active=10  # user activity every 10 seconds
scrcpy --keep-active     # default value (4 seconds for example)
```

anotheruserofgithub:
I can't reproduce https://github.com/Genymobile/scrcpy/issues/6787 with the Genymotion emulator. :/ But this feature looks really great!

rom1v:
Replaced by #6792.

PR #6792, opened 2026-04-26
Add --keep-active

### LINKED ISSUE #6787: Virtual displays go black after being idle for 10sec when screen is off

rom1v:
An an option to simulate user activity to prevent the screen from turning off due to inactivity.

Fixes #6787
Supersedes #6789

---

Contrary to what I proposed here: https://github.com/Genymobile/scrcpy/pull/6789#issuecomment-4319038666
I didn't add a parameter to specify the interval duration. I think 4 seconds (less than 5) will work for all use cases, and the call is cheap, so there is no good reason to configure it. I'll add it later if necessary.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/control/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/control/Controller.java
@@ -70,10 +70,14 @@ public class Controller implements AsyncProcessor, VirtualDisplayListener {
     // control_msg.h values of the pointerId field in inject_touch_event message
     private static final int POINTER_ID_MOUSE = -1;
 
+    // Interval between simulated user activity events
+    private static final long KEEP_ACTIVE_INTERVAL_MS = 4000;
+
     private static final ScheduledExecutorService EXECUTOR = Executors.newSingleThreadScheduledExecutor();
     private ExecutorService startAppExecutor;
 
     private Thread thread;
+    private Thread keepActiveThread;
 
     private UhidManager uhidManager;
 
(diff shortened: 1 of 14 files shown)


server/src/main/java/com/genymobile/scrcpy/control/Controller.java
```diff
@@ -259,6 +293,9 @@ public void start(TerminationListener listener) {
 
     @Override
     public void stop() {
+        if (keepActiveThread != null) {
+            keepActiveThread.interrupt();
```
anotheruserofgithub, line 297: Should you also call `keepActiveThread.join()` in `Controller.join()`? Or you don't because it's a daemon thread?

rom1v: I don't because it's a daemon. There's no need to wait.

PR #6802, opened 2026-04-30
Allow window position or size setting to fail

rom1v:
Setting the window position can be denied, as observed in logs from https://github.com/Genymobile/scrcpy/pull/6772#issuecomment-4344753479:

    ERROR: Could not set window position: wayland cannot position non-popup windows

Also do not assume window size setting succeeds.

Refs <https://github.com/Genymobile/scrcpy/pull/6216#issuecomment-3399353701>

### DIFF
--- a/app/src/util/sdl.c
+++ b/app/src/util/sdl.c
@@ -79,8 +79,7 @@ void
 sc_sdl_set_window_size(SDL_Window *window, struct sc_size size) {
     bool ok = SDL_SetWindowSize(window, size.width, size.height);
     if (!ok) {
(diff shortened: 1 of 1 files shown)


rom1v:
In response to https://github.com/Genymobile/scrcpy/pull/6772#issuecomment-4353169629 (@anotheruserofgithub):

> However, you should always be able to set the window size, and in fact it's quite important for flex displays.

I agree that it should always work, but I will still remove the assertion just in case it is denied for some reason. This has no real impact since we ignore the failure anyway.

Issue #222, opened 2018-08-10
Could Not Create Texture.


Jenis667:
Info: Initial Texture : 1080 × 2160
Critical: Could Not Create Texture: Texture Dimentions are limited to 2048 × 2048

Invalid Return value 0 for stream protocol (15x)

Error Could not Open Video Stream

Redmi Note 5(India) - Vince

rom1v:
scrcpy -m 2048

PR #6807, opened 2026-05-05
Add option to change the background color

### LINKED ISSUE #5855: Instead of black background, provide an option to make it 1.transparent 2.user-defined color
- [ ] I have checked that a similar [feature request](https://github.com/Genymobile/scrcpy/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

Surrounding background "outside android" is solid black.

1. Being able to make it trasnparent (to see wallpaper behind while app is stretched)
2. Being able to pick a color to match rest of the user's desktop
Something like mpv's "background=none" and "background-color=0.0/0.0/0.0/0.0"

rom1v:
Introduce a `--background-color` option that accepts hexadecimal color codes (3-digit or 6-digit formats):

```bash
scrcpy --background-color=#234567
scrcpy --background-color=234567  # leading '#' is optional
scrcpy --background-color=#567    # equivalent to #556677
```
    
Fixes #5855
Refs https://github.com/Genymobile/scrcpy/pull/6772#issuecomment-4374943892

Also set the default background to dark gray (`#222`).


app/src/cli.c
```diff
@@ -253,6 +254,14 @@ static const struct sc_option options[] = {
                 "Unit suffixes are supported: 'K' (x1000) and 'M' (x1000000).\n"
                 "Default is 8M (8000000).",
     },
+    {
+        .longopt_id = OPT_BACKGROUND_COLOR,
+        .longopt = "background-color",
+        .argdesc = "hexcolor",
+        .text = "Set the background color, encoded as hexadcimal color code "
```
anotheruserofgithub:
```suggestion
        .text = "Set the background color, encoded as hexadecimal color code "
```

app/src/cli.c
```diff
@@ -253,6 +254,14 @@ static const struct sc_option options[] = {
...
+                "(#RGB or #RRGGBB).\n"
```
anotheruserofgithub, line 262: Do you want to allow transparency? It's a feature request from https://github.com/Genymobile/scrcpy/issues/5855.

anotheruserofgithub (commented):
TBD the renaming of "black borders" in docs and code comments (see https://github.com/Genymobile/scrcpy/pull/6772#issuecomment-4379183568 & https://github.com/Genymobile/scrcpy/pull/6772#issuecomment-4381382778 & https://github.com/Genymobile/scrcpy/pull/6772#issuecomment-4381731162).

app/src/cli.c
rom1v: No, I don't want to allow transparency. And it will not work everywhere.

rom1v:
> TBD the renaming of "black borders" in docs

I think I will keep "black borders" for the doc, the alternatives are not clear enough for users IMO.

app/src/cli.c
anotheruserofgithub: OK!

anotheruserofgithub:
Another alternative could be "background" or "background borders".

rom1v:
"extra space", "surrounding space"? (suggested by an LLM, after I said that the term "letterboxing" was too technical)

anotheruserofgithub:
Anyway, the default color is very dark so I think "black" is OK.

Ah, maybe "dark borders" then?

anotheruserofgithub:
> "extra space"

Why not. I think borders/margins is clearer than extra space but they are all good alternatives.

rom1v:
> Anyway, the default color is very dark so I think "black" is OK.

Yeah, I think I will keep that. 😅

AminurAlam:
how does one set the bg to be transparent? there is no way to tell it what alpha channel to use

seems like only alpha channels are supported

rom1v:
> how does one set the bg to be transparent?

The background cannot be set transparent: https://github.com/Genymobile/scrcpy/pull/6807#discussion_r3191750904

PR #6816, opened 2026-05-09
Replace run-on-main-thread implementation

rom1v:
Use `SDL_RunOnMainThread()` introduced in SDL3.

### DIFF
--- a/app/src/events.c
+++ b/app/src/events.c
@@ -5,6 +5,9 @@
 #include "util/log.h"
 #include "util/thread.h"
 
+static sc_mutex mutex;
+static bool stopped;
+
 bool
 sc_push_event_impl(uint32_t type, void *ptr, const char *name) {
     SDL_Event event = {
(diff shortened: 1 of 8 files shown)


app/src/main.c
```diff
@@ -13,9 +13,9 @@
 #ifdef HAVE_USB
 # include "usb/scrcpy_otg.h"
 #endif
+#include "events.h"
 #include "util/log.h"
 #include "util/net.h"
-#include "util/thread.h"
```
anotheruserofgithub, line 18: Can this include be removed in the previous-to-last commit already?

app/src/events.h
anotheruserofgithub: Is it possible now to remove this user event and also in `scrcpy.c` L143 `case SC_EVENT_RUN_ON_MAIN_THREAD`?

app/src/events.c
```diff
@@ -23,38 +26,41 @@ sc_push_event_impl(uint32_t type, void *ptr, const char *name) {
...
-sc_reject_new_runnables(void) {
-    assert(sc_thread_is_main());
+sc_main_thread_stop(void) {
+    sc_mutex_lock(&mutex);
+    stopped = true;
+    sc_mutex_unlock(&mutex);
 
-    SDL_SetEventFilter(task_event_filter, NULL);
+    // Run all remaining runnables on the main thread
+    SDL_PumpEvents();
```
anotheruserofgithub, line 65: I'm not sure but it looks like `SDL_PumpEvents()` doesn't run the pending events, it just places them on the queue, right? And you already returned from `event_loop()` when you call `sc_main_thread_stop()` so they won't be caught and executed. But I may be mistaken.

app/src/events.h
rom1v: Oh yes, of course. Thank you.

app/src/events.c
rom1v: It calls `SDL_RunMainThreadCallbacks();` internally:

https://github.com/libsdl-org/SDL/blob/cbe3fbe9f367340dcd924de29c225c9f4ffea1f5/src/events/SDL_events.c#L1509

app/src/main.c
```diff
@@ -13,6 +13,7 @@
 #ifdef HAVE_USB
 # include "usb/scrcpy_otg.h"
 #endif
+#include "events.h"
```
anotheruserofgithub: This is not in alphabetical order, should be at L11.

app/src/events.c
anotheruserofgithub: Ahhh, OK, so, main thread callbacks are different from events, yet they are executed when "pumping" events. Thank you for the link! ;)

Issue #3693, opened 2023-01-26
SCRCPY Crashing Sometimes When the Screen is Rotated


eeriemyxi:
- [☑️] I have read the [FAQ](https://github.com/Genymobile/scrcpy/blob/master/FAQ.md).
 - [☑️] I have searched in existing [issues](https://github.com/Genymobile/scrcpy/issues).

**Environment**
 - OS: Microsoft Windows [Version 10.0.19045.2546]
 - scrcpy version: 1.25
 - installation method: Windows release.
 - device model: RMX1971.
 - Android version: 11

**Describing the Bug**
Sometimes when I try to rotate the screen using `MOD + R`, SCRCPY crashes with an exception but when I restart the server the screen rotation is as I'd expect after triggering the hotkey.

The following terminal output is the result of me rotating the screen several times on the same app and on the same app activity screen. I did find an open issue (#1700) also addressing server crashes sometimes when the screen is rotated, but that error is slightly different, the observer was emulating some device virtually (which may have contained unfinished encoders and decoders to enhance the performance), and collaborator's reply was that the device does not support quick encoding or something which causes the errors but in my case it happens even when I wouldn't rotate screen for several minutes before the error-causing attempt to rotate the screen.

If this isn't something that could be prevented by configurations or modifications of the code, is there some way to auto reconnect the device on error?

```
scrcpy -b 2500k -m 1080 -S -w --power-off-on-close
scrcpy 1.25 <https://github.com/Genymobile/scrcpy>
C:\Users\Myxi\Documents\scrcpy\scrcpy-server: 1 fi...ushed, 0 skipped. 0.6 MB/s (42151 bytes in 0.065s)
[server] INFO: Device: realme RMX1971 (Android 11)
INFO: Renderer: direct3d
INFO: Initial texture: 496x1080
[server] INFO: Device screen turned off
[server] INFO: Device rotation requested: landscape
INFO: New texture: 1080x496
[server] INFO: Device rotation requested: portrait
INFO: New texture: 496x1080
[server] INFO: Device rotation requested: landscape
INFO: New texture: 1080x496
[server] INFO: Device rotation requested: portrait
INFO: New texture: 496x1080
[server] INFO: Device rotation requested: landscape
INFO: New texture: 1080x496
[server] INFO: Device rotation requested: portrait
INFO: New texture: 496x1080
[server] INFO: Device rotation requested: landscape
INFO: New texture: 1080x496
[server] INFO: Device rotation requested: portrait
[...]

PR #6822, opened 2026-05-10
Restore retry-on-error mechanism

rom1v:
First, restore retry on spurious error.
    
MediaCodec may fail spuriously, typically when stopping an encoding and starting a new one immediately (for example on device rotation).
    
This was initially fixed by 07806ba9159b1f4dbd2ede77b110e254af1fd7f6 (#3693), but the fix was removed by 4f97e2e30b9b02ff43d64e940165acafd7512a23 (#6766), along with the removal of the mechanism to retry capture at a lower resolution on error.

---

Then, restore retry capture at lower resolution.
    
The mechanism to retry capture at a lower resolution on error (#2947) was removed in favor of using video encoder capabilities instead (#6766).
    
However, an encoder configuration that is compatible with the declared capabilities may still fail, either because the declared capabilities are incorrect or because some resources are unavailable.

Restore the downsize-on-error behavior for such cases.

This reverts commit 4f97e2e30b9b02ff43d64e940165acafd7512a23 (#6766).

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
@@ -314,6 +314,12 @@ public class NewDisplayCapture extends SurfaceCapture {
         return videoSize;
     }
 
+    @Override
+    protected boolean applyNewVideoConstraints(VideoConstraints videoConstraints) {
+        setVideoConstraints(videoConstraints); // with synchronization
+        return true;
+    }
+
     private static int scaleDpi(Size initialSize, int initialDpi, Size size) {
         int den = initialSize.getMax();
         int num = size.getMax();
@@ -349,6 +355,7 @@ public class NewDisplayCapture extends SurfaceCapture {
 
     private synchronized void triggerResize(Size size) {
         if (virtualDisplay != null) {
+            size = size.constrain(videoConstraints); // in case the constraints have changed
             int displayId = virtualDisplay.getDisplay().getDisplayId();
             DisplayInfo displayInfo = ServiceManager.getDisplayManager().getDisplayInfo(displayId);
             int displayRotation = displayInfo.getRotation();
--- a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
@@ -191,6 +191,12 @@ public class ScreenCapture extends SurfaceCapture {
         return videoSize;
     }
 
+    @Override
+    protected boolean applyNewVideoConstraints(VideoConstraints videoConstraints) {
+        this.videoConstraints = videoConstraints;
+        return true;
+    }
+
     private static IBinder createDisplay() throws Exception {
         // Since Android 12 (preview), secure displays could not be created with shell permissions anymore.
         // On Android 12 preview, SDK_INT is still R (not S), but CODENAME is "S".
--- a/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
@@ -32,6 +32,8 @@ public class SurfaceEncoder implements AsyncProcessor {
     private static final int REPEAT_FRAME_DELAY_US = 100_000; // repeat after 100ms
     private static final String KEY_MAX_FPS_TO_ENCODER = "max-fps-to-encoder";
 
+    // Keep the values in descending order
+    private static final int[] MAX_SIZE_FALLBACK = {2560, 1920, 1600, 1280, 1024, 800};
     private static final int MAX_CONSECUTIVE_ERRORS = 3;
 
     private final SurfaceCapture capture;
@@ -41,6 +43,7 @@ public class SurfaceEncoder implements AsyncProcessor {
     private final int videoBitRate;
     private final int maxSize;
     private final float maxFps;
+    private final boolean downsizeOnError;
     private final int minSizeAlignment;
 
     private boolean firstFrameSent;
@@ -59,6 +62,7 @@ public class SurfaceEncoder implements AsyncProcessor {
         this.maxFps = options.getMaxFps();
         this.codecOptions = options.getVideoCodecOptions();
         this.encoderName = options.getVideoEncoder();
+        this.downsizeOnError = options.getDownsizeOnError();
         this.minSizeAlignment = options.getMinSizeAlignment();
     }
 
@@ -102,11 +106,19 @@ public class SurfaceEncoder implements AsyncProcessor {
 
             streamer.writeVideoHeader();
 
+            int retainedResetReasons = 0;
+
             do {
                 int resetReasons = captureControl.consumeReset();
                 if ((resetReasons & CaptureControl.RESET_REASON_TERMINATED) != 0) {
                     break;
                 }
+                if (retainedResetReasons != 0) {
+                    // The reasons for the previous failed encoding must be preserved when retrying
+                    resetReasons |= retainedResetReasons;
+                    retainedResetReasons = 0;
+                }
+
                 capture.prepare();
                 Size size = capture.getSize();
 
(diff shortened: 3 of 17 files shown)


server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
```diff
@@ -209,4 +209,10 @@ private static void setDisplaySurface(IBinder display, Surface surface, Rect dev
             SurfaceControl.closeTransaction();
         }
     }
+
+    @Override
+    protected boolean applyNewVideoConstraints(VideoConstraints videoConstraints) {
```
anotheruserofgithub: It would be nice if this was moved right below `getSize()` like previously.

server/src/main/java/com/genymobile/scrcpy/video/VideoConstraints.java
```diff
@@ -73,4 +73,14 @@ public Size getMaxCodecPortraitSize() {
     public int getMinCodecSize() {
         return minCodecSize;
     }
+
+    /**
+     * Return the video constraints with the provided max size.
+     *
+     * @param maxSize the max requested size
+     * @return new video constraints
```
anotheruserofgithub:
```suggestion
     * @return the new video constraints
```

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -102,11 +106,23 @@ private void streamCapture() throws IOException, ConfigurationException {
...
                 if ((resetReasons & CaptureControl.RESET_REASON_TERMINATED) != 0) {
                     break;
                 }
+                if (resetReasons != 0) {
+                    // No frame has been produced since the last reset
+                    firstFrameSent = false;
+                }
+                if (retainedResetReasons != 0) {
+                    // The reasons for the previous failed encoding must be preserved when retrying
+                    resetReasons |= retainedResetReasons;
```
anotheruserofgithub, line 118: I'm not sure this is always true but as usual I may be mistaken. For instance, if the user resizes the window of a flex display to a new valid size, this starts a new encoding session with the `CLIENT_RESIZED` flag as reset reason. Then if the user manually resets the video capture (with <kbd>MOD</kbd>+<kbd>Shift</kbd>+<kbd>r</kbd>), this starts yet another encoding session with `CLIENT_RESET` as reset reason, however the retained reset reasons would still contain `CLIENT_RESIZED` (am I right or not?) thus the new session packet would have a wrong client resized flag. The same reasoning applies to anything else than `CLIENT_RESET`.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -102,11 +106,23 @@ private void streamCapture() throws IOException, ConfigurationException {
...
+            int retainedResetReasons = 0;
+
             do {
                 int resetReasons = captureControl.consumeReset();
                 if ((resetReasons & CaptureControl.RESET_REASON_TERMINATED) != 0) {
                     break;
                 }
+                if (resetReasons != 0) {
+                    // No frame has been produced since the last reset
+                    firstFrameSent = false;
```
anotheruserofgithub: So the retry-on-error stuff will apply if the user resizes a flex dispay to a too big size? That's great as it avoids the explicit `--max-size` workaround in case of incorrect video encoder capabilities.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
rom1v: > however the retained reset reasons would still contain CLIENT_RESIZED (am I right or not?) thus the new session packet would have a wrong client reset flag.

The flags are retained until at least one frame is produced in a new session.

In theory, a "reset" should not change the frame size, so it should not impact the resizing behavior.

If the client resizes and then resets, once the client receives the first frame after resizing was requested, it should be considered as a "client resize" IMO (not resizing the client window).

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
rom1v: > So the retry-on-error stuff will apply if the user resizes a flex display to a too big size?

Ah, that was the intent, but in the end I wanted to remove that part. The max size should not change in the middle of a session (there's even a comment about this):

```
// Downsizing on error is only enabled if an encoding failure occurs before the first frame (downsizing later could be surprising)
```

Moreover, it breaks the "retry on spurious error" mechanism that could recover without downsizing.

anotheruserofgithub: OK, got it!

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
anotheruserofgithub: OK, let me try another case. If display properties change (e.g. rotation with <kbd>MOD</kbd>+<kbd>r</kbd>), the current session is interrupted so `retainedResetReasons` will contain anything in the current `resetReasons` (assume 0) and `resetReasons` will consume `DISPLAY_PROPERTIES_CHANGED` for the new encoding session. Now if the client resizes the window, this interrupts the session and `retainedResetReasons` copies the current `resetReasons` so it will contain `DISPLAY_PROPERTIES_CHANGED`. Hence for the next session `resetReasons` will have both `CLIENT_RESIZED` from `consumeReset()` and `DISPLAY_PROPERTIES_CHANGED` from `retainedResetReasons`, which means that `isClientResize` will be `false` even though this reset was triggered by a client resize request, thus a bad flag in the session packet.

What I don't understand is where this happens:
> The flags are retained until at least one frame is produced in a new session.

anotheruserofgithub: Oh, sorry, I think I get it. When there is a new session due to any of the flags in `CaptureControl`, then `encode()` finishes gracefully by EOS and `retainedResetReasons` does _not_ copy `resetReasons`. The new `catch` in `streamCapture()` is only for retries when `encode()` fails for another reason, e.g., bad size, in other words when there were no frames produced yet in this session (or a spurious failure). Am I correct now?

EDIT: I must be tired, that was obvious since `catch` is newly (re)introduced and it calls `prepareRetry()`.

anotheruserofgithub: So this is all right. ;) Sorry, I didn't have the full picture in mind.

rom1v: > Hence for the next session `resetReasons` will have both `CLIENT_RESIZED` from `consumeReset()` and `DISPLAY_PROPERTIES_CHANGED` from `retainedResetReasons`, which means that `isClientResize` will be `false` even though this reset was triggered by a client resize request, thus a bad flag in the session packet.

Yes. But there is an inherent race condition between the display rotation on the server side and the client side resize anyway.

> The new `catch` in `streamCapture()` is only for retries when `encode()` fails for another reason, e.g., bad size, in other words when there were no frames produced yet in this session (or a spurious failure). Am I correct now?

Yes :+1:

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
```diff
@@ -349,6 +355,7 @@ private synchronized void setVideoConstraints(VideoConstraints videoConstraints)
 
     private synchronized void triggerResize(Size size) {
         if (virtualDisplay != null) {
+            size = size.constrain(videoConstraints); // in case the constraints have changed
```
anotheruserofgithub, line 358:
```suggestion
            size = size.constrain(videoConstraints, false); // in case the constraints have changed
```

anotheruserofgithub (commented):
An error I spotted while reviewing https://github.com/Genymobile/scrcpy/pull/6859 (how new video constraints are handled). See f2ee5b3d0646c77bdc61c1a4a63d2b9cfd7fbc38.

server/src/main/java/com/genymobile/scrcpy/video/NewDisplayCapture.java
rom1v: Good catch :+1:

PR #6825, opened 2026-05-11
Set terminal tab title to scrcpy and device name

zxmpg41:
# Summary

Set the terminal tab/window title to `scrcpy` when the program starts, and update it to `scrcpy - <device name>` once a device is connected. This makes it easy to identify scrcpy tabs when multiple tabs are open in a terminal multiplexer or tabbed terminal emulator (e.g. Windows Terminal, macOS Terminal, tmux).

scrcpy_title  

The title respects `--window-title`: if a custom window title is provided, the terminal tab title becomes `scrcpy - <custom title>` instead of using the device name.

The original terminal title is restored (cleared) when scrcpy exits, allowing the shell to reset its own title on the next prompt.

# Changes

- **`app/src/util/term.h`** — declared new function `sc_term_set_title(const char *title)`
- **`app/src/util/term.c`** — implemented `sc_term_set_title()`: only emits output when stdout is a TTY (`isatty()` on Unix/macOS, `_isatty()` on Windows), so piping (e.g. `scrcpy | tee logs.txt`) is not affected; emits an OSC 0 escape sequence (`\033]0;<title>\007`) to stdout, supported by all modern terminal emulators; on Windows additionally calls `SetConsoleTitleA()` unconditionally as a fallback for legacy console hosts (`cmd.exe`, `conhost`) that do not process VT sequences
- **`app/src/main.c`** — calls `sc_term_set_title("scrcpy")` immediately on startup; calls `sc_term_set_title("")` on exit to restore the terminal title
- **`app/src/scrcpy.c`** — calls `sc_term_set_title("scrcpy - <name>")` after the device handshake completes (using `--window-title` if provided, otherwise the device name reported by the server); calls `sc_term_set_title("")` on exit
- **`app/src/usb/scrcpy_otg.c`** — same behaviour in OTG mode: updates the terminal tab title to `scrcpy - <USB product name>` once the device is known; calls `sc_term_set_title("")` on exit

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -63,6 +63,7 @@ _scrcpy() {
         --no-mipmaps
         --no-mouse-hover
         --no-power-on
+        --no-terminal-title
         --no-vd-destroy-content
         --no-vd-system-decorations
         --no-video
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -69,6 +69,7 @@ arguments=(
     '--no-mipmaps[Disable the generation of mipmaps]'
     '--no-mouse-hover[Do not forward mouse hover events]'
     '--no-power-on[Do not power on the device on start]'
+    '--no-terminal-title[Disable automatic terminal title updates]'
     '--no-vd-destroy-content[Disable virtual display "destroy content on removal" flag]'
     '--no-vd-system-decorations[Disable virtual display system decorations flag]'
     '--no-video[Disable video forwarding]'
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -421,6 +421,10 @@ Do not forward mouse hover (mouse motion without any clicks) events.
 Do not power on the device on start.
 
 .TP
+.B \-\-no\-terminal\-title
+Disable automatic terminal title updates.
+
+.TP
 .B \-\-no\-vd\-destroy\-content
 Disable virtual display "destroy content on removal" flag.
 
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -106,6 +106,7 @@ enum {
     OPT_CAMERA_ZOOM,
     OPT_MIN_SIZE_ALIGNMENT,
     OPT_NO_WINDOW_ASPECT_RATIO_LOCK,
+    OPT_NO_TERMINAL_TITLE,
     OPT_KEEP_ACTIVE,
     OPT_BACKGROUND_COLOR,
     OPT_RENDER_FIT,
@@ -658,6 +659,11 @@ static const struct sc_option options[] = {
         .longopt_id = OPT_NO_POWER_ON,
         .longopt = "no-power-on",
         .text = "Do not power on the device on start.",
+    },
+    {
+        .longopt_id = OPT_NO_TERMINAL_TITLE,
+        .longopt = "no-terminal-title",
+        .text = "Disable automatic terminal title updates.",
     },
     {
         .longopt_id = OPT_NO_VD_DESTROY_CONTENT,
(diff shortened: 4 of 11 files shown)


rom1v:
Thank you, interesting. :+1:

I think it should be printed only when the output is a tty (using `isatty()` I guess), otherwise with `scrcpy | tee logs.txt` the logs would include these special characters.

Also, please rebase onto `dev`.

zxmpg41:
Good point. I've added `isatty()` and `_isatty()` for legacy windows check and rebased onto `dev`.  
Ready for another look whenever you get a chance.

zxmpg41:
@rom1v can you have another look?

rom1v:
Yes, I'll probably integrate this in scrcpy 4.1. Will review later.

I'd also like an option to disable it in case it breaks some workflows (`--no-terminal-title` or similar).

Here is an example of commit adding a client-only option: 65153a09c367d7a8534a152241247d143bbae72e

rom1v:
Thank you. LGTM overall.

I refactored a bit and made a few changes (branch [`terminal-title.4`](https://github.com/Genymobile/scrcpy/commits/terminal-title.4/)):
 - limited length to 128 and truncated at UTF-8 code points (and avoided `strncpy()`)
 - replaced `SetConsoleTitleA()` with `SetConsoleTitleW()`
 - added a commit to restore the previous terminal title
 
Not tested on Windows.

Please review and test on Windows, including with special characters like `--window-title=中国`.

rom1v:
I'll test my branch on Windows and merge soon. If you get the chance to review/test, please don't hesitate.

zxmpg41:
Sorry took me a while to find time.
So regarding your branch I have two open points other than that everything works.

### Open points
1. Fix includes for `term.c`: https://github.com/Genymobile/scrcpy/blob/1ba2711158ca4920c132b39655871a4c32f9b8f7/app/src/util/term.c#L3
```
#include <assert.h>
+ #include <stdio.h>
+ #include "util/log.h"
+ #include "util/str.h" 
```
2. Wrong variable `varname` -> `title`: https://github.com/Genymobile/scrcpy/blob/1ba2711158ca4920c132b39655871a4c32f9b8f7/app/src/util/term.c#L71
`varname` should be `title`

After these changes it compiles and works

### Tested combinations of parameters
System: Windows 11 Enterprise 23H2
- [x] scrcpy without any paramters
- [x] --window-title=zxmpg41
- [x] --no-terminal-title
- [x] --no-terminal-title --window-title=zxmpg41

rom1v:
Thank you. Please test [`terminal-title.6`](https://github.com/Genymobile/scrcpy/commits/terminal-title.6/) with the fixes.

Could you also please test with special chars on Windows? `--window-title=é中国`

zxmpg41:
Oh my bad I forgot to add it to the documentation. But I did test it:

```
./scrcpy.exe
./scrcpy.exe --window-title=zxmpg41
./scrcpy.exe --window-title=é中国
./scrcpy.exe --no-terminal-title
./scrcpy.exe --no-terminal-title --window-title=zxmpg41
```
scrcpy

Branch `terminal-title.6` looks good on my machine

rom1v:
Thank you for your tests.

I just tested on windows in cmd, the title is correctly set, but it prints `←[22;2t` in the terminal (due to my last commit). I don't know how to detect if we should print or not. To be investigated.

rom1v:
> I just tested on windows in cmd, the title is correctly set, but it prints `←[22;2t` in the terminal (due to my last commit). I don't know how to detect if we should print or not. To be investigated.

Thanks to Claude, I think I have a solution: [`terminal-title.8`](https://github.com/Genymobile/scrcpy/commits/terminal-title.8/).

Could you please test both on your terminal and on `cmd.exe`?

zxmpg41:
Rebuilt scrcpy using `terminal-title.8` and tested all above mentioned combinations using
- Windows PowerShell
- cmd.exe
- Terminal with MINGW64 shell

Issue from previous branch was reproducible and is now fixed

rom1v:
Thank you, merged into `dev` :rocket:

zxmpg41:
Closing the PR as it is merged into `dev`. Looking forward to the next release :)

PR #6769, opened 2026-04-13
Add support for VP8 and VP9 video encoders

pngocthach:
### **Summary**
This PR adds support for **VP8** and **VP9** video codecs. This is primarily intended as a fallback for devices where the manufacturer has intentionally disabled H.264 (AVC) and H.265 (HEVC) encoders in the firmware.

### **The Problem (Onyx Boox Note X case)**
On recent Onyx Boox devices (e.g., Note X with current firmware), the vendor has disabled both hardware and software H.264/H.265 encoders. This makes `scrcpy` unusable as it cannot find any compatible video encoder.

### **Evidence**
In `/vendor/etc/media_codecs.xml`, all H.264/H.265 encoders are commented out:

<details>
<summary><b>/vendor/etc/media_codecs.xml</b></summary>

```xml
<MediaCodecs>
    <Include href="media_codecs_google_audio.xml" />
    <Include href="media_codecs_google_telephony.xml" />
    <Settings>
        <Setting name="max-video-encoder-input-buffers" value="11" />
    </Settings>
    <Encoders>
        <!-- Video Hardware  -->
        <!-- Disabled: H.264/AVC encoder -->
        <!--
        <MediaCodec name="OMX.qcom.video.encoder.avc" type="video/avc" >
            <Quirk name="requires-allocate-on-input-ports" />
            <Quirk name="requires-allocate-on-output-ports" />
            <Quirk name="requires-loaded-to-idle-after-allocation" />
            <Limit name="size" min="128x128" max="1920x1088" />
            <Limit name="alignment" value="2x2" />
            <Limit name="block-size" value="16x16" />
            <Limit name="blocks-per-second" min="64" max="244800" />
            <Limit name="bitrate" range="1-60000000" />
            <Limit name="frame-rate" range="1-120" />
            <Limit name="concurrent-instances" max="16" />
            <Limit name="performance-point-1920x1080" value="30" />
            <Limit name="performance-point-1280x720" value="60" />
            <Limit name="performance-point-720x480" value="120" />
        </MediaCodec>
        -->
        <!-- Disabled: HEVC encoder -->
        <!--
        <MediaCodec name="OMX.qcom.video.encoder.hevc" type="video/hevc" >
            <Quirk name="requires-allocate-on-input-ports" />
            <Quirk name="requires-allocate-on-output-ports" />
            <Quirk name="requires-loaded-to-idle-after-allocation" />
            <Limit name="size" min="128x128" max="1920x1088" />
[...]

PR #6849, opened 2026-05-18
Add --ignore-video-encoder-constraints

### LINKED ISSUE #6829: Resolution doesn't match the command request (Texture Resolution)
I previously used Scrcpy version 3.3.4. I ran scrcpy by adding the command `--max-size=1280 --max-fps=60`, and with that version, it ran smoothly and as expected (the expectation being a longer resolution size of 1280).  
After I updated Scrcpy to version 4.0 and ran it with the same additional command, the message

**INFO: texture: 224x496**

and a scrcpy window appeared that did not match the size I requested.
Is this some kind of bug or did I miss something?  

The device I am using:  
Redmi Note 11 4G NFC (spesn)  
Android 16 Custom Rom Crdroid  
Screen Resolution 2400x1080

Image

### LINKED ISSUE #6848: scrcpy v4.0 detects slightly different resolution
## Environment

 - **OS:** macOS 12.7.6 Monterey (also tested on macOS 10.14.6 Mojave)
 - **Scrcpy version:** v4.0
 - **Installation method:** binary from the Releases section
 - **Device model:** Samsung Galaxy A7 (2018) - model number SM-A750FN
 - **Android version:** 10

## Describe the bug

I noticed an inconsistency in how v4.0 detects my phone's display resolution. Releases up to v3.3.4 detected a resolution of 1080x2216, while v4.0 detects a resolution of 1050x2160. Neither is really correct, as my phone is supposed to have a resolution of 1080x2220, but what's more confusing is the discrepancy, especially since the aspect ratio between the two is slightly different.

It also gives a slightly different resolution when constraining the maximum size to 1064, but it's the same when constraining it to 1056.

| Max Size | v3.3.4 | v4.0 |
| --- | --- | --- |
| no restriction | 1080x2216 | 1050x2160 |
| `-m1064` | 520x1064 | 516x1064 |
| `-m1056` | 512x1056 | 512x1056 |

Renderer is metal and encoder left to the default in all cases.

I have tried playing with `--min-size-alignment` values, but the results did not change.

rom1v:
Some encoders report incorrect capabilities. Add an option to ignore these constraints.
    
Fixes #6829
Fixes #6848

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -34,6 +34,7 @@ _scrcpy() {
         -G
         --gamepad=
         -h --help
+        --ignore-video-encoder-constraints
         -K
         --keep-active
         --keyboard=
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -41,6 +41,7 @@ arguments=(
     '-G[Use UHID/AOA gamepad \(same as --gamepad=uhid or --gamepad=aoa, depending on OTG mode\)]'
     '--gamepad=[Set the gamepad input mode]:mode:(disabled uhid aoa)'
     {-h,--help}'[Print the help]'
+    '--ignore-video-encoder-constraints[Do not consider video encoder capabilities]'
     '-K[Use UHID/AOA keyboard \(same as --keyboard=uhid or --keyboard=aoa, depending on OTG mode\)]'
     '--keep-active[Keep the screen on by simulating user activity]'
     '--keyboard=[Set the keyboard input mode]:mode:(disabled sdk uhid aoa)'
--- a/app/scrcpy.1
+++ b/app/scrcpy.1
@@ -236,6 +236,14 @@ Also see \fB\-\-keyboard\fR and R\fB\-\-mouse\fR.
 Print this help.
 
 .TP
+.B \-\-ignore\-video\-encoder\-constraints
+Do not consider video encoder capabilities.
+
+This is useful if the reported capabilities are incorrect.
+
+It may help to force a value for \fB\-\-min\-size\-alignment\fR.
+
+.TP
 .B \-K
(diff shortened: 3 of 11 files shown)


rom1v:
When `--ignore-video-encoder-constraints` is set, maybe I should change the default value for `--min-size-alignment` to 8 if it is not explicitly set, otherwise the video will often fail due to misalignment… Or maybe not…

anotheruserofgithub:
Or maybe this option should only apply to max/min size, not alignment?

I guess the alignment reported by video capabilities will always work (except e.g. for emulators).

But I'm even wondering if the supported size is really wrong. In https://github.com/Genymobile/scrcpy/pull/6769#issuecomment-4501772790 as well as in other issues you're assuming that the encoder reports incorrect capabilities. It seems more probable that the way you compute the maximum possible size is not appropriate, as additional logs in https://github.com/Genymobile/scrcpy/pull/6769#issuecomment-4501157604 tend to suggest (IMHO).

What I mean is that I wouldn't be surprised if the encoder could handle only a limited number of pixels in total or a limited aspect ratio, something like that. Thus if you first retrieve the maximum allowed width and the encoder returns 2048, then if you ask it the maximum possible height it can handle for this width it will return 448. On the other hand, if you ask what is the maximum height for a width of, say, 1600, it would return 1200 or higher. In other words, maybe it can handle larger areas for big squares than for thin rectangles, but anyway the more you increase one dimension, the more the other dimension has to be limited. Or maybe not? How would you know a priori? You should probably test with more values on a real device.

As a consequence, the decision for the maximum minor dimension should depend dynamically on the current (capped) major dimension in `Size.constrain()`.

I don't know, just a guess. At least, did I manage to make myself clear?

rom1v:
@anotheruserofgithub Thank you, you're right! I completely reworked the size constraint algorithm: #6859 (which replaces this PR, so I'm closing).

Issue #6851, opened 2026-05-19
Opt-in --media-scan to make pushed images/videos appear in the gallery


eret9616:
- [x] I have checked that a similar feature request does not already exist.

**Is your feature request related to a problem?**

Dragging a file into the scrcpy window runs `adb push` and stops there. The gallery never sees it because:

1. `adb push` is just a filesystem write — MediaStore is never notified.
2. Most vendor galleries (MIUI, ColorOS, OneUI, Huawei) only show files from a fixed directory list. `/sdcard/Download/` (scrcpy's default push target) isn't on that list; `/sdcard/Pictures/` and `/sdcard/Movies/` usually are.

End result: even a freshly pushed image stays invisible until the user opens a file manager and triggers a scan by hand.

**Solution**

An opt-in `--media-scan` flag that:

- routes image extensions (jpg, png, heic, webp, …) to `/sdcard/Pictures/` and video extensions (mp4, mkv, mov, …) to `/sdcard/Movies/` when `--push-target` is not set; other files keep `/sdcard/Download/`
- after each successful push runs `cmd media_scanner scan-file <remote>` (Android 8+), falling back to `am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file://<remote>` on ROMs where the shell service is missing (OnePlus Android 13 / ColorOS confirmed — returns `cmd: Can't find service: media_scanner`)
- explicit `--push-target` keeps full control of the destination, scan still triggered there

Default off — no behavior change for existing users.

**Alternatives considered**

- Drag-into-window → paste image: blocked. scrcpy runs as shell uid on the device, can't register a ContentProvider, so the clipboard route isn't available.
- Always scan regardless of flag: risks unexpected side effects on `Download/` contents. Opt-in keeps it predictable.

**Implementation**

Working branch on `dev`: https://github.com/eret9616/scrcpy/tree/feat/media-scan-on-push
Tested on OnePlus PGZ110 / Android 13. Will send PR if direction is OK.

Related: #6200

Issue #6200, opened 2025-07-02
Why pictures not showing in gallery after dragging into scrcpy android？


hide-and-seek123:
I could see the pictures in /sdcard/DCIM/Camera, to where I transferred pictures from my PC, but I just can't find them in my gallery. Even after renaming them to something like IMG_xxx, they still don't show up, I also can't find them when when trying to send these pictures in apps. I think the fact that they're missing from the gallery explains this.

eret9616:
Vendor galleries only index a fixed set of folders and need MediaStore to know about the file. `adb push` does neither, so dragged-in images never show up.

Sent a patch as an opt-in `--media-scan`:

- routes images to `/sdcard/Pictures/` and videos to `/sdcard/Movies/` when `--push-target` isn't set
- runs `cmd media_scanner scan-file` after push, with `am broadcast MEDIA_SCANNER_SCAN_FILE` as fallback (OnePlus / ColorOS strips the `media_scanner` shell service — first call returns `Can't find service`)
- explicit `--push-target` keeps current behavior

Tested on OnePlus Android 13. Default off.

Branch: https://github.com/eret9616/scrcpy/tree/feat/media-scan-on-push

PR #6852, opened 2026-05-19
Add --media-scan to register pushed files with MediaStore

eret9616:
Refs #6851 (also discussed in #6200).

Pushed files (especially images dragged into the window) don't show in the gallery: MediaStore isn't notified and most vendor galleries filter `/sdcard/Download/` out. See the linked issue for the full reasoning.

This adds an opt-in `--media-scan`:

- routes image/video extensions to `/sdcard/Pictures/` and `/sdcard/Movies/` when `--push-target` is not set
- runs `cmd media_scanner scan-file` after each push (Android 8+), with `am broadcast MEDIA_SCANNER_SCAN_FILE` as fallback for ROMs that strip the shell service (OnePlus / ColorOS confirmed — first call returns `Can't find service`)
- explicit `--push-target` keeps full control
- default off, no behavior change for existing users

Tested on OnePlus PGZ110 (Android 13), drag-and-drop a `.png` shows up in the gallery within seconds.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/control/Controller.java
+++ b/server/src/main/java/com/genymobile/scrcpy/control/Controller.java
@@ -3,6 +3,7 @@ package com.genymobile.scrcpy.control;
 import com.genymobile.scrcpy.AndroidVersions;
 import com.genymobile.scrcpy.AsyncProcessor;
 import com.genymobile.scrcpy.CleanUp;
+import com.genymobile.scrcpy.FakeContext;
 import com.genymobile.scrcpy.Options;
 import com.genymobile.scrcpy.device.Device;
 import com.genymobile.scrcpy.display.DisplayInfo;
@@ -23,7 +24,10 @@ import com.genymobile.scrcpy.wrappers.InputManager;
 import com.genymobile.scrcpy.wrappers.ServiceManager;
 
 import android.content.Intent;
+import android.net.Uri;
 import android.os.Build;
+import android.os.Bundle;
+import android.provider.MediaStore;
 import android.os.SystemClock;
 import android.util.Pair;
 import android.view.InputDevice;
@@ -334,6 +338,9 @@ public class Controller implements AsyncProcessor, VirtualDisplayListener {
         switch (type) {
             case ControlMessage.TYPE_RESET_VIDEO:
                 resetVideo();
+                return true;
+            case ControlMessage.TYPE_SCAN_FILE:
+                scanFile(msg.getText());
                 return true;
             default:
                 // fall through
@@ -874,5 +881,21 @@ public class Controller implements AsyncProcessor, VirtualDisplayListener {
     private void resizeDisplay(int width, int height) {
         NewDisplayCapture newDisplayCapture = (NewDisplayCapture) surfaceCapture;
         newDisplayCapture.requestResize(width, height);
+    }
+
+    private void scanFile(String path) {
+        if (path == null || path.isEmpty()) {
+            return;
+        }
+        try {
+            // MediaStore.scanFile() is @hide; call the same provider method directly.
+            // Equivalent to: adb shell content call --uri content://media --method scan_file --arg <path>
+            Bundle out = FakeContext.get().getContentResolver()
+                    .call(MediaStore.AUTHORITY, "scan_file", path, null);
+            Uri uri = (out != null) ? out.<Uri>getParcelable(Intent.EXTRA_STREAM) : null;
+            Ln.i("MediaStore scan " + path + " -> " + uri);
+        } catch (Throwable t) {
+            Ln.e("MediaStore scan failed for " + path, t);
+        }
     }
 }
--- a/app/src/adb/adb.c
+++ b/app/src/adb/adb.c
@@ -353,6 +353,47 @@ sc_adb_install(struct sc_intr *intr, const char *serial, const char *local,
 }
 
 bool
+sc_adb_media_scan(struct sc_intr *intr, const char *serial,
+                  const char *remote_path, unsigned flags) {
+    assert(serial);
+    assert(remote_path);
+
+    // First try `cmd media_scanner scan-file` (stock Android 8.0+).
+    {
+        const char *const argv[] =
+            SC_ADB_COMMAND("-s", serial, "shell", "cmd", "media_scanner",
+                           "scan-file", remote_path);
+        sc_pid pid = sc_adb_execute(argv, flags | SC_ADB_SILENT);
+        if (process_check_success_intr(intr, pid,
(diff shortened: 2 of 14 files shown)


rom1v:
On a Pixel 8 with Android 16:

```console
$ adb shell cmd media_scanner 
cmd: Can't find service: media_scanner
```

eret9616:
Huh, didn't realize stock Android dropped it too — thought it was just OEM ROMs. I'll fix the description and the comment in the code.

The PR already falls back to `am broadcast MEDIA_SCANNER_SCAN_FILE` if the first call fails — that's how it works on my OnePlus. But broadcast got deprecated in Android 12, so I don't know if it still works on Pixel 8 / Android 16, and I don't have either device.

@rom1v 

Could you try this on the Pixel 8 and see if the gallery picks it up?

```
adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Pictures/foo.jpg
```

If that's also dead, only path I see is going through MediaStore from the server side (`content insert` or similar). Wanted to confirm the fallback status first.

yume-chan:
https://developer.android.com/reference/android/media/MediaScannerConnection#scanFile(android.content.Context,%20java.lang.String[],%20java.lang.String[],%20android.media.MediaScannerConnection.OnScanCompletedListener)

```java
public static void scanFile (Context context, 
                String[] paths, 
                String[] mimeTypes, 
                MediaScannerConnection.OnScanCompletedListener callback)
```

The API itself was added in Android 2.2, I only tested it on Andorid 13 and Android 16. `context` can be fake context, `mimeTypes` and `callback` can be `null` (in fact `mimeTypes` is straightly ignored in recent Android versions)

Also mentioned by `Context#getExternalFilesDir`: https://developer.android.com/reference/android/content/Context#:~:text=Though%20these%20files%20are%20not%20automatically%20scanned%20by%20the%20media%20scanner%2C%20you%20can%20explicitly%20add%20them%20to%20the%20media%20database%20with%20MediaScannerConnection.scanFile.

eret9616:
@yume-chan thanks, went with your suggestion. Server now handles a new SCAN_FILE control message and calls `ContentResolver.call(MediaStore.AUTHORITY, "scan_file", ...)` through the FakeContext. (`MediaStore.scanFile()` itself is hide on SDK 36, so calling the provider method directly — same path internally.)

Works on my OnePlus 13. @rom1v mind giving it another go on the Pixel when you get a chance?

rom1v:
This works on a Pixel 8:

```
adb push file.jpg /sdcard/Download
adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Download
```

I also found an old commit https://github.com/Genymobile/scrcpy/commit/190a90f45dba7654660a4ec820d62d475f639d44 (on branch [`scan_media`](https://github.com/Genymobile/scrcpy/commits/scan_media/)). I don't remember why I abandoned it, probably because it did not work reliably across devices.

But if it works now using `MediaStore.scanFile()`, great :+1:

However, I would prefer not to auto-detect images/videos to change the target directory, just push to "push-target" whatever it is, then scan this directory.

Also, if there are no downsides, I think an option `--media-scan` is not necessary, just do it unconditionally.

eret9616:
Done in fda146c — both the routing and the --media-scan flag are gone, scan runs unconditionally now.

Tested on a Pixel 8 emulator (Android 17) and the OnePlus PGZ110 from the original report, both fine.

Doc updated accordingly.

server/src/main/java/com/genymobile/scrcpy/control/Controller.java
```diff
@@ -875,4 +882,20 @@ private void resizeDisplay(int width, int height) {
...
         newDisplayCapture.requestResize(width, height);
     }
+
+    private void scanFile(String path) {
+        if (path == null || path.isEmpty()) {
+            return;
+        }
+        try {
+            // MediaStore.scanFile() is @hide; call the same provider method directly.
+            // Equivalent to: adb shell content call --uri content://media --method scan_file --arg <path>
```
anotheruserofgithub, line 892: Why not execute this command directly from the client?

rom1v:
Thank you. :+1:

I made some changes, the resulting branch is [`scan-media.3`](https://github.com/Genymobile/scrcpy/commits/scan-media.3/).

Firstly, I removed the `build_remote_path()` to guess the target file, since the actual file path depends on the directories on the device.

For example:

```
adb push a /sdcard/Download/a
```

sometimes creates a file `/sdcard/Download/a`, but if `/sdcard/Download/a` was already an existing directory, it creates `/sdcard/Download/a/a`.

Instead, for simplicity, I always request to scan the "push target" directory (by default `/sdcard/Download`).

I also added unit tests for the control message serialization.

---

One remaining problem on:

```java
Bundle out = FakeContext.get().getContentResolver().call(MediaStore.AUTHORITY, "scan_file", path, null);
```

> Call requires API level 29 (current min is 21): android.content.ContentResolver#call

so it does not work on older versions (tested on Android 6).

We should probably use (the equivalent of):

```
adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Download
```

(like in 190a90f45dba7654660a4ec820d62d475f639d44)

as a fallback, or even as the main method (since it still works on Android 16).

server/src/main/java/com/genymobile/scrcpy/control/Controller.java
rom1v: @anotheruserofgithub I prefer to execute this on the server side, the server knows best how to request a "scan" (it's an implementation detail that it could also be called from adb).

anotheruserofgithub:
On `scan-media.3`:
- `control_msg.c` L350: `scan media` -> `scan file`
- `file_pusher.h` L27: ` SCAN_MEDIA` -> `SCAN_FILE`

rom1v:
I replaced the MediaStore calls with a broadcast: [`scan-media.6`](https://github.com/Genymobile/scrcpy/commits/scan-media.6/)

Please review and test on your devices :wink:

eret9616:
Tested `scan-media.6` on two devices:

- **Pixel emulator — Android 17 (API 37)**
- **OnePlus PGZ110 — Android 13 (API 33)**

In both cases, files dropped onto the scrcpy window land in `/sdcard/Download/` and show up in the gallery right away. MediaStore picks them up, and there's no `MediaStore scan failed` in the logs. 👍

rom1v:
Thank you. Merged into `dev`:
 - b8a3fca786fc7a998602670a27dc53e5b65e1b45
 - 1b0fc0c0c821ab5b692ced5fe1aa5eaa38bf2f87

Issue #5820, opened 2025-02-05
Windows ARM


siniux:
Any plans for a ARM version for windows?

PR #6857, opened 2026-05-22
Add Windows ARM 64 support

talynone:
With the rising popularity of Windows ARM 64 devices, such as the latest Surface / Snapdragon X laptops, recently released Snapdragon X2 devices and soon to be released Nvidia N1X SOC users are increasingly expecting native applications.

Built and tested the GitHub build Windows ARM artifacts from my fork using my Snapdragon X2 Elite Extreme based Asus Zenbook A16 running Windows 11 ARM 26H1.

This should close off:
https://github.com/Genymobile/scrcpy/issues/5820

And supersede:
https://github.com/Genymobile/scrcpy/pull/3252

Here's a screenshot showing scrcpy.exe running natively on my system:
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -172,25 +172,62 @@ jobs:
           name: build-win32-intermediate
           path: release/work/build-win32/dist-tar/
 
-  build-win64:
-    runs-on: ubuntu-latest
+  build-windows:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: '64'
+            target: win64
+            runs-on: ubuntu-latest
+            shell: bash
+          - arch: arm64
+            target: winarm64
+            runs-on: windows-11-arm
+            shell: 'msys2 {0}'
+    runs-on: ${{ matrix.runs-on }}
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
 
-      - name: Install dependencies
+      - name: Install dependencies (Linux cross-build)
+        if: matrix.target == 'win64'
         run: |
           sudo apt update
           sudo apt install -y meson ninja-build nasm \
              mingw-w64 mingw-w64-tools libz-mingw-w64-dev
 
+      - name: Setup MSYS2 (native Windows ARM64 build)
+        if: matrix.target == 'winarm64'
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: CLANGARM64
+          update: true
+          install: >-
+            base-devel
+            unzip
+            wget
+            autoconf
+            automake
+            libtool
+            make
+            mingw-w64-clang-aarch64-toolchain
+            mingw-w64-clang-aarch64-meson
+            mingw-w64-clang-aarch64-ninja
+            mingw-w64-clang-aarch64-cmake
+            mingw-w64-clang-aarch64-pkgconf
+            mingw-w64-clang-aarch64-zlib
+
       - name: Build
-        run: release/build_windows.sh 64
+        run: release/build_windows.sh ${{ matrix.arch }}
 
       # upload-artifact does not preserve permissions
       - name: Tar
         run: |
-            cd release/work/build-win64
+            cd release/work/build-${{ matrix.target }}
             mkdir dist-tar
             cd dist-tar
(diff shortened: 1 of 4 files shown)


arter97:
Friendly ping @rom1v :)
Cleanly merges.

rom1v:
You don't need an arm64 `adb.exe`?

rom1v:
Thank you for the PR :+1:

For consistency, could you please use a separate build-winarm64 recipe in addition to build-win64, rather than having a common recipe with a matrix, especially given that it requires specific branches (`if: matrix.target == 'winarm64'`)?

talynone:
> You don't need an arm64 `adb.exe`?

It's not required on modern Windows 11 ARM which has a very capable transparent x64 emulator (PRISM). I have tested an extracted native adb.exe from a msys2 environment for testing, but doing that as part of this PR is a bit out of scope and the speed advantages for just adb would probably be imperceptible.

talynone:
> Thank you for the PR 👍
> 
> For consistency, could you please use a separate build-winarm64 recipe in addition to build-win64, rather than having a common recipe with a matrix, especially given that it requires specific branches (if: matrix.target == 'winarm64')?

Sure, I'll try to find time this weekend. Every repo maintainer seems to have a difference preference for this, almost everyone else prefers a shared matrix since doing updates requires changing fewer locations. 

Edit: Do you also want separate build_windows.sh script for Windows ARM or keep the shared one?

rom1v:
> It's not required on modern Windows 11 ARM which has a very capable transparent x64 emulator (PRISM).

:ok_hand: 

> Sure, I'll try to find time this weekend.

Thank you :+1: 

> Every repo maintainer seems to have a difference preference for this, almost everyone else prefers a shared matrix since doing updates requires changing fewer locations.

In theory, matrix is probably better, but:
 - it's currently not done that way for the other recipes (they could be changed all at the same time later if necessary, I prefer keeping the consistency)
 - it's easier to handle target-specific differences
 - especially this one where there's a big `if: matrix.target == 'winarm64'`

> Edit: Do you also want separate build_windows.sh script for Windows ARM or keep the shared one?

Keep the shared one.

talynone:
> Keep the shared one.

Separated the Windows x64/ARM64 builds, tested to make sure it builds/runs via my feature branch (https://github.com/talynone/scrcpy/actions/runs/33988637696) and squashed the commits for a clean merge.

.github/workflows/release.yml
```diff
@@ -201,6 +201,53 @@ jobs:
           name: build-win64-intermediate
           path: release/work/build-win64/dist-tar/
 
+  build-winarm64:
+    runs-on: windows-11-arm
```
rom1v, line 205: Btw, `build-win32` and `build-win64` runs on `ubuntu-latest`, couldn't we do the same for `build-winarm64`?

(And this would allow to build all windows releases from Linux.)

rom1v (commented):
Thank you.

One remaining question inline.

.github/workflows/release.yml
talynone: That would be a major refactor with it's own challenges, probably would have to switch it to llvm and don't know if there would be any blockers/more fixes for the dependencies based on that. Most maintainers prefer native that's why I took that approach. I've done several cross compile PRs and everyone one has asked me to refactor it to use the native runner.

talynone: If you really want me to do that I can, but not this weekend.

rom1v: Thank you for your answer.

> Most maintainers prefer native that's why I took that approach. I've done several cross compile PRs and everyone one has asked me to refactor it to use the native runner.

The more I can build from Linux the better it is for me. Without the CI, I am able to build releases for Linux and win32/win64 (because they're cross-compiled from Linux). That's why I have a preference to cross-build from Linux.

However, it seems it's more complicated for Windows 64 ARM. Out of curiosity, I asked claude to add support for Windows ARM 64: [`claude.winarm64.1`](https://github.com/Genymobile/scrcpy/commits/claude.winarm64.1/).
But as you said it requires a separate toolchain, and it adds complexity with zlib… So the native build is probably easier.

I'll review your MR later.

Thank you.

talynone: @rom1v If you need help testing let me know, I see you have some llvm based code from Claude merged, but not built yet.

PR #6859, opened 2026-05-22
Fix size constraints algorithm

### LINKED ISSUE #6829: Resolution doesn't match the command request (Texture Resolution)

### LINKED ISSUE #6848: scrcpy v4.0 detects slightly different resolution

rom1v:
The algorithm used to constrain a size according to the video capabilities, implemented in #6766, was too simplistic: it computed the maximum landscape size by retrieving the maximum width and its associated maximum height (and vice versa for the maximum portrait size).

In addition to width and height, encoders are also constrained by a maximum surface area (in blocks). As a result, it is possible for the maximum width to be 4096 with an associated maximum height of 512, while the encoder still supports 1600x1200.

In that case, the previous algorithm incorrectly considered 512 to be the absolute maximum height in landscape mode.

Instead, determine the optimal supported size, defined as the largest size with the same aspect ratio supported by the encoder. If preserving the aspect ratio is not required (for virtual displays), then the dimensions derived from this optimal size can be extended.
    
Refs https://github.com/Genymobile/scrcpy/pull/6849#issuecomment-4502114417
Refs #6766
Fixes #6829
Fixes #6848

---

~Also add an option to ignore video encoder constraints.~

Also apply encoder size constraints only after failure. See https://github.com/Genymobile/scrcpy/pull/6859#issuecomment-4534940031

---

Binary built by the CI: https://github.com/rom1v/scrcpy/actions/runs/26405024621

<details><summary>old</summary>

https://github.com/rom1v/scrcpy/actions/runs/26306965950

</details>


server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
+                maxMinor = minor;
+            }
+            if (maxSize > 0 && maxMinor > maxSize) {
+                maxMinor = maxSize;
+            }
+            bestMinor = align(maxMinor, alignment);
+        }
+
+        assert bestMajor >= minMajor : "The dimensions must not be lower than the minimum codec size";
+        int minMinor = landscape ? caps.getSupportedHeights().getLower() : caps.getSupportedWidths().getLower();
```
anotheruserofgithub, line 141:
```suggestion
        int minMinor = alignUp(landscape ? caps.getSupportedHeights().getLower() : caps.getSupportedWidths().getLower(), alignment);
```
For below assertions to work.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
+            maxMajor = major;
+        }
+        if (maxSize > 0 && maxMajor > maxSize) {
+            maxMajor = maxSize;
         }
 
-        w = align(w, alignment);
-        h = align(h, alignment);
+        int minBlock = (minMajor + alignment - 1) / alignment;
+        int maxBlock = maxMajor / alignment;
```
anotheruserofgithub, line 94:
```suggestion
        minMajor = alignUp(minMajor, alignment);
        maxMajor = align(maxMajor, alignment);

        int minBlock = minMajor / alignment;
        int maxBlock = maxMajor / alignment;
```
Well, this suggestion is not optimal and not necessary even for assertions to be correct. But maybe a bit more readable?

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
+            bestMinor = align(minor * bestMajor / major, alignment);
+        } else {
+            int maxMinor = landscape ? caps.getSupportedHeightsFor(bestMajor).getUpper() : caps.getSupportedWidthsFor(bestMajor).getUpper();
+            if (maxMinor > minor) {
+                maxMinor = minor;
+            }
+            if (maxSize > 0 && maxMinor > maxSize) {
+                maxMinor = maxSize;
+            }
+            bestMinor = align(maxMinor, alignment);
```
anotheruserofgithub, line 123:
```suggestion
            maxMinor = align(maxMinor, alignment);
            bestMinor = maxMinor;
```
To match the previous comment, but this may be a ridiculous suggestion.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
-        int maxWidth = maxCodecSize.width;
-        int maxHeight = maxCodecSize.height;
-        if (maxSize > 0) {
-            if (maxWidth > maxSize) {
-                maxWidth = maxSize;
+        if (caps == null) {
+            if (maxSize == 0) {
+                // No constraints
+                return this;
             }
```
anotheruserofgithub: In this case the size should still be aligned, shouldn't it?

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
+        boolean landscape = width >= height;
+        int major = landscape ? width : height;
+        int minor = landscape ? height : width;
 
-        int maxWidth = maxCodecSize.width;
-        int maxHeight = maxCodecSize.height;
-        if (maxSize > 0) {
-            if (maxWidth > maxSize) {
-                maxWidth = maxSize;
+        if (caps == null) {
```
anotheruserofgithub: Notice that `landscape`, `major` and `minor` are not used in the scope of `if (caps == null)` so they could be moved down.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
-                h = height;
+                w = Math.min(width, maxSize);
+                h = Math.min(height, maxSize);
             }
-        } else {
-            w = Math.min(width, maxWidth);
-            h = Math.min(height, maxHeight);
+
+            w = align(w, alignment);
+            h = align(h, alignment);
```
anotheruserofgithub:
```suggestion
            w = Math.max(align(w, alignment), alignment);
            h = Math.max(align(h, alignment), alignment);
```
To not miss the case where `w` or `h` would be lower than `alignment` (thus 0 upon return), which should be the minimum allowed size in this case I think (in other words, `int minSize = alignUp(1, alignment);`, similar to the previous `minCodecSize` but without video capabilities).

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,99 @@ public Size constrain(VideoConstraints constraints) {
...
+        int maxBlock = maxMajor / alignment;
 
-        assert w <= maxWidth : "The width cannot exceed maxWidth";
-        assert h <= maxHeight : "The height cannot exceed maxHeight";
+        int bestBlock = BinarySearch.findLastTrue(minBlock, maxBlock, block -> {
+            int pixels = block * alignment;
+            int w = align(width * pixels / major, alignment);
+            int h = align(height * pixels / major, alignment);
+            return caps.isSizeSupported(w, h);
+        });
```
anotheruserofgithub: I guess you could avoid binary search entirely in most cases if you first check whether `maxMajor` with its corresponding minor is a supported size. Would be equivalent to adding `if (predicate.test(high)) return high;` at the top of `BinarySearch.findLastTrue()`, but it could probably be optimized if done from `Size.constrain()`.

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.function.Predicate;
+
+public final class BinarySearch {
+    private BinarySearch() {
+        // not instantiable
+    }
+
+    /**
+     * Find the last value for which the predicate is true.
```
anotheruserofgithub:
```suggestion
     * Find the highest value for which the predicate is true.
```

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
```diff
@@ -0,0 +1,45 @@
...
+     * - if it returns {@code true} for a given value, it must return {@code true} for lower values.
+     *
+     * @param low       the lowest value
+     * @param high      the highest value
+     * @param predicate the predicate
+     * @return the highest value for which {@code predicate.test(value)} is {@code true}, or {@code low - 1} if not found.
+     */
+    // This function is used for virtual displays, so we don't need to support older versions anyway
+    @TargetApi(AndroidVersions.API_24_ANDROID_7_0) // for Predicate<T>
+    public static int findLastTrue(int low, int high, Predicate<Integer> predicate) {
```
anotheruserofgithub:
```suggestion
    public static int findHighestTrue(int low, int high, Predicate<Integer> predicate) {
```
What do you think of renaming? `@return` uses "highest". And "last" makes me think of linear search intuitively, but that's not a strong argument.

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
```diff
@@ -0,0 +1,45 @@
...
+        int result = low - 1; // low-1 means "not found"
+        while (low <= high) {
+            int mid = low + (high - low + 1) / 2;
+            if (predicate.test(mid)) {
+                result = mid;
+
+                // predicate holds, go right
```
anotheruserofgithub, line 43:
```suggestion
                // predicate holds, go up
```

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
```diff
@@ -0,0 +1,45 @@
...
+                low = mid + 1;
+            } else {
+                // predicate false, go left
```
anotheruserofgithub, line 46:
```suggestion
                // predicate false, go down
```
low <-> high, thus down <-> up, what do you think?

app/src/cli.c
```diff
@@ -428,6 +429,13 @@ static const struct sc_option options[] = {
         .longopt = "help",
         .text = "Print this help.",
     },
+    {
+        .longopt_id = OPT_IGNORE_VIDEO_ENCODER_CONSTRAINTS,
+        .longopt = "ignore-video-encoder-constraints",
+        .text = "Do not consider video encoder capabilities.\n"
+                "This is useful if the reported capabilities are incorrect.\n"
+                "It may help to force a value for --min-size-alignment.",
+    },
```
anotheruserofgithub: Do you think it would be worth adding a note about this option in `doc/video.md#size`?

server/src/main/java/com/genymobile/scrcpy/model/Size.java
anotheruserofgithub: At first sight, in the case where `preserveAspectRatio` is `false`, I think this strategy works when the major dimension is the first to reach the encoder's limits, but maybe not when it's the minor dimension that must be constrained.

Imagine the limits are always 500x1000 (max size is zero) and the alignment is 1px. Assume the previous size was 400x800.

On the one hand, if only the width is increased to 600 while the height remains at 800, then `maxBlock = major = height = 800` and `minor = width = 600`. The binary search thus yields a best size of 500x666 in order to respect the highest supported size while preserving the aspect ratio. Final size 500x666 (I would have expected 500x800).

On the other hand, if trying to go to 400x1200, the best major is 1000 (so a size of 333x1000 if the aspect ratio would have to be preserved), however the best minor can extend afterward up to the original 400 since this remains below the limit 500. Final size 400x1000 (as expected).

Put differently:
- when I pull the major dimension beyond its limit, it just saturates to the highest in-bound value as expected (the minor dimension remains untouched);
- when I pull the minor dimension beyond its limit, it correctly saturates at the limit, however it also shrinks the major dimension while I would expect to just keep the previous value in this direction (think of resizing a flex display).

Note that I elaborated this example just mentally so I may be plain wrong. I will have to test how it behaves in reality.

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
rom1v: In my mental model, the binary search algorithm applies on an horizontal array. It's arbitrary, but I think it is commonly presented that way, so intuitively it's clearer for me with left/right :)

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
rom1v: (same as the other comment, I prefer left/right)

server/src/main/java/com/genymobile/scrcpy/model/Size.java
rom1v: Will think about it later :slightly_smiling_face:

rom1v: Did it in `findHighestTrue()`.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
rom1v: (I prefer to keep it like that)

server/src/main/java/com/genymobile/scrcpy/model/Size.java
rom1v: I prefer the current version.

anotheruserofgithub: OK!

server/src/main/java/com/genymobile/scrcpy/model/Size.java
anotheruserofgithub: OK!

server/src/main/java/com/genymobile/scrcpy/model/Size.java
anotheruserofgithub: Great!

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
anotheruserofgithub: OK!

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
anotheruserofgithub: Fair enough.

doc/video.md
```diff
@@ -31,9 +31,9 @@ If encoding fails, scrcpy automatically tries again with a lower definition
 For camera mirroring, the `--max-size` value is used to select the camera source
 size instead (among the available resolutions).
 
-The size is rounded to a multiple of the _alignment_ required by the encoder, a
-power-of-2 value (1, 2, 4, 8 or 16) that the video width and height must be
-multiples of.
+The size is constrainted by the encoder capabilities, and rounded to a multiple
```
anotheruserofgithub:
```suggestion
The size is constrained by the encoder capabilities, and rounded to a multiple
```

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
```diff
@@ -0,0 +1,53 @@
...
+     * - if it returns {@code true} for a given value, it must return {@code true} for lower values.
+     *
+     * @param low       the lowest value
+     * @param high      the highest value
+     * @param predicate the predicate
+     * @return the highest value for which {@code predicate.test(value)} is {@code true}, or {@code low - 1} if not found.
+     */
+    // This function is used for virtual displays, so we don't need to support older versions anyway
+    @TargetApi(AndroidVersions.API_24_ANDROID_7_0) // for Predicate<T>
+    public static int findHighestTrue(int low, int high, Predicate<Integer> predicate) {
```
anotheruserofgithub, line 28: Notice that it would be trivial to locally define the required interface:
```suggestion
    public interface Predicate {
        boolean test(int value);
    }

    public static int findHighestTrue(int low, int high, Predicate predicate) {
```

rom1v: Yes, sure :+1: but not necessary.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
rom1v: > I think this strategy works when the major dimension is the first to reach the encoder's limits, but maybe not when it's the minor dimension that must be constrained.

:+1:

Since we can't know in advance, I propose to extend the minor first, then the major: 5148947f09b9ba9fbfcba6f90d64f5f5f1164f2c

Does it seem ok to you?

anotheruserofgithub: Yes it looks good. I will test for real later.

I'm just wondering what happens when the encoder constraints are not fixed (like in my simple example) but can vary. Would you need to adjust the size iteratively? I mean, as the following loop:
```java
        if (preserveAspectRatio) {
            bestMinor = […]
        } else {
            int lastBestMajor;
            do {
                lastBestMajor = bestMajor;

                […]
                bestMinor = […]

                […]
                bestMajor = […]
            } while (bestMajor != lastBestMajor);
        }
```

server/src/main/java/com/genymobile/scrcpy/util/BinarySearch.java
anotheruserofgithub: OK!

server/src/main/java/com/genymobile/scrcpy/model/Size.java
rom1v: I wondered the same thing this morning, but I think a single step is sufficient.

Initially, we have a supported size; let's call it `(initialBestMajor, associatedBestMinor)`, derived from `bestBlock`.

First, we compute `bestMinor`:

```
bestMinor = computeBestMinorFromMajor(initialBestMajor)
```

Then `bestMajor`:

```
bestMajor = computeBestMajorFromMinor(bestMinor)
```

Since it is the highest possible _major_ (within contraints) associated with `bestMinor` that is supported, it is necessarily greater or equal to the initial best major (`bestMajor >= initialBestMajor`).

If we try to compute again:

```
newBestMinor = computeBestMinorFromMajor(bestMajor)
```

We know that `bestMajor >= initialBestMajor`, so **assuming that the capabilities are consistent** (the smaller the major, the greater the minor, which is a reasonable assumption), it follows that `newBestMinor <= bestMinor` .

Since `(bestMinor, bestMajor)` is a supported size, we know that at least `bestMinor` is supported, so `newBestMinor >= bestMinor`.

Therefore, `newBestMinor == bestMinor`. QED

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,107 @@ public Size constrain(VideoConstraints constraints) {
...
+            if (maxMajor > major) {
+                maxMajor = major;
+            }
+            if (maxSize > 0 && maxMajor > maxSize) {
+                maxMajor = maxSize;
+            }
+            bestMajor = align(maxMajor, alignment);
         }
 
+        assert bestMajor >= minMajor : "The dimensions must not be lower than the minimum codec size";
```
anotheruserofgithub: This assert may not be satisfied when binary search fails, either because `maxSize < minMajor` or when `maxMajor - minMajor < alignment - 1`. These are corner cases that should not be observed in practice, but that are valid in theory.

For instance, consider a codec alignment of 1px but a user-requested `alignment = 4`, and assume `minMajor = 5` while `maxMajor = 7`. Therefore `minBlock` is rounded up to `2` and `maxBlock` is rounded down to `1`. Since `minBlock > maxBlock`, binary search returns `minBlock - 1` hence `bestBlock = maxBlock`, thus `bestMajor = 4` which is lower than `minMajor = 5`, and the assert fails.

This can be solved by checking for the lower bound explicitly, like for `bestMinor`:
```suggestion
        minMajor = alignUp(minMajor, alignment);
        if (bestMajor < minMajor) {
            bestMajor = minMajor;
        }
```

server/src/main/java/com/genymobile/scrcpy/model/Size.java
anotheruserofgithub: I suppose you meant `newBestMinor == bestMinor` in the last line.

Your reasoning seems correct and the assumption is reasonable indeed. Thank you.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,111 @@ public Size constrain(VideoConstraints constraints) {
...
+                        h = height * maxSize / width;
+                    } else {
+                        w = width * maxSize / height;
+                        h = maxSize;
+                    }
                 } else {
-                    w = width * maxHeight / height;
-                    h = maxHeight;
+                    w = Math.min(width, maxSize);
+                    h = Math.max(height, maxSize);
```
anotheruserofgithub:
```suggestion
                    h = Math.min(height, maxSize);
```
Typo introduced in https://github.com/Genymobile/scrcpy/compare/7a054d4227f70774931d00a318e3b30122468ae9..31fa71c3fd830a44c0f3bf65621d73a39beccbb3.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,111 @@ public Size constrain(VideoConstraints constraints) {
...
+            // The major dimension can potentially be extended
+            maxMajor = landscape ? caps.getSupportedWidthsFor(bestMinor).getUpper() : caps.getSupportedHeightsFor(bestMinor).getUpper();
+            if (maxMajor > major) {
+                maxMajor = major;
+            }
+            if (maxSize > 0 && maxMajor > maxSize) {
+                maxMajor = maxSize;
+            }
+            bestMajor = align(maxMajor, alignment);
+        }
```
anotheruserofgithub, line 134: The commit says `then the second
dimension derived from this optimal size can be extended`, but now it should be "then both dimensions […] can be extended" (potentially), shouldn't it?

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,111 @@ public Size constrain(VideoConstraints constraints) {
...
-        if (w < minCodecSize) {
-            w = minCodecSize;
+        int minBlock = (minMajor + alignment - 1) / alignment;
+        int maxBlock = maxMajor / alignment;
+
+        int bestBlock = BinarySearch.findHighestTrue(
+                minBlock, maxBlock, block -> {
+                    int pixels = block * alignment;
+                    int w = align(width * pixels / major, alignment);
+                    int h = align(height * pixels / major, alignment);
```
anotheruserofgithub, line 100: This could be optimized a bit I think:
```suggestion
                    int testMajor = block * alignment;
                    int testMinor = align(minor * testMajor / major, alignment);
                    int w = landscape ? testMajor : testMinor;
                    int h = landscape ? testMinor : testMajor;
```
That would match current lines 109, 113, 147-148.

rom1v: It could, but I like that the predicate code does not depend on `landscape`, it's simpler like that.

anotheruserofgithub: As you prefer. It depends on `major`, so indirectly it depends on `landscape` already.

rom1v:
The commit `Fix size constraints algorithm` is necessary to correctly compute size constraints.

However, I removed the second commit that added the `--ignore-video-encoder-constraints` option.
Instead, video encoder capabilities no longer constrain the video size unless encoding fails at the requested size.

This provides the best of both worlds:
 - By default, it works at the expected size, even if the capabilities report it as unsupported.
 - If encoding fails, then the video encoder capabilities are used to constrain the size to a valid one.
 - This avoids introducing a new option.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
```diff
@@ -38,61 +42,111 @@ public Size constrain(VideoConstraints constraints) {
     public Size constrain(VideoConstraints constraints, boolean preserveAspectRatio) {
         int maxSize = constraints.getMaxSize();
         int alignment = constraints.getAlignment();
+        MediaCodecInfo.VideoCapabilities caps = constraints.getEncoderCapabilities();
 
         assert maxSize >= 0 : "Max size may not be negative";
         assert alignment > 0 : "Alignment must be positive";
         assert (alignment & (alignment - 1)) == 0 : "Alignment must be a power-of-two";
```
anotheruserofgithub, line 49: Maybe you should add `assert caps != null : "Blah blah";` in the first commit and remove it in the last (as in `VideoConstraints`)?

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -66,40 +68,24 @@ public SurfaceEncoder(SurfaceCapture capture, Streamer streamer, Options options
...
         assert caps != null; // caps cannot be null for a video codec
-        VideoConstraints constraints = createVideoConstraints(maxSize, minSizeAlignment, caps);
+        int alignment = Math.max(caps.getWidthAlignment(), caps.getHeightAlignment());
+        Ln.d("Video codec size alignment requirement: " + alignment + "px");
+        if (alignment < minSizeAlignment) {
+            alignment = minSizeAlignment;
+            Ln.d("Actual video size alignment: " + alignment + "px");
+        }
+
+        // Do not constraint by the declared video encoder capabilities before encoding actually fails
```
anotheruserofgithub:
```suggestion
        // Do not constrain by the declared video encoder capabilities before encoding actually fails
```

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -193,25 +179,35 @@ private void streamCapture() throws IOException, ConfigurationException {
...
             // Must fail immediately
             return false;
         }
 
-        // Downsizing on error is only enabled if an encoding failure occurs before the first frame (downsizing later could be surprising)
+        if (videoConstraints.getEncoderCapabilities() == null) {
+            Ln.i("Applying video encoder constraints");
+            videoConstraints = videoConstraints.withCapabilities(caps);
+            capture.applyNewVideoConstraints(videoConstraints);
+            return true;
```
anotheruserofgithub: I don't know if you should / want to check the return status here?
```suggestion
            if (capture.applyNewVideoConstraints(videoConstraints)) {
                return true;
            }
```

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -193,25 +179,35 @@ private void streamCapture() throws IOException, ConfigurationException {
...
+        }
+
+        if (consecutiveErrors >= MAX_CONSECUTIVE_ERRORS) {
+            // Definitively fail
+            return false;
+        }
```
anotheruserofgithub: Could you explain why you had to move this condition? Is it possible that the encoder sends a few frames before failing due to a bad size, and only then you want to enforce video capabilities? Surely I'm missing the point.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
```diff
@@ -163,7 +149,7 @@ private void streamCapture() throws IOException, ConfigurationException {
                         throw e;
                     }
                     Ln.e("Capture/encoding error: " + e.getClass().getName() + ": " + e.getMessage());
-                    if (!prepareRetry(constraints, size)) {
+                    if (!prepareRetry(caps, size)) {
```
anotheruserofgithub: BTW: Does this retry mechanism affect camera or other `SurfaceCapture` subclasses?

anotheruserofgithub (commented):
> However, I removed the second commit that added the `--ignore-video-encoder-constraints` option.
> Instead, video encoder capabilities no longer constrain the video size unless encoding fails at the requested size.

That's the best solution. A bit sad that some encoders report wrong capabilities.

server/src/main/java/com/genymobile/scrcpy/model/Size.java
rom1v: I could have done that. But nevermind, not important.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
rom1v: Again this one :smile: My non-native English is leaking.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
rom1v: Good catch :+1:

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
rom1v: Previously, downsizing on error was only enabled if an encoding failure occurred before the first frame (because downsizing later could be surprising).

But with both:
 - flex displays
 - video encoder capabilities not enforced initially

this is no longer possible, because resizing the window of a flex display to a size bigger than the supported size would automatically make it fail without retrying.

The now strategy is almost the same, except that we accept "downsizing on error" once to apply video encoder size constraints.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
rom1v: Yes.

server/src/main/java/com/genymobile/scrcpy/video/SurfaceEncoder.java
anotheruserofgithub: Ah yes, I missed the case of flex displays, obviously it makes sense now, thanks. :)

zz64:
Hi,
this fix not included in master (from today) ?
```
usr/local/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 1085.1 MB/s (732226 bytes in 0.001s)
[server] INFO: Device: [Xiaomi] Xiaomi MiTV-MOOQ0 (Android 10)
INFO: Renderer: opengl
INFO: OpenGL version: 4.6 (Compatibility Profile) Mesa 25.0.7-2
INFO: Texture: 398x224
```
2026-06-13-165409_727x428_scrot

Tech-Tac:
it was merged into dev, it will arrive in master later

zz64:
> it was merged into dev, it will arrive in master later

Where can get "dev" source code ?

Tech-Tac:
```bash
git checkout dev
```

zz64:
Build ok, nothing change. bad texture ...
need option or something ? like --ignore-video-encoder-constraints ...

rom1v:
There's a binary built by the CI in this PR.

Tech-Tac:
As rom1v said, you can try the prebuilt binaries.

~~There is is a possibility that something from the old branch may have stuck, it happened to me once, try cloning the repo to a new folder and building again. If that still doesn't work then the issue is still present.~~

zz64:
I used :
```
rm -rf scrcpy
git clone https://github.com/Genymobile/scrcpy.git -b dev
cd scrcpy
./install_release.sh
```

rom1v:
install_release.sh installs the prebuilt server from the latest release, so it won't include the fix.

zz64:
Ok, downgraded to 3.3.4 and all... fine.
Thanks.

PR #6872, opened 2026-05-28
feat: add --shortcut-key option to customize shortcut key bindings

muxueqz:
Introduce a new CLI option `--shortcut-key=action=key` that allows users
to override individual shortcut keys (e.g., `--shortcut-key=paste=d`,
`--shortcut-key=home=j`). This also solves the issue where shortcut keys
conflict with app shortcuts like Gboard's Alt+V — since an unmatched key
now falls through to the device instead of being swallowed by the catch-all
return.

Changes:
- Add `struct sc_shortcut_key_bindings` in options.h with per-action key fields
- Set default bindings in options.c matching the original hardcoded values
- Convert the large switch statement in input_manager.c to a chain of
  `if (sdl_keycode == keys->xxx)` blocks via `sc_shortcut_key_bindings`
- Remove the catch-all `return;` so unrecognized shortcut keys pass through
  to the device
- Add `--shortcut-key` parsing in cli.c with conflict detection
- Add tests in test_cli.c for the new option
- Update documentation in shortcuts.md and shell completions

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -88,6 +88,7 @@ _scrcpy() {
         -s --serial=
         -S --turn-screen-off
         --screen-off-timeout=
+        --shortcut-key=
         --shortcut-mod=
         --start-app=
         -t --show-touches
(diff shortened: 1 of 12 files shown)


aalmanasir:
done

aalmanasir (approved):
done

aalmanasir:
done

aalmanasir (approved):

PR #6873, opened 2026-05-28
Fix initial gamepad open with SDL joystick IDs

awsms:
`SDL_GetJoysticks()` returns `SDL_JoystickID` values, but the initial gamepad-added event used the loop index as event.gdevice.which.

When joystick IDs differ from their array index, scrcpy may try to open the wrong device at startup, causing `SDL_OpenGamepad()` to fail with errors like "Couldn't find mapping for device (0)".

Use the actual `SDL_JoystickID` when pushing the synthetic `SDL_EVENT_GAMEPAD_ADDED` event.


rom1v:
Thank you, but already fixed by #6853 on `dev` branch.

awsms:
> Thank you, but already fixed by #6853 on `dev` branch.

Merci pour ta rapidité ! :-)

PR #6880, opened 2026-05-29
Fix AbstractMethodError on Android 16 (IDisplayWindowListener)

Lycorecoil:
Android 16 added onDisplayAnimationsDisabledChanged(int, boolean) to IDisplayWindowListener. Without an implementation, the binder dispatches to an abstract method and throws AbstractMethodError, crashing the server.

Add the method to the AIDL stub and provide an empty implementation in DisplayWindowListener to keep it compatible with Android 16+.

Tested on Pixel 6 Pro (Android 16).


rom1v:
It should not be necessary since e11399aff0562a45e397beb9a98743be1c411710.

Lycorecoil:
Oh ok, thanks! I missed that commit, sorry for the noise.

PR #6883, opened 2026-06-01
fix(audio): separate --no-audio-playback from audio capture

### LINKED ISSUE #6846: --no-audio-playback disables audio capture in addition to playback
## Environment

 - **OS:** Pop!_OS 24.04 LTS
 - **Scrcpy version:** 4.0
 - **Installation method:** static build
 - **Device model:** Samsung Galaxy A16 5G
 - **Android version:** 16

## Describe the bug

`--no-audio-playback` disables audio capture in addition to playback. Without `--no-audio-playback`, there are both audio capture and playback.  

Terminal output with `--no-audio-playback`:
```
scrcpy --v4l2-sink=/dev/video9 --video-source=camera --camera-id=1 --camera-size=1920x1080 --no-audio-playback
scrcpy 4.0 <https://github.com/Genymobile/scrcpy>
No audio playback, no recording: audio disabled
INFO: ADB device found:
INFO:     -->   (usb)  <serial>                     device  SM_A166B
/home/<user>/.local/share/scrcpy/scrcpy-server: 1 file pushed, 0 skipped. 360.8 MB/s (732226 bytes in 0.002s)
[server] INFO: Device: [samsung] samsung SM-A166B (Android 16)
INFO: Renderer: opengl
INFO: OpenGL version: 4.6 (Compatibility Profile) Mesa 25.2.8-0ubuntu0.24.04.1
[server] INFO: Using camera '1'
INFO: v4l2 sink started to device: /dev/video9
INFO: Texture: 1920x1080
```

mfolofy:
Fixes #6846. When --no-audio-playback is specified, only local playback should be disabled - audio capture must still function for use cases like AI processing.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -2977,11 +2977,6 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
         opts->video = false;
     }
 
-    if (opts->audio && !opts->audio_playback && !opts->record_filename) {
-        LOGI("No audio playback, no recording: audio disabled");
-        opts->audio = false;
-    }
-
     if (!opts->video && !opts->audio && !opts->control && !otg) {
         LOGE("No video, no audio, no control, no OTG: nothing to do");
         return false;


rom1v:
> When --no-audio-playback is specified, only local playback should be disabled

In scrcpy, if there is no audio sink at all (no playback, no record), then audio capture is disabled on purpose (otherwise nothing will consume the audio packets, and they will accumulate)

> audio capture must still function for use cases like AI processing.

Sure, but scrcpy has no AI processing feature. If someone modifies scrcpy to add another audio sink, then it must not disable audio when that sink is enabled. But this change alone has no reason to be merged in scrcpy.

mfolofy:
Understood — closing per maintainer guidance. The change doesn't make sense for scrcpy's scope. Thanks for the explanation.

mfolofy:
Revisiting this with a sharper argument.

You're right that scrcpy has no built-in audio sink beyond playback and file recording today. But the architectural issue is that `--no-audio-playback` conflates two concerns:

1. **Don't play audio through local speakers** (the user's intent)
2. **Don't capture audio at all** (the side effect)

The user already has `--no-audio` if they want to fully disable capture. The current behavior means `--no-audio-playback` is secretly `--no-audio` unless `--record-filename` is also passed — which violates the principle of least surprise.

A concrete example: `--no-audio-playback --record-filename=-` pipes audio to stdout. The code allows this. But if a third-party tool wants to consume audio from scrcpy without the user hearing it locally AND without piping through stdout+ffmpeg, there's no way to do it without modifying the source — even though `--no-audio-playback` says "just mute local audio."

The fix is clean: separate `no_audio_playback` from `disable_capture`. `--no-audio-playback` = suppress local output only. `--no-audio` = full capture disable. The packet accumulation concern is already handled — if someone passes `--no-audio-playback` with no recorder and no external consumer, the pipeline simply drops decoded frames instead of capturing nothing at all.

This matches how the feature request #6846 was filed: the flag name says "no playback" not "no capture."

rom1v:
> But the architectural issue is that `--no-audio-playback` conflates two concerns

No, `--no-audio-playback` initially only disables audio playback (that's why it's a separate option).

Only when it sees that nothing can consume the audio (or the video), it disables it. It's like `--no-window` implies `--no-video-playback` (and others dependencies like that).

> But if a third-party tool wants to consume audio from scrcpy

How? There's no way to do that without changing the code.

Issue #6905, opened 2026-06-19
[BUG] Scrcpy crashes - divide-by-zero


profucius:
## Environment

- **OS:** Windows 11 Pro Build 26100
- **Scrcpy version:** 4.0
- **Installation method:** Windows release (portable)
- **Device model:** OnePlus 13
- **Android version:** Android 16

## Describe the bug

scrcpy v4.0 crashes immediately on launch with a divide-by-zero exception (`STATUS_INTEGER_DIVIDE_BY_ZERO`, code `0xc0000094`) before any window appears. The fault occurs at a fixed offset within `scrcpy.exe` itself (`0x0000000000019f7e`), consistently across every crash instance, suggesting a deterministic code path rather than a race condition.

The crash was discovered via the Windows Application Event Log (Event ID 1000). scrcpy was being launched by an automation script with no attached console, so no terminal output was captured. The event log entries are provided below as the equivalent.

The crash appears to occur during early initialization, possibly when the target device is not yet available or ADB hasn't enumerated it — scrcpy exits within seconds of launch each time. The automation retries every ~25 seconds, producing bursts of 3–9 crashes per session.

**Launch arguments in use:**
```
scrcpy.exe -w --window-title="DEVICE" --max-size=RES
```

**39 crash instances observed over 7 days (2026-06-13 to 2026-06-19).**

**Windows Application Event Log (Event ID 1000) — representative entry:**
```
Faulting application name: scrcpy.exe, version: 0.0.0.0, time stamp: 0x6a037e5b
Faulting module name: scrcpy.exe, version: 0.0.0.0, time stamp: 0x6a037e5b
Exception code: 0xc0000094
Fault offset: 0x0000000000019f7e
Faulting application path: D:\Portable\scrcpy\scrcpy.exe
Faulting module path: D:\Portable\scrcpy\scrcpy.exe
```

The fault offset `0x0000000000019f7e` is identical across all 39 instances. DLLs present in the installation directory: `SDL3.dll`, `avcodec-62.dll`, `avformat-62.dll`, `avutil-60.dll`, `swresample-6.dll`, `libusb-1.0.dll`.

rom1v:
> scrcpy was being launched by an automation script with no attached console, so no terminal output was captured.

Can you reproduce manually with an attached console?

profucius:
**Update:** Found the reproducing condition.

[...]

PR #6911, opened 2026-06-22
Fix data race

rom1v:
Fields in `struct sc_screen` were written from the decoder thread and read from the main thread without proper synchronization.
    
Contrary to what the comment stated, pushing an SDL event does not provide the synchronization required to make these writes visible to the main thread.
    
Pass the session size as data attached to the SDL event to fix the issue.

CI: https://github.com/rom1v/scrcpy/actions/runs/27986960479

### DIFF
--- a/app/src/demuxer.c
+++ b/app/src/demuxer.c
@@ -244,6 +244,12 @@ run_demuxer(void *data) {
         session = &session_data;
         sc_demuxer_parse_session(header, session);
 
+        if (!session_data.video.width || !session_data.video.height) {
+            LOGE("Invalid session video size: %" PRIu32 "x%" PRIu32,
+                 session_data.video.width, session_data.video.height);
+            goto finally_free_context;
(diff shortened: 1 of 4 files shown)


rom1v:
@profucius Could you please confirm that this fixes #6905?

profucius:
I just had a chance to catch-up here. I think it may be working now, but I need a couple days to test. I'll report back

Tested twice today, worked both times. Need another day or two, it's hard to know when it will happen.

rom1v:
@profucius I added a commit to check zero-sized video (it must be done regardless of your issue). I will merge that.

It might "fix" (report an explicit error instead of crashing) the issue on your specific device: https://github.com/Genymobile/scrcpy/issues/6905#issuecomment-4838538459

Binaries for the latest version are available here: https://github.com/rom1v/scrcpy/actions/runs/28935088279

PR #6946, opened 2026-07-15
Add fuzzy matching for --start-app

### LINKED ISSUE #6945: Fuzzy search for --start-app
Hi, I really like the flex display mode. It is very helpful when testing windowed apps, much nicer interaction than through the android studio. I use `--start-app` flag to launch my app, but also like to use it to quickly compare with other apps, how they behave. Here comes my idea - support fuzzy search, so that I don't have to copy paste full name or package name, to launch specific 3rd party app, but figure out the best match based on the app name I fat fingered with typos and incompletely :).

Pawloland:
Fixes https://github.com/Genymobile/scrcpy/issues/6945#issue-4894777758

### DIFF
--- a/doc/device.md
+++ b/doc/device.md
@@ -185,6 +185,12 @@ For convenience, it is also possible to select an app by its name, by adding a
 scrcpy --start-app=?firefox
 ```
 
+When the name of the app is too long, fuzzy matching can be used by prefixing with a `%` to find the best matching app name. 
+It handles typos and partial matches, for example "Remote Desktop Manager" can be started like so:
+```bash 
+scrcpy --start-app=%remotedesk
+```
+
 But retrieving app names may take some time (sometimes several seconds), so
 passing the package name is recommended.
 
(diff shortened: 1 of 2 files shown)


Pawloland:
[scrcpy-server.zip](https://github.com/user-attachments/files/30102477/scrcpy-server.zip)
If anyone wants to quickly test the server (me until it is not merged, from other devices), this is the built done, from [5fbee8f](https://github.com/Genymobile/scrcpy/pull/6946/commits/5fbee8f3dfa8396b88840699feb47162e7ea7663) - (based on 4.1 release). Just download the server, and replace the one alongside scrcpy binary (remove `.zip` as it is `.apk` file, but github doesn't allow uploading them)
```
Algorithm : SHA256
Hash      : 909DE3CC6707CEBFE860058CD1BC5E4E1383645DA2B00D46C54261CE8ACF15C0
Path      : scrcpy-server.zip
```

PR #6954, opened 2026-07-20
Add --hwaccel for VideoToolbox hardware-accelerated video decoding on macOS

mo2g:
VideoToolbox hardware acceleration reduces CPU usage on Apple Silicon Macs
from ~50% to ~20% during fast screen updates (e.g. swipe scrolling).

### Changes

- Add `--hwaccel` CLI option (disabled by default, opt-in)
- New `hwaccel.c` / `hwaccel.h`: VideoToolbox device setup via
  `av_hwdevice_ctx_create(AV_HWDEVICE_TYPE_VIDEOTOOLBOX)`
- Zero-copy rendering: `CVPixelBufferRef` from hardware frames is wrapped
  as a Metal texture using `SDL_PROP_TEXTURE_CREATE_METAL_PIXELBUFFER_POINTER`,
  avoiding any CPU-side pixel copy
- `sc_texture_set_from_frame()`: VIDEOTOOLBOX path uses early-return to
  isolate from the software NV12/YUV420P path
- Software decoding path is fully preserved — no behavior change without
  `--hwaccel`

### Scope

| Aspect | Detail |
|--------|--------|
| Platform | macOS only — returns `false` on other platforms |
| Renderer | Metal required — warns and disables if non-Metal renderer is explicitly set |
| Default | Disabled — software decoding is the default |
| V4L2 | Incompatible — explicit error on `--hwaccel --v4l2-sink=...` |
| Server | No server-side changes — all modifications in `app/src/` |

### Performance (Apple M1, macOS 14)

| Scenario | CPU (soft) | CPU (hwaccel) |
|----------|-----------|---------------|
| Idle mirroring | ~10% | ~8% |
| Fast swipe scrolling | ~53% | ~24% |

Tested with Homebrew FFmpeg and SDL3.

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -20,6 +20,7 @@ src = [
     'src/file_pusher.c',
     'src/fps_counter.c',
     'src/frame_buffer.c',
+    'src/hwaccel.c',
     'src/input_manager.c',
     'src/keyboard_sdk.c',
     'src/mouse_capture.c',
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -111,6 +111,7 @@ enum {
     OPT_RENDER_FIT,
     OPT_IGNORE_VIDEO_ENCODER_CONSTRAINTS,
     OPT_NO_TERMINAL_TITLE,
+    OPT_HWACCEL,
 };
 
 struct sc_option {
@@ -672,6 +673,14 @@ static const struct sc_option options[] = {
         .longopt_id = OPT_NO_TERMINAL_TITLE,
         .longopt = "no-terminal-title",
         .text = "Disable terminal title updates.",
+    },
+    {
+        .longopt_id = OPT_HWACCEL,
+        .longopt = "hwaccel",
+        .text = "Enable hardware-accelerated video decoding.\n"
+                "Currently only supported on macOS\n"
+                "using VideoToolbox. Requires Metal\n"
+                "renderer.",
     },
     {
         .longopt_id = OPT_NO_VD_DESTROY_CONTENT,
@@ -2944,6 +2953,9 @@ parse_args_with_getopt(struct scrcpy_cli_args *args, int argc, char *argv[],
                 break;
             case OPT_NO_TERMINAL_TITLE:
                 opts->update_terminal_title = false;
+                break;
+            case OPT_HWACCEL:
+                opts->hwaccel = true;
                 break;
             default:
                 // getopt prints the error message on stderr
--- a/app/src/demuxer.c
+++ b/app/src/demuxer.c
@@ -5,6 +5,7 @@
 #include <libavcodec/avcodec.h>
 #include <libavutil/channel_layout.h>
 
+#include "hwaccel.h"
 #include "packet_merger.h"
 #include "util/binary.h"
 #include "util/log.h"
@@ -252,7 +253,10 @@ run_demuxer(void *data) {
 
         codec_ctx->width = session_data.video.width;
         codec_ctx->height = session_data.video.height;
-        codec_ctx->pix_fmt = AV_PIX_FMT_YUV420P;
+
+        if (!demuxer->hwaccel || !sc_hwaccel_setup(codec_ctx)) {
+            codec_ctx->pix_fmt = AV_PIX_FMT_YUV420P;
+        }
 
     } else {
         // Hardcoded audio properties
@@ -362,6 +366,8 @@ sc_demuxer_init(struct sc_demuxer *demuxer, const char *name, sc_socket socket,
     demuxer->name = name; // statically allocated
     demuxer->socket = socket;
     sc_packet_source_init(&demuxer->packet_source);
+
+    demuxer->hwaccel = false;
 
     assert(cbs && cbs->on_ended);
 
--- a/app/src/demuxer.h
+++ b/app/src/demuxer.h
@@ -17,6 +17,8 @@ struct sc_demuxer {
     sc_socket socket;
     sc_thread thread;
 
+    bool hwaccel;
+
     const struct sc_demuxer_callbacks *cbs;
     void *cbs_userdata;
 };
--- a/app/src/hwaccel.c
+++ b/app/src/hwaccel.c
@@ -1,0 +1,53 @@
+#include "hwaccel.h"
+
+#include <libavcodec/avcodec.h>
+#include <libavutil/hwcontext.h>
+#include <SDL3/SDL.h>
+
+#include "util/log.h"
+
+bool
+sc_hwaccel_is_supported(void) {
+#ifdef __APPLE__
+    const char *driver = SDL_GetHint(SDL_HINT_RENDER_DRIVER);
+    // Hardware decoding uses CVPixelBuffer-backed Metal textures, which
+    // only the Metal renderer supports. Default (NULL) means Metal on
+    // macOS — only reject explicitly non-Metal backends.
+    if (driver && strncmp(driver, "metal", 5)) {
+        return false;
+    }
+    return true;
+#else
+    return false;
+#endif
+}
+
+bool
+sc_hwaccel_setup(AVCodecContext *ctx) {
+#ifdef __APPLE__
+    assert(ctx);
+
+    // Idempotent: already configured by a prior call
+    if (ctx->hw_device_ctx) {
+        return true;
+    }
+
+    AVBufferRef *hw_dev = NULL;
+    if (av_hwdevice_ctx_create(&hw_dev, AV_HWDEVICE_TYPE_VIDEOTOOLBOX,
+                                NULL, NULL, 0) >= 0) {
+        ctx->hw_device_ctx = av_buffer_ref(hw_dev);
+        av_buffer_unref(&hw_dev);
+        if (!ctx->hw_device_ctx) {
+            LOG_OOM();
+            return false;
+        }
+        LOGI("VideoToolbox hardware acceleration enabled");
+        return true;
+    }
+    LOGW("VideoToolbox hardware acceleration unavailable");
+    return false;
+#else
+    (void) ctx;
+    return false;
+#endif
+}
--- a/app/src/hwaccel.h
+++ b/app/src/hwaccel.h
@@ -1,0 +1,28 @@
+#ifndef SC_HWACCEL_H
+#define SC_HWACCEL_H
+
+#include <stdbool.h>
+
+struct AVCodecContext;
+
+/**
+ * Check whether the current renderer supports hardware acceleration.
+ *
+ * Currently requires a Metal renderer; returns false for
+ * OpenGL or unknown renderers.
+ */
+bool
+sc_hwaccel_is_supported(void);
+
+/**
+ * Configure AVCodecContext for hardware-accelerated decoding.
+ *
+ * Must be called after avcodec_alloc_context3() and before
+ * avcodec_open2(). On macOS, configures ctx->hw_device_ctx
+ * for VideoToolbox. On other platforms, returns false
+ * immediately (no-op).
+ */
+bool
+sc_hwaccel_setup(struct AVCodecContext *ctx);
+
+#endif
--- a/app/src/options.c
+++ b/app/src/options.c
@@ -89,6 +89,7 @@ const struct scrcpy_options scrcpy_options_default = {
     .window_aspect_ratio_lock = true,
     .window_borderless = false,
     .mipmaps = true,
+    .hwaccel = false,
     .stay_awake = false,
     .force_adb_forward = false,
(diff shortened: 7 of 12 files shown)


rom1v:
Thank you for the MR. (AI-generated?)

Hardware decoding (on all platforms) is the next big subject I want to implement, cf footnote in #6216:

> I'd like to implement hardware decoding first (#3800 | https://github.com/libsdl-org/SDL/issues/5405#issuecomment-1754956025 | https://github.com/libsdl-org/SDL/pull/8366), and if everything works as expected, release a new major version (scrcpy 4.0).
>
> EDIT: hardware decoding will not be for 4.0. Flex display was implemented instead (#6772).

This MR will help for the macOS part :wink:

mo2g:
> Thank you for the MR. (AI-generated?)
> 
> Hardware decoding (on all platforms) is the next big subject I want to implement, cf footnote in #6216:
> 
> > I'd like to implement hardware decoding first (#3800 | [libsdl-org/SDL#5405 (comment)](https://github.com/libsdl-org/SDL/issues/5405#issuecomment-1754956025) | [libsdl-org/SDL#8366](https://github.com/libsdl-org/SDL/pull/8366)), and if everything works as expected, release a new major version (scrcpy 4.0).
> > EDIT: hardware decoding will not be for 4.0. Flex display was implemented instead (#6772).
> 
> This MR will help for the macOS part 😉

Thanks for taking a look! Yes, AI (Claude Code) assisted with the
implementation — the design decisions (CVPixelBuffer zero-copy
via SDL3 Metal pixelbuffer API, early-return in
`sc_texture_set_from_frame`, keeping the software path untouched)
were my own, and AI helped write the code and review the diff.

The motivation came from personal use: I first noticed high CPU
(~50%) on my M1 MacBook when using QtScrcPy to play mobile games
with my kid. I managed to optimize that with VideoToolbox, then
realized scrcpy had the same issue during fast swipes, which led
to this PR.

I'm aware hardware decoding is on your roadmap — happy to adapt
the implementation however you see fit once you get to a detailed
review. No rush at all. 😄

Ran-Xing:
It looks really cool. I tried to optimize scrcpy to achieve the ability to play FPS games at high speed, but since I'm making an Android app, I suddenly wanted to laugh because I'm controlling Android from within Android.

[Screen-Remote](https://github.com/XRSec/Screen-Remote)

Could you give me some advice? I'm currently using `c2.qti.avc.decoder.low_latency` `720/1080/1920` MAX_SIZE `1M/2M/4M` BITRATE `60/90/120` FPS

video rendering and control are already at their maximum.

mo2g:
> It looks really cool. I tried to optimize scrcpy to achieve the ability to play FPS games at high speed, but since I'm making an Android app, I suddenly wanted to laugh because I'm controlling Android from within Android.
> 
> [Screen-Remote](https://github.com/XRSec/Screen-Remote)
> 
> Could you give me some advice? I'm currently using `c2.qti.avc.decoder.low_latency` `720/1080/1920` MAX_SIZE `1M/2M/4M` BITRATE `60/90/120` FPS
> 
> video rendering and control are already at their maximum.

# MTK Encoder Optimization: Key Takeaways for Screen-Remote

> Reply to [Screen-Remote](https://github.com/XRSec/Screen-Remote) (scrcpy#6954)

Screen-Remote is a really cool project — same-device Android casting is something scrcpy fundamentally can't do, and bridging that gap for FPS gaming is genuinely useful. When MTK users inevitably report blur, start them on `video-qp-max=35 + bitrate-mode=1 + i-frame-interval=10`. That triplet fixes ~90% of cases. The other 10% need the auto-reset mechanism described below.

I primarily work on MTK device optimization for scrcpy-derived tools. Below are the key lessons from the Mediatek trenches.

---

## 1. MTK Encoder Has "Sticky QP" Degradation

MTK's `c2.mtk.avc.encoder` (Dimensity chips) has a nasty trait: when scene complexity spikes — gunfights, racing, explosions — the encoder pushes QP up and it **sticks there**. Quality degrades in stair-step fashion: sharp → blurry → nearly black. It's not thermal. It's not GPU contention. It's an asymmetric hysteresis in MTK's rate control algorithm: QP goes up fast, comes down glacially slow (or not at all).

**The single most effective fix** — cap the QP ceiling:

```
video-qp-max=35    # Standard Android key, API 31+. Works on Qualcomm too.
```

Default QP range is 0–51. Without a cap, the encoder can push QP all the way to 51 on complex scenes. Locked at 35, it simply cannot enter the catastrophic low-quality state.

## 2. Short GOP Breaks the QP Ratchet

I-frames (IDR) reset the encoder's prediction chain, giving QP a fresh start. The default 30s is way too long — MTK will climb several QP steps in that window:

```
i-frame-interval=10    # Don't go beyond 10s for gaming
```

## 3. VBR Mode + Don't Handcuff the Encoder

Two MTK vendor keys I found actively harmful when combined:

```
vendor.mtk.ext.venc.qpvbr.upper-enable:int=0   # "no overshoot allowed"
vendor.mtk.ext.venc.qpvbr.lower-enable:int=0    # "no quality drop allowed"
```

Both at zero tells the encoder "you can't exceed budget AND you can't drop quality." It has nowhere to go except QP=51. Remove them or at least keep `upper-enable=1`:

```
bitrate-mode=1    # VBR — let the encoder breathe on complex frames
```

## 4. Battle-Tested Parameter Set

This config survived hours of 1080p@60fps FPS gaming on MTK Dimensity. For your `CodecOptions`:

```
profile=1          # Baseline: CAVLC + no 8×8 transform
max-bframes=0      # Kill B-frames → zero reorder latency
i-frame-interval=10
priority=0         # Realtime encoding thread
bitrate-mode=1     # VBR
video-qp-max=35    # QP ceiling lock
```

## 5. Auto-Recovery Is Still Necessary

Even with the above, MTK will occasionally glitch. Not often, but not never. I built an `EncoderMonitor` module that detects P-frame cliff-drops in real time and auto-rebuilds the encoder. Without it, users see transient blur that doesn't self-heal.

**Per-frame fast path** (60 checks/sec):

```
Ring buffer of last 30 P-frame sizes.
If P-frame drops below 2KB (normal avg ~8KB) for 3 consecutive frames → trigger reset.
```

**Per-second confirmation path** (15s sliding window, 3 confirmations):

```
pMin < 35% of window avg + (frame-freeze OR bitrate-collapse OR iframe-stall) → trigger reset.
```

## 6. Bitrate: Don't Starve the Encoder

1M–4M at 1080p@120fps is way too low for FPS gaming. Complex motion at high frame rates needs headroom:

| Resolution | FPS | Minimum recommended |
|-----------|-----|---------------------|
| 720p | 60 | 8 Mbps |
| 1080p | 60 | 12 Mbps |
| 1080p | 120 | 20 Mbps |

Screen-Remote is local Android-to-Android (essentially loopback), so there's no network bottleneck. Don't be stingy with bitrate — more bits means lower QP means fewer encoder panics.

## 7. MTK Low-Power Mode (If You Want to Chase Latency)

MTK has a vendor key for encoder operating mode. Low Power mode tames the encoder's internal compute aggressiveness, which can help latency at the margin:

```
vendor.mtk.ext.venc.highquality.feature-on:int=1
vendor.mtk.ext.venc.highquality.mode:int=0    # 0=Low Power, 1=High Quality
```

Note: some MTK firmware versions don't accept these through MediaFormat passthrough — they need to be set directly on the codec. Test on device.

Ran-Xing:
@mo2g doing -> does 

Thanks for the detailed recommendations! One concern I have is that many non-Google devices seem to impose unusual ADB throughput limits. In practice, USB ADB can sometimes be slower than Wi-Fi ADB, while some devices also appear to throttle ADB over Wi-Fi.

Do your minimum bitrate recommendations account for devices with these limitations? For example, could 8/12/20 Mbps cause additional latency or instability when the ADB transport cannot sustain that throughput?

That said, if a device has extremely poor USB ADB performance, it may be reasonable to assume that it is not really suitable for the game mode use case anyway—its overall performance is probably limited as well.

Ran-Xing:
@mo2g I've updated the README; hopefully you can get to know it again.

mo2g:
@Ran-Xing 

Thanks for updating README.md.

Before diving in, an important scope note: **my testing is exclusively on MTK Dimensity devices (8200/8300 series) over 5GHz WiFi and USB.** I haven't tested on Qualcomm, Exynos, Tensor, or other SoCs. I also haven't tested on 2.4GHz WiFi or under significant RF congestion. Everything below is what I've observed in my specific setup — YMMV on different hardware or network conditions.

## USB: I haven't hit limits on my test devices

On Dimensity 8200/8300 devices over USB, ADB reliably sustains 30–50 Mbps in my testing. For the bitrate range I use (8–24 Mbps), this leaves comfortable headroom.

That said, I've seen *reports* (not my own testing) that some OEM kernels — particularly Xiaomi/Redmi MTK variants and certain Samsung Exynos models — appear to cap ADB over USB at surprisingly low rates. If true, the 8–12–20 Mbps recommendations could cause problems on those specific devices. I can't confirm or deny from direct experience.

If you're unsure about a device's ADB throughput, scrcpy already prints a rough indicator when pushing the server JAR — the transfer speed line like `564.7 MB/s (733706 bytes in 0.001s)`. It's not a sustained-throughput benchmark (the file is tiny), but a dramatically low number there is a red flag. Beyond that, the real test is just running your target bitrate and watching for stutter — if the pipe can't keep up, TCP backpressure will show up as frame drops.

## WiFi: the issue I've actually seen is RTT jitter, not throughput

To be clear: my WiFi testing is on **5GHz** with the access point in the same room or one wall away. I haven't tested on 2.4GHz or under heavy channel congestion, so I can't speak to those conditions.

What I *have* observed on 5GHz WiFi: the problem isn't a throughput ceiling — it's **RTT jitter**. Even on a relatively clean 5GHz channel, ping can spike from <5ms to 50–100ms in bursts, especially when other devices on the network start streaming or downloading. TCP over ADB doesn't handle these transient spikes gracefully:
- Retransmissions cascade, briefly saturating the link
- The encoder's pacing assumes a stable drain on the output pipe; it isn't designed for bursty transport
- The decoder starves momentarily, producing visible stutter

This stutter is qualitatively different from the MTK sticky-QP blur I described earlier — it's intermittent frame drops rather than persistent quality degradation — but equally problematic for FPS gaming where every frame counts.

**A tentative suggestion for WiFi** (based on my limited testing, not a universal prescription):

| Parameter | USB (my testing) | WiFi 5GHz (my testing) |
|-----------|-----------------|------------------------|
| bitrate | 12–20 Mbps | 4–8 Mbps (conservative starting point) |
| bitrate-mode | 1 (VBR) | 2 (CBR) — more predictable output pacing |
| i-frame-interval | 10s | 5s — faster recovery if packets do drop |
| priority | 0 (realtime) | 0 (realtime) |
| profile | 1 (Baseline) | 1 (Baseline) |

The WiFi column is a starting point, not gospel. The trade-off is visual quality vs. transport resilience: CBR with modest bitrate won't look as sharp, but it produces a steadier data stream that's less likely to provoke TCP congestion control. **However, I haven't validated this on 2.4GHz or congested networks**, and the optimal settings could be very different on those links.

## For Screen-Remote's use case: correction — transport IS a constraint

I need to correct my earlier assumption. Screen-Remote is **Device A remotely controlling Device B** — not same-device local loopback. The video path goes:

```
Device B (game) → MediaCodec encoder → ADB/TCP/WiFi → Device A → MediaCodec decoder → display
```

This means there **is** a transport bottleneck, same as scrcpy. Phone-to-phone adds another wrinkle: both devices on the same WiFi channel may double airtime contention, and phone WiFi chipsets are weaker than PC counterparts. The bitrate table above is transport-aware — on phone-to-phone WiFi, be more conservative than PC-to-phone. I don't have test data for this topology, so validate against your own measurements.

## When the transport *is* the bottleneck

Since your video stream does traverse ADB or TCP between devices, transport constraints are real. From my MTK experience:

1. **Limit bitrate first** — find the max stable throughput, use ~50% to leave TCP overhead headroom
2. **Drop resolution before framerate** — 720p@60fps beats 1080p@30fps for FPS
3. **Consider CBR** — VBR bursting can trigger TCP congestion control on marginal links
4. **Shorter GOP** (5s) — faster recovery if packets do drop

Again, educated guesses from one platform. Different SoCs, different networks — your mileage will vary.

Ran-Xing:
🤣  

@mo2g 

After some effort, I can now barely play FPS games.

For example, I added a game tag to the app and used the phone's built-in game accelerator.

Thank you very much for your valuable suggestions.

```xml
<application android:appCategory="game"
```

I know a software called [tcgames](https://www.sigma-rt.com/en/tcgames/), but it might be using software to speed things up.

PR #6961, opened 2026-07-26
Add mjpeg video codec as a MediaCodec-free fallback

twaik:
## Why

On some devices MediaCodec is simply broken and scrcpy cannot mirror at all. 
Concrete example: Bliss OS (x86) running as a Dell Inc. 08NPPY (Android 13) guest. Both video and audio MediaCodec calls fail, and scrcpy dies with a fairly cryptic error:

```
/home/twaik/Downloads/scrcpy-linux-x86_64-v3.3.3/scrcpy-server: 1 file pushed, 0 skipped. 629.3 MB/s (732226 bytes in 0.001s)
[server] INFO: Device: [Dell Inc.] Dell Inc. 08NPPY (Android 13)
[server] INFO: Retrying with -m1600...
[server] ERROR: Exception on thread Thread[audio-in,5,main]
android.media.MediaCodec$CodecException: Error 0xe
        at android.media.MediaCodec.native_queueInputBuffer(Native Method)
        at android.media.MediaCodec.queueInputBuffer(MediaCodec.java:2671)
        at com.genymobile.scrcpy.audio.AudioEncoder.inputThread(AudioEncoder.java:113)
        at com.genymobile.scrcpy.audio.AudioEncoder.lambda$encode$1$com-genymobile-scrcpy-audio-AudioEncoder(AudioEncoder.java:240)
        at com.genymobile.scrcpy.audio.AudioEncoder$$ExternalSyntheticLambda2.run(D8$$SyntheticClass:0)
[server] ERROR: MediaCodec error
android.media.MediaCodec$CodecException: Error 0xffffffe0
[server] ERROR: Capture/encoding error: java.lang.IllegalStateException: null
Killed 
ERROR: Could not retrieve device information
```

There's no working combination of `--video-codec`/`--video-encoder` that avoids MediaCodec entirely, so on this kind of device scrcpy currently just can't be used, period.

## What

Adds a new `mjpeg` video codec that never touches `MediaCodec`. It reuses the exact same capture pipeline as `SurfaceEncoder` (`ScreenCapture`/`VirtualDisplay`, crop/rotation, `PositionMapper` for touch mapping), but instead of feeding a MediaCodec input `Surface`, it targets an `ImageReader` (RGBA_8888), grabs raw frames, and JPEG-compresses them on the CPU (`JpegEncoder`). Frames are streamed with a new `mjpg` fourcc that the client demuxes as `AV_CODEC_ID_MJPEG` — FFmpeg already decodes Motion JPEG natively, so no client-side rendering changes were needed.

This is deliberately a last-resort fallback, not a replacement for a real video codec:

- No RFB/VNC, no extra native `.so`, no new protocol — just one more codec option that reuses scrcpy's existing streaming/control infrastructure end to end (touch/key injection keeps working unchanged).
- JPEG per frame instead of raw ARGB8888 specifically so it stays usable over a slow/remote link, not just on localhost — quality is tunable (`--video-codec-options=quality=NN`, default 80), and frame rate is throttled via the existing `--max-fps` (default 4 fps if unset), so bandwidth/CPU stay bounded on weak devices too.
- Usage: `--video-codec=mjpeg` (optionally with `--max-fps=` and `--video-codec-options=quality=`).

## Limitations

- Recording (`--record`) with `mjpeg` is rejected for now, not tested against the muxers.
- JPEG-per-frame is obviously far less efficient than H264/H265/AV1 — this is only meant for devices where those aren't an option at all.

Tested manually against a real device with `--video-codec=mjpeg --max-fps=2 --video-codec-options=quality=70`: mirrors correctly and touch input keeps working.

### DIFF
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -982,8 +982,11 @@ static const struct sc_option options[] = {
         .longopt_id = OPT_VIDEO_CODEC,
         .longopt = "video-codec",
         .argdesc = "name",
-        .text = "Select a video codec (h264, h265, av1, vp8 or vp9).\n"
-                "Default is h264.",
+        .text = "Select a video codec (h264, h265, av1, vp8, vp9 or mjpeg).\n"
+                "Default is h264.\n"
+                "The mjpeg codec does not use MediaCodec at all: it captures "
+                "raw frames and compresses them to JPEG on the CPU. Use it as "
+                "a fallback on devices where MediaCodec is broken.",
     },
     {
         .longopt_id = OPT_VIDEO_CODEC_OPTIONS,
@@ -2038,7 +2041,11 @@ parse_video_codec(const char *optarg, enum sc_codec *codec) {
         *codec = SC_CODEC_VP9;
         return true;
     }
-    LOGE("Unsupported video codec: %s (expected h264, h265, av1, vp8 or vp9)", optarg);
+    if (!strcmp(optarg, "mjpeg")) {
+        *codec = SC_CODEC_MJPEG;
+        return true;
+    }
+    LOGE("Unsupported video codec: %s (expected h264, h265, av1, vp8, vp9 or mjpeg)", optarg);
     return false;
 }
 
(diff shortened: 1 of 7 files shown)


twaik:
Wild guess, but it may be a fine workaround for a lot of people who faced the same issue I did.
https://github.com/Genymobile/scrcpy/issues?q=is%3Aissue%20state%3Aopen%20android.media.MediaCodec%24CodecException&page=3

robertkirkman:
> Recording (--record) with mjpeg is rejected for now, not tested against the muxers.

that makes it sound as though it might work with recording if enabled and then tested,

if you enable recording could I test that, or is it hard to enable recording?

twaik:
Even if it will work, I am not sure storing barely compressed stream is a good idea.

robertkirkman:
Yes, it is better to record the window with OBS.

PR #6967, opened 2026-08-02
Fix display capture regression on original Oculus Quest

### LINKED ISSUE #5913: Scrpcy is not working on Meta Quest (Meta Horizon OS 74)
I tested scrcpy on Quest 2 and the Quest 3 with the latest OS version (74) and I have now a black screen. Do you have any news about Meta blocking tools such as scrcpy ? If so, that really sucks.

doyle31:
The workaround that @rom1v implemented in d3ffab9d7489af6cc0533adeb43e3cfa5b128c81 to fix #5913 broke the original Quest as it has not received updates from Meta in multiple years.

This PR restricts the workaround so it only applies to newer device generations (e.g. the Quest Pro, Quest 2, Quest 3/S) by excluding legacy hardware whose model identifier lacks a numerical suffix.

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
@@ -134,7 +134,7 @@ public class ScreenCapture extends SurfaceCapture {
                     .createVirtualDisplay("scrcpy", inputSize.getWidth(), inputSize.getHeight(), displayId, surface);
             Ln.d("Display: using DisplayManager API");
         } catch (Exception displayManagerException) {
-            if (Build.BRAND.equalsIgnoreCase("oculus") && Build.MODEL.startsWith("quest") && !Build.MODEL.equalsIgnoreCase("quest")) {
+            if (Build.BRAND.equalsIgnoreCase("oculus") && Build.MODEL.toLowerCase(Locale.ROOT).startsWith("quest ")) {
                 // Workaround for buggy createVirtualDisplay on Quest
                 // Extra workaround to fix the original Quest
                 try {
--- a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
@@ -134,8 +134,9 @@ public class ScreenCapture extends SurfaceCapture {
                     .createVirtualDisplay("scrcpy", inputSize.getWidth(), inputSize.getHeight(), displayId, surface);
             Ln.d("Display: using DisplayManager API");
         } catch (Exception displayManagerException) {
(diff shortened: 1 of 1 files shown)


doyle31:
The latest tagged release completely breaks the original Quest, so it would also probably be worth considering a patch release once this is merged.

eomanis:
I am affected by this currently, would be cool if this could be released.

doyle31:
I should add that this has been tested on both an original Quest and a Quest 3.

server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
```diff
@@ -134,8 +134,9 @@ public void start(Surface surface) throws IOException {
                     .createVirtualDisplay("scrcpy", inputSize.getWidth(), inputSize.getHeight(), displayId, surface);
             Ln.d("Display: using DisplayManager API");
         } catch (Exception displayManagerException) {
-            if (Build.BRAND.equalsIgnoreCase("oculus") && Build.MODEL.toLowerCase(Locale.ROOT).startsWith("quest")) {
+            if (Build.BRAND.equalsIgnoreCase("oculus") && Build.MODEL.startsWith("quest") && !Build.MODEL.equalsIgnoreCase("quest")) {
```
rom1v: If `Build.MODEL` is `"Quest 3"`, then the condition will not match, because `startsWith("quest")` is `false`.

rom1v:
It does not look correct to me for `"Quest 2"` and `"Quest 3"`: https://github.com/Genymobile/scrcpy/pull/6967#discussion_r3750227836

To only ignore the quest 1, I suggest:

```diff
diff --git a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
index 0768181a4..2a525c46f 100644
--- a/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
+++ b/server/src/main/java/com/genymobile/scrcpy/video/ScreenCapture.java
@@ -134,7 +134,7 @@ public class ScreenCapture extends SurfaceCapture {
                     .createVirtualDisplay("scrcpy", inputSize.getWidth(), inputSize.getHeight(), displayId, surface);
             Ln.d("Display: using DisplayManager API");
         } catch (Exception displayManagerException) {
-            if (Build.BRAND.equalsIgnoreCase("oculus") && Build.MODEL.toLowerCase(Locale.ROOT).startsWith("quest")) {
+            if (Build.BRAND.equalsIgnoreCase("oculus") && Build.MODEL.toLowerCase(Locale.ROOT).startsWith("quest ")) {
                 // Workaround for buggy createVirtualDisplay on Quest
                 try {
                     virtualDisplay = (VirtualDisplay) VirtualDisplay.class.getDeclaredConstructors()[0].newInstance(null, null, null, surface);
```

doyle31:
Oh yea, thanks. Sorry I overlooked that.

PR #6973, opened 2026-08-06
fix: the cleanup in CleanUp.java

anupamme:
## Summary
Fix critical severity security issue in `server/src/main/java/com/genymobile/scrcpy/CleanUp.java`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `server/src/main/java/com/genymobile/scrcpy/CleanUp.java:125` |
| **Assessment** | Likely exploitable |

**Description**: The CleanUp.java class sets the CLASSPATH environment variable to Server.SERVER_PATH without validation. If Server.SERVER_PATH can be influenced by an attacker (via environment variables or filesystem manipulation), it could cause the cleanup process to load and execute arbitrary Java code.

## Evidence

**Exploitation scenario**: An attacker with filesystem write access or control over environment variables at server startup can manipulate SCRCPY_SERVER_PATH to point to a malicious JAR file, leading to arbitrary code.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

## Changes
- `server/src/main/java/com/genymobile/scrcpy/Server.java`
- `server/src/main/java/com/genymobile/scrcpy/CleanUp.java`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
+++ b/server/src/main/java/com/genymobile/scrcpy/CleanUp.java
@@ -124,19 +124,16 @@ public final class CleanUp {
 
     private void run(int displayId, int restoreStayOn, boolean disableShowTouches, boolean powerOffScreen, int restoreScreenOffTimeout,
             int restoreDisplayImePolicy) throws IOException {
-        String[] cmd = {
+        ProcessBuilder builder = new ProcessBuilder(
                 "app_process",
                 "/",
-                CleanUp.class.getName(),
-                String.valueOf(displayId),
-                String.valueOf(restoreStayOn),
-                String.valueOf(disableShowTouches),
-                String.valueOf(powerOffScreen),
-                String.valueOf(restoreScreenOffTimeout),
-                String.valueOf(restoreDisplayImePolicy),
-        };
-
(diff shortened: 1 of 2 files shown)


rom1v:
> Threat Model Context
>
> This is a Java service - vulnerabilities in servlets/controllers are remotely exploitable.

This does not apply.

anupamme:
Thanks for the clarification. You’re right; my threat-model assumption was incorrect. I treated the server as a remotely reachable Java service, whereas scrcpy’s server is launched locally on the Android device through ADB and the classpath/server JAR is controlled as part of that startup flow.

I agree that the reported classpath injection does not represent a remotely exploitable vulnerability in scrcpy’s threat model, and that canonicalising SERVER_PATH does not provide a meaningful security boundary here. I’ll consider this finding a false positive.

Issue #4380, opened 2023-10-26
Alternative audio capture API


yume-chan:
I tried to use Audio Policy API to mirror audio instead. Audio Policy API uses the same underlying API as AudioPlaybackCapture API, but requires `MODIFY_AUDIO_ROUTING` permission instead of showing a popup every time. (Use it to capture audio is the counterpart of https://github.com/Genymobile/scrcpy/issues/3880#issuecomment-1595722119)

Comparing to the Remote Submix API we are currently using, Audio Policy API has both pros and cons.

Pros:

* Can choose to continue playing audio on device
* Not affected by connected headsets
* Not affected by system volume
* If needed, can capture audio from each app separately

Cons:

* Requires Android 13, in which the Shell app has `MODIFY_AUDIO_ROUTING` permission added
* Apps can opt out from being captured. But IMO if Netflix shows a black screen in Scrcpy, not playing its audio also makes sense

Both methods can't capture notification, ringtone and alarm sounds.

---

The code is pretty simple, but again I'm using https://github.com/Reginer/aosp-android-jar/tree/main to unhide framework APIs:

```kotlin
package moe.chensi.tango

import android.annotation.SuppressLint
import android.app.ActivityThread
import android.content.Context
import android.media.AudioAttributes
import android.media.AudioFormat
import android.media.AudioManager
import android.media.audiopolicy.AudioMix
import android.media.audiopolicy.AudioMixingRule
import android.media.audiopolicy.AudioPolicy
import android.os.Build
import android.system.ErrnoException
import android.system.Os
import android.system.OsConstants
import org.joor.Reflect
import org.joor.ReflectException
import java.io.FileDescriptor
import java.io.IOException
import java.nio.ByteBuffer

class Test {
    @SuppressLint("StaticFieldLeak")
    private var _systemContext: Context? = null
    private val systemContext: Context
        @SuppressLint("DiscouragedPrivateApi", "PrivateApi") get() {
            if (_systemContext == null) {
                try {
                    // Hide warnings on XiaoMi devices
                    Reflect.onClass("android.content.res.ThemeManagerStub").set("sResource", null)
                } catch (_: ReflectException) {
                }

                _systemContext = ActivityThread.systemMain().systemContext
            }
            return _systemContext as Context
        }

    private fun writeFully(fd: FileDescriptor, from: ByteBuffer) {
        // ByteBuffer position is not updated as expected by Os.write() on old Android versions, so
        // count the remaining bytes manually.
[...]

PR #6975, opened 2026-08-07
Capture more audio usages with --audio-dup

Jayatubi:
## Summary

`--audio-dup` (audio playback capture) only captures players declaring
`USAGE_MEDIA`. This silently drops game audio, which declares `USAGE_GAME`.
This PR makes the set of captured usages configurable via an optional argument,
keeping the current behavior when no value is passed.

## Problem

While mirroring a game, sounds going through the audio engine were missing from
the forwarded stream, while other sounds from the very same app were forwarded
correctly. Using the default `--audio-source=output` captured everything.

`adb shell dumpsys audio` shows the app (same uid/pid) playing two streams that
differ only by their usage:

```
AudioPlaybackConfiguration piid:16695 deviceIds:[3] type:OpenSL ES AudioPlayer (Buffer Queue) u/pid:10499/3001 state:started attr:AudioAttributes: usage=USAGE_MEDIA ...
AudioPlaybackConfiguration piid:16711 deviceIds:[3] type:AAudio                              u/pid:10499/3001 state:started attr:AudioAttributes: usage=USAGE_GAME  ...
```

Only the first one was captured.

## Root cause

`AudioPlaybackCapture` adds a single mixing rule:

```java
AudioAttributes attributes = new AudioAttributes.Builder().setUsage(AudioAttributes.USAGE_MEDIA).build();
addMixRuleMethod.invoke(audioMixingRuleBuilder, ruleMatchAttributeUsageConstant, attributes);
```

`RULE_MATCH_ATTRIBUTE_USAGE` compares the usage for strict equality, and there
is no notion of category, so `USAGE_GAME` does not match `USAGE_MEDIA` (they are
only grouped together for *volume* control, both mapping to `STREAM_MUSIC`,
which makes this easy to miss).

Games are expected to declare `USAGE_GAME`: the Android
[low latency audio guidelines][1] require `AAUDIO_USAGE_GAME` for latency
optimizations, and audio engines such as Wwise, FMOD, Unity and Unreal follow
that. So they are doing the right thing and get filtered out.

For reference, Android's own public playback capture API documents that only
`USAGE_UNKNOWN`, `USAGE_GAME` and `USAGE_MEDIA` can ever be captured by
non-privileged apps, i.e. `USAGE_GAME` is one of the three main cases. scrcpy
registers its policy as shell (with `MODIFY_AUDIO_ROUTING`), so it is not
restricted to those three.

[1]: https://developer.android.com/games/sdk/oboe/low-latency-audio

## Changes

`--audio-dup` now accepts an optional comma-separated list of extra usages:

```bash
scrcpy --audio-dup                           # media only (unchanged)
scrcpy --audio-dup=game                      # media + game
scrcpy --audio-dup=game,voice-communication
scrcpy --audio-dup=all
```

Possible values are `all`, `unknown`, `game`, `alarm`, `notification`,
`assistant`, `sonification`, `accessibility`, `navigation` and
`voice-communication`. `USAGE_MEDIA` is always captured, so passing no value is
strictly identical to the current behavior. `USAGE_VIRTUAL_SOURCE` is not
exposed, since it would capture the loopback itself.

Several "match" rules of the same kind are OR-ed together by the audio policy,
so this is just a matter of calling `addMixRule()` once per usage. Names are
validated client-side, so a typo fails immediately instead of on the device.

A separate commit fixes `voiceCommunicationCaptureAllowed(true)`, which was
called *after* `AudioMixingRule.Builder.build()` and therefore only mutated the
builder, leaving the built rule untouched. It is required for
`USAGE_VOICE_COMMUNICATION` to be captured at all. Its absence on ROMs that do
not expose the hidden method is now tolerated with a warning instead of failing
the whole audio capture.

Related to #4380.

## Testing

- Verified on a OnePlus PJZ110 (Android 16) that `--audio-dup=game` captures the
  `USAGE_GAME` AAudio stream shown above, which was silent before.
- Checked that plain `--audio-dup` still produces the same result as before.
- Unit-tested the option parser against valid lists, unknown names, wrong case,
  empty values and malformed separators (leading, trailing and doubled commas).

## Notes

Some players may still not be captured after requesting their usage if they use
the AAudio MMAP exclusive low latency path, which bypasses the mixer and cannot
be intercepted by a dynamic audio policy. This is documented in `doc/audio.md`
along with how to identify a player's usage via `dumpsys audio`.

### DIFF
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -8,6 +8,7 @@ _scrcpy() {
         --audio-codec=
         --audio-codec-options=
         --audio-dup
+        --audio-dup=
         --audio-encoder=
         --audio-source=
         --audio-output-buffer=
@@ -169,6 +170,11 @@ _scrcpy() {
             ;;
         --pause-on-exit)
             COMPREPLY=($(compgen -W 'true false if-error' -- "$cur"))
+            return
+            ;;
+        --audio-dup)
+            # Only auto-complete a single usage
+            COMPREPLY=($(compgen -W 'all media unknown game alarm notification assistant sonification accessibility navigation voice-communication' -- "$cur"))
             return
             ;;
         -r|--record)
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -14,7 +14,7 @@ arguments=(
     '--audio-buffer=[Configure the audio buffering delay \(in milliseconds\)]'
     '--audio-codec=[Select the audio codec]:codec:(opus aac flac raw)'
     '--audio-codec-options=[Set a list of comma-separated key\:type=value options for the device audio encoder]'
-    '--audio-dup=[Duplicate audio]'
+    '--audio-dup=-[Duplicate audio, with an optional comma-separated list of extra usages to capture]:usages:(all media unknown game alarm notification assistant sonification accessibility navigation voice-communication)'
     '--audio-encoder=[Use a specific MediaCodec audio encoder]'
     '--audio-source=[Select the audio source]:source:(output playback mic mic-unprocessed mic-camcorder mic-voice-recognition mic-voice-communication voice-call voice-call-uplink voice-call-downlink voice-performance)'
     '--audio-output-buffer=[Configure the size of the SDL audio output buffer (in milliseconds)]'
--- a/app/src/cli.c
+++ b/app/src/cli.c
@@ -198,9 +198,33 @@ static const struct sc_option options[] = {
     {
         .longopt_id = OPT_AUDIO_DUP,
         .longopt = "audio-dup",
+        .argdesc = "usages",
+        .optional_arg = true,
         .text = "Duplicate audio (capture and keep playing on the device).\n"
-                "This feature is only available with --audio-source=playback."
-
+                "This feature is only available with --audio-source=playback.\n"
+                "Audio playback capture matches each player by its audio usage, "
+                "and only \"media\" is captured by default. Games using an audio "
+                "engine such as Wwise declare \"game\" instead, so their sound is "
+                "not captured unless it is requested explicitly.\n"
+                "An optional comma-separated list of extra usages to capture may "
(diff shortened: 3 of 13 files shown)


rom1v:
Can we make all these capture modes non-optional (make `--audio-dup` behave like your `--audio-dup=all`, without options)?

Jayatubi:
> Can we make all these capture modes non-optional (make `--audio-dup` behave like your `--audio-dup=all`, without options)?

This PR was intended to add new features without modifying the original behavior. Of course, that could make `all` the default, if that were allowed.

rom1v:
Does `all` cause issues on some device where it works with the current default behavior?

If not, just use the `all` behavior without options. :wink:

Jayatubi:
> Does `all` cause issues on some device where it works with the current default behavior?
> 
> If not, just use the `all` behavior without options. 😉

Sorry but I only test on my own device and app. That's another reason why I didn't make it default.

rom1v:
Thank you :+1:

To keep things simple, I think I'll enable all usages by default, without an option. We can add an option if it turns out we need to enable/disable usages.

Please review/test branch `pr6975.1`:
 - 321ebd98b4c2f0f98df43f017274ea9f47d07cdd
 - a4d19bcbd9de39fb76dd88b6267ccf5f2bd940dd

numToStr:
@rom1v I tested the branch `pr6975.1` and it seems to fix an issue that i was facing with in-game mic usage.

On v4.1, using `--audio-source=playback`-- In PUBG mobile, whenever i opened mic to talk, my teammates voices always came out from the mobile device, but in-game music went to the host. But with `pr6975.1` branch, all audio goes to the host.

This issue was particularly frustrating, so I'm now happy that it will be addressed in upcoming release. Thanks @rom1v @Jayatubi

rom1v:
Merged the "simple" version where all the capture modes are enabled, without option:
 - 8b05a6163c789f27da0ced27bec466f12f813b2e
 - e71cdb8782f524d8ec4deaa1b1e4930c728acab0

Thank you :rocket:

PR #6987, opened 2026-08-11
feat(display): add in-window rotation control and shortcut with auto-hide

aashishrajput9838:
## Summary

Adds an interactive client-side display rotation control and keyboard shortcut (`MOD+l`) to cycle through display orientations.

## Features

- **In-window rotation control** — Adds a compact rotation button in the top-right corner of the scrcpy window.
- **Keyboard shortcut** — `MOD+l` cycles the display orientation by 90°.
- **Auto-hide** — The rotation control automatically hides after 2 seconds of mouse inactivity and reappears when the cursor moves.
- **Input isolation** — Button interactions are consumed locally and are not forwarded as Android touch events.
- **Drag cancellation** — Pressing the button and releasing outside cancels the action cleanly.
- **Client-side only** — No Android/server-side changes are required.
- **Procedural rendering** — The control is rendered using native SDL3 primitives without external runtime assets.

## Implementation

The feature reuses scrcpy's existing orientation handling and rendering pipeline rather than introducing a separate rotation mechanism.

Orientation cycle:

`0° → 90° → 180° → 270° → 0°`

## Verification

- Built successfully with Meson/Ninja.
- Tested with Direct3D11 and OpenGL rendering backends.
- Tested on a physical Android device.
- Verified portrait/landscape transitions.
- Verified button auto-hide/reveal behavior.
- Verified Android input pass-through when the control is hidden.
- Verified button press/drag cancellation behavior.

## Scope

No server-side or Android-side files were modified.

## Demo

rotate_button_pr_demo


rom1v:
There are already key shortcuts to rotate the display: https://github.com/Genymobile/scrcpy/blob/master/doc/shortcuts.md

Issue #6983, opened 2026-08-10
Right Shift key does not work with UHID keyboard (`-K`) since scrcpy 4.0


ucbear:
## Environment

* **OS:** Windows 11
* **Scrcpy version:** 4.0 and later
* **Installation method:** Windows release
* **Device model:** Galaxy Tab S11 Ultra
* **Android version:** 16
* **Keyboard:** K995p
* **Keyboard layout:**Korean / US

## Describe the bug

When using scrcpy with UHID keyboard mode (`-K` / `--keyboard=uhid`), the **Right Shift key does not work as a Shift modifier**, while the **Left Shift key works correctly**.

This appears to be a regression introduced in **scrcpy 4.0**. The same PC, keyboard, and Android device worked correctly with scrcpy versions before 4.0.

Since scrcpy 4.0 migrated from SDL2 to SDL3, this might be related to SDL3 keyboard/scancode handling, but I have not confirmed the root cause.

### Steps to reproduce

1. Start scrcpy with UHID keyboard mode and verbose logging:

```bash
scrcpy -K -Vverbose
```

2. Hold **Left Shift** and press some keys.
3. Release Left Shift.
4. Hold **Right Shift** and press the same keys.

### Expected behavior

Both Shift keys should work as Shift modifiers.

For UHID keyboard input, I would expect:

* Left Shift: HID scancode `0xe1`, modifier `0x02`
* Right Shift: HID scancode `0xe5`, modifier `0x20`

### Actual behavior

Left Shift works correctly. For example, while holding Left Shift:

```text
VERBOSE: keys[51] = true
VERBOSE: hid keyboard: key down scancode=51 (81) mod=02
VERBOSE: input: UHID input [1] 02 00 51 00 00 00 00 00
VERBOSE: keys[51] = false
VERBOSE: hid keyboard: key up   scancode=51 (81) mod=02
VERBOSE: input: UHID input [1] 02 00 00 00 00 00 00 00
VERBOSE: keys[52] = true
VERBOSE: hid keyboard: key down scancode=52 (82) mod=02
VERBOSE: input: UHID input [1] 02 00 52 00 00 00 00 00
VERBOSE: keys[52] = false
VERBOSE: hid keyboard: key up   scancode=52 (82) mod=02
VERBOSE: input: UHID input [1] 02 00 00 00 00 00 00 00
VERBOSE: hid keyboard: key up   scancode=e1 (225) mod=00
[...]

PR #6990, opened 2026-08-16
Fix dropped keys in UHID mode for some keyboard layouts

Vishnu070803:
Subject: Fix dropped keys in UHID mode for some keyboard layouts

Description: This PR fixes Issue #6983 where the Right Shift key is ignored on Korean layouts when using the UHID keyboard (-K).

Background & Testing: We successfully reproduced this issue by testing the Right Shift key under the English (India) layout, which successfully emitted the e5 scancode. However, switching to the Korean IME in Windows caused the Right Shift key to completely fail in scrcpy.

Root Cause Analysis: This is a regression introduced during the SDL3 migration in c8479fe8 ("Discard unknown SDL events"). Because the Korean IME hooks the Right Shift key, SDL3 does not assign it a standard virtual keycode (SDLK_UNKNOWN), but it still faithfully reports the physical hardware scancode (e5).

Because of the strict filter introduced in c8479fe8, scrcpy was dropping the key event entirely whenever the virtual keycode was unknown. This broke the contract of -K (UHID mode), which is designed to bypass the OS layout and rely purely on the physical scancode.

The Fix: This PR relaxes the filter in input_manager.c. It now only drops the key event if both the virtual keycode AND the physical scancode are unknown. This safely allows physical keys intercepted by OS layouts to still be passed through to the Android device in UHID mode, while still protecting the injection layer from garbage events

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,4 @@ build/
 /x/
 local.properties
 /scrcpy-server
-Start-Wireless-Scrcpy.bat
-issue6983.json
+
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ build/
 /x/
 local.properties
 /scrcpy-server
+Start-Wireless-Scrcpy.bat
+issue6983.json
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ build/
 /x/
 local.properties
 /scrcpy-server
-
--- a/app/src/input_manager.c
+++ b/app/src/input_manager.c
@@ -707,12 +707,9 @@ sc_input_manager_process_key(struct sc_input_manager *im,
     }
 
     enum sc_keycode keycode = sc_keycode_from_sdl(sdl_keycode);
-    if (keycode == SC_KEYCODE_UNKNOWN) {
-        return;
-    }
-
     enum sc_scancode scancode = sc_scancode_from_sdl(event->scancode);
-    if (scancode == SC_SCANCODE_UNKNOWN) {
+
+    if (keycode == SC_KEYCODE_UNKNOWN && scancode == SC_SCANCODE_UNKNOWN) {
         return;
     }
 


app/src/input_manager.c
```diff
@@ -707,12 +707,9 @@ sc_input_manager_process_key(struct sc_input_manager *im,
...
 
     enum sc_keycode keycode = sc_keycode_from_sdl(sdl_keycode);
-    if (keycode == SC_KEYCODE_UNKNOWN) {
-        return;
-    }
-
     enum sc_scancode scancode = sc_scancode_from_sdl(event->scancode);
-    if (scancode == SC_SCANCODE_UNKNOWN) {
+
+    if (keycode == SC_KEYCODE_UNKNOWN && scancode == SC_SCANCODE_UNKNOWN) {
```
rom1v, line 712: There's no need to check `SC_KEYCODE_UNKNOWN` then:
 - I think `scancode == SC_SCANCODE_UNKNOWN` implies `keycode == SC_KEYCODE_UNKNOWN`
 - even if this was not the case, we already process unknown keycodes when the scancode is known, so it is harmless to process it when the scancode is unknown

.gitignore
```diff
@@ -8,3 +8,5 @@ build/
 /x/
 local.properties
 /scrcpy-server
+Start-Wireless-Scrcpy.bat
+issue6983.json
```
rom1v: Note: you should put your local ignores in `.git/info/exclude`

rom1v (commented):
Thank you for the analysis.

I made some changes, please review 01f368358e9017402f45d0387e19a1f26a38d10b.

.gitignore
Vishnu070803: > Note: you should put your local ignores in `.git/info/exclude`

Noted, this was a mistake

rom1v:
(no need to make further changes on your PR, please just test and review 01f368358e9017402f45d0387e19a1f26a38d10b :wink:)

rom1v:
Thank you, merged: b80edbc71c9e5bfb9aefd140b90af45337b4dbdb :rocket:

PR #6996, opened 2026-08-24
Upgrade actions/setup-java to v6

brunoborges:
## Summary

Upgrade all three release workflow references from actions/setup-java v5 to v6.0.0.

## Validation

Reviewed the workflow-only diff and verified no setup-java reference older than v6 remains in active workflow YAML.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #1894, opened 2020-11-09
[RFC] Initial draft for Linux hardware-accelerated decoding with VA-API

arter97:
Addresses #1672

This pull-request adds support for hardware-accelerated decoding with VA-API on Linux platform.

I'm about 70% certain that this code can be better optimized, but even with this initial draft I'm seeing much lower CPU usage overall.

Few limitations here:
* VA-API/NV12 is hardcoded; we need a switch and auto-detection method.
* Rotation is broken.
* !SCRCPY_LAVF_HAS_NEW_ENCODING_DECODING_API is not taken care of.
* From a quick glimpse, it looks like it's easy to extend this to other H/W accelerated methods such as VDPAU or CUDA.
* CPU usage is still a bit high despite the fact that it's using VA-API.
* Reducing memory copy? I'm not sure if this is zero-copy or not, or whether zero-copy is even possible under VA-API and SDL.

Any suggestions would be nice :)

PR #7001, opened 2026-08-26
Add zero-copy hardware decoding for Linux, Windows and macOS

arter97:
https://github.com/libsdl-org/SDL/pull/8366/commits was used as a reference and tested on linux-x86_64, win64, winarm64, macos-x86_64, macos-aarch64.

Target device: Galaxy S25
linux-x86_64-1: Intel Core Ultra X7 358H on Ubuntu 26.04 (Wayland, GNOME)
linux-x86_64-2: Intel Atom E3950 on Ubuntu 24.04 (KDE, X11)
win64: AMD Ryzen 5 5600G
winarm64: Xiaomi Mi Pad 5
macos-x86_64: AMD Ryzen 5 5600G (OpenCore)
macos-aarch64: Apple M3 Pro (Mac15,7)

| Platform        | Before | After | Improvement |
| --------------- | -----: | ----: | ----------: |
| linux-x86_64-1 |  48.5% |  3.0% |       93.8% |
| linux-x86_64-1 | 120.0% | 16.7% |       86.1% |
| win64           | 103.2% |  3.6% |       96.5% |
| winarm64        | 112.0% |  8.0% |       92.9% |
| macos-x86_64    | 120.0% | 10.8% |       91.0% |
| macos-aarch64   |  50.2% | 19.8% |       60.6% |

On Windows, CPU usage reported by Task Manager (arm64) or TMX was multiplied by the number of cores for consistencies with other platforms.

This is additionally tested on following devices with both H264 and H265 encoders (when available by the device):
OnePlus 3T (ONEPLUS A3003) - Snapdragon 821
OnePlus 5 (ONEPLUS A5000) - Snapdragon 835
OnePlus 6 (ONEPLUS A6003) - Snapdragon 845
Samsung Galaxy A03s (SM-A037U) - MediaTek Helio P35
UMIDIGI Power 7 (UMIDIGI MP06) - Unisoc Tiger T610
Samsung Galaxy A Quantum (SM-A716S) - Exynos 980
OnePlus 8T (KB2003) - Snapdragon 865
Samsung Galaxy S20+ 5G (SM-G986N) - Snapdragon 865
Samsung Galaxy M32 5G (SM-M326B) - MediaTek Dimensity 720
OnePlus 9 Pro (OnePlus 9 Pro) - Snapdragon 888
Samsung Galaxy S21+ 5G (SM-G996N) - Exynos 2100
Samsung Galaxy Quantum3 (SM-M536S) - MediaTek Dimensity 900
DOOGEE N50 Pro (DOOGEE N50Pro) - Unisoc Tiger T606
ASUS Zenfone 9 (ASUS_AI2202) - Snapdragon 8+ Gen 1
Nothing Phone (2a) (Nothing A142) - MediaTek Dimensity 7200 Pro
Samsung Galaxy A55 5G (SM-A556E) - Exynos 1480
Samsung Galaxy S25 Edge (SM-S937N) - Snapdragon 8 Elite for Galaxy
Samsung Galaxy S25+ (SM-S936N) - Snapdragon 8 Elite for Galaxy

Assisted-by: Codex:gpt-5.6-sol

### DIFF
--- a/app/meson.build
+++ b/app/meson.build
@@ -106,6 +106,75 @@ if v4l2_support
     src += [ 'src/v4l2_sink.c' ]
 endif
 
+cc = meson.get_compiler('c')
+
+static = get_option('static')
+
+avformat_dep = dependency('libavformat', version: '>= 57.33', static: static)
+avcodec_dep = dependency('libavcodec', version: '>= 57.37', static: static)
+avutil_dep = dependency('libavutil', static: static)
+swresample_dep = dependency('libswresample', static: static)
+sdl_dep = dependency('sdl3', version: '>= 3.2.0', static: static)
+
+dependencies = [
+    avformat_dep,
+    avcodec_dep,
+    avutil_dep,
+    swresample_dep,
+    sdl_dep,
+]
+
+hwaccel_opt = get_option('hardware_decoding')
+hwaccel_support = false
+if host_machine.system() == 'linux' and not hwaccel_opt.disabled()
+    has_vaapi_api = cc.has_header_symbol(
+        'libavutil/hwcontext.h', 'AV_HWDEVICE_TYPE_VAAPI',
+        dependencies: avutil_dep)
+    has_vaapi_api = has_vaapi_api and cc.has_header_symbol(
+        'libavutil/pixfmt.h', 'AV_PIX_FMT_DRM_PRIME',
+        dependencies: avutil_dep)
+    libdrm_dep = dependency('libdrm', required: hwaccel_opt)
+    hwaccel_support = has_vaapi_api and libdrm_dep.found()
+    if hwaccel_support
+        conf.set('SC_HAVE_VAAPI', true)
+        src += [ 'src/vaapi.c' ]
+    elif hwaccel_opt.enabled()
+        error('The installed FFmpeg does not provide the required VA-API API')
+    endif
+elif host_machine.system() == 'windows' and not hwaccel_opt.disabled()
+    hwaccel_support = cc.has_header_symbol(
+        'libavutil/hwcontext.h', 'AV_HWDEVICE_TYPE_D3D11VA',
+        dependencies: avutil_dep)
+    hwaccel_support = hwaccel_support and cc.has_header_symbol(
+        'libavutil/pixfmt.h', 'AV_PIX_FMT_D3D11',
+        dependencies: avutil_dep)
+    hwaccel_support = hwaccel_support and cc.has_header(
+        'libavutil/hwcontext_d3d11va.h', dependencies: avutil_dep)
+    if hwaccel_support
+        conf.set('SC_HAVE_D3D11', true)
+    elif hwaccel_opt.enabled()
+        error('The installed FFmpeg does not provide the required D3D11 API')
+    endif
+elif host_machine.system() == 'darwin' and not hwaccel_opt.disabled()
+    hwaccel_support = cc.has_header_symbol(
+        'libavutil/hwcontext.h', 'AV_HWDEVICE_TYPE_VIDEOTOOLBOX',
+        dependencies: avutil_dep)
+    hwaccel_support = hwaccel_support and cc.has_header_symbol(
+        'libavutil/pixfmt.h', 'AV_PIX_FMT_VIDEOTOOLBOX',
+        dependencies: avutil_dep)
+    if hwaccel_support
+        conf.set('SC_HAVE_VIDEOTOOLBOX', true)
+    elif hwaccel_opt.enabled()
+        error('The installed FFmpeg does not provide the required ' +
+              'VideoToolbox API')
+    endif
+elif hwaccel_opt.enabled()
+    error('Hardware decoding is only supported on Linux, Windows, and macOS')
+endif
+if hwaccel_support
+    src += [ 'src/hwaccel.c' ]
+endif
+
 usb_support = get_option('usb')
 if usb_support
     src += [
@@ -118,17 +187,9 @@ if usb_support
     ]
 endif
 
-cc = meson.get_compiler('c')
-
-static = get_option('static')
-
-dependencies = [
-    dependency('libavformat', version: '>= 57.33', static: static),
-    dependency('libavcodec', version: '>= 57.37', static: static),
-    dependency('libavutil', static: static),
-    dependency('libswresample', static: static),
-    dependency('sdl3', version: '>= 3.2.0', static: static),
-]
+if hwaccel_support and host_machine.system() == 'linux'
+    dependencies += [libdrm_dep]
+endif
 
 if v4l2_support
     dependencies += dependency('libavdevice', static: static)
@@ -182,6 +243,9 @@ conf.set('SERVER_DEBUGGER', get_option('server_debugger'))
 
 # enable V4L2 support (linux only)
 conf.set('HAVE_V4L2', v4l2_support)
+
+# enable platform hardware video decoding
+conf.set('HAVE_HWACCEL', hwaccel_support)
 
 # enable HID over AOA support (linux only)
 conf.set('HAVE_USB', usb_support)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,7 @@ jobs:
              libasound2-dev libpulse-dev \
              libaudio-dev libfribidi-dev libjack-dev libsndio-dev libx11-dev libxext-dev \
              libxrandr-dev libxcursor-dev libxfixes-dev libxi-dev libxss-dev libxtst-dev \
-             libxkbcommon-dev libdrm-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
+             libxkbcommon-dev libdrm-dev libva-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
              libegl1-mesa-dev libdbus-1-dev libibus-1.0-dev libudev-dev libthai-dev \
              libpipewire-0.3-dev libwayland-dev libdecor-0-dev liburing-dev
 
@@ -122,7 +122,7 @@ jobs:
              libasound2-dev libpulse-dev \
              libaudio-dev libfribidi-dev libjack-dev libsndio-dev libx11-dev libxext-dev \
              libxrandr-dev libxcursor-dev libxfixes-dev libxi-dev libxss-dev libxtst-dev \
-             libxkbcommon-dev libdrm-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
+             libxkbcommon-dev libdrm-dev libva-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
              libegl1-mesa-dev libdbus-1-dev libibus-1.0-dev libudev-dev libthai-dev \
              libpipewire-0.3-dev libwayland-dev libdecor-0-dev liburing-dev
 
--- a/app/data/bash-completion/scrcpy
+++ b/app/data/bash-completion/scrcpy
@@ -60,6 +60,7 @@ _scrcpy() {
         --no-cleanup
         --no-clipboard-autosync
         --no-downsize-on-error
+        --no-hardware-decoding
         --no-key-repeat
         --no-mipmaps
         --no-mouse-hover
--- a/app/data/zsh-completion/_scrcpy
+++ b/app/data/zsh-completion/_scrcpy
@@ -66,6 +66,7 @@ arguments=(
     '--no-cleanup[Disable device cleanup actions on exit]'
     '--no-clipboard-autosync[Disable automatic clipboard synchronization]'
     '--no-downsize-on-error[Disable lowering definition on MediaCodec error]'
+    '--no-hardware-decoding[Disable hardware video decoding on the computer]'
     '--no-key-repeat[Do not forward repeated key events when a key is held down]'
     '--no-mipmaps[Disable the generation of mipmaps]'
     '--no-mouse-hover[Do not forward mouse hover events]'
--- a/app/deps/ffmpeg.sh
+++ b/app/deps/ffmpeg.sh
@@ -56,7 +56,6 @@ else
         --disable-network
         --disable-everything
         --disable-vulkan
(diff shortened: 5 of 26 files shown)


arter97:
6 years later, this has come to a full circle. Since #1894, LLMs have made it much easier to achieve this.

The code looks sane. It's now truly zero-copy.

numToStr:
Thanks for the working on this ❤️.

But i have a stupid question to ask, is it possible to assign GPU? i.e. between Intel UHD 630 and Intel ARC A380. I'm thinking of setup where UHD-630 decodes scrcpy stream and use A380 to livestream on YT.

arter97:
You would need another patch on top of this to tell FFmpeg's VA-API code to use a specific device other than /dev/dri/renderD128. It's not a trivial-enough-to-be-user-friendly-enough task.

Also to be honest, it really shouldn't matter. Modern GPUs can handle multiple streams of high-def video decodes. You may be paying more in copying overhead if you try to do this on a GPU that doesn't have the monitor you're viewing it on connected to.

app/meson.build
```diff
@@ -106,6 +106,23 @@ if v4l2_support
     src += [ 'src/v4l2_sink.c' ]
 endif
 
+vaapi_opt = get_option('vaapi')
+vaapi_support = false
+if host_machine.system() == 'linux' and not vaapi_opt.disabled()
+    egl_dep = dependency('egl', required: vaapi_opt)
```
rom1v: Use EGL via SDL to avoid a dependency: `SDL_EGL_GetProcAddress` instead of `eglGetProcAddress` for example.

rom1v:
Thank you :+1: And it works :tada:

---

Currently, it's just VAAPI, but in the end, there will be NVDEC/CUDA, VideoToolbox (#6954), D3D11VA… (https://github.com/libsdl-org/SDL/pull/8366/commits is a good example)

So I think we need some hwdec abstraction (without _ifdefs_ everywhere). Claude suggested:

> :warning: AI generated suggestion, without review :warning:
> **The abstraction to build**
>
> The two things a HW backend changes are (a) decoder setup and (b) turning a
> hardware AVFrame into a renderable texture. Model exactly those, register
> backends in a probe-order list, and make every stage runtime-probed with
> layered fallback:
>
> ```c
> struct sc_hwaccel {
>     const char *name;                 // "vaapi", "videotoolbox", "nvdec"
>     enum AVHWDeviceType hw_type;
>     enum AVPixelFormat  hw_pix_fmt;   // AV_PIX_FMT_VAAPI, …
>
>     // Runtime probe: create AVHWDeviceContext; return false if unavailable.
>     bool (*init)(struct sc_hwaccel *);
>     // Set codec_ctx->hw_device_ctx and the get_format callback.
>     bool (*configure)(struct sc_hwaccel *, AVCodecContext *);
>     // Zero-copy import of a HW frame → SDL-usable texture.
>     bool (*map_frame)(struct sc_hwaccel *, const AVFrame *, struct sc_texture *);
>     void (*destroy)(struct sc_hwaccel *);
> };
> ```
>
> - Probe at runtime, in order; first that init()s wins; else software. Compile
> each backend behind #ifdef HAVE_VAAPI etc., but the selection is runtime —
> that's what gives you "same binary, feature present or not."

(I have not reviewed the code yet, and in the end the design/code will be done manually, but AI is great for prototyping.)

Also, the static binary for scrcpy should work as expected (the `libva` and `libva-drm` client libs should probably built statically), and gracefully error if not available.

I'll dive into the subject later, feel free to keep the prototype as it is or to "improve" it.

arter97:
I didn't pay much attention on building scrcpy throughout the years, but didn't you specifically encourage Linux users to always build from source to avoid headaches with static libs?

I recall using libva* that doesn't match the system's can cause problems back when I was trying to setup Steam's In-house streaming like 9-10 years ago. Steam was doing its own chroot-like environment (kinda like how snap/flatpak these days work I guess), and hardware acceleration only worked when I coped libva* from the system to Steam's own lib directory.
This is a very old anecdotal memory that could be very wrong now, but still.

lilydjwg:
> the `libva` and `libva-drm` client libs should probably built statically

Will it work? The VA-API driver libraries come separately for different cards, one or two for each vendor. I don't know if you can mix different versions of mesa and VA-API drivers that comes with it. Or, you statically-link mesa? That sounds not great.

rom1v:
> I didn't pay much attention on building scrcpy throughout the years, but didn't you specifically encourage Linux users to always build from source to avoid headaches with static libs?

But there's an official static release now (since #5515), and it works great: https://github.com/Genymobile/scrcpy/blob/master/doc/linux.md#from-the-official-release

Both should work (static or building from sources).


> Will it work? The VA-API driver libraries come separately for different cards, one or two for each vendor.

No idea (I have never played with libva), here is what Claude said:

> :warning: AI-generated :warning:
>
> Static-embed the client libs (libva.a into scrcpy) → no DT_NEEDED, the
> binary always starts. libva's internal dlopen of the driver fails gracefully
> on hosts without a GPU/driver → software fallback. ✓

> Two refinements that shrink the dependency surface further:
>
> - EGL: don't link it at all. SDL already brings up the GL/EGL stack. Get every
> entry point you need — including the dma-buf import extensions
> (eglCreateImageKHR, glEGLImageTargetTexture2DOES) — via SDL_GL_GetProcAddress
> / eglGetProcAddress. Those extension functions are proc-address-resolved
> regardless, so this is the natural path and it sidesteps the GLVND-static
> problem entirely.
> - libdrm: header-only for this use. vaExportSurfaceHandle() gives you a
> VADRMPRIMESurfaceDescriptor (dma-buf fds + fourcc + modifiers). You consume it
> with the DRM_FORMAT_* macros from drm_fourcc.h (a header) and feed the fds to
> EGL. You don't call libdrm ioctls, so there's no libdrm.so runtime dependency
> — just the header at build time.
>
> **Net: the only thing you actually static-embed is libva + libva-drm (both MIT,
> static-buildable via meson -Ddefault_library=static). Everything else is
> proc-address or headers.**
>
> The FFmpeg gotcha you need to watch
>
> Here's the trap. FFmpeg's VA-API hwaccel (hwcontext_vaapi.c, the vaapi
> decoders) hard-links libva — there's no dlopen wrapper in FFmpeg for VA-API.
> So if you build static FFmpeg with --enable-vaapi and then link -lva (shared)
> into scrcpy, FFmpeg's object code re-introduces the DT_NEEDED libva.so.2 and
> your binary won't start on a host without libva — right back to problem (1).
>
> Static-embedding (strategy 2) is exactly what resolves this: link libva.a +
> libva-drm.a at the final scrcpy link. FFmpeg's undefined libva symbols resolve
> into the binary, no DT_NEEDED, always starts. Transitive needs are tiny
> (-ldl, maybe -lm).
>
> One deployment caveat: a bundled libva uses its own compiled-in driver search
> path. In practice VA_MAJOR_VERSION has been stable at 2 for the entire VA-API
> 1.x era, so a bundled libva 2.x loads host *_drv_video.so of the same major
> fine; just make sure its default LIBVA_DRIVERS_PATH matches typical distro
> layout (or document the env override).
>
> Good news for the other backends: only VA-API needs this care
>
> Design the abstraction knowing that each backend has a different linkage
> model, but most are free:
>
>```
> ┌───────────────┬─────────────────────────────────────┬──────────────────┐
> │    Backend    │        Who links the HW lib         │   Optionality    │
> │               │                                     │    mechanism     │
> ├───────────────┼─────────────────────────────────────┼──────────────────┤
> │               │                                     │ static-embed     │
> │ VAAPI (Linux) │ FFmpeg hard-links libva             │ libva (the work  │
> │               │                                     │ above)           │
> ├───────────────┼─────────────────────────────────────┼──────────────────┤
> │ NVDEC/CUDA    │ FFmpeg dlopens                      │ free graceful    │
> │ (Linux/Win)   │ libcuda.so.1/libnvcuvid.so.1 via    │ fallback         │
> │               │ header-only ffnvcodec               │                  │
> ├───────────────┼─────────────────────────────────────┼──────────────────┤
> │ VideoToolbox  │ system framework, always present    │ weak-link,       │
> │ (macOS)       │                                     │ nothing to do    │
> ├───────────────┼─────────────────────────────────────┼──────────────────┤
> │ D3D11VA       │ system DLL, always present          │ delay-load,      │
> │ (Windows)     │                                     │ nothing to do    │
> └───────────────┴─────────────────────────────────────┴──────────────────┘
>```
>
> So VA-API is the one awkward case; the rest either dlopen inside FFmpeg
> already or are guaranteed-present system components.

arter97:
So I see 2 issues with static libva. (I'm surprised your Claude response didn't catch this.)

1. VA-API's ABI is only forward-compatible.

We build on ubuntu-22.04 and it ships with only v1.14 ABI.
(Do `apt show libva2`, it'll show bunch of "`Provides: libva-driver-abi-*`" entries.)
26.04 ships with v1.23, and its iHD_drv_video.so is v1.22, radeonsi_drv_video.so is v1.23, meaning statically linking libva while building on 22.04 will break newer distros.
Even if we choose to always build latest libva, we then add a maintenance burden where we'd have to keep updating scrcpy matching libva's update and moreover:

2. libva is built with hardcoded drivers path.

For Ubuntu, VA_DRIVERS_PATH is set to /usr/lib/x86_64-linux-gnu/dri.
For Fedora, it's /usr/lib64/dri, for Arch, it's /usr/lib/dri.

I see 3 viable paths for us:

1. Just ship dynamically linked libva2.

It's ubiquitous on modern distros as this point. It's not like they do `libva2.14`. Linking `libva2` provides enough ABI compatibility for us.
It's a simple package manager install away. Even for unsupported system like nvdec, user can still have libva2 installed and have it fallback to software (or have https://github.com/elFarto/nvidia-vaapi-driver installed).

2. Patch FFmpeg to lazily dlopen libva2.

Needs a custom patch to FFmpeg and add maintenance burden.

3. Distribute 2 variants for Linux: vanilla and libva.

Ew.

I'd vote for 1.

arter97:
I took the liberty of extending this to all platforms.

https://github.com/libsdl-org/SDL/pull/8366/commits was used as a reference and tested on linux-x86_64, win64, winarm64, macos-x86_64, macos-aarch64.

Target device: Galaxy S25
linux-x86_64-1: Intel Core Ultra X7 358H on Ubuntu 26.04 (Wayland, GNOME)
linux-x86_64-2: Intel Atom E3950 on Ubuntu 24.04 (KDE, X11)
win64: AMD Ryzen 5 5600G
winarm64: Xiaomi Mi Pad 5
macos-x86_64: AMD Ryzen 5 5600G (OpenCore)
macos-aarch64: Apple M3 Pro (Mac15,7)

| Platform        | Before | After | Improvement |
| --------------- | -----: | ----: | ----------: |
| linux-x86_64-1 |  48.5% |  3.0% |       93.8% |
| linux-x86_64-2 | 120.0% | 16.7% |       86.1% |
| win64           | 103.2% |  3.6% |       96.5% |
| winarm64        | 112.0% |  8.0% |       92.9% |
| macos-x86_64    | 120.0% | 10.8% |       91.0% |
| macos-aarch64   |  50.2% | 19.8% |       60.6% |

On Windows, CPU usage reported by Task Manager (arm64) or TMX was multiplied by the number of cores for consistencies with other platforms.

Full transparency: While I guided the AI in a best-effort manner, >95% of the code is written by GPT 5.6 Sol or Claude Opus. Fable was used to audit/review.

The winarm64 testing was done by pulling #6857.
(Would be nice to see that merged.)
Windows needed some extra work to maintain --video-buffer compatibility. Even so, when a big enough value is used, software decode should be used.

arter97:
Prebuilt binaries:
https://github.com/arter97/scrcpy/releases/tag/hwaccel-test-1

rom1v:
> I took the liberty of extending this to all platforms.

Awesome :+1:  Thank you.

> Full transparency: While I guided the AI in a best-effort manner, >95% of the code is written by GPT 5.6 Sol or Claude Opus. Fable was used to audit/review.

Thank you for the transparency. It's great to have a first working codebase :+1:

I'll review/rework anyway (LLM or not).

> Windows needed some extra work to maintain --video-buffer compatibility

Yes, I had that in mind. I know what to do (store "receive date" for each packet, forward it to the decoded frame, and use this date instead of "now" in `video_regulator.c`). With that, you should not need to add an explicit limit in video buffer.

arter97:
> > I took the liberty of extending this to all platforms.
> 
> Awesome 👍 Thank you.
> 
> > Full transparency: While I guided the AI in a best-effort manner, >95% of the code is written by GPT 5.6 Sol or Claude Opus. Fable was used to audit/review.
> 
> Thank you for the transparency. It's great to have a first working codebase 👍
> 
> I'll review/rework anyway (LLM or not).
> 
> > Windows needed some extra work to maintain --video-buffer compatibility
> 
> Yes, I had that in mind. I know what to do (store "receive date" for each packet, forward it to the decoded frame, and use this date instead of "now" in `video_regulator.c`). With that, you should not need to add an explicit limit in video buffer.

Glad I kept it as a separate commit :)

I'm not familiar with video_regulator.c, but if it works after the decode, wouldn't that balloon the memory consumption if large enough --video-buffer is used? For 1080p, post-decode is a 3MB-ish bitmap and just a couple of those would increase scrcpy's memory usage rather drastically.

If it's pre-decode, it wouldn't matter much and that's what I personally chose for my app that uses scrcpy code with the hopeful assumption that the hardware decoder will mostly have even pacing/latency for decoding each frame and that works quite nicely.

rom1v:
> but if it works after the decode, wouldn't that balloon the memory consumption if large enough --video-buffer is used?

Yes, it is designed for "small enough" values. The goal is to compensate for jitter, to play the decoded frame at the best time (#2464).

But this cause a specific problem with hardware decoding: it is not possible to hold an arbitrary number of decoded frames, so some calls becomes blocking, causing the buffering timestamps to be wrong. That's why we need to explicitly pass the "receive date".

TomBeckett:
@rom1v Out of curiosity, once this abstraction has settled, would you be open to a follow-up PR adding Raspberry Pi 5 support?

The Pi 5 exposes HEVC decoding through the V4L2 stateless/Request API rather than VA-API, so it would need another Linux backend. I’d be happy to work on and test this.

PR #7006, opened 2026-08-30
Update dependencies

arter97:
Especially for dav1d, the existing URL was blocking GitHub workflows from working correctly.

### DIFF
--- a/app/deps/adb_linux.sh
+++ b/app/deps/adb_linux.sh
@@ -2,9 +2,9 @@
 set -ex
 . $(dirname ${BASH_SOURCE[0]})/_init "$@"
 
-VERSION=37.0.0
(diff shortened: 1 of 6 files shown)


rom1v:
Thank you.

But I usually upgrade dependencies just before a release.

> Especially for dav1d, the existing URL was blocking GitHub workflows from working correctly.

Why? It works for me.

About "Fetch dav1d from videolan FTP instead of code.videolan.org", I prefer using the sources directly from the git repo than a release that may be different (see be3d357a6dcdf0f3fccabd795b869a30e1e5bee1).

arter97:
Huh, weird. It spit out a 10kB html. Upon retrial, it's working now. I guess it was rate-limited or something. Sorry!

Issue #6492, opened 2025-11-22
Replace Reflection in scrcpy with Hidden Puppet Classes


mengyanshou:
## Using Hidden Classes

Sorry, I still feel that the extensive use of reflection in scrcpy makes the code very verbose.

I closed the previous issue because much of it had drifted off-topic.

If we use hidden classes, like the Shizuku Hidden API approach, we can simplify a lot of reflection calls.

I’m not sure if this should be called a “puppet class”—that might not be accurate—but that’s roughly what I mean.

This pattern has been somewhat validated; my open-source library [AAS](https://github.com/nightmare-space/android_api_server) also uses it extensively.

Could we try replacing some parts of the code first? I can test on emulators across multiple Android versions.

The rough changes are as follows:

Define a puppet class

Create classes according to the package name, similar to the existing android.content.IContentProvider

```java
public class ActivityThread {

    public ActivityThread() {
    }

    public static ActivityThread sCurrentActivityThread;

    public Boolean mSystemThread;
}
```

If we end up with many such puppet classes, we can create an inline library and import it into the server project; this is also how my project does it.

Call the APIs directly

```java
ActivityThread activityThread = new ActivityThread();
```

Take `Workarounds` as an example

Before
```java

public final class Workarounds {
    static {
        try {
            ACTIVITY_THREAD_CLASS = Class.forName("android.app.ActivityThread");
            Constructor<?> activityThreadConstructor = ACTIVITY_THREAD_CLASS.getDeclaredConstructor();
            activityThreadConstructor.setAccessible(true);
            ACTIVITY_THREAD = activityThreadConstructor.newInstance();
            Field sCurrentActivityThreadField = ACTIVITY_THREAD_CLASS.getDeclaredField("sCurrentActivityThread");
            sCurrentActivityThreadField.setAccessible(true);
            sCurrentActivityThreadField.set(null, ACTIVITY_THREAD);
            Field mSystemThreadField = ACTIVITY_THREAD_CLASS.getDeclaredField("mSystemThread");
            mSystemThreadField.setAccessible(true);
            mSystemThreadField.setBoolean(ACTIVITY_THREAD, true);
        } catch (Exception e) {
            throw new AssertionError(e);
        }
    }
}
```

After
```java
public final class Workarounds {
    static {
        try {
            ActivityThread activityThread = new ActivityThread();
            ActivityThread.sCurrentActivityThread = activityThread;
            activityThread.mSystemThread = true;
        } catch (Exception e) {
[...]

PR #7011, opened 2026-08-31
Call `IContentProvider.call()` directly through a compile-only stub

### LINKED ISSUE #7010: Use a compile-only framework stub for `IContentProvider.call()`
## Background

The scrcpy server uses reflection to access Android framework APIs that are hidden from the public SDK. Reflection is still necessary for private members and for APIs whose runtime compatibility cannot be trusted, but some framework APIs are callable at runtime and are only absent from the SDK `android.jar`.

I previously opened [#6492](https://github.com/Genymobile/scrcpy/issues/6492) to discuss using local framework declarations for these cases. At the time, I did not have enough time to prepare a series of small, testable changes. I would now like to revisit the idea step by step, starting with one limited call site. If you think this approach is reasonable, I would consider separate, focused proposals and PRs for other suitable APIs later.

The first candidate is `IContentProvider.call()` in the settings provider wrapper. scrcpy already exposes the framework object as `IContentProvider`, but the wrapper stores it as `Object` and invokes `call()` through reflection.

## Problem

For this call site, the current reflection layer:

- discards an existing useful framework type;
- performs reflective lookup and argument assembly for four historical signatures;
- introduces reflection-specific exception handling around a Binder call;
- packages the local `IContentProvider` declaration in the server DEX, even though the runtime definition must come from the Android framework.

The implementation must continue to support the historical signatures used from Android 9 and earlier through Android 12 and later. Compatibility with vendor-modified frameworks is also important, so older signatures should still be selected by runtime capability rather than only by `SDK_INT`.

## Proposed direction

- Move the local `IContentProvider` declaration to a small compile-only module.
- Keep the provider strongly typed and call `IContentProvider.call()` directly.
- Preserve the existing pre-Android 12 fallback order and cache the selected signature on the provider instance.
- Keep the public `SettingsException` contract unchanged.
- Ensure that the stub is available during compilation but is not defined in the final server DEX.

Direct calls produce Binder, runtime, and linkage failures instead of reflection-specific exceptions. The implementation should use only a missing intermediate method to select an older signature; other failures should retain their meaning and continue to be reported through `SettingsException`.

## Scope

This proposal only covers `IContentProvider.call()` in the settings provider wrapper. It does not attempt to replace reflection used for private fields, private constructors, or unrelated framework wrappers.

## Acceptance criteria

- `ContentProvider` no longer uses reflection for `IContentProvider.call()`.
- All four existing framework signatures remain supported.
- The pre-Android 12 runtime fallback and per-provider cache are preserved.
- The final server DEX references `IContentProvider` but does not define it.
- Both Gradle and non-Gradle server builds work.
- Runtime behavior is verified on representative Android versions before merging.

Laurie-Lin:
Closes #7010  
Related to [#6492](https://github.com/Genymobile/scrcpy/issues/6492)

## Summary

This change replaces reflection in the settings `ContentProvider` wrapper with typed calls to the hidden framework interface `android.content.IContentProvider`.

The local framework declaration is moved to a new `server-hidden-api` module and referenced by the server with `compileOnly`. It is therefore available during compilation but is not defined in the final server DEX. The non-Gradle build follows the same rule by compiling hidden API stubs into a separate classpath directory.

This PR is intentionally limited to one hidden interface and one call site. Reflection used for private framework members and unrelated wrappers is unchanged.

## Implementation

`ActivityManager` already returned an `IContentProvider`, but `ContentProvider` stored it as `Object`, found one of four `call()` overloads with `getMethod()`, assembled an `Object[]`, and invoked the selected `Method` reflectively.

The provider now remains strongly typed:

```java
private final IContentProvider provider;
```

The compile-only stub declares the same four overloads that the reflection code previously searched for. Android 12 and later use the `AttributionSource` signature. Older systems preserve the existing runtime fallback order:

```text
calling package + attribution tag + authority
    -> calling package + authority
        -> calling package only
```

This is intentionally capability-based rather than a strict `SDK_INT` mapping so that the existing behavior remains available for older or vendor-modified frameworks. Once an overload succeeds, its version is cached on the `ContentProvider` instance and subsequent calls on that provider reuse it.

Representative definitions in the official Android source:

- Android 12: [`AttributionSource` signature, lines 123-124](https://cs.android.com/android/platform/superproject/+/android-12.0.0_r1:frameworks/base/core/java/android/content/IContentProvider.java;drc=cebf5c06997b64f4e47a1611edb5f97044509d76;l=123)
- Android 11: [attribution-tag signature, lines 118-119](https://cs.android.com/android/platform/superproject/+/android-11.0.0_r1:frameworks/base/core/java/android/content/IContentProvider.java;drc=34a1b9c951c38537ab96b69bc308f6e0884823f5;l=118)
- Android 10: [authority signature, lines 82-83](https://cs.android.com/android/platform/superproject/+/android-10.0.0_r1:frameworks/base/core/java/android/content/IContentProvider.java;drc=57bb140be9e48cf08acba131f7463e461777bb8e;l=82)
- Android 9: [legacy signature, lines 60-62](https://cs.android.com/android/platform/superproject/+/android-9.0.0_r1:frameworks/base/core/java/android/content/IContentProvider.java;drc=6549309f6c473b792ec62d1aebadec62bcf07827;l=60)

## Compile-only packaging

The stub is needed to compile direct method references, but its runtime definition must be supplied by the Android framework:

```gradle
compileOnly project(':server-hidden-api')
```

The resulting DEX contains references to `IContentProvider` and its `call()` methods, but no `android.content.IContentProvider` class definition. `app_process` resolves those references against the framework class from the boot class path.

For `build_without_gradle.sh`, hidden API stubs are compiled into `hidden-api-classes`. That directory is used only on the javac and D8 classpaths, while actual server and generated AIDL classes remain in the program classes directory passed to D8.

## Exception handling

Removing reflection changes which failures are visible. The new catches preserve the old public behavior while retaining the real direct-call cause:

| Failure | Purpose and behavior |
| --- | --- |
| `NoSuchMethodError` | A missing intermediate pre-Android 12 overload selects the next historical signature. A missing final overload is not recovered. |
| `RemoteException` | Represents the checked Binder communication failure declared by `IContentProvider.call()`. |
| `RuntimeException` | Preserves framework failures such as `SecurityException`, which are exposed directly instead of being wrapped by `Method.invoke()`. |
| `LinkageError` | Reports a final incompatibility between the compile-time stub and runtime framework, such as `NoSuchMethodError`, `IllegalAccessError`, or `IncompatibleClassChangeError`. |

Only `NoSuchMethodError` around the two intermediate old-signature calls triggers fallback. Other linkage failures are logged and propagated without being treated as an Android-version difference.

The private wrapper logs direct invocation failures without changing their type:

```java
catch (RemoteException | RuntimeException | LinkageError e) {
    Ln.e("Could not call content provider", e);
    throw e;
}
```

At the public boundary, `getValue()` and `putValue()` continue to report `SettingsException`, with the direct failure retained as the cause. `LinkageError` is listed separately because it is an `Error`, not an `Exception`:

```java
catch (Exception | LinkageError e) {
    throw new SettingsException(..., e);
}
```

## Verification

Completed locally:

- [x] `git diff --check`
- [x] `./gradlew :server-hidden-api:check`
- [x] `./gradlew :server:check`
- [x] `./gradlew :server:assembleDebug`
- [x] `./gradlew :server:assembleRelease`
- [x] Debug build through `server/scripts/build-wrapper.sh`
- [x] Manual build through `server/build_without_gradle.sh` with Android platform 36 and build-tools 36.0.0
- [x] Unit tests, checkstyle, and Android lint
- [x] Debug and Release duplicate-class checks
- [x] D8 packaging
- [x] Release DEX contains `IContentProvider` references but no `IContentProvider` class definition

### Runtime compatibility

The settings `ContentProvider` path was tested against these Android Emulator system images:

- Android 5.0.2 (API 21), 5.1.1 (API 22), 6 (API 23), 7 (API 24), 7.1.1 (API 25), 8 (API 26), 8.1 (API 27), and 9 (API 28);
- Android 10 (API 29) and Android 11 (API 30);
- Android 12 (API 31), 12L (API 32), 13 (API 33), 14 (API 34), 15 (API 35), 16 (API 36), and 17 (API 37).

| Runtime API | Successful `IContentProvider.call()` signature |
| --- | --- |
| API 21-28 | calling package only |
| API 29 | calling package + authority |
| API 30 | calling package + attribution tag + authority |
| API 31-37 | `AttributionSource` |

For every image, the GET and PUT operations used by `--show-touches`, `--stay-awake`, and `--screen-off-timeout` completed successfully and the settings were restored after exit. No uncaught `NoSuchMethodError`, `LinkageError`, `RemoteException`, or `Could not call content provider` failure was observed.

After restoring the per-provider signature cache, signature selection and subsequent cache reuse were revalidated on Android 9, 10, 11, and 12, covering all four overload branches.

This matrix covers Android Emulator images only. Vendor-modified ROMs may change this part of the framework, so affected physical devices will still need observation. The runtime fallback for older signatures is retained for that reason.

## AI assistance

AI assistance was used for the runtime testing workflow and for drafting this PR description. The test procedure, test results, and final PR text were manually reviewed and verified.

### DIFF
--- a/server-hidden-api/build.gradle
+++ b/server-hidden-api/build.gradle
@@ -1,0 +1,9 @@
+apply plugin: 'com.android.library'
+
+android {
+    namespace = 'com.genymobile.scrcpy.hiddenapi'
+    compileSdk = 36
+    defaultConfig {
+        minSdkVersion 21
+    }
+}
--- a/server-hidden-api/src/main/AndroidManifest.xml
+++ b/server-hidden-api/src/main/AndroidManifest.xml
@@ -1,0 +1,1 @@
+<manifest />
--- a/server-hidden-api/src/main/java/android/content/IContentProvider.java
+++ b/server-hidden-api/src/main/java/android/content/IContentProvider.java
@@ -1,0 +1,17 @@
+package android.content;
+
+import android.os.Bundle;
+import android.os.IInterface;
+import android.os.RemoteException;
+
+// android.content.IContentProvider is hidden, this is a fake one to expose the type to the project
+public interface IContentProvider extends IInterface {
+
+    Bundle call(AttributionSource attributionSource, String authority, String method, String arg, Bundle extras) throws RemoteException;
+
+    Bundle call(String callingPackage, String attributionTag, String authority, String method, String arg, Bundle extras) throws RemoteException;
+
+    Bundle call(String callingPackage, String authority, String method, String arg, Bundle extras) throws RemoteException;
+
+    Bundle call(String callingPackage, String method, String arg, Bundle extras) throws RemoteException;
+}
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -26,6 +26,7 @@ android {
 }
 
 dependencies {
+    compileOnly project(':server-hidden-api')
     testImplementation 'junit:junit:4.13.2'
 }
 
--- a/server/build_without_gradle.sh
+++ b/server/build_without_gradle.sh
@@ -21,8 +21,10 @@ BUILD_TOOLS_DIR="$ANDROID_HOME/build-tools/$BUILD_TOOLS"
 
 BUILD_DIR="$(realpath ${BUILD_DIR:-build_manual})"
 CLASSES_DIR="$BUILD_DIR/classes"
+HIDDEN_API_CLASSES_DIR="$BUILD_DIR/hidden-api-classes"
 GEN_DIR="$BUILD_DIR/gen"
-SERVER_DIR=$(dirname "$0")
+SERVER_DIR="$(cd "$(dirname "$0")" && pwd)"
+HIDDEN_API_SRC="$SERVER_DIR/../server-hidden-api/src/main/java"
 SERVER_BINARY=scrcpy-server
 ANDROID_JAR="$PLATFORM_TOOLS/android.jar"
 ANDROID_AIDL="$PLATFORM_TOOLS/framework.aidl"
@@ -32,8 +34,9 @@ echo "Platform: android-$PLATFORM"
 echo "Build-tools: $BUILD_TOOLS"
 echo "Build dir: $BUILD_DIR"
 
-rm -rf "$CLASSES_DIR" "$GEN_DIR" "$BUILD_DIR/$SERVER_BINARY" classes.dex
+rm -rf "$CLASSES_DIR" "$HIDDEN_API_CLASSES_DIR" "$GEN_DIR" "$BUILD_DIR/$SERVER_BINARY" classes.dex
 mkdir -p "$CLASSES_DIR"
+mkdir -p "$HIDDEN_API_CLASSES_DIR"
 mkdir -p "$GEN_DIR/com/genymobile/scrcpy"
 
 << EOF cat > "$GEN_DIR/com/genymobile/scrcpy/BuildConfig.java"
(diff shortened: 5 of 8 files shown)


eyadmkv:
This is amazing!!!!

Issue #5622, opened 2024-12-06
I couldn't get the FPS to 120 anyway, even with 120 fps support turned on on my Android device, even with max_fps=120 set


cprogpb:
## Environment

- **OS:** Windows 10.0.26100
- **Scrcpy version:** 3.0.2                                                                                                                                                                                                                                                                                                                        
![微信图片_20241206164554](https://github.com/user-attachments/assets/62e88a31-4d1f-4f6e-afc1-54261fd276b9)

- **Installation method:** Windows release
- **Device model:** Redmi Turbo 3
- **Android version:** HyperOS 1.0.18.unpcnxm

## Describe the bug

Unable to achieve 120 FPS with Redmi Turbo 3 despite device supporting 120Hz refresh rate and proper configuration.

### Setup
- Device display settings: 120Hz enabled in HyperOS display settings
- USB connection: USB 3.0
- Launch parameter: `max_fps=120`

### Observed behavior
Performance metrics from logs show:
Performance stats:
Current FPS: 58.5-60.0
Avg frame interval: 16.68-17.08ms
Avg decode time: 0.03-0.14ms
Last receive time: 13.58-18.98ms

### Steps to reproduce
1. Enable 120Hz in HyperOS display settings
2. Enable USB debugging in developer options
3. Set USB configuration to maximum performance mode
4. Connect Redmi Turbo 3 via USB 3.0
5. Launch scrcpy with `max_fps=120`
6. Monitor FPS using built-in performance logging

### Expected behavior
FPS should reach close to 120 FPS since:
1. Redmi Turbo 3 hardware supports 120Hz display
2. Using USB 3.0 connection which has sufficient bandwidth
3. Decode times are very low (0.03-0.14ms) showing no performance bottleneck
4. HyperOS display settings are set to 120Hz

### Additional Information
1. Device specific:
   - Running HyperOS (MIUI successor)
   - USB debugging is enabled
   - USB configuration is set to maximum performance

2. Performance analysis:
   - Frame intervals are locked at ~16.7ms (60Hz)
   - Decoding is very efficient (0.03-0.14ms)
   - Receive times (13.58-18.98ms) suggest no bandwidth issues

### Questions
1. Is there any specific configuration needed for HyperOS devices?
2. Are there known issues with high refresh rates on Xiaomi/Redmi devices?
3. Is there a compatibility issue between current scrcpy version and HyperOS?

### What I've tried
1. Confirmed 120Hz is enabled in system settings
[...]

PR #7013, opened 2026-09-01
Request source refresh rate for mirrored displays

shanshulin:
## Summary

On some Android 14+ devices, a mirrored virtual display created through the legacy `DisplayManager.createVirtualDisplay()` API defaults to 60Hz even when the physical display is currently running at a higher refresh rate.

This makes the encoder receive at most about 60 frames per second. Passing `--max-fps=120` does not help, because `--max-fps` is only an encoder-side limiter.

This change:

- reads the current refresh rate of the source display;
- requests that same rate through `VirtualDisplayConfig.setRequestedRefreshRate()` for mirrored displays above 60Hz on Android 14+;
- keeps the existing virtual-display creation path for Android versions below 14, 60Hz displays, and cases where the refresh rate cannot be read.

`--max-fps` keeps its existing semantics and is not used to select the virtual display refresh rate.

## Motivation

Test device:

- HONOR AAK-AN00
- Android 16
- physical display running at 120.00001Hz
- scrcpy 4.1 macOS arm64 client

Before this change, the mirrored scrcpy virtual display was created at 60Hz. After requesting the source refresh rate, `dumpsys display` reports:

```text
source renderFrameRate=120.00001
scrcpy renderFrameRate=120.00001
mRequestedRefreshRate=120.00001
```

Packet timestamps from a 15-second recording while scrolling showed a peak one-second bin of 116 frames, instead of being capped around 60.

## Compatibility and limiter behavior

The legacy path is preserved for a 60Hz source (reported as `60.000004` on the test device):

```text
source renderFrameRate=60.000004
scrcpy renderFrameRate=60.0
mRequestedRefreshRate=0.0
```

With a 120Hz source and `--max-fps=60`, the virtual display remains at 120Hz while the encoded recording is limited to a peak one-second bin of 60 frames:

```text
source renderFrameRate=120.00001
scrcpy renderFrameRate=120.00001
mRequestedRefreshRate=120.00001
```

## Difference from #5505

#5505 added an arbitrary refresh-rate setting for new displays and connected it to `max_fps`. This PR only affects mirroring, automatically uses the refresh rate reported by the source display, and does not add a user-facing option or change `--max-fps` behavior.

Related: #5622

## Tests

- `server/build_without_gradle.sh`
- `./gradlew -p server check`
- Android 16 device test with a 120Hz source display
- Android 16 device test with a 60Hz source display
- Android 16 device test with a 120Hz source and `--max-fps=60`

### DIFF
--- a/server/src/main/java/com/genymobile/scrcpy/display/DisplayInfo.java
+++ b/server/src/main/java/com/genymobile/scrcpy/display/DisplayInfo.java
@@ -10,10 +10,12 @@ public final class DisplayInfo {
     private final int flags;
     private final int dpi;
     private final String uniqueId;
+    private final float refreshRate;
 
     public static final int FLAG_SUPPORTS_PROTECTED_BUFFERS = 0x00000001;
 
-    public DisplayInfo(int displayId, Size size, int rotation, int layerStack, int flags, int dpi, String uniqueId) {
+    @SuppressWarnings("checkstyle:ParameterNumber")
+    public DisplayInfo(int displayId, Size size, int rotation, int layerStack, int flags, int dpi, String uniqueId, float refreshRate) {
         this.displayId = displayId;
         this.size = size;
         this.rotation = rotation;
@@ -21,6 +23,7 @@ public final class DisplayInfo {
         this.flags = flags;
         this.dpi = dpi;
         this.uniqueId = uniqueId;
+        this.refreshRate = refreshRate;
     }
 
     public int getDisplayId() {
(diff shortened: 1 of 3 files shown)


rom1v:
Hi,

In https://github.com/Genymobile/scrcpy/pull/5505#issuecomment-2489369956 I implemented a version where we could specify a frame rate for a new display. But it did not work as expected.

I just rebased ([`display_capture_pts.3`](https://github.com/Genymobile/scrcpy/commits/display_capture_fps.3/)) and retested on Android 17: it works a bit better, it correctly limits the frame rate to a lower value, but it is still intrinsically limited by the device frame rate (if I configure 60Hz for the main display, I cannot create a virtual display at 120Hz, but if I configure 120Hz for the main display I can).

This MR just changes the way the frame rate is selected, it does not change the problems. I prefer keeping `display_capture_pts.3` somewhere, in case it works correctly in the future.

PR #7016, opened 2026-09-03
harden: use bounded strlcpy/snprintf in v4l2_sink.c...

anupamme:
## Summary
Harden input handling in `app/src/v4l2_sink.c` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | c.lang.security.insecure-use-string-copy-fn.insecure-use-string-copy-fn |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `c.lang.security.insecure-use-string-copy-fn.insecure-use-string-copy-fn` |
| **File** | `app/src/v4l2_sink.c:203` |
| **Assessment** | Defensive hardening |

**Description**: Finding triggers whenever there is a strcpy or strncpy used. This is an issue because strcpy does not affirm the size of the destination array and strncpy will not automatically NULL-terminate strings. This can lead to buffer overflows, which can cause program crashes and potentially let an attacker inject code in the program. Fix this by using strcpy_s instead (although note that strcpy_s is an optional part of the C11 standard, and so may not be available).

## Threat Model Context

## Changes
- `app/src/v4l2_sink.c`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---

### DIFF
--- a/app/src/v4l2_sink.c
+++ b/app/src/v4l2_sink.c
@@ -200,8 +200,8 @@ sc_v4l2_sink_open(struct sc_v4l2_sink *vs, const AVCodecContext *ctx,
         goto error_avformat_free_context;
     }
 #else
-    strncpy(vs->format_ctx->filename, vs->device_name,
-            sizeof(vs->format_ctx->filename));
+    sc_strncpy(vs->format_ctx->filename, vs->device_name,
+               sizeof(vs->format_ctx->filename));
 #endif
 
     AVStream *ostream = avformat_new_stream(vs->format_ctx, encoder);


rom1v:
Thank you, merged: c29743af4bca0e4a2aa2b30b258099445ed91b16 :rocket: 
I also added: 6ce3d289145bf8acf4eeeaef0ce334f4d448acd7

(there is no real security issue here, but it must still be fixed)

PR #7029, opened 2026-09-06
Fix timeout startup race

cuishuang:
`sc_timeout_start()` created the timeout thread before initializing the data read by `run_timeout()`.

`SDL_CreateThread()` may start the worker before it returns, so `run_timeout()` could read `timeout->deadline` before it was assigned and access the callback fields before they were initialized. This creates a real startup race and could make `--time-limit` use an invalid deadline or callback state.

Validate the callback and initialize `deadline`, `cbs`, and `cbs_userdata` before creating the timeout thread.

This guarantees that the worker only starts after all data it needs has been initialized.


rom1v:
Thank you, good catch :+1:

Merged into `dev`: 7e5d47038857127f73204ecfbf9770a012aabc5d

Issue #7042, opened 2026-09-18
Question: Android controller client scope


Moa1er:
Hello. I have implemented an Android controller client that connects to a target Android device over ADB TCP, embeds scrcpy video and control, supports multiple saved connections, direct app launch, locked-target control, screen-off on disconnect, and encrypted local ADB key storage. Before asking for review, could you confirm whether this functionality belongs in the scrcpy project? The proposed implementation is available here: https://github.com/Moa1er/scrcpy/tree/android-controller. If it is out of scope, guidance on the preferred repository or architecture would be appreciated.

PR #7043, opened 2026-09-18
Add Android controller client

Moa1er:
## Summary

- Add an Android controller client for scrcpy.
- Connect to multiple saved Android targets over ADB TCP.
- Stream H.264 video and send touch, back, app-launch, and disconnect controls.
- Support direct launch of a configured target app, including while the target display is locked.
- Turn off the target display when disconnecting.
- Store the ADB authentication key encrypted with Android Keystore and exclude it from backups.
- Add automatic video resolution adaptation between the configured maximum and 720p without reconnecting the ADB session.
- Add manual maximum video sizes of 720p, 960p, 1280p, and 1440p.
- Add Android unit tests, an instrumentation smoke test, lint, R8 release configuration, and API 35 emulator CI.

## Validation

- `./gradlew check` passes.
- Android debug and R8 release APKs build successfully.
- JVM tests pass.
- Android and server lint checks pass with zero errors.
- The debug APK was installed and launched on a physical Android 15 device.

## Security and limitations

ADB-over-Wi-Fi is plaintext and does not cryptographically authenticate the target device. This limitation is documented in the app and documentation. The release APK is intentionally unsigned and requires signing by the distributor.

This PR follows the scope question in #7042.


Moa1er:
Hi @rom1v, could you please take a look at this proposal when you have time?

It adds an experimental Android controller client for controlling another Android device over ADB TCP, including video, touch input, saved connections, app launching, and adaptive resolution.

The Android workflow passes unit tests, lint, release compilation, and emulator smoke tests. I would especially appreciate feedback on whether the architecture and scope fit scrcpy, even if changes are needed before merge.

Thank you !!

rom1v:
Hi,

Thank you for sharing this. However, an Android client is out of scope for scrcpy.
This is a completely separate app, that does not share code with the desktop client (which would be complicated anyway, since the constraints are not the same).

I cannot (and don't want to) maintain a separate app, with thousands of (vibe-coded) lines of code.

You can publish this app in a separate git repo (under a name that does not contain "scrcpy", to avoid confusing it with scrcpy).

Moa1er:
Thanks for the quick review, and for taking the time to look at the proposal.

I thought reusing scrcpy's existing `:server` module, packaging it as `scrcpy-server.jar`, and implementing the Android client around the scrcpy ADB, video, and control protocols would make this sufficiently connected to scrcpy to be interesting as part of the project. The Android-side transport, rendering, controls, connection management, and UI are new, though, so I understand your point that this would still be a separate application to maintain.

I will create a separate repository for it under a name that does not contain "scrcpy". Thanks again!

Moa1er:
The Android controller has been published as an independent project at https://github.com/Moa1er/android-remote-controller, as discussed.

